PR #5, opened 2024-07-17
Add block height

patimen:
Started by editing the proto file and adding the BlockHeight parameter as an int64.
Then the command:
`ignite generate proto-go` to regenerate the proto files
then simply assigning the value when we process the Start event
**Note:** Couldn't find an easy way to test multiple values for BlockHeight in the Context, but it seems so trivial I'm not too worried.


inference-chain/proto/inference/inference/inference.proto
```diff
@@ -4,17 +4,17 @@ package inference.inference;
...
+  string promptHash = 3;
+  string promptPayload = 4;
+  string responseHash = 5;
+  string responsePayload = 6;
+  uint64 promptTokenCount = 7;
+  uint64 completionTokenCount = 8;
+  string receivedBy = 9;
+  string executedBy = 10;
+  string status = 11;
+  int64 blockHeight = 12;
```
DimaOrekhovPS: What do you think about having both `blockHeightReceived` and `blockHeightFinished`?

DimaOrekhovPS (approved):

PR #7, opened 2024-07-23
Add a basic participant:

patimen:
Initial command:`ignite scaffold map participant address reputation:int weight:int joinTime:uint joinHeight:int lastInferenceTime:uint --index index --module inference --no-message`

But lots of changes after that, most of them changes in the proto file.
ALso added the SubmitNewParticipant command and added logic to add finished inference token totals to the Participant record.


inference-chain/proto/inference/inference/participant.proto
```diff
@@ -0,0 +1,25 @@
...
+  string address = 2;
+  int32 reputation = 3;
+  int32 weight = 4;
+  int64 joinTime = 5;
+  int64 joinHeight = 6;
+  int64 lastInferenceTime = 7;
+  string inferenceUrl = 8;
+  repeated string models = 9;
+  ParticipantStatus status = 10;
+  map<string, uint64> promptTokenCount = 11;
```
DimaOrekhovPS, line 23: Why do you think we need to store token counts?

patimen: I wanted something to show the history of the participant for now, and since we didn't have a firm idea of how to manage reputation and weight I thought I'd store the raw input.

PR #8, opened 2024-07-26
Dockerize

DimaOrekhovPS:
Set up a docker-compiose.yml file for conveniently launching both services locally. Later it will evolve into production compose file.

### DIFF
--- a/decentralized-api/main.go
+++ b/decentralized-api/main.go
@@ -38,7 +38,7 @@ func main() {
 	go func() {
 		StartInferenceServerWrapper(*recorder, config)
 	}()
-	go func() {
-		StartValidationScheduledTask(*recorder, config)
-	}()
+	//go func() {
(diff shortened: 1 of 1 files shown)


decentralized-api/config-docker.yaml
```diff
@@ -0,0 +1,11 @@
+nodes:
+  - id: node1
+    url:  http://34.171.235.205:8080/ # Deployed inference node
+    max_concurrent: 500
+    models:
+      - unsloth/llama-3-8b-Instruct
+chain_node:
```
DimaOrekhovPS, line 7: I'm more used to kebab-case in config files :)
At the same time I think consistency is what matters, so I went and checked Tendermint (config.toml) and Cosmos (app.toml) configs. Tendermint uses snake_case and cosmos uses kebab 🤦 
So I guess we can choose whatever, let's stick to snake_case then

PR #10, opened 2024-08-01
Validation

DimaOrekhovPS:


decentralized-api/cosmos-client.go
```diff
@@ -117,3 +118,19 @@ func (icc *InferenceCosmosClient) FinishInference(transaction *inference.MsgFini
 	println(response.Data)
 	return nil
 }
+
+func (icc *InferenceCosmosClient) Validation(transaction *inference.MsgValidation) error {
```
patimen: Probably something like a verb here? RecordValidation?

decentralized-api/cosmos-client.go
```diff
@@ -117,3 +118,19 @@ func (icc *InferenceCosmosClient) FinishInference(transaction *inference.MsgFini
...
+	transaction.Creator = icc.address
+	return icc.sendTransaction(transaction)
+}
+
+func (icc *InferenceCosmosClient) sendTransaction(msg sdk.Msg) error {
```
patimen, line 113: Might as well refactor Finish and Start Inference to use this.

decentralized-api/validation.go
```diff
@@ -26,25 +31,164 @@ func StartValidationScheduledTask(transactionRecorder InferenceCosmosClient, con
...
+		originalLogprobs = append(originalLogprobs, o.Logprob)
+		validationLogprobs = append(validationLogprobs, v.Logprob)
+	}
+
+	cosSimValue := cosineSimilarity(originalLogprobs, validationLogprobs)
+
+	return &CosineSimilarityValidationResult{BaseValidationResult: baseComparisonResult, Value: cosSimValue}
+}
+
+func cosineSimilarity(a, b []float64) float64 {
```
patimen, line 186: It's a nit, but:
Probably want to check that a and b have same length
Also want to check that neither magnitude is zero.

patimen (approved):
Just a few nits, let's go ahead and move forward.

decentralized-api/validation.go
DimaOrekhovPS: We check length upstream so I think it's ok to omit it here since it's a private function.

I also thought about checking that neither of vectors is a zero vector, but I'm not sure that we can do a better error handling than what's already happening?

In my understanding we'll just get some panic error and whole method will be aborted.

Let me leave a comment to think about it later

DimaOrekhovPS: Moreover cosine similarity won't stay with us for long

patimen: Sure, if it isn't staying. I just would say in general we should strive to implement mathematical functions like this as general purpose as can reasonably be done, so we can safely reuse them (which we very may well as we go).

PR #11, opened 2024-08-02
Rudimentary reputation 

patimen:
Very simple for now:
Assume a False POsitive Rate (hardcoded for now), measure the z-score (distance from expected in Standard Deviation), assume greater than 1 std-dev should be validated for every inference, lower only 1 in 10.

Also add a Ramping for new Participants, since we should be checking every inference until they reach some minimum.

I derive the minimum needed from the FPR, using the minimum value that a single failure would be considered OK as a guess. So, the lower the FPR, the more inferences needed to say they're good to go. Not perfect, but better than an arbitrary number


inference-chain/proto/inference/inference/participant.proto
```diff
@@ -7,6 +7,8 @@ enum ParticipantStatus {
   UNSPECIFIED = 0;
   ACTIVE = 1;
   INACTIVE = 2;
+  INVALID = 3;
+  RAMPING = 4;
```
DimaOrekhovPS, line 11: What's the effective difference between `INVALID` and `RAMPING`? Can we just leave `INVALID`?

Because technically we want to validate all requests of both new and non-trusted participants

inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -10,8 +11,79 @@ import (
...
+	creator.InferenceCount++
+	if passed {
+		creator.ValidatedInferences++
+	} else {
+		creator.InvalidatedInferences++
+	}
+	// Where will we get this number? How much does it vary by model?
+	falsePositiveRate := 0.05
+
+	creator.Status = calculateStatus(falsePositiveRate, creator)
```
DimaOrekhovPS: `creator` is the person validating the request, we need to change not their status, but the status of the original inference executor.

You should use `MsgValidation.InerenceId` to obtain inference entity and then `executedBy`

inference-chain/proto/inference/inference/participant.proto
patimen: I just think it describes things better, and I prefer to have accuracy in my naming. Maybe we'll treat them differently in the future?

inference-chain/proto/inference/inference/inference.proto
```diff
@@ -14,7 +21,7 @@ message Inference {
   uint64 completionTokenCount = 8;
   string receivedBy = 9;
   string executedBy = 10;
-  string status = 11;
+  InferenceStatus status = 11;
```
DimaOrekhovPS, line 24: Does it serialize/deserialize ok? Is it backward compatible? Just curious. Of course it's ok to just reset the DB for now

inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -2,16 +2,107 @@ package keeper
 
 import (
 	"context"
+	"math"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/productscience/inference/x/inference/types"
 )
 
+const FalsePositiveRate = 0.05
```
DimaOrekhovPS: I also saw a syntax like this for constants:
```
const(
    FalsePositiveRate = 0.05
    MinRampUpMeasurements = 10
    ...
)
```

DimaOrekhovPS (approved):

inference-chain/proto/inference/inference/inference.proto
patimen: I don't think it is backwards compatible. But I think that's ok.

inference-chain/x/inference/keeper/msg_server_validation.go
patimen: OK, when in Rome.

PR #14, opened 2024-08-12
First stab at payments

patimen:
In theory:
1. Calculate cost in "icoin" of inference by multiplying total tokens by 1000.
2. Move total cost to escrow account on inference requested
3. Use a DEFAULT_MAX_TOKENS for now instead of calculating from the request. That's what gets put in escrow.
4. Had to make sure all the test addresses were valid
5. For now, keep track of coin balance. Add the reconciliation/transfer event next.
6. Still need unit tests for the actual transfer to Escrow

Mocks using go.uber.org gomock, which seems pretty standard. Not the new Makefile target for rebuilding mocks if needed.

For now, just want to get this working in the most basic form. NOTE:
We are using ignite 28.x.x and Cosmos SDK 0.50+. This means the checkers tutorial is wrong in several ways. I'll try and enumerate:
1. `app.go` is v2, not v1. This means it relies on injection, not on explicit assignment. Definition of Keeper struct is a similar change, but adding the BankEscrowKeeper is done through the injection for ModuleInputs (thus the add of the Keeper there). I think.
2. permissions seem to be in the moduleAccPerms (module level), instead of maccPerms as in the tutorial. I am still not feeling 100% on this, we'll have to test.
3. Interface definition seems pretty much the same.
4. Mocking for tests is very similar, but uses the latest version of the mocks (go.uber.org version) 5.


inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -44,11 +45,20 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 	} else {
 		inference.Status = types.InferenceStatus_INVALIDATED
 		executor.InvalidatedInferences++
+		executor.CoinBalance -= CalculateCost(inference)
+		// We need to refund the cost, so we have to lookup the person who paid
+		payer, found := k.GetParticipant(ctx, inference.ReceivedBy)
+		if !found {
+			return nil, types.ErrParticipantNotFound
+		}
+		payer.CoinBalance += CalculateCost(inference)
```
DimaOrekhovPS: Can we just transfer them the money directly? Or are we concerned that the validation itself may be wrong and we're still waiting for some "end of the period"?

DimaOrekhovPS (approved):

PR #17, opened 2024-08-30
Staking v0

patimen:
- Participant needs the PubKey when submitted if it wants to get any power, so we add ValidatorKey to both Participant and the MsgSubmitNewParticipant message. It's a string of the encoded version of the ed2519 PubKey type. I tried transmitting directly as a PubKey instead of a string, but the transition through HTTP and then RPC and then into the code just proved too tricky ATM.

fund_accounts.py submits this correctly as an example.

- SetComputeResults takes a list of ComputeResult (containing key,power and operator address)
- For this moment (for testing) we are faking the SetComputeResults call whenever we submit a new participant. Ultimately this will ge done at the end of the PoC sequence.
-

- Point to local Cosmos version (needs to be cleaned for both dev and deploy scenarios)
- Also fixed up how the dependencies worked between decentralized-api and inference-chain. go mod tidy was failing, and something about all the changes here and in cosmos caused issues.


decentralized-api/server.go
```diff
@@ -446,6 +452,45 @@ func getParticipants(recorder InferenceCosmosClient, w http.ResponseWriter, requ
...
+	// Index validators by address
+	validatorMap := make(map[string]types2.Validator)
+	for _, v := range validators.Validators {
+		log.Printf("-Validator info:")
+		log.Printf("Validator: %s", v.Address)
+		// Use public key... account is based on this anyhow
+		s := v.PubKey.Address().String()
+		log.Printf("PubKey: %s", s)
+		log.Printf("VotingPower: %d", v.VotingPower)
+		accAddress := sdk.AccAddress(v.PubKey.Address()).String()
```
DimaOrekhovPS, line 485: Were you testing/debugging something here? The PubKey can't be all these 3 things at once, right?

inference-chain/x/inference/keeper/msg_server_submit_new_participant.go
```diff
@@ -2,19 +2,23 @@ package keeper
...
 func (k msgServer) SubmitNewParticipant(goCtx context.Context, msg *types.MsgSubmitNewParticipant) (*types.MsgSubmitNewParticipantResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
-
+	logger := k.Logger().With("op", "submitNew")
 	newParticipant := types.Participant{
 		Index:                msg.Creator,
 		Address:              msg.Creator,
 		Reputation:           1,
-		Weight:               1,
+		Weight:               int32(ctx.BlockHeight()),
```
DimaOrekhovPS, line 21: Should we put -1/nil until they participate in their first PoC round?

inference-chain/x/inference/keeper/msg_server_submit_new_participant.go
```diff
@@ -23,13 +27,46 @@ func (k msgServer) SubmitNewParticipant(goCtx context.Context, msg *types.MsgSub
...
 		CompletionTokenCount: make(map[string]uint64),
+		ValidatorKey:         msg.ValidatorKey,
 	}
-
 	for _, model := range msg.Models {
 		newParticipant.PromptTokenCount[model] = 0
 		newParticipant.CompletionTokenCount[model] = 0
 	}
 	k.SetParticipant(ctx, newParticipant)
+	response, err := k.updateComputeResults(ctx, logger)
```
DimaOrekhovPS, line 37: Could you please explain your approach here? Is it for debug purposes, updating the power map as soon as someone joins?

inference-chain/x/inference/module/module.go
```diff
@@ -212,5 +218,9 @@ func ProvideModule(in ModuleInputs) ModuleOutputs {
 		in.BankKeeper,
 	)
 
-	return ModuleOutputs{InferenceKeeper: k, Module: m}
+	return ModuleOutputs{
+		InferenceKeeper: k,
+		Module:          m,
+		Hooks:           stakingtypes.StakingHooksWrapper{StakingHooks: StakingHooksLogger{}},
```
DimaOrekhovPS, line 224: What do hooks do in Cosmos? I've never used them so far

EDIT: Oh, I see below. So you mostly hook for observability into what's happening?

DimaOrekhovPS (approved):

decentralized-api/go.mod
```diff
@@ -163,6 +166,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace inference => ../inference-chain
-
-replace github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
+replace (
+	github.com/cosmos/cosmos-sdk => ../../ps-fork/cosmos-sdk
```
patimen, line 170: This is where the dependency to a local fork of the cosmos-sdk should go. I put it in a subfolder because I had a current download of the main repo already

decentralized-api/server.go
patimen: Yeah, experimenting with how the sdk uses pubkeys to generate addresses.

patimen: For debugging/demoing, we can leave these in for now.

decentralized-api/server.go
```diff
@@ -487,3 +534,36 @@ func getParticipants(recorder InferenceCosmosClient, w http.ResponseWriter, requ
 	w.WriteHeader(http.StatusOK)
 	w.Write(responseJson)
 }
+
+func getVotingPower(recorder InferenceCosmosClient, address string, validatorMap map[string]types2.Validator) int64 {
```
patimen, line 538: This is not currently working right. We should probably pull voting power from the actual tendermint data, not from our local cache of power.

inference-chain/x/inference/module/module.go
patimen: Yeah, this was helpful for debugging. Remove, you think?

DimaOrekhovPS: Let's keep

PR #18, opened 2024-09-04
PoC: orchestration, execution, weight updates

DimaOrekhovPS:


decentralized-api/event_listener.go
```diff
@@ -63,27 +28,45 @@ func StartEventListener(nodeBroker *broker.Broker, transactionRecorder Inference
...
 		}
 
-		go func() {
-			SampleInferenceToValidate(txEvent.Result.Events["inference_finished.inference_id"], transactionRecorder, nodeBroker)
-		}()
+		switch event.Result.Data.Type {
+		case "tendermint/event/NewBlock":
+			log.Printf("New block event received. type = %s", event.Result.Data.Type)
+			pow.ProcessNewBlockEvent(powOrchestrator, &event, transactionRecorder)
+		default:
```
patimen: Maybe for consistency we should be specific here and catch the inference event? We'll probably add more soon enough.

decentralized-api/pow/orchestrator.go
```diff
@@ -0,0 +1,244 @@
...
+
+type ProofOfWork struct {
+	BlockHeight int64
+	BlockHash   string
+	PubKey      string
+	Nonce       string
+	ProofHash   string
+}
+
+func NewPowOrchestrator(pubKey string, difficulty int) *POWOrchestrator {
```
patimen: Should be consistently PoW OR POW (or PoC, that's another discussion)

decentralized-api/pow/orchestrator.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+}
+
+func (o *POWOrchestrator) acceptHash(hash string) bool {
+	return proofofcompute.AcceptHash(hash, o.difficulty)
+}
+
+// startProcessing is the function that starts when a start event is triggered
+func (o *POWOrchestrator) startProcessing(event StartPowEvent) {
+	o.mu.Lock()
```
patimen: Regarding the mutex and locking...
This is probably ok for now. But it would definitely be more idiomatic and safe to use channels here, with a channel for adding/getting results and running managed by one goroutine that the other components send messages to, almost like an Actor model. This is especially true since our architecture is already fundamentally message based.

decentralized-api/pow/orchestrator.go
patimen: We should standardize between:
PoW
POW
PoC
I'd go with PoC...

decentralized-api/apiconfig/config.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type ChainNodeConfig struct {
+	Url            string `koanf:"url"`
+	AccountName    string `koanf:"account_name"`
+	KeyringBackend string `koanf:"keyring_backend"`
+	KeyringDir     string `koanf:"keyring_dir"`
+}
+
+func ReadConfig() Config {
```
patimen, line 29: Good cleanup to put this into it's own code section.

inference-chain/x/inference/keeper/msg_server_submit_pow.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+func (k msgServer) SubmitPow(goCtx context.Context, msg *types.MsgSubmitPow) (*types.MsgSubmitPowResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	currentBlockHeight := ctx.BlockHeight()
+	startBlockHeight := msg.BlockHeight
+
+	if !proofofcompute.IsStartOfPocStage(startBlockHeight) {
+		errMsg := fmt.Sprintf("start block height must be divisible by 240. msg.BlockHeight = %d", startBlockHeight)
```
patimen: Probably want 240 to reference config?

inference-chain/x/inference/proofofcompute/epoch.go
```diff
@@ -0,0 +1,24 @@
+package proofofcompute
+
+const epochLength = 240
+
+func IsStartOfPocStage(blockHeight int64) bool {
+	return isNotZeroEpoch(blockHeight) && blockHeight%epochLength == 0
+}
+
+func IsEndOfPocStage(blockHeight int64) bool {
+	return isNotZeroEpoch(blockHeight) && blockHeight%epochLength == 60
```
patimen: Let's put all the consts together at the top of the file so we can easily see how the window looks.
Of course, this should ultimately be a matter up for a consensus vote/change by the chain, right?

PR #20, opened 2024-09-10
Load Balancing and Auth Keys

patimen:
## Auth Keys:
1. Lightweight call to get a single participants URL and address. (might not be needed once we have LB integrated)
2. Add Transfer/Executor modes to the dApp based on the headers.
3. Transfer nodes get the Authorization header (the signature for the payload based on the Requesters PrivKey) and the identity of the Requester (X-Requester-Address)
4. Executor nodes get the Authorization header, the PubKey for the Requester, the ID of the Inference and the random Seed to use (since we cannot change the payload, but need to know the seed to create the Inference)
5. Added CLI commands to create and verify signatures of arbitrary texts based on the primary Local signature (or any address specified). See `inferenced signature`.
6. Made all modifications we make to the inference request body go into ModifyBodyRequest


decentralized-api/server.go
```diff
@@ -81,36 +98,302 @@ func wrapGetCompletion(recorder cosmos_client.InferenceCosmosClient) func(w http
...
+	log.Print("SHA of bytes (verification):" + encoding)
+
+	valid := actualKey.VerifySignature(request.Body, keyBytes)
+	if !valid {
+		return errors2.New("invalid signature")
+	}
+	return nil
+}
+
+func processGetInferenceParticipantByAddress(w http.ResponseWriter, request *http.Request, recorder cosmos_client.InferenceCosmosClient) {
```
DimaOrekhovPS, line 363: I guess soon we'll need to split this server file into dedicated endpoints

inference-chain/x/inference/keeper/query_inference_participant.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func (k Keeper) InferenceParticipant(goCtx context.Context, req *types.QueryInferenceParticipantRequest) (*types.QueryInferenceParticipantResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO: Process the query
```
DimaOrekhovPS, line 20: Remove TODO?

DimaOrekhovPS (approved):

PR #21, opened 2024-09-13
Testermint initial checkin

patimen:
## Testermint
LocalInferencePair is the two local docker containers that have matching ApplicationAPI (decentralizedapi) and ApplicationCLI (inferenced) instances.
ApplicationCLI is a wrapper over docker commands sent to inferenced. 
ApplicationAPI is a wrapper over the REST endpoints for the decentralizedapi.
It's not perfect, but it's a lot better than mucking around with Python.
## Logging improvements
Instead of dumping everything to the console, which shows up as a generic log level, use the Cosmos Logger to log with proper log levels, etc. Added some methods to make this fluent/easy.

## Account Settling
For every Epoch, at the same time we change power we also settle all amounts owed or refunds owed.
- Participants track CoinsOwed (owed for work done) and RefundsOwed (overpayments for inference in escrow) separately
- At Epoch end:
  - Pay Owed
  - Issue Refunds
  - Calculate reward coins, based on the amount of work done (NOT refunds).

## Things left to do:
1. Handle rounding issues. The rewards might have a few coins left over (we round down), those coins should be put somehwere.
2. Renaming - As part of a future event I'm calling "The Great Rename", we should take a look at how we've named these amounts, they aren't quite expressive enough.
3. Logging improvements - with the new logging levels, we should revisit logging in various places so levels make sense, especially for debugging tests.
4.


inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -44,9 +44,9 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
 	refundAmount := existingInference.EscrowAmount - existingInference.ActualCost
 	if refundAmount > 0 {
```
DimaOrekhovPS, line 45: Can it potentially be < 0? Or the executor will stop the execution if so?

inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -52,9 +52,9 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 		if payer.Address == executor.Address {
 			// It is possible that a participant returns an invalid
```
DimaOrekhovPS, line 53: But why is that a separate case? And why don't you follow up with a SetParticipant call?

decentralized-api/server.go
```diff
@@ -652,6 +652,7 @@ type ParticipantDto struct {
 	Url         string   `json:"url"`
 	Models      []string `json:"models"`
 	CoinsOwed   uint64   `json:"coins_owed"`
+	RefundsOwed uint64   `json:"refunds_owed"`
```
DimaOrekhovPS, line 655: To me the naming is a bit ambiguous:

1. Is it the amount I as a participant owe?
2. Is it the amount owed to me?

My idea would be `ToBeRefunded`, but I guess it's also ambiguous haha. `ExpectsRefunds`. `PendingRefundsAmount`

decentralized-api/server.go
```diff
@@ -652,6 +652,7 @@ type ParticipantDto struct {
 	Url         string   `json:"url"`
 	Models      []string `json:"models"`
 	CoinsOwed   uint64   `json:"coins_owed"`
```
DimaOrekhovPS, line 654: CoinsOwed -- the amount that was withdrawn from me and put into escrow during the current epoch
RefundsOwed -- the amount that will be refunded to me at the end end of epoch

So it's true that RefundOwed alway less than CoinsOwed?
Just trying to understand what's going on :) Solidifying my understanding

decentralized-api/server.go
patimen: I think we need an entire PR with nothing but renames soon. Once we release we won't be able to really change the names, and our naming has been haphazard at best so far. But I support we can start here.

decentralized-api/server.go
patimen: CoinsOwed - amount owed for the inference work I've actually done
RefundsOwed - amount owed for overpayment of escrow.

inference-chain/x/inference/keeper/msg_server_validation.go
patimen: The difference IS the setParticipant. If I remember correctly, we originally didn't have this branch and things failed when we tried to SetParticipant for the same participant twice. Since in this branch executor == payer, we do SetParticipant later on line 64.

inference-chain/x/inference/keeper/payment_handler.go
```diff
@@ -16,9 +16,60 @@ func (k *Keeper) PutPaymentInEscrow(ctx context.Context, inference *types.Infere
...
+		k.LogError("Error sending coins to escrow", "error", err)
+		return 0,
+			sdkerrors.Wrapf(err, types.ErrRequesterCannotPay.Error())
 	}
+	k.LogInfo("Sent coins to escrow", "inference", inference.InferenceId, "coins", cost)
 	return cost, nil
 }
+
+func (k *Keeper) MintRewardCoins(ctx context.Context, newCoins uint64) error {
+	return k.bank.MintCoins(ctx, types.ModuleName, getCoins(newCoins))
```
DimaOrekhovPS, line 30: To where are they getting allocated? Module balance? So then we can use them in SendCoinsFromModuleToAccount?

inference-chain/x/inference/keeper/msg_server_finish_inference.go
patimen: It's a uint, so no, it cannot be < 0.

inference-chain/x/inference/keeper/payment_handler.go
patimen: Yes, we mint the rewards coins into the same escrow account, so that when we pay out earnedCoins + refundCoins + rewardCoins there will be enough.

DimaOrekhovPS (approved):

inference-chain/x/inference/keeper/msg_server_finish_inference.go
DimaOrekhovPS: But I mean technically. 
What happens if inference costs to the executor more than we initially estimated?

decentralized-api/server.go
DimaOrekhovPS: ok, so at the moment `Owed` postfix alway means Owed to me (this participant)

inference-chain/x/inference/keeper/msg_server_finish_inference.go
patimen: The idea is it can't. If it somehow does we should cap it. By setting a limit for tokens, the requester is setting a limit for how much they'll pay.
It does raise an issue that if a requester want's a potentially HUGE response, they may not want to put the ENTIRE amount in escrow each time, but that's perhaps an issue for another day.

decentralized-api/server.go
patimen: Yes, exactly. Since with escrow there should be no way for the Owe to go the other way.

PR #22, opened 2024-09-18
Streaming

DimaOrekhovPS:


decentralized-api/server.go
```diff
@@ -316,26 +411,48 @@ func handleExecutorRequest(w http.ResponseWriter, request *ChatRequest, nodeBrok
...
-	// Copy the response back to the client
-	for key, values := range respWithBody.Response.Header {
-		// Skip Content-Length, because we're modifying body
-		if key == "Content-Length" {
-			continue
-		}
-		for _, value := range values {
-			w.Header().Add(key, value)
-		}
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
patimen, line 434: So, we lose the reason for the error? This could be frustrating from the end-user perspective.

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -36,17 +37,24 @@ func (am AppModule) SendNewValidatorWeightsToStaking(ctx context.Context, blockH
...
 		}
-		log.Printf("Setting compute validator: %v", r)
+		am.LogInfo("Setting compute validator: %v", r)
 		computeResults = append(computeResults, r)
 	}
 
+	am.keeper.RemoveAllPower(ctx)
+
+	if len(computeResults) == 0 {
+		am.LogWarn("No compute validators to set. Keeping validators and active participants the same.")
```
patimen, line 47: But didn't we just RemoveAllPower?

testermint/src/main/kotlin/ApplicationAPI.kt
```diff
@@ -69,3 +88,45 @@ fun logResponse(reqData: Triple<Request, Response, Result<*, FuelError>>) {
...
+    // Check response code
+    val responseCode = connection.responseCode
+    if (responseCode == HttpURLConnection.HTTP_OK) {
+        // Read the event stream line by line
+        val reader = BufferedReader(InputStreamReader(connection.inputStream))
+        var line: String?
+
+        // Continuously read from the stream
+        while (reader.readLine().also { line = it } != null) {
+            println(line)
```
patimen: Could we use TinyLog here instead? Logger.Debug or Logger.Trace?

inference-chain/x/inference/module/chainvalidation.go
DimaOrekhovPS: Yea, it's a cleanup of power values at the end of epoch. But before the clean-up we create in-memory `computeResults` map. So if there were any power in the storage then computeMap shouldn't be empty regardless of the cleanup.

Or am I missing something here?

decentralized-api/server.go
DimaOrekhovPS: You think it'd be better to just proxy the whole thing? I'm not sure if the end-user needs to be exposed with the error coming from an inference node managed by an executor.

inference-chain/x/inference/module/chainvalidation.go
patimen: OK, that makes sense.

decentralized-api/server.go
patimen: But if there is something they got wrong in the inference request, or the model they are calling doesn't support a feature that other models do, they are left totally in the dark as to what happened.

PR #25, opened 2024-09-25
Add an endpoint to add unfunded accounts

patimen:
- Instead of requiring a new account to get money sent to them, enable an existing member to add them so they can start serving requests even without money.
- Lots of little further improvements to logging. Guessing at all the log levels, we can adjust.
- Switched testermint initialize to use the new method instead of funding.
- Dockerjava breaks up all log messages into 1024 byte chunks. Added the timestamp and try to use that as a way to append them back together for better logs. Not perfect, but better.
- NOTE: For testermint, we should probably split not on unfunded/funded but instead registered/unregistered.


decentralized-api/server.go
```diff
@@ -53,6 +53,7 @@ func StartInferenceServerWrapper(nodeBroker *broker.Broker, transactionRecorder
 	http.HandleFunc("/v1/chat/completions", wrapChat(nodeBroker, transactionRecorder, config))
 	http.HandleFunc("/v1/validation", wrapValidation(nodeBroker, transactionRecorder))
 	http.HandleFunc("/v1/participants", wrapSubmitNewParticipant(transactionRecorder))
+	http.HandleFunc("/v1/participants/unfunded", wrapSubmitUnfundedNewParticipant(transactionRecorder))
 	http.HandleFunc("/v1/participant/", wrapGetInferenceParticipant(transactionRecorder))
```
DimaOrekhovPS, line 57: Hm I think it should be `v1/participants/{id}` to conform to REST conventions. And this way `v1/participants/unfunded` would be problematic to use. Maybre `v1/participants/{id}` and `v1/unfunded-participnats`

inference-chain/x/inference/keeper/msg_server_submit_new_unfunded_participant.go
```diff
@@ -0,0 +1,43 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+func (k msgServer) SubmitNewUnfundedParticipant(goCtx context.Context, msg *types.MsgSubmitNewUnfundedParticipant) (*types.MsgSubmitNewUnfundedParticipantResponse, error) {
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	k.LogInfo("Adding new account directly", "address", msg.Address)
+	// First, add the account
+	newAccount := k.AccountKeeper.NewAccountWithAddress(ctx, sdk.MustAccAddressFromBech32(msg.Address))
```
DimaOrekhovPS, line 17: Should we check before if an account already exists? Or is the operation idempotent?

DimaOrekhovPS (approved):

decentralized-api/server.go
patimen: Will address in the subsequent PR.

inference-chain/x/inference/keeper/msg_server_submit_new_unfunded_participant.go
patimen: I'll experiment. Definitely should be idempotent.

PR #26, opened 2024-09-30
Registration Next Steps

patimen:
1. Add/Manage nodes via API - this way we can add the inference nodes after the chain and api have started, better for setup, and we needed to be able to manage nodes outside of the config file anyhow.
2. Created and changed Docker and docker-compose files to allow single-step deployments.
3. Added much documentation on how all of this works.
4. Added endpoints and code needed to have someone join, get funded (some "experiment money") and make requests without running a node in the chain.

The documents are meant to really explain this in detail, take a look.


decentralized-api/broker/broker.go
```diff
@@ -13,9 +13,14 @@ type Broker struct {
 }
 
 type NodeState struct {
-	LockCount     int
-	Operational   bool
-	FailureReason string
+	LockCount     int    `json:"lock_count"`
```
DimaOrekhovPS, line 16: We probably don't need to treat locking too serious for now, because concurrency capabilities of the inference framework we use far exceeds any load we'll be getting in the foreseeable future

DimaOrekhovPS (approved):

PR #33, opened 2024-11-19
EpochGroup logic with excluding Invalid participants

patimen:
Move all the logic for EpochGroup into a separate package. This will help both how we plan and how we test all of this.
EpochGroup now handles::
1. Storing weights from PoC
2. GetRandomMember for finding Executors
3. All the logic for handling voting
4. Created a new data called EpochGroupData which will store the vital info about each EpochGroup
5. 
Move all logic for handling power into the EpochGroups.
UpcomingEpochGroup is where we dump the incoming PoC results.
Now we simply switch over the the UpcomingEpochGroup when it's time to change weights.
ConsecutiveInvalidInferences allows us to move a bad participant as soon as possible.
Moved all the logic for Epoch End rewards/transfers into some pure functions, making testing quite easy and robust.
Added logic for excluding Invalid Participants from Epoch rewards and for distrubuting their other rewards based on work done.

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_submit_poc.go
+++ b/inference-chain/x/inference/keeper/msg_server_submit_poc.go
@@ -88,7 +88,7 @@ func (k msgServer) SubmitPoC(goCtx context.Context, msg *types.MsgSubmitPoC) (*t
 		PocStageStartBlockHeight: startBlockHeight,
 		ReceivedAtBlockHeight:    currentBlockHeight,
 	})
-	group, err := k.Keeper.GetEpochGroup(ctx, startBlockHeight)
+	group, err := k.Keeper.GetEpochGroup(ctx, uint64(startBlockHeight))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


inference-chain/proto/inference/inference/epoch_group_data.proto
```diff
@@ -0,0 +1,14 @@
+syntax = "proto3";
+package inference.inference;
+
+option go_package = "github.com/productscience/inference/x/inference/types";
+
+message EpochGroupData {
+  uint64 pocStartBlockHeight = 1; 
+  uint64 epochGroupId = 2;
+  string epochPolicy = 3; 
+  uint64 effectiveBlockHeight = 4;
```
DimaOrekhovPS, line 10: What's the difference between the effective and last?
Could you leave a comment with an explanation

patimen: Effective means the height at which it becomes "current", last is the last block it was effective.

PR #37, opened 2024-11-27
Consistent random seed for Validation

patimen:
1. Create a random seed while submitting PoC, sign it, send signature.
2. Use random seed for all validation decisions.
3. Store seed signatures for validation later
4. Store removal of members so seed will make sense.

### DIFF
--- a/inference-chain/proto/inference/inference/epoch_group_data.proto
+++ b/inference-chain/proto/inference/inference/epoch_group_data.proto
@@ -4,12 +4,16 @@ package inference.inference;
 option go_package = "github.com/productscience/inference/x/inference/types";
 
 message EpochGroupData {
-  uint64 pocStartBlockHeight = 1; 
+  uint64 pocStartBlockHeight = 1;
   uint64 epochGroupId = 2;
-  string epochPolicy = 3; 
-  uint64 effectiveBlockHeight = 4; 
-  uint64 lastBlockHeight = 5; 
-  map<string, string> memberSeedSignatures = 6;
-  map<string, uint64> removalBlockHeights = 7;
+  string epochPolicy = 3;
+  uint64 effectiveBlockHeight = 4;
+  uint64 lastBlockHeight = 5;
+  repeated SeedSignature memberSeedSignatures = 6;
+}
+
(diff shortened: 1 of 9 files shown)


decentralized-api/poc/orchestrator.go
```diff
@@ -236,21 +238,33 @@ func getBlockHash(data map[string]interface{}) (string, error) {
 	return hash, nil
 }
 
+var Seed int64 = 0
+
 func createSubmitPoCCallback(transactionRecorder cosmosclient.InferenceCosmosClient) func(proofs *ProofOfComputeResults) {
```
DimaOrekhovPS, line 243: I already implemented PoC v2, signature seed submission needs to be adapted

Could you please make submitting the seed a totally separate message from PoC?

inference-chain/proto/inference/inference/epoch_group_data.proto
```diff
@@ -4,11 +4,16 @@ package inference.inference;
...
+  uint64 pocStartBlockHeight = 1;
   uint64 epochGroupId = 2;
-  string epochPolicy = 3; 
-  uint64 effectiveBlockHeight = 4; 
-  uint64 lastBlockHeight = 5; 
-  
+  string epochPolicy = 3;
+  uint64 effectiveBlockHeight = 4;
+  uint64 lastBlockHeight = 5;
+  repeated SeedSignature memberSeedSignatures = 6;
```
DimaOrekhovPS, line 12: You decided to put it here instead of group metadata?
I also think it's better to put it here, but jsut want to formalize when we're putting something as metadata vs part of epochgroupdata proto

decentralized-api/poc/orchestrator.go
patimen: I can do that later down the PR cycle?

inference-chain/proto/inference/inference/epoch_group_data.proto
patimen: Group metadata is just a string, so really just one thing can be stored there unless we want a whole json schema or something.

PR #38, opened 2024-12-03
Reward by request

patimen:
Move to claiming rewards only be request
On Epoch end, calculate the rewards and store their values
On ClaimRewards, for now, simply rewards the actual values with no verification.

### DIFF
--- a/decentralized-api/poc/orchestrator.go
+++ b/decentralized-api/poc/orchestrator.go
@@ -198,6 +198,20 @@ func ProcessNewBlockEvent(orchestrator *PoCOrchestrator, event *chainevents.JSON
 		orchestrator.StopProcessing(createSubmitPoCCallback(transactionRecorder))
 		return
 	}
+	// once the new stage has started, request our money!
+	if proofofcompute.IsSetNewValidatorsStage(blockHeight - 1) {
+		go func() {
+			slog.Info("IsSetNewValidatorsStage: sending ClaimRewards transaction", "blockHeight", PreviousPocHeight,
+				"seed", PreviousSeed)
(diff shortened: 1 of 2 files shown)


decentralized-api/poc/orchestrator.go
```diff
@@ -198,6 +198,20 @@ func ProcessNewBlockEvent(orchestrator *PoCOrchestrator, event *chainevents.JSON
 		orchestrator.StopProcessing(createSubmitPoCCallback(transactionRecorder))
 		return
 	}
+	// once the new stage has started, request our money!
+	if proofofcompute.IsSetNewValidatorsStage(blockHeight - 1) {
```
DimaOrekhovPS, line 202: How about creating a `proofofcompute.IsClaimRewardStage` function?

inference-chain/proto/inference/inference/genesis.proto
```diff
@@ -19,5 +20,6 @@ message GenesisState {
   repeated Inference      inferenceList      = 2 [(gogoproto.nullable) = false] ;
   repeated Participant    participantList    = 3 [(gogoproto.nullable) = false] ;
   repeated EpochGroupData epochGroupDataList = 4 [(gogoproto.nullable) = false] ;
+  repeated SettleAmount   settleAmountList   = 5 [(gogoproto.nullable) = false] ;
```
DimaOrekhovPS, line 23: Probably shouldn't be part of GenesisState, because there's nothing to settle at genesis

probably Inference should also be dropped from GenesisState.

At the same time it's a good idea to have EpochGroupData and Participant as part of genesis. We probably need to modify InitGenesis or some other method to create an epoch 0 EpochGroup

PR #41, opened 2024-12-06
Final validation of validations

patimen:
This should finish the block.
Now we actually run the comparison of expected vs actual validations when a claim comes in.
We had to add a few more things to the store to make it work:

ValidationWeights: These are the validation weights at the START of the Epoch, unadjusted. They are used for determining validation THROUGHOUT the epoch, even after a member has been removed. They are also used after the fact to validation the validations.

InferenceDetail - not enough just to store the ID. While we could lookup the executor ID in the ID DB, might as well store it here so we don't have a hard dependency. But we also need executor reputation to figure things out at the end too.


decentralized-api/validation.go
```diff
@@ -134,13 +126,12 @@ func logInferencesToValidate(toValidate []types.Inference) {
 	slog.Info("Validation: Inferences to validate", "inferences", ids)
 }
 
-func ShouldValidate(executor *types.Participant, currentAccountAddress string, currentAccountPower uint32, totalPower uint32, seed int64) (bool, float32) {
-	if executor.Index == currentAccountAddress {
+func ShouldValidate(executorAddress string, executorReputation float32, currentAccountAddress string, currentAccountPower uint32, totalPower uint32, seed int64) (bool, float32) {
```
DimaOrekhovPS: To be deleted

decentralized-api/validation.go
```diff
@@ -80,11 +80,12 @@ func SampleInferenceToValidate(ids []string, transactionRecorder cosmosclient.In
 
 	var toValidate []types.Inference
 	for _, inferenceWithExecutor := range r.InferenceWithExecutor {
-
-		inferenceSeed := hashStringToInt64(inferenceWithExecutor.Inference.InferenceId)
-
-		shouldValidate, _ := ShouldValidate(&inferenceWithExecutor.Executor, transactionRecorder.Address,
-			r.ValidatorPower, r.TotalPower-inferenceWithExecutor.CurrentPower, inferenceSeed+poc.CurrentSeed.Seed)
+		shouldValidate := keeper.ShouldValidate(
```
DimaOrekhovPS, line 85: keeper is a bit confusing, it feels like a call of a keeper method instead of a function from the keeper namespace

DimaOrekhovPS (approved):

PR #42, opened 2024-12-10
Make our client scalable

patimen:
1. Break down the steps inside the ignite cosmosclient into its components inside our code for better control. Specifically, we send without waiting for the transaction to be processed and we add some checks for the account sequence if it isn't incremented from where we know it was.
2. Introduced an interface for the message sending client. Not strictly needed at the end of the process, but useful for trying different clients in the future.
3. Fixed a bug in how and when we were sending seeds and claims.

### DIFF
--- a/decentralized-api/cosmosclient/cosmosclient.go
+++ b/decentralized-api/cosmosclient/cosmosclient.go
@@ -5,8 +5,11 @@ import (
 	"decentralized-api/apiconfig"
 	"errors"
 	"fmt"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/tx"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/tx/signing"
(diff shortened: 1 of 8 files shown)


decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -151,22 +167,87 @@ func (icc *InferenceCosmosClient) ClaimRewards(transaction *inference.MsgClaimRe
 	return icc.sendTransaction(transaction)
 }
 
+func (icc *InferenceCosmosClient) BankBalances(ctx context.Context, address string) ([]sdk.Coin, error) {
+	return icc.Client.BankBalances(ctx, address, nil)
+}
+
 var sendTransactionMutex sync.Mutex = sync.Mutex{}
```
DimaOrekhovPS, line 174: So we're still using the mutex, but now transaction execution itself is faster? (since we don't wait for it to be processed completely?)

patimen: Yes, that keeps us from getting race conditions with the account sequence. From my testing the lock is VERY short lived.

PR #47, opened 2025-01-13
Finish merge, add PoC mocks and fix tests

patimen:
Final fixups from the new PoC merge:
## Consensus fixes
Two breaks in consensus fixed:
1. Random Guid in state calculation. Moved to API and submitted with request.
2. Using a map in creating a list for the state. Maps have indeterminate order
3. Added docs on how to deal with consensus issues

## Cosmovisor fixes
1. Fixed the custom build step for generating upgrade builds (and their SHA)
2. Removed logging in config loading (breaks json parsing for cosmovisor)
4. Added a test so that won't happen again. (PITA to debug)

## Test fixes
1. Added PoC mocks so we can run locally.
2. Added some retries for some tests when randomness is involved.
3. Added better logging for unit test failures in Github

## Misc
1. Update to latest version of our Cosmos branch (several fixes in there, most notable when a participant has 0 power from their PoC.
2. Some simplification of test2.sh and launch_chain.sh

### DIFF
--- a/decentralized-api/Makefile
+++ b/decentralized-api/Makefile
@@ -82,7 +82,7 @@ docker-push:
 	@echo "--> pushing decentralized-api docker image"
 	@docker push gcr.io/decentralized-ai/api
 
-build-v2-test:
+build-for-upgrade:
 	$(eval PLATFORM=linux/amd64)
 	$(eval GOOS=linux)
 	$(eval GOARCH=amd64)
--- a/docs/upgrades.md
+++ b/docs/upgrades.md
@@ -1,0 +1,91 @@
+# Upgrade Strategy for Cosmos Blockchain App Using Cosmovisor
+
+## Overview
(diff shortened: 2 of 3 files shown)


decentralized-api/Makefile
```diff
@@ -34,6 +34,23 @@ define DOCKER_BUILD
...
+	@echo "    	BUILD_FLAGS: $(BUILD_FLAGS)"
+	@echo "    	GOOS: $(GOOS)"
+	@echo "    	GOARCH: $(GOARCH)"
+	@docker build \
+		--platform $(PLATFORM) \
+		--build-arg BUILD_FLAGS="$(BUILD_FLAGS)" \
+		--build-arg GOOS=$(GOOS) \
+		--build-arg GOARCH=$(GOARCH) \
+		-f $(DOCKER_FILE) \
+		--output type=local,dest=./output \
```
DimaOrekhovPS, line 49: Is there some convenient way to make it a parameter of the DOCKER_BUILD define?

PR #48, opened 2025-01-14
Get rid of complex logic in launch_chain.sh

DimaOrekhovPS:
Now any "complex" operations are happening inside backed in init-docker.sh scripts and Go code.

1. Download genesis is an `inferenced` command
2. Set seeds is an `inferenced` command
3. Reading nodes from `node_config.json` happens on `decentralized-api` start-up
4. Participant registration requests happens on `decentralized-api` start-up

### DIFF
--- a/decentralized-api/apiconfig/config_manager.go
+++ b/decentralized-api/apiconfig/config_manager.go
@@ -171,6 +171,10 @@ func readConfig(provider koanf.Provider) (Config, error) {
 		config.ChainNode.AccountName = keyName
 	}
 
+	if err := loadNodeConfig(&config); err != nil {
+		log.Fatalf("error loading node config: %v", err)
+	}
+
 	return config, nil
 }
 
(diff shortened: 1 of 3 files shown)


decentralized-api/apiconfig/config_manager.go
```diff
@@ -168,6 +171,12 @@ func readConfig(provider koanf.Provider) (Config, error) {
 		slog.Info("Setting config.ChainNode.AccountName to env var", "AccountName", keyName)
 		config.ChainNode.AccountName = keyName
 	}
+
+	err = loadNodeConfig(&config)
```
DimaOrekhovPS: @patimen Could you please confirm it will work properly?

DimaOrekhovPS: I have a concern that since loadNodeConfig is not idempotent we may try to populate with the same nodes twice?

1. Read original config with empty nodes
2. call loadNodeConfig
3. Save config on disk
4. Reload from disk and try to merge with the same loadNodeConfig again

Is my understanding correct? In that case we could prioritize nodes from the original config file, while only adding nodes in `loadNodeConfig` if there are no nodes with such id

patimen: So, is the intent here to have two separate files to load the node config from? What's the logic?
Load config is at present only called on startup, from there we work in memory and WRITE the values...
However, won't we be writing the values from NODE_CONFIG_PATH to the main path?

PR #55, opened 2025-01-27
Fix casing for our proto files

patimen:
The result of the misnaming is a very hard to parse state/genesis file, where our stuff is camelCase and everything else is snake_case. Plug perhaps some other issues.


DimaOrekhovPS (approved):
Changes seem to be quite straightforward, but I didn't get why they're necessary. Could you give me an example of what was wrong before, when things were named in camelCase?

PR #56, opened 2025-01-28
Setup parameters needed for Tokenomics

patimen:
Two kinds of parameters for this:
Add the votable parameters (rate of payout for the coins) to Params (votable) 

Add supply parts to GenesisOnlyParams - cannot be voted on, but ARE part of the genesis file (thus modifiable for tests and "on the record" so to speak).


inference-chain/proto/inference/inference/params.proto
```diff
@@ -14,6 +14,22 @@ message Params {
   EpochParams epoch_params = 1;
   ValidationParams validation_params = 2;
   PocParams poc_params = 3;
+  TokenomicsParams tokenomics_params = 4;
+}
+
+message GenesisOnlyParams {
```
DimaOrekhovPS, line 20: Is it intended for GenesisOnlyParams being present only in the Genesis and not in the Params?

Do I understand correctly that Params structure is mutable (by voting), while Genesis can't be changed? And this is why you separated the two sets of params

DimaOrekhovPS (approved):

inference-chain/proto/inference/inference/params.proto
patimen: Yes, that's it exactly.

PR #57, opened 2025-01-29
Allocate coin supply for tokenomics

patimen:
1. Expand denom to mcicoin (millions of icoin)
1. Add StandardRewardsAmount (600 mcicoin) to GenesisOnlyParams
2. Create reward accounts in InitGenesis (run at chain start)
3. Create a Product Science account during init-docker-genesis.sh (cannot be done in chain) to hold originator funds

Still not 100% sure the denom stuff will work the way it seems like it should, but we will see.


inference-chain/x/inference/keeper/query_inference_participant.go
```diff
@@ -15,6 +15,7 @@ func (k Keeper) InferenceParticipant(goCtx context.Context, req *types.QueryInfe
 		return nil, status.Error(codes.InvalidArgument, "invalid request")
 	}
 
+	k.LogInfo("InferenceParticipant", "address", req.Address)
```
DimaOrekhovPS: Logging probably not needed

DimaOrekhovPS (approved):

PR #59, opened 2025-01-31
Implement tokenomics

patimen:
1. Add TokenomicsData to store how much subsidy we have given out.
2. Store GenesisOnlyParams in the state (was not before)
3. Calculation of subsidy (including crossover where Epoch starts in one rate and ends in another) is all in `SettleParameters`
4. Removed logic that distributed rewards for invalidated participants to everyone else. No longer applicable, invalidated participants get 0 rewards.
5. Send the BaseCoin denom metadata to sdk, allowing us to use sdk.NormalizeCoins to convert denominations. MUST be done manually, Cosmos will NOT convert for us.
6. Handle floating point calculations with great care to keep bookkeeping and chain consistent and testable/predictable.


inference-chain/x/inference/keeper/tokenomics_data.go
```diff
@@ -0,0 +1,53 @@
...
+	current.TotalFees = current.TotalFees + tokenomicsData.TotalFees
+	current.TotalSubsidies = current.TotalSubsidies + tokenomicsData.TotalSubsidies
+	current.TotalRefunded = current.TotalRefunded + tokenomicsData.TotalRefunded
+	k.SetTokenomicsData(ctx, current)
+	newData, _ := k.GetTokenomicsData(ctx)
+	k.LogInfo("Tokenomics data added", "tokenomicsData", newData)
+}
+
+// RemoveTokenomicsData removes tokenomicsData from the store
+func (k Keeper) RemoveTokenomicsData(ctx context.Context) {
```
DimaOrekhovPS: Should be deleted

PR #63, opened 2025-02-07
Allows creation of working clusters inside testermint

patimen:
Uses the DockerClient to create the network of nodes for testing. Also, get rid of the http server, as Wiremock can serve the exact same purpose.

### DIFF
--- a/testermint/src/main/kotlin/DockerGroup.kt
+++ b/testermint/src/main/kotlin/DockerGroup.kt
@@ -1,18 +1,15 @@
 package com.productscience
 
 import com.github.dockerjava.api.DockerClient
-import com.github.dockerjava.api.command.CreateContainerResponse
-import com.github.dockerjava.api.model.Bind
-import com.github.dockerjava.api.model.ExposedPort
-import com.github.dockerjava.api.model.HostConfig
-import com.github.dockerjava.api.model.Ports
-import com.github.dockerjava.api.model.Volume
 import com.github.dockerjava.core.DockerClientBuilder
 import java.nio.file.Files
 import java.nio.file.Path
 import kotlin.io.path.ExperimentalPathApi
 import kotlin.io.path.copyToRecursively
 import kotlin.io.path.deleteRecursively
+
+const val GENESIS_COMPOSE_FILE = "docker-compose-local-genesis.yml"
+const val NODE_COMPOSE_FILE = "docker-compose-local.yml"
 
 data class DockerGroup(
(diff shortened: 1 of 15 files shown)


testermint/src/main/kotlin/DockerGroup.kt
```diff
@@ -0,0 +1,282 @@
...
+        }
+    }
+
+    private fun getCommonEnv(): List<String> {
+        return listOfNotNull(
+            "KEY_NAME=$keyName",
+            "NODE_HOST=$keyName-node",
+            "DAPI_API__POC_CALLBACK_URL=$pocCallbackUrl",
+            "DAPI_API__PUBLIC_URL=$publicUrl",
+            "DAP_CHAIN_NODE__IS_GENESIS=${isGenesis.toString().lowercase()}",
```
DimaOrekhovPS: DAP >> DAPI?

testermint/src/main/kotlin/DockerGroup.kt
```diff
@@ -0,0 +1,282 @@
...
+) {
+    fun init() {
+        tearDownExisting()
+        setupFiles()
+        val chain = createChainNodeContainer()
+        val api = createApiContainer()
+        val mock = createWiremockContainer()
+        if (!dockerClient.listNetworksCmd().exec().any { it.name == "chain-public" })
+            dockerClient.createNetworkCmd().withName("chain-public").exec()
+        dockerClient.startContainerCmd(chain.id).exec()
```
DimaOrekhovPS: Could we just use docker-compose here instead of spinning up each container individually? That would greatly reduce logic duplication you mentioned

DimaOrekhovPS: We expect the compose file to be the main way for us to deploy on test net and even probably for our first miners to deploy on their hardware. 

I think our test process has to reflect that

DimaOrekhovPS (changes requested):
I think we should keep using docker compose. Could we make DockerGroup.kt spin up clusters using docker compose?

testermint/src/main/kotlin/DockerGroup.kt
patimen: Yeah... in the end I bit the bullet and just call `docker compose` directly using ProcessBuilder. It feels hacky, and introduces environment dependecies that are hard to control, but it saves a ton of work and reduces duplication.

DimaOrekhovPS (approved):

PR #65, opened 2025-02-12
Top miner rewards

patimen:
All needed structs to store TopMiner as a separate entry (perf wise, much faster than tacking on to the Participant list)
Core logic implemented in top_miner_calculations, tested well
Moved to Go 1.22.8 to fix issues with running unit tests in GoLand.


inference-chain/proto/inference/inference/top_miner.proto
```diff
@@ -0,0 +1,22 @@
+syntax = "proto3";
+package inference.inference;
+
+option go_package = "github.com/productscience/inference/x/inference/types";
+import "gogoproto/gogo.proto"; 
+import "cosmos/base/v1beta1/coin.proto";
+
+message TopMiner {
```
DimaOrekhovPS, line 8: We'll have both TopMiner entity + Participant entity per one top miner, right?

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -29,6 +29,8 @@ func (k msgServer) ClaimRewards(goCtx context.Context, msg *types.MsgClaimReward
 		return nil, err
 	}
 
+	//k.updateTopMiners(ctx, msg.Creator, settleAmount)
```
DimaOrekhovPS: Commented?

inference-chain/x/inference/keeper/top_miner.go
```diff
@@ -0,0 +1,70 @@
...
+	storeAdapter := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
+	store := prefix.NewStore(storeAdapter, types.KeyPrefix(types.TopMinerKeyPrefix))
+	b := k.cdc.MustMarshal(&topMiner)
+	store.Set(types.TopMinerKey(
+		topMiner.Address,
+	), b)
+}
+
+// GetTopMiner returns a topMiner from its index
+func (k Keeper) GetTopMiner(
```
DimaOrekhovPS, line 23: Check out some of my keeper_utils functions

inference-chain/proto/inference/inference/top_miner.proto
patimen: Yes, more performant than searching through the entire Participant records

PR #66, opened 2025-02-14
1. Add epoch shift to epoch params. 2. Track node catching up

DimaOrekhovPS:
1. Adding shift to epoch params allows for easier debugging and development. Shift controls when the first POC start, so even if epoch length is high we can start poc almost right away (e.g. epoch_length = 2000, epoch_shift = 1990)

2.  PoC lifecycle is controlled by listening to NewBlock events, this is a problem when a node is joining an existing network and catching up since we don't distinguish replayed blocks from up-to-date blocks. 

Implemented a simple solution of checking node status every 5 seconds. If the chain node is still catching up then most of then event processing is skipped. @patimen I kept upgrade event processing, because wasn't sure, can we skip it as well? I think ideally if node is catching up we can turn off the whole API

### DIFF
--- a/launch_chain.sh
+++ b/launch_chain.sh
@@ -60,10 +60,16 @@ fi
 echo "project_name=$project_name"
 
 # Set up wiremock
-mkdir -p "./prod-local/wiremock/$KEY_NAME/mappings/"
-mkdir -p "./prod-local/wiremock/$KEY_NAME/__files/"
-cp ./testermint/src/main/resources/mappings/*.json "./prod-local/wiremock/$KEY_NAME/mappings/"
-cp -r ./public-html/* "./prod-local/wiremock/$KEY_NAME/__files/"
+if [ "$mode" == "local" ]; then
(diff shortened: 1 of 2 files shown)


inference-chain/x/inference/types/epoch_params.go
```diff
@@ -1,97 +1,95 @@
...
 }
 
 func (p *EpochParams) GetEndOfPoCStage() int64 {
-	return 10 * p.EpochMultiplier
+	return p.GetStartOfPoCStage() + (10 * p.EpochMultiplier)
 }
 
 func (p *EpochParams) GetPoCExchangeDeadline() int64 {
-	return (p.GetEndOfPoCStage() + 2) * p.EpochMultiplier
+	return p.GetEndOfPoCStage() + (2 * p.EpochMultiplier)
```
DimaOrekhovPS, line 66: @patimen FYI fixed order of operations here, it feels more clear this way what we control

PR #67, opened 2025-02-14
Use a worker pool for event listening instead of creating a goroutine each time

DimaOrekhovPS:


decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -135,28 +144,37 @@ func StartEventListener(
...
+// processEvent is the worker function that processes a JSONRPCResponse event.
+func processEvent(
+	event *chainevents.JSONRPCResponse,
+	nodeBroker *broker.Broker,
+	transactionRecorder cosmosclient.InferenceCosmosClient,
+	configManager *apiconfig.ConfigManager,
+	nodePocOrchestrator *poc.NodePoCOrchestrator,
+) {
+	switch event.Result.Data.Type {
+	case "tendermint/event/NewBlock":
```
0xBECEDA: I would use constants here instead of strings

PR #68, opened 2025-02-18
Enable snapshots

0xBECEDA:
1. Enable creating and fetching snapshots: implemented commands to edit config.toml to allow fetch snapshots, edited sh scripts to support changes, added test (temporary written on go)
2. Removed outdated scripts, refactored existing


Makefile
```diff
@@ -10,6 +10,7 @@ api-build-docker:
 node-build-docker:
 	@make -C inference-chain build-docker
 
+# TODO 'build and push'
```
DimaOrekhovPS, line 13: What does this TODO mean?

inference-chain/internal/rpc/client.go
```diff
@@ -0,0 +1,99 @@
...
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+)
+
+const (
+	defaultTrustedBlocksPeriod = 1000
+)
+
```
DimaOrekhovPS, line 14: Another good candidate to place in this file is downloadGenesis func defined in download_genesis.go at the moment

inference-chain/scripts/prod/init-docker.sh
```diff
@@ -49,21 +49,38 @@ $APP_NAME config set client chain-id $CHAIN_ID
 $APP_NAME config set client keyring-backend $KEYRING_BACKEND
 $APP_NAME config set app minimum-gas-prices "0$COIN_DENOM"
 
+# create snapshots every N blocks and keep last M snapshots
+SNAPSHOT_INTERVAL=${SNAPSHOT_INTERVAL:-10}
+SNAPSHOT_KEEP_RECENT=${SNAPSHOT_KEEP_RECENT:-5}
+
+$APP_NAME config set app state-sync.snapshot-interval $SNAPSHOT_INTERVAL
```
DimaOrekhovPS, line 56: What happens if different nodes define different snapshot intervals?

Or they don't need to be in sync across the chain? Since when node joins they can choose which node (rpc server address) to use to fetch a snapshot?

DimaOrekhovPS (approved):

inference-chain/cmd/inferenced/cmd/sync_snapshots.go
```diff
@@ -0,0 +1,151 @@
...
+func updateTrustedBlock(configPath string, blockHeight uint64, blockHash string) error {
+	data, err := os.ReadFile(configPath)
+	if err != nil {
+		return fmt.Errorf("error reading file %s: %w", configPath, err)
+	}
+
+	// Compile a regex that looks for a line starting with:
+	// trust_hash = "some_value" || some_value || ""
+	// Using (?m) to enable multiline matching of ^ and $
+	re := regexp.MustCompile(`(?m)^\s*trust_hash\s*=\s*"?[^"\n]*"?\s*$`)
```
DimaOrekhovPS, line 138: Could we reuse some library for working with tomls? I'd assume base cosmossdk app would already have a lib like that as its dependency, because it also uses toml for config management

inference-chain/scripts/prod/init-docker.sh
0xBECEDA: 1. As I got, node requests snapshots from given rpc-servers and chooses nearest snapshot to set up trusted_block. If few nodes have same snapshots, requesting node will download snapshot from less busy node. So generally it should be no problem if nodes are creating snapshots with different intervals. 

But! If snapshots have not compatible states, it can be a problem. There shouldn't be any crash of node, but if snapshots can't be applied, node will restore older snapshots (if it exists) and will reproduce blocks from there. 

2. User still has options. Sync with snapshots can be completely disabled if required parameters in **config.toml** are not set: node will reproduce all blocks. User also can disable **creating** snapshots also if snapshot-interval = 0 (by default it is so). 

User can set up by himself any rpc-servers (normally provider of nodes publish this info) and take snapshots from there (but for testnet I  think it will be good to have some persistent peers, which will always create snapshots and set up rpc_servers by default to their rpc addresses)

Makefile
0xBECEDA: It means "fix this target" because in both subrepos is no target "build-and-push-docker" in Makefiles

inference-chain/cmd/inferenced/cmd/sync_snapshots.go
0xBECEDA: Sorry, didn't find it, may be you know exact name of lib?

DimaOrekhovPS: I think they use github.com/spf13/viper 
It's in our go.mod

PR #78, opened 2025-03-01
PoC fixes: 1. Count only unique nonces towards participant weight 2. PoC validation accounts for validator weights

DimaOrekhovPS:


inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -113,31 +116,65 @@ func (am AppModule) ComputeNewWeights(ctx context.Context, upcomingGroupData *ty
...
+			uniqueNonces[nonce] = struct{}{}
+		}
 	}
-	return weight
+
+	return int64(len(uniqueNonces))
 }
 
-func getActiveAddressSet(activeParticipants *types.ActiveParticipants) *map[string]struct{} {
+func getActiveParticipantsWeights(activeParticipants *types.ActiveParticipants) *map[string]int64 {
```
0xBECEDA: In Go map is a reference type. So you do not need to create pointer of map, because it always will be a pointer by default. Let's use map without pointer everywhere.

DimaOrekhovPS: Good call, thank you! Fixed

PR #80, opened 2025-03-05
Cluster deployment

0xBECEDA:
- fixed bug of finding less busy node for inference
- pass unique node number as parameter to init/generate request to make ML node generate unique nonces
- some refactoring


decentralized-api/broker/broker.go
```diff
@@ -119,13 +137,20 @@ func (b *Broker) getNodes(command GetNodesCommand) {
 }
 
 func (b *Broker) registerNode(command RegisterNode) {
+	curNodesAmount := len(b.nodes)
```
DimaOrekhovPS: I see a scenario when it won't work properly

1. We register a node `A` when `len(b.nodes) == 5`. Node `A` gets assigned `NodeNum = 6`.
2. Then a node that is not node `A` gets deleted
3. Then we register a new node `B`, `len(b.nodes)` again equals 5.  Node `B` gets assigned `NodeNum = B`.

I see following alternatives:

1. Use a hash code of node id/node host with a collision resolution step (increment by one until there are no collisions)

2. OR just find maximum current NodeNum and assign max + 1

decentralized-api/config-local.yaml
```diff
@@ -1,16 +1,41 @@
...
+nodes:
+    - hardware: []
+      host: 34.171.235.205:8080/
+      id: node1
+      inference_port: 0
+      max_concurrent: 500
+      models:
+        - unsloth/llama-3-8b-Instruct
+      poc_port: 0
+previous_seed:
```
DimaOrekhovPS: Do you think it's more explicit to have all the configurable properties in the config file? Even if they are default ones

decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -24,9 +24,10 @@ import (
 	"time"
 )
 
+// TODO idea for refactoring: create EventListener struct, which will contain this global vars and all params passed to StartEventListener as fields fields
```
DimaOrekhovPS, line 27: Yes, 100%

docker-compose-local-genesis.yml
```diff
@@ -26,7 +27,7 @@ services:
       - NODE_HOST=${KEY_NAME}-node
       - DAPI_API__POC_CALLBACK_URL=${POC_CALLBACK_URL}
       - DAPI_API__PUBLIC_URL=${PUBLIC_URL}
-      - DAPI_CHAIN_NODE__IS_GENESIS=true
+      - IS_GENESIS=true
```
DimaOrekhovPS, line 30: Why?

DimaOrekhovPS (changes requested):
Looks good, but seems like the NodeNum assignment algorithm needs to be tweaked a little bit

docker-compose-local-genesis.yml
0xBECEDA: I do not know why, but DAPI_CHAIN_NODE__IS_GENESIS wasn't recognized properly

decentralized-api/config-local.yaml
0xBECEDA: Hm.. I didn't make changes manually.. May be it is migrated from other PRs or smth else.. Reverted changes.

DimaOrekhovPS (approved):

PR #82, opened 2025-03-10
Training assignment

DimaOrekhovPS:
## Training

Two main abstractions for training. The rest of the code is boilerplate (like messages and queries) for enabling them

1. `Assigner` -- its goal is to monitor newly submitted unassigned training task. A participant then can claim a task like that for assignment. Then they will have some time to find actual executors for that training task, executors that match the tasks's hardware requirements.

2. `Executor` -- its goal is to monitor if a task got assigned to the participant who's running the node and start its execution. In the future it will also be monitoring progress of training task and free resources when a task is finished.

## Node broker

Some changes to node broker:

1. Monitor status of ml nodes using /status endpoint
2. Introduce intended status property for node status
3. If intended status doesn't match actual status try to repair/reboot an ML node

## Refactorings

1. Introduced mlnodeclient.Client, the goal is to define all REST endpoints of the ML node service there.
2. Now StartPoc and InferenceUpAll are node broker commands, this way we make sure that status updates are thread safe


0xBECEDA (approved):

decentralized-api/api/training.go
```diff
@@ -94,3 +111,20 @@ func handleGetTrainingJob(cosmosClient cosmosclient.CosmosMessageClient, id stri
 
 	RespondWithJson(w, task)
 }
+
+// FIXME: Needs some kind of a proof that the requester is the assigner
```
patimen, line 115: So when do we have this planned to fix?

decentralized-api/broker/broker_test.go
```diff
@@ -17,7 +17,7 @@ func TestSingleNode(t *testing.T) {
 		Id:            "node1",
```
patimen, line 17: Shouldn't there be some new tests?

decentralized-api/training/assigner.go
```diff
@@ -0,0 +1,416 @@
...
+
+	httpClient := utils.NewHttpClient(120 * time.Second)
+	for _, p := range selectedParticipants {
+		participant, err := queryClient.Participant(a.ctx, &types.QueryGetParticipantRequest{Index: p.participant})
+		if err != nil {
+			slog.Error(logTag+"Error querying for participant", "participant", p.participant, "err", err)
+			return
+		}
+
+		err = confirmAvailability(httpClient, participant.Participant.InferenceUrl, task.Id, p.nodeIds)
```
patimen, line 174: So, we're locking the node at this point, right?
What happens if there is some issue with sending the Message later? Will the lock timeout?

decentralized-api/internal/validation/inference_validation.go
```diff
@@ -465,7 +465,7 @@ func customDistance(
 		}
 		distance += posDistance
 	}
-	totalLogprobs := len(originalLogprobs) * len(originalLogprobs[0].TopLogprobs)
+	totalLogprobs := max(100, len(originalLogprobs)) * len(originalLogprobs[0].TopLogprobs)
```
patimen, line 468: Why this change?

decentralized-api/training/assigner_test.go
```diff
@@ -0,0 +1,9 @@
+package training_test
+
+import (
+	"testing"
+)
+
+func Test(t *testing.T) {
```
patimen, line 7: I feel like this test might not be comprehensive. :)

patimen (approved):
I'm going to approve this for now... I get we need to merge it in.
However, the testing is essentially none. Few if any unit tests, and only one testermint test in all of this.
We need to track that debt somewhere and address it soon.

decentralized-api/internal/validation/inference_validation.go
DimaOrekhovPS: This change was made by Gleb, we were releasing testnet from this branch. He needed to tweak this formula

decentralized-api/api/training.go
DimaOrekhovPS: my current block

decentralized-api/training/assigner.go
DimaOrekhovPS: yes, timeout. Locking isn't implemented atm, it's just always return ok. I'm still not sure if i want to go the lock route or manage node availability fully on-chain

PR #84, opened 2025-03-12
Convert Params to Decimal (and int64), model Upgrade

patimen:
## Convert from double/float/uints
Doubles and floats are notoriously imprecise, and are heavily cautioned against in BlockChain usage (for anything on the chain). So we make a (very) simple Decimal type for chain usage. We could have used a library, but the code is so simple it seemed much simpler to just do it directly.

uints should not be used anywhere we have math. One mistake and we have an overrun to uint64.Max.
1. Convert all floats or doubles in Params to Decimal. It uses a coefficient and and exponent (like other similar libraries).
2. Convert uint64 to int64 when math is used on the value (so, everything but IDs.)
3. Had to convert both Params and EpochGroupData (which contans ValidationParams.
4. Lots of scattered changes to support this.

## Upgrade example
While we won't need this specific upgrade going forward, this commit demos our approach to upgrading the data in the chain:
1. Copy the unmodified .proto files for the data we are changing. Rename them with a V1 in front (or V2 if there is already a V1)
2. Change the original proto files to whatever new format we want
3. Write a custom getter for the OLD values (V1)
4. Register a change handler that gets the old values and then writes the new ones.

### DIFF
--- a/docs/upgrades.md
+++ b/docs/upgrades.md
@@ -45,6 +45,7 @@ Since our system includes two binaries (chain app and decentralized API), we nee
         - Constants defining the upgrade name (in `constants.go`).
         - A `CreateUpgradeHandler` function.
         - Invocation in `setupUpgradeHandlers` (found in `upgrades.go`).
+        - If the upgrade changes the data in the state, increment `ConsensusVersion` in `module.go`
 
 2. **Build the Binaries**:
     - Generate and publish binaries with the `build-for-upgrade` Makefile target.


decentralized-api/api/pricing.go
```diff
@@ -50,7 +50,7 @@ func getPricing(w http.ResponseWriter, cosmosClient cosmosclient.CosmosMessageCl
 
 	models := make([]model.ModelPriceDto, len(modelsResponse.Model))
 	for i, m := range modelsResponse.Model {
-		pricePerToken := m.UnitsOfComputePerToken * unitOfComputePrice
+		pricePerToken := m.UnitsOfComputePerToken * uint64(unitOfComputePrice)
```
DimaOrekhovPS, line 53: But shouldn't  `pricePerToken` be `int64` as well? Since we've made the decision to never use uints?

decentralized-api/api/pricing.go
```diff
@@ -59,7 +59,7 @@ func getPricing(w http.ResponseWriter, cosmosClient cosmosclient.CosmosMessageCl
 	}
 
 	var responseBody = &model.PricingDto{
-		Price:  unitOfComputePrice,
+		Price:  uint64(unitOfComputePrice),
```
DimaOrekhovPS, line 62: Same question here

inference-chain/app/upgrades.go
```diff
@@ -1,9 +1,11 @@
 package app
 
 import (
+	v2 "github.com/productscience/inference/app/upgrades/v2"
 	"github.com/productscience/inference/app/upgrades/v2test"
 )
 
 func (app *App) setupUpgradeHandlers() {
 	app.UpgradeKeeper.SetUpgradeHandler(v2test.UpgradeName, v2test.CreateUpgradeHandler(app.ModuleManager, app.Configurator()))
+	app.UpgradeKeeper.SetUpgradeHandler(v2.UpgradeName, v2.CreateUpgradeHandler(app.ModuleManager, app.Configurator(), app.InferenceKeeper))
```
DimaOrekhovPS, line 10: So this upgrade mechanism is a cosmos builtin, not our invention, right?

inference-chain/app/upgrades/v2/upgrades.go
```diff
@@ -0,0 +1,96 @@
+package v2
```
DimaOrekhovPS, line 1: So any other migration we'll have in the future will require creating a `vX` subpackage? For example it will be `v3` for the next one

inference-chain/proto/inference/inference/params.proto
```diff
@@ -54,21 +54,27 @@ message EpochParams {
...
+  Decimal miss_requests_penalty = 12;
 }
 
 message PocParams {
   option (gogoproto.equal) = true;
-  uint32 default_difficulty = 1;
+  int32 default_difficulty = 1;
+}
+
+message Decimal {
```
DimaOrekhovPS, line 76: Should we put it in a separate file? I expect we'd like to export it and use not only for `params.proto`

inference-chain/x/inference/module/module.go
```diff
@@ -147,7 +147,7 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 // ConsensusVersion is a sequence number for state-breaking change of the module.
 // It should be incremented on each consensus-breaking change introduced by the module.
 // To avoid wrong/empty versions, the initial version should be set to 1.
-func (AppModule) ConsensusVersion() uint64 { return 1 }
+func (AppModule) ConsensusVersion() uint64 { return 2 }
```
DimaOrekhovPS, line 150: Should it be mentioned in `docs/upgrades.md`, that we have to increment this version?

decentralized-api/api/pricing.go
patimen: I decided not to push more uint conversions at the moment, to keep this PR a bit more focused.

inference-chain/app/upgrades.go
patimen: Yes, this is the standard upgrade mechanism

inference-chain/proto/inference/inference/params.proto
patimen: Maybe, will do as a follow up

inference-chain/x/inference/module/module.go
patimen: Excellent point, will do.

PR #86, opened 2025-03-20
Increase base fund

0xBECEDA:
- increase base fund for not active participant (without inference nodes) to 5000 tokens for 1000 requests
- by re-run container if volume is not cleaned, node will use same key

### DIFF
--- a/inference-chain/scripts/prod/init-docker.sh
+++ b/inference-chain/scripts/prod/init-docker.sh
@@ -47,62 +47,65 @@ fi
 
 if [ "$ACCOUNT_EXISTS" = true ]; then
     echo "Node is already configured, skip configuration"
-else
-    echo "Configure node"
-    echo "Current directory: $(pwd)"
 
-    # Init the chain:
-    # I'm using prod-sim as the chain name (production simulation)
-    #   and icoin (intelligence coin) as the default denomination
-    #   and my-node as a node moniker (it doesn't have to be unique)
-    $APP_NAME init \
-      --overwrite \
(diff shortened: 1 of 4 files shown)


inference-chain/scripts/prod/init-docker.sh
```diff
@@ -33,62 +32,79 @@ CHAIN_ID="prod-sim"
...
+   echo "❌ Account $KEY_NAME does not exist!"
+else
+   echo "✅ Account $KEY_NAME found. Using existing account."
+   ACCOUNT_EXISTS=true
+fi
 
-echo "Using genesis file: $GENESIS_FILE"
-cp "$GENESIS_FILE" $STATE_DIR/config/genesis.json
+if [ "$ACCOUNT_EXISTS" = true ]; then
+    echo "Node is already configured, skip configuration"
```
DimaOrekhovPS, line 49: can we just add `exit 0` so the long intialization below could be put outside of the else branch:

1. easier to read code, less nestedness
2. easier to read digg

DimaOrekhovPS (changes requested):
What about initialization route for `decentralized-api`. Don't we need to adapt it as well?

inference-chain/scripts/prod/init-docker.sh
0xBECEDA: We can't just use exit 0, because we have to run cosmovisor anyway. 
But I have made small update to make dig shorter

DimaOrekhovPS (approved):

PR #89, opened 2025-03-24
gm/genesis

DimaOrekhovPS:


decentralized-api/internal/poc/node_orchestrator.go
```diff
@@ -106,6 +107,20 @@ var DevTestParams = Params{
 	SeqLen:           4,
 }
 
+var TestNetParams = Params{
```
patimen, line 110: Do we want/need to make these votable, official params?

decentralized-api/internal/poc/node_orchestrator.go
```diff
@@ -222,7 +237,7 @@ func (o *NodePoCOrchestrator) sendInferenceUpRequest(node *broker.Node) (*http.R
 	inferenceUpDto := InferenceUpDto{
 		Model: model,
 		Dtype: "float16",
-		Args:  []string{"--enforce-eager"},
+		Args:  []string{"--quantization", "fp8"},
```
patimen: Again, params? Maybe we ought to start making a list of things to make into params.

decentralized-api/internal/server/validation.go
```diff
@@ -400,32 +402,100 @@ func compareLogits(
...
 		o := originalLogits[i]
 		v := validationLogits[i]
 		if o.Token != v.Token {
 			return &DifferentTokensValidationResult{baseComparisonResult}
 		}
-
-		originalLogprobs = append(originalLogprobs, o.Logprob)
-		validationLogprobs = append(validationLogprobs, v.Logprob)
 	}
+	similarity := customSimilarity(originalLogits, validationLogits)
```
patimen: Do we have documentation on this change? On the math behind how we're comparing our inferences?

patimen (approved):
I have some questions, but otherwise feel we should get this merged to main ASAP.
But, can you merge in main and run the tests first? Preferably locally, I'm still gettng a few errors on images that I can't quite fix.

decentralized-api/internal/poc/node_orchestrator.go
DimaOrekhovPS: These are per model params. I think Sofia recently made it configurable via node config

decentralized-api/internal/poc/node_orchestrator.go
DimaOrekhovPS: I think in the future, but not necessarily now

PR #90, opened 2025-03-24
update packages

0xBECEDA:


inference-chain/scripts/prod/init-docker.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
-set -e
```
patimen, line 2: Are we sure about this? It seems this could end with us having the chain in some bad state because some step here failed.

patimen (approved):
If this is passing (most) of our testermint tests I'm ok with it. Might be a tricky merge, though.

inference-chain/scripts/prod/init-docker.sh
DimaOrekhovPS: As far as I understand it's done so the account check doesn't kill the whole script:
```
ACCOUNT_CHECK=$($APP_NAME keys show...
```

We need to keep `set -e`, then quickly unset it before the check, then set it back:

```
#!/bin/sh
 set -e
 
 ...
 set +e
 ACCOUNT_CHECK=$($APP_NAME keys show "$KEY_NAME" --keyring-backend "$KEYRING_BACKEND" --keyring-dir "$STATE_DIR" 2>&1)
 set -e
 
 ...
```

@0xBECEDA

PR #91, opened 2025-03-26
Node self-request connection refusal

0xBECEDA:
- refactoring: created 'Server' object and assign methods to it (reduce passed parameters), extracted inference validation process to separate package (both server and event listener using it)
- fix bug of selfrouting


patimen (approved):
Looks good, I like the refactorings. In the future, if you've got a LOT of refactoring and one substantive change, try and call that out with a comment so reviewers can be sure to examine that one change more carefully, I spent too long looking for it. :)

PR #92, opened 2025-03-27
 Block 75: Modify node-config.json to support additional inference args

0xBECEDA:


decentralized-api/broker/broker.go
```diff
@@ -343,11 +351,17 @@ func convertInferenceNodeToHardwareNode(in *NodeWithState) *types.HardwareNode {
 			Count: hw.Count,
 		})
 	}
+
+	modelsNames := make([]string, 0)
```
patimen, line 378: The order of this list will now be indeterministic. In other words, each time we call `convertInferenceNodeToHardwareNode`, it could return a different result.

patimen (changes requested):
Is there really enough perf involved to make Models a map? This introduces a lot of indeterminism, and while it LOOKS like it would be constrained to the API side of things, we could be wrong and that could change, or someone could not recognize it.

patimen (approved):
Approving per our conversation.
Be sure to merge in main and run the tests.

decentralized-api/broker/broker.go
0xBECEDA: added sort of models names

PR #97, opened 2025-04-03
Toml overrides

patimen:
Two mechanisms:
1. ENV variables (injected likely through a .env file). CONFIG_ followed by the name of the value to override, with "." repaced with "__". So CONFIG_p2p__max_num_outbound_peers=15 would change p2p.max_num_outbound_peers to 15
2. Patching with files. If the file config_overrides.toml exists, it will be merged into the main config.

The second is better described as a patch than a merge. All formatting and comments will be preserved in the first file, and ONLY values will be merged from the second.
The code is... functional. I couldn't find a simple way to do this, especially with the need to preserve comments and needing to process the entire file (complete with the embedded table tags and multi-line strings).


inference-chain/cmd/inferenced/cmd/toml_handling.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+// Below here, All AI generated (FWIW, openAI o3-mini-high, with a once over by Claude 3.7 Sonnet.
+// But tested
+
+// MergeTomlFiles patches toml1 (the base) with values from toml2 (the override).
+// toml1's comments and formatting are preserved; keys in toml2 override those in toml1.
+// Any keys in toml2 not found in toml1 are appended at the end of their section.
+// Note: Multi-line values are detected for triple-quoted strings.
+func MergeTomlFiles(toml1, toml2 string) (string, error) {
```
DimaOrekhovPS, line 59: What do you think of instead reading from the override file value by value and for each value running something like

`inferenced config set config <key> <value>`

So the approach is similar to the env var, but all values are nicely structured in an override file

DimaOrekhovPS (approved):

PR #98, opened 2025-04-04
Separate access to API's 

0xBECEDA:
- public, admin and mlnode APIs as separate server
- mlnode supports old and new url paths to send generated/validated batches
- support changes in docker compose files and launching test local network script


docker-compose-local-genesis.yml
```diff
@@ -24,11 +24,17 @@ services:
         - ./${NODE_CONFIG}:/root/node_config.json
     ports:
       - "${PORT}:8080"
```
DimaOrekhovPS: Do we keep this one?

decentralized-api/internal/server/public/server.go
```diff
@@ -0,0 +1,68 @@
...
+	e := echo.New()
+	s := &Server{
+		e:             e,
+		nodeBroker:    nodeBroker,
+		configManager: configManager,
+		recorder:      recorder,
+	}
+
+	e.Use(middleware.LoggingMiddleware)
+	g := e.Group("/public/v1/")
```
DimaOrekhovPS: I think we should drop the word `public` from public API paths. Since it's the default API and the one intended for external user and we'd like to keep paths shorter, clear and organized.

decentralized-api/main.go
```diff
@@ -103,8 +106,24 @@ func main() {
...
 
-	s := server.NewServer(nodeBroker, config, validator, recorder)
-	s.Start()
+	addr := fmt.Sprintf(":%v", config.GetApiConfig().PublicServerPort)
+	logging.Info("start public server on addr", types.Server, "addr", addr)
+
+	publicServer := pserver.NewServer(nodeBroker, config, recorder)
+	publicServer.Start(addr)
+
+	addr = fmt.Sprintf(":%v", config.GetApiConfig().MLServerPort)
```
DimaOrekhovPS, line 125: Shouldn't we also create config parameters like:
```
publicApiHost
mlNodeHost
adminApiHost
```
?

And then we would set
`publicApiHost  = "0.0.0.0"` so it could listen to all incoming connections.

`adminApiHost = "localhost"` so it listens only to local connections

`mlNodeHost = "10.0.0.1"` so it listens only to connections over a private net

DimaOrekhovPS: Make admin and mlNode apis unreachable from the outside internet in one of the main reasons for this feature. Ideally we'd need to set up some auth keys in the future, but for now we should at least protect it with specifying web interfaces it should listen to.

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -0,0 +1,349 @@
...
+		ResponseHash:         transaction.ResponseHash,
+		ResponsePayload:      transaction.ResponsePayload,
+		PromptTokenCount:     transaction.PromptTokenCount,
+		CompletionTokenCount: transaction.CompletionTokenCount,
+		ExecutedBy:           accountName,
+	}
+
+	// Submit to the blockchain effectively AFTER we've served the request. Speed before certainty.
+	go func() {
+		// PRTODO: delete me and probably introduce retries if FinishInference returns not found
```
patimen, line 251: The correct TODO item here, (which I'm fine with us not doing immediately but should be noted) is that we update Finish and Start inference message handling to support them coming in out of order.
@tcharchian Can you make sure that gets on a roadmap? I know I've mentioned it before.
Summary: Ensure MsgFinishInference and MsgStartInference can come in out of order

decentralized-api/internal/server/public/get_participants_handler.go
```diff
@@ -193,16 +161,53 @@ func getParticipants(epochOrNil *uint64, w http.ResponseWriter, chainNodeUrl str
...
+	}
+
+	participants := make([]ParticipantDto, len(r.Participant))
+	for i, p := range r.Participant {
+		balances, err := s.recorder.BankBalances(ctx.Request().Context(), p.Address)
+		pBalance := int64(0)
+		if err == nil {
+			for _, balance := range balances {
+				// TODO: surely there is a place to get denom from
+				if balance.Denom == "nicoin" {
```
patimen, line 188: @tcharchian Another minor thing I'd like on a roadmap somewhere.
I'd put it as part of a Settings cleanup task, where we make sure we aren't unecessarily hard coding anything.

patimen (approved):
OK, I'm content this didn't have any subtle errors. It was an interesting LLM exercise for sure.
Couple of comments on todo items, and a couple thumbs up on Dima's comments.

DimaOrekhovPS (approved):

PR #100, opened 2025-04-08
A better fix for event queues

patimen:
Channels simply hang when they are at capacity until they are drained.
This produces a deadlock situation when the mainEvent queues are waiting for the blockEvents to get us out of the height race condition.
This is meant to be a complete solution:
Introduce UnboundedQueue. It is thread safe, takes input and produces output via channels, for minimal changes and a fairly clear implementation.
Thing coming `In` just get placed in the `items` slice.
Things in the `items` slice get fed into the `Out` channel.
There is no connection. The `items` slice will get as large as it needs to until the `Out` channel drains the events. Thus, no deadlock.

Replace other channels with these. One for mainEvent and one for blockEvent.


decentralized-api/internal/event_listener/unbounded_queue.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+
+		select {
+		case item := <-q.input:
+			// Store new item from producer
+			items = append(items, item)
+
+		case out <- first:
+			// First item was consumed, remove it
+			items = items[1:]
```
DimaOrekhovPS, line 68: What happens when a slice gets reassigned like this? Will it be resized properly? Will garbage collector delete the poped item?

patimen: No it won't. That's a great catch. Let me fix that.

patimen: Oddly, with testing it shows this is not a problem. The allocated memory goes away, as does the capacity of the slice.
I think the reason is the way `append` works by creating a new (appropriately sized) slice each time we add an item.

decentralized-api/internal/event_listener/unbounded_queue.go
```diff
@@ -0,0 +1,92 @@
...
+		var out chan T
+		var first T
+
+		if len(items) > 0 {
+			out = q.output
+			first = items[0]
+		}
+
+		select {
+		case item := <-q.input:
```
DimaOrekhovPS, line 62: Could you please confirm my understanding of this select statement.

1. If there are items in `q.input` then we always hit the first case

2. Seconds case happens when we can put first to out. My concern is, what if `len(items) <= 0` and `q.input` is empty, why won't we get an NPE when trying to eval case # 2?

3. only happens when after a shutdown signal

patimen: OK, the key is to remember that `select` doesn't really function like `switch`.
1. The order is random if multiple channels are ready.
2. If len(items) isn't greater than 0, `out` is nil and won't be selected. (see the `if` statement)
So, to talk through your scenario:
If there are no items in the queue, we will block at `select` until either something comes in to `q.input` or we get `done`.
Once there is an item, we add it to items and go back to our `for` loop.
At that point, if there are more items we will either handle adding more items or send some to `out`
NOTE: All of these cases are handled in the tests as well.
Does that all make sense? It's fairly idiomatic Go, from what I've seen.

DimaOrekhovPS (approved):

decentralized-api/internal/event_listener/unbounded_queue.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// NewUnboundedQueue creates a new unbounded queue that exposes channels
+func NewUnboundedQueue[T any]() *UnboundedQueue[T] {
+	input := make(chan T, 100)  // Buffer size is just for performance
+	output := make(chan T, 100) // Buffer size is just for performance
+	done := make(chan struct{})
+
+	q := &UnboundedQueue[T]{
+		In:     input,  // Public producer channel (send-only)
```
0xBECEDA, line 29: Great implementation!

Just a small thing: instead of using same channels for public and private access, I would have these channels only as private fields of struct and would add methods to push and read elements in/from queue. 

This is, of course, a matter of taste and won’t affect performance in any way, but I think it will make the code easier to read.

decentralized-api/internal/event_listener/unbounded_queue.go
```diff
@@ -0,0 +1,92 @@
...
+			out = q.output
+			first = items[0]
+		}
+
+		select {
+		case item := <-q.input:
+			// Store new item from producer
+			items = append(items, item)
+
+		case out <- first:
```
0xBECEDA, line 66: I I see one potential issue here (though I might be wrong). If no one is reading from the `Out` channel, the `items` slice won’t get cleared and will keep growing uncontrollably.

If output is full, I would suggest at least add log message, because otherwise we might have problem with memory usage and will spend a lot of time trying to find it.

decentralized-api/internal/event_listener/unbounded_queue.go
```diff
@@ -0,0 +1,92 @@
...
+// immediately after the count is returned
+func (q *UnboundedQueue[T]) Size() int {
+	// This is just an approximation based on channel buffer lengths
+	return len(q.input) + len(q.output)
+}
+
+// Close shuts down the queue and waits for the manager to exit
+func (q *UnboundedQueue[T]) Close() {
+	defer func() {
+		_ = recover() // Ignore panic from closing already closed channels
```
0xBECEDA: Why can't we avoid closing already closed channels? It can be a little bit dangerous hack, because we do not really control which procedure closes channels

decentralized-api/internal/event_listener/unbounded_queue.go
patimen: I found the channel method a bit more "Go" like, and it allowed for minimal changes elsewhere in the code.

decentralized-api/internal/event_listener/unbounded_queue.go
patimen: I think if the queue is growing like that we're already in an essentially unrecoverable state. Maybe we can revisit at some point.

decentralized-api/internal/event_listener/unbounded_queue.go
patimen: Close() should be idempotent, but maybe this is a bit of a hacky way to do it. I'll find a more idiomatic method.

PR #101, opened 2025-04-10
Api security merge

DimaOrekhovPS:


decentralized-api/internal/server/public/training_jobs_handlers.go
```diff
@@ -56,3 +66,16 @@ func (s *Server) getTrainingJob(ctx echo.Context) error {
...
+	var body LockTrainingNodesDto
+	if err := ctx.Bind(&body); err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, err)
+	}
+
+	if err := s.trainingExecutor.PreassignTask(body.TrainingTaskId, body.NodeIds); err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, err)
+	}
+
+	return nil
```
DimaOrekhovPS, line 80: If we return nil, will it just return empty body with code 200?

PR #103, opened 2025-04-16
Ethereum Bridge Receipts COSMWASM Update

GLiberman:


decentralized-api/internal/server/public/entities.go
```diff
@@ -139,3 +142,26 @@ type ModelPriceDto struct {
...
+type BridgeReceipt struct {
+	ContractAddress string `json:"contract"`     // Address of the smart contract on the origin chain
+	OwnerAddress    string `json:"owner"`        // Address of the token owner on the origin chain
+	OwnerPubKey     string `json:"publicKey"`    // Public key of the token owner on the origin chain
+	Amount          string `json:"amount"`       // Amount of tokens to be bridged
+	ReceiptIndex    string `json:"receiptIndex"` // Index of the transaction receipt in the block
+}
+
+// BlockQueue manages a queue of blocks to be processed
+type BridgeQueue struct {
```
DimaOrekhovPS, line 162: should be moved to `bridge_handlers.go` since this file is for serializable data transfer objects only

decentralized-api/internal/server/public/bridge_handlers.go
```diff
@@ -0,0 +1,289 @@
...
+
+	slog.Info("Added block to pending queue",
+		"blockNumber", block.BlockNumber,
+		"originChain", block.OriginChain,
+		"receiptsCount", len(block.Receipts),
+		"queueLength", len(q.pendingBlocks))
+
+	// Trigger processing if we have enough blocks
+	if len(q.pendingBlocks) >= q.minBlocksBeforeProcessing {
+		go q.processBlocks()
```
DimaOrekhovPS, line 51: A better solution here would be to not spawn a goroutine each time, but to post an event to a `chan`

Then we already have our main processing loop:
`func (q *BridgeQueue) init()`

And we can make it to wake up either every 60 seconds or when an event comes in like:

```
       for {
		select {
		case <-ticker.C:
			// Process blocks if we have enough
			if len(q.pendingBlocks) >= q.minBlocksBeforeProcessing {
				q.processBlocks()
			}
		}
                case blk := <-q.blockCh: q.processBlocks()
	}
```

decentralized-api/internal/server/public/server.go
```diff
@@ -60,6 +61,13 @@ func NewServer(
 
 	g.GET("debug/pubkey-to-addr/:pubkey", s.debugPubKeyToAddr)
 	g.GET("debug/verify/:height", s.debugVerify)
+
+	g.POST("bridge/block", s.postBlock)
+	g.GET("bridge/status", s.getBridgeStatus)
+
+	// Initialize block queue
+	blockQueue = NewBlockQueue(recorder)
```
DimaOrekhovPS, line 69: Please move this intialization to `main.go`

local-test-net/docker-compose-local-genesis.yml
```diff
@@ -5,6 +5,7 @@ services:
     image: ghcr.io/product-science/inferenced
     volumes:
       - ./prod-local/${KEY_NAME}:/root/.inference
+      - ./my_contract:/contract
```
DimaOrekhovPS, line 8: @tcharchian please create a ticket to update our deploy scripts and k8s net to mimic Gabriel's changes here

inference-chain/app/app.go
```diff
@@ -280,171 +242,170 @@ func New(
...
+	kv, ok := app.UnsafeFindStoreKey(storeKey).(*storetypes.KVStoreKey)
 	if !ok {
 		return nil
 	}
-	return kvStoreKey
+	return kv
 }
 
-// GetMemKey returns the MemoryStoreKey for the provided store key.
+// GetMemKey returns a MemoryStoreKey from the runtime store
```
DimaOrekhovPS, line 349: There are a lot of changes like this in this file, how did they happen, what's their goal? Hard to understand if something actually important changed

inference-chain/app/legacy.go
```diff
@@ -1,12 +1,21 @@
 package app
 
 import (
```
DimaOrekhovPS, line 3: Why was the file renamed to legacy.go?

inference-chain/app/sim_test.go
```diff
@@ -133,7 +136,9 @@ func TestAppImportExport(t *testing.T) {
 	appOptions[flags.FlagHome] = app.DefaultNodeHome
 	appOptions[server.FlagInvCheckPeriod] = simcli.FlagPeriodValue
 
-	bApp, err := app.New(logger, db, nil, true, appOptions, fauxMerkleModeOpt, baseapp.SetChainID(SimAppChainID))
+	var emptyWasmOpts []wasmkeeper.Option
+
+	bApp, err := app.New(logger, db, nil, true, appOptions, emptyWasmOpts, fauxMerkleModeOpt, baseapp.SetChainID(SimAppChainID), fauxMerkleModeOpt)
```
DimaOrekhovPS, line 141: Is it ok that we pass `fauxMerkleModeOpt` two times?

inference-chain/proto/inference/inference/bridge_transaction.proto
```diff
@@ -0,0 +1,27 @@
+syntax = "proto3";
+package inference.inference;
+
+option go_package = "github.com/productscience/inference/x/inference/types";
+
+enum BridgeTransactionStatus {
+  BRIDGE_PENDING = 0;
+  BRIDGE_COMPLETED = 1;
```
DimaOrekhovPS, line 8: Could it be `FAILED`?

inference-chain/proto/inference/inference/contracts.proto
```diff
@@ -0,0 +1,27 @@
...
+  uint64 cw20_code_id = 2;
+}
+
+// Stores mapping between external chain tokens and their CW20 counterparts
+message BridgeContract {
+  string external_chain = 1;           // e.g. "ethereum" 
+  string external_contract = 2;        // original chain contract address
+  string cw20_contract = 3;           // our CW20 contract address
+  string name = 4;                    // token name
+  string symbol = 5;                  // token symbol
```
DimaOrekhovPS, line 20: Symbol is denomination? (nicoin, uicoin, ...)

inference-chain/proto/inference/inference/token_contract.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+package inference.inference;
+
+option go_package = "github.com/productscience/inference/x/inference/types";
+
+// Stores mapping between external chain tokens and their CW20 counterparts
+message TokenContract {
```
DimaOrekhovPS, line 7: What's the difference from `BridgeContract`? they look identical if I'm not missing smth

inference-chain/scripts/init-docker.sh
```diff
@@ -84,6 +84,11 @@ fi
 echo "Configure node"
 echo "Current directory: $(pwd)"
```
DimaOrekhovPS, line 86: We also have `init-docker-genesis.sh`
Do we need to change anything there?

inference-chain/x/inference/keeper/bridge_transaction.go
```diff
@@ -0,0 +1,49 @@
...
+func (k Keeper) SetBridgeTransaction(ctx context.Context, tx *types.BridgeTransaction) {
+	slog.Info("Setting bridge transaction",
+		"index", tx.Index,
+		"status", tx.Status,
+		"validationCount", tx.ValidationCount)
+
+	storeAdapter := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
+	store := prefix.NewStore(storeAdapter, []byte(BridgeTransactionKeyPrefix))
+	bz := k.cdc.MustMarshal(tx)
+	store.Set([]byte(tx.Index), bz)
```
DimaOrekhovPS, line 34: is `tx.Index` same as `txHash` in other funcs?

inference-chain/x/inference/keeper/msg_server_bridge_exchange.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	if !pubKey.VerifySignature([]byte(signatureData), signatureBytes) {
+		return fmt.Errorf("signature verification failed")
+	}
+
+	return nil
+}
+
+func (k msgServer) BridgeExchange(goCtx context.Context, msg *types.MsgBridgeExchange) (*types.MsgBridgeExchangeResponse, error) {
```
DimaOrekhovPS, line 51: Just a general comment about message handlers, because we've had a few issues with them before:

Please make sure it's deterministic: no randomness, no external dependencies, it always should produce identical results for the same input. A common pitfall I fell into a few times is iterating over maps which have indetermenistic element order in Go, any maps should be converted to array and sorted if data from them is stored on chain

DimaOrekhovPS (commented):
Please provide some MD file describing our approach here, it would be great to have it for both us and AI agents

inference-chain/proto/inference/inference/bridge_transaction.proto
GLiberman: In theory currently its quite impossible for us to get a transaction which doesn't exist  - except somebody intentionally spams "bad" blocks in that scenario we need to punish such behavior, but im not sure that we currently have direct way to punish users. Or I'm wrong?

inference-chain/app/legacy.go
GLiberman: Well all modules which doesn't update their code to accept newer version of cosmos with support depinject are considered as legacy modules which need manual integration. All other chain examples of cosmos which uses wasm store such integration with such naming

decentralized-api/internal/server/public/bridge_handlers.go
GLiberman: Thanks, I actually forget to fix that part

inference-chain/app/app.go
GLiberman: You're right - After I fix all other comments I will def-ly redo this branch and will clean up changes made within app.go

inference-chain/proto/inference/inference/contracts.proto
GLiberman: Sort of. Symbol - typically an abbreviation of the token's name

PR #116, opened 2025-04-29
Create new "sanity" github action (just a few tests)

patimen:
Also add AI guidelines for better agent behavior and fix one weird edge case.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -10,9 +10,22 @@ on:
       - 'decentralized-api/**'
       - 'local-test-net/**'
       - 'testermint/**'
+  issue_comment:
+    types: [created]
 
 jobs:
+  check-comment:
+    runs-on: ubuntu-latest
+    if: github.event.issue.pull_request && contains(github.event.comment.body, '/run-integration')
+    steps:
+      - name: Check comment
+        run: echo "Comment contains /run-integration, triggering integration tests"
+    outputs:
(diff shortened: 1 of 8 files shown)


.github/workflows/integration.yml
```diff
@@ -5,56 +5,21 @@ on:
   push:
     branches:
       - main
+    paths:
```
DimaOrekhovPS, line 8: Can we exclude .md files and such? I expect a lot of changes in .md files in the upcoming months

.junie/guidelines.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+/client-libs        # Client script to interact with the chain
+/cosmovisor         # Cosmovisor binaries
+/decentralized-api  # Api node
+/dev_notes          # Chain developer knowledge base
+/docs               # Documentation on specific aspects of the chain
+/inference-chain    # Chain node
+/prepare-local      # Scripts and configs for running local chain
+/testermint         # Integration tests suite
```
DimaOrekhovPS, line 24: /local-test-net # Contains scripts and other files for running a local test net consisting of 3 nodes

Something like that

.github/workflows/integration.yml
patimen: There is no reliable way to exclude .md files in already included paths. There were some pretty hacky workarounds, I might look into them later.

DimaOrekhovPS (approved):

PR #125, opened 2025-05-03
Compute new weights uses EpochGroup, not active participants

patimen:
1. Break into functional/imperative parts (calculate functions should be pure functions)
2. extract what we were getting from ActiveParticipants into a method that just returned participant/weight pairs
3. Replace that method with current EpochGroup instead
Nearly impossible to screw up that way.


inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -4,29 +4,78 @@ import (
...
+		Seeds:                   seeds,
+		EpochStartBlockHeight:   epochStartBlockHeight,
+		Logger:                  logger,
+	}
+}
+
+// getCurrentValidatorWeights gets the active participants for the previous epoch and returns a map of weights
+func (am AppModule) getCurrentValidatorWeights(ctx context.Context, epochGroupId uint64) map[string]int64 {
+	if epochGroupId <= 1 {
+		return nil
```
DimaOrekhovPS: Can we do a call to staking module here and get list of genesis validators?

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -4,29 +4,78 @@ import (
...
+
+	currentGroup, err := am.keeper.GetCurrentEpochGroup(ctx)
+	if err != nil {
+		am.LogError("getCurrentValidatorWeights: Error getting current epoch group", types.PoC, "error", err)
+		return nil
+	}
+	currentMembers, err := currentGroup.GetGroupMembers(ctx)
+	if err != nil {
+		am.LogError("getCurrentValidatorWeights: Error getting current group members", types.PoC, "error", err)
+		return nil
```
DimaOrekhovPS: I think we should propagate the error on top, make the func return `(map[string]int64, error)`

We need to fail the whole thing, if we can't get current validators (when we expect to have them)

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -4,29 +4,78 @@ import (
...
+	}
+
+	weights := make(map[string]int64)
+	for _, member := range currentMembers {
+		weight, err := strconv.ParseInt(member.Member.Weight, 10, 64)
+		if err != nil {
```
DimaOrekhovPS, line 73: Also an illegal state error, not sure what our strategy should be. Not sure if we should panic and halt the app, but definitely something is wrong here if we allow to write non-int weights

DimaOrekhovPS (approved):

PR #130, opened 2025-05-07
Fix empty streamed response

DimaOrekhovPS:


decentralized-api/internal/validation/inference_validation.go
```diff
@@ -276,15 +276,49 @@ type UnmarshalledResponse struct {
 
 func (r *UnmarshalledResponse) GetEnforcedStr() (string, error) {
 	if r.JsonResponse != nil {
-		return r.JsonResponse.Choices[0].Message.Content, nil
+		if len(r.JsonResponse.Choices) == 0 {
```
patimen, line 279: What is the impact of having these errors? Is it going to make this code more brittle?

patimen (approved):

decentralized-api/internal/validation/inference_validation.go
DimaOrekhovPS: It would just stop validation

PR #136, opened 2025-05-14
Make FinishInference and StartInference work out of order

patimen:
There are edge cases where a StartInference could end up after a FinishInference. This fixes it.
This was done by Junie entirely, though I of course reviewed it. I feel there are a lot of smaller issues like this we can move faster on, having Junie (or your AI Agent of choice) run, check tests, retry, etc.


inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -49,11 +73,15 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
...
 
-	refundAmount := existingInference.EscrowAmount - existingInference.ActualCost
-	if refundAmount > 0 {
-		err = k.IssueRefund(ctx, uint64(refundAmount), requester.Address, "inference_refund:"+existingInference.InferenceId)
-		if err != nil {
-			k.LogError("Unable to Issue Refund for finished inference", types.Payments, err)
+	// Only issue a refund if we have requester information and an escrow amount
+	// (if FinishInference came before StartInference, we won't have this information yet)
+	if existingInference.RequestedBy != "" && existingInference.EscrowAmount > 0 {
+		refundAmount := existingInference.EscrowAmount - existingInference.ActualCost
```
DimaOrekhovPS, line 79: And how are fees handled in this case, when finish comes before start?

patimen: Fees are applied when the StartInference comes in, using the amount from Finish. Instead of putting the entire max amount in escrow and then refunding in Finish.

PR #141, opened 2025-05-16
k8s

DimaOrekhovPS:


tmkms/docker/Dockerfile_import_keys
```diff
@@ -12,19 +12,19 @@ RUN cargo install tmkms --features=softsign --version=0.12.0
 
 WORKDIR /root
 
-RUN tmkms init .tmkms
+RUN tmkms init /app/tmkms_init_data
```
DimaOrekhovPS, line 15: When building an image we initialize tmkms with state, config and keys using .tmkms directory. Then when running a container we mount a volume into that directory since we want to persist the key. Docker just copies the pre-existing content of the directory into the volume, but kubernetes overrides it.

Here's a more robust solution that makes the container work independently of the details of how a volume is mounted

PR #142, opened 2025-05-19
Support explorer

0xBECEDA:
Here is first steps to run network node with ping-pub explorer.


decentralized-api/internal/server/public/server.go
```diff
@@ -33,10 +38,18 @@ func NewServer(
...
+		logging.Error("Failed to parse explorer url", types.Server, "error", err, "url", explorerUrl)
+	} else {
+		s.explorerTargetUrl = explorerUrlParsed
+	}
+
 	e.Use(middleware.LoggingMiddleware)
 	g := e.Group("/v1/")
 
 	g.GET("status", s.getStatus)
+	g.Any("explorer/*", s.getExplorerUI)
```
DimaOrekhovPS, line 52: Are we proxying only because on the 6blocks testnet it's hard to open additional ports?

inference-chain/scripts/init-docker-genesis.sh
```diff
@@ -15,6 +15,25 @@ APP_NAME="inferenced"
 CHAIN_ID="gonka-testnet-3"
 COIN_DENOM="nicoin"
 STATE_DIR="/root/.inference"
+
+update_configs_for_explorer() {
+  if [ "$WITH_EXPLORER" = true ]; then
+    echo "Updating configs for enable explorer..."
+    sed -i 's/^enable *= *false/enable = true/' "$STATE_DIR/config/app.toml"
```
DimaOrekhovPS, line 22: Please let's use `inferenced config set app` ... here instead of sed

Sed is a bit unreliable, especially when we search by `enable`, it's to general, we can accidentally enable something we didn't intend

local-test-net/docker-compose-local-genesis-with-explorer.yml
```diff
@@ -0,0 +1,69 @@
+services:
+  explorer:
```
DimaOrekhovPS, line 2: Instead of creating a separate file, let's use features provided by docker compose for merging multiple compose files.

We can use our existing local docker compose as a base and then in this additional compose define only the new `explorer` service and additional env vars for `api`

You can provide multiple files to docker compose command 

`docker compose -f docker-compose.yml -f docker-compose.override.yml up`

I used it in a different project, should work

tmkms/Makefile
```diff
@@ -35,10 +35,10 @@ docker-push:
 
 	@if [ "$(SET_LATEST)" = "1" ]; then \
 		@echo "Setting latest tag..."; \
-		@docker tag gcr.io/decentralized-ai/inferenced:$(VERSION) gcr.io/decentralized-ai/inferenced:latest; \
+		@docker tag gcr.io/decentralized-ai/tmkms-softsign-with-keygen:$(VERSION) gcr.io/decentralized-ai/tmkms-softsign-with-keygen:latest; \
 		@echo "Pushing latest tag to Google Cloud Artifact Registry"; \
-		@docker push gcr.io/decentralized-ai/inferenced:latest; \
+		@docker push gcr.io/decentralized-ai/tmkms-softsign-with-keygen:latest; \
```
DimaOrekhovPS, line 40: In one my recent PRs I deleted all references to gcr.io, let's use ghcr only

PR #143, opened 2025-05-19
Fix streamed responses

DimaOrekhovPS:


patimen (approved):
Seems like it should work. Maybe we can add a testermint test later, but at least we've got a unit test that seems close to what we'll see.
**however**, I would like to see adding the include_usage flag into the payload. We cannot rely on customers to add it.

PR #144, opened 2025-05-20
Fix multi models

patimen:
Properly calculate both the on-chain and the on-api amounts.
Add tests that actually fail when things aren't done right.


decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -254,7 +256,7 @@ func (el *EventListener) handleMessage(event *chainevents.JSONRPCResponse, name
 	//	}
 	//}
 	switch action {
-	case finishInferenceAction:
+	case startInferenceAction, finishInferenceAction:
```
DimaOrekhovPS, line 259: right, makes sense now :)

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -75,6 +76,9 @@ func (s *Server) handleTransferRequest(ctx echo.Context, request *ChatRequest) e
 
 	go func() {
 		logging.Debug("Starting inference", types.Inferences, "id", inferenceRequest.InferenceId)
+		if s.configManager.GetApiConfig().TestMode && request.OpenAiRequest.Seed == 8675309 {
```
DimaOrekhovPS, line 79: Hm, now I'm thinking that maybe a special header for the request would do the trick

inference-chain/scripts/init-docker-genesis.sh
```diff
@@ -49,6 +49,10 @@ sed -Ei 's/^laddr = ".*:26657"$/laddr = "tcp:\/\/0\.0\.0\.0:26657"/g' \
 sed -Ei "s/^seeds = .*$/seeds = \"\"/g" \
   $STATE_DIR/config/config.toml
 #sed -Ei 's/^log_level = "info"$/log_level = "debug"/g' $STATE_DIR/config/config.toml
+#if [ -n "${DEBUG-}" ]; then
```
DimaOrekhovPS, line 52: Why commented if it's already inside an `if` clause?

inference-chain/scripts/init-docker.sh
```diff
@@ -104,6 +104,10 @@ $APP_NAME config set app state-sync.snapshot-keep-recent $SNAPSHOT_KEEP_RECENT
 sed -Ei 's/^laddr = ".*:26657"$/laddr = "tcp:\/\/0\.0\.0\.0:26657"/g' \
   $STATE_DIR/config/config.toml
 
+#if [ -n "${DEBUG-}" ]; then
```
DimaOrekhovPS, line 107: same question

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -155,38 +155,99 @@ func (k msgServer) getValidatedInferences(ctx sdk.Context, msg *types.MsgClaimRe
...
-		k.LogError("Validator not found in weight map", types.Claims, "validator", msg.Creator)
+		k.LogError("Validator not found in main weight map", types.Claims, "validator", msg.Creator)
 		return nil, types.ErrParticipantNotFound
 	}
+
+	// Get sub models from the main epoch data
+	for _, subModelId := range mainEpochData.SubGroupModels {
+		_, subWeightMap, subTotalWeight, found := k.getEpochGroupWeightData(ctx, msg.PocStartHeight, subModelId)
+		if !found {
+			continue
```
DimaOrekhovPS, line 209: Isn't it some kind of abnormality? Do we need a warning/error here?

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -155,38 +155,99 @@ func (k msgServer) getValidatedInferences(ctx sdk.Context, msg *types.MsgClaimRe
...
+		}
+
+		modelWeightMaps[subModelId] = subWeightMap
+		modelTotalWeights[subModelId] = subTotalWeight
+	}
+
 	mustBeValidated := make([]string, 0)
-	finishedInferences := k.GetInferenceValidationDetailsForEpoch(ctx, epochData.EpochGroupId)
+	finishedInferences := k.GetInferenceValidationDetailsForEpoch(ctx, mainEpochData.EpochGroupId)
```
DimaOrekhovPS, line 217: Looks expensive, I wonder if we need to rethink how we index inferences. I mean how keys in KV storage look for them

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -155,38 +155,99 @@ func (k msgServer) getValidatedInferences(ctx sdk.Context, msg *types.MsgClaimRe
...
 		if inference.ExecutorId == msg.Creator {
 			continue
 		}
+
+		// Determine which model this inference belongs to
+		modelId := inference.Model
+		weightMap, exists := modelWeightMaps[modelId]
+		if !exists {
+			// If no specific model weight map exists, use the main one
+			weightMap = mainWeightMap
```
DimaOrekhovPS: Why isn't it an error?

DimaOrekhovPS (commented):
Listed my questions

inference-chain/scripts/init-docker-genesis.sh
patimen: It just produced too much noise, even for debugging. Very hard to make sense of it.

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
patimen: These aren't inferences, they're InferenceValidationDetails, and they are indexed by Epoch already.

decentralized-api/internal/server/public/post_chat_handler.go
patimen: I see this is a touch hacky... but propagating a header all the way down here, etc... just for this one scenario? Maybe if (when) we have more scenarios.

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
patimen: I think a warning... I think there's some circumenstances where we might not have inferences for a model during an epoch? In any event, this would cause issues downstream and a wanring would clue us in.

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
patimen: Yeah, an error.

PR #150, opened 2025-05-22
Fix streamed response validation

DimaOrekhovPS:


decentralized-api/completionapi/openai.go
```diff
@@ -90,35 +92,112 @@ func UnmarshalEvent(event string) (*Response, error) {
 }
 
 type JsonOrStreamedResponse struct {
```
patimen: Seems a bit cludgy in the end. Lots of conditionals where you would normally expect a class/interface type solution? Did you look into having a more object oriented/inheritance based structure?

Additionally, I'd say this should be named "CompletionResponse"

decentralized-api/completionapi/openai.go
```diff
@@ -90,35 +92,112 @@ func UnmarshalEvent(event string) (*Response, error) {
...
+	StreamedResponse *struct {
+		Lines []string
+		Resp  StreamedResponse
+	}
+}
+
+func NewJsonOrStreamedResponseFromBytes(bytes []byte) (*JsonOrStreamedResponse, error) {
+	var response Response
+	if err := json.Unmarshal(bytes, &response); err != nil {
+		logging.Error("Failed to unmarshal json response into completionapi.Response", types.Inferences, "responseString", string(bytes), "err", err)
```
patimen: Probably want to limit the length of the log somehow. This could be some gigantic thing, couldn't it?

decentralized-api/internal/validation/inference_validation.go
```diff
@@ -272,26 +272,26 @@ func (s *InferenceValidator) Validate(inference types.Inference, inferenceNode *
 
 func GetEnforcedStr(r *completionapi.JsonOrStreamedResponse) (string, error) {
```
patimen: Like here. GetEnforcedStr would be a function in a CompletionResponse interface, with JsonResponse and StreamedResponse with their own implementation.

patimen (approved):
Take the suggestions if you'd like. This DOES seem like it should use an interface, not a bunch of if statements.

PR #159, opened 2025-05-30
Fix cosmovisor for API

patimen:
1. Make sure the test fails if the upgrade fails.
2. Make sure `decentralized-api` stops properly when we hit the upgrade height.
3. Make sure `decentralized-api` returns an error on upgrade so cosmovisor will upgrade correctly
4. Update to new cosmovisor that gets upgrade height the old way (via inferenced status).


gmorgachev (approved):
I tested - upgrade is successful

DimaOrekhovPS (approved):

PR #160, opened 2025-05-30
MLNode PoC state reconciliation

DimaOrekhovPS:


decentralized-api/broker/node_worker_commands.go
```diff
@@ -0,0 +1,328 @@
...
+	WorldSize      int
+}
+
+func (c StartTrainingNodeCommand) Execute(ctx context.Context, worker *NodeWorker) NodeResult {
+	result := NodeResult{
+		OriginalTarget: types.HardwareNodeStatus_TRAINING,
+	}
+
+	if ctx.Err() != nil {
+		result.Succeeded = false
```
patimen, line 269: Any patterns you can think of to reduce the size of the error handling here? It takes up the bulk of the code and kind of obscures the actual main like.
Even just a helper method or two?

decentralized-api/broker/node_worker_commands.go
```diff
@@ -0,0 +1,328 @@
...
+type StopNodeCommand struct{}
+
+func (c StopNodeCommand) Execute(ctx context.Context, worker *NodeWorker) NodeResult {
+	result := NodeResult{
+		OriginalTarget: types.HardwareNodeStatus_STOPPED,
+	}
+
+	if ctx.Err() != nil {
+		result.Succeeded = false
+		result.Error = ctx.Err().Error()
```
patimen, line 27: Are we sure the ctx.Err().Error() will be enough context?

decentralized-api/broker/node_worker.go
```diff
@@ -0,0 +1,137 @@
...
+	go worker.run()
+	return worker
+}
+
+// run is the main event loop for the worker
+func (w *NodeWorker) run() {
+	for {
+		select {
+		case item := <-w.commands:
+			result := item.cmd.Execute(item.ctx, w)
```
patimen, line 47: Do we need a log here to know when this starts?

decentralized-api/broker/node_worker.go
```diff
@@ -0,0 +1,137 @@
...
+				Response: make(chan bool, 1),
+			}
+			if err := w.broker.QueueMessage(updateCmd); err != nil {
+				logging.Error("Failed to queue node result update command", types.Nodes,
+					"node_id", w.nodeId, "error", err)
+			}
+			// We don't wait for the response from updateCmd, the worker's job is done.
+			w.wg.Done()
+		case <-w.shutdown:
+			// Drain remaining commands before shutting down
```
patimen, line 62: And maybe a log here too.

decentralized-api/broker/node_worker_commands.go
```diff
@@ -0,0 +1,328 @@
...
+		logging.Error(err.Error(), types.Training, "node_id", worker.nodeId)
+		result.Succeeded = false
+		result.Error = err.Error()
+		result.FinalStatus = types.HardwareNodeStatus_FAILED
+		return result
+	}
+
+	// Stop node first
+	if err := worker.mlClient.Stop(ctx); err != nil {
+		logging.Error("Failed to stop node for training", types.Training, "node_id", worker.nodeId, "error", err)
```
patimen, line 288: Do we want to log something with types.Node here as well? And in other places? I'm just thinking about being able to see the entire story by filtering by subsystem=Node

DimaOrekhovPS: Can we specify multiple subsystems?

PR #168, opened 2025-06-05
Explorer missing metrics

0xBECEDA:


decentralized-api/apiconfig/config.go
```diff
@@ -95,6 +95,7 @@ type ApiConfig struct {
 	MLServerPort          int    `koanf:"ml_server_port"`
 	AdminServerPort       int    `koanf:"admin_server_port"`
 	MlGrpcServerPort      int    `koanf:"ml_grpc_server_port"`
+	ExplorerUrl      string `koanf:"explorer_url"`
```
patimen: alignment?

decentralized-api/internal/server/admin/entity.go
```diff
@@ -7,5 +7,9 @@ type UnitOfComputePriceProposalDto struct {
 
 type RegisterModelDto struct {
 	Id                     string `json:"id"`
+	ContextInAiTokens      uint64 `json:"context_in_ai_tokens"`
```
patimen, line 10: Context window? I'm assuming, let's be explicit.

patimen: And let's just call it "ContextWindow". Tokens is the only unit I've every seen for that, and I think the only one possible with LLMs?

decentralized-api/internal/server/admin/entity.go
```diff
@@ -7,5 +7,9 @@ type UnitOfComputePriceProposalDto struct {
...
+	Quantization           string `json:"quantization"`
+	InputPriceInAiTokens   uint64 `json:"input_price_in_ai_tokens"`
```
patimen, line 12: How about "CoinsPerInputToken" and "CoinsPerOutputToken"?

inference-chain/scripts/init-docker-genesis.sh
```diff
@@ -31,6 +31,25 @@ APP_NAME="inferenced"
 CHAIN_ID="gonka-testnet-3"
 COIN_DENOM="nicoin"
 STATE_DIR="/root/.inference"
+
+update_configs_for_explorer() {
+  if [ "$WITH_EXPLORER" = true ]; then
+    echo "Updating configs for enable explorer..."
+    sed -i 's/^enable *= *false/enable = true/' "$STATE_DIR/config/app.toml"
```
patimen: Maybe we could specify this more specifically? This seems dangerous to enable anything.

patimen: Let's see if we can use patch-toml.

inference-chain/x/inference/keeper/developer_stats_aggregation.go
```diff
@@ -0,0 +1,436 @@
...
+	if val := byTimeStore.Get(timeKey); val != nil {
+		k.LogInfo("record found by time key", types.Stat, "inference_id", inference.InferenceId, "developer", inference.RequestedBy)
+		k.cdc.MustUnmarshal(val, &statsByTime)
+		prevEpochId = statsByTime.EpochId
+
+		statsByTime.EpochId = currentEpochId
+		statsByTime.Inference.Status = inference.Status
+		statsByTime.Inference.AiTokensUsed = tokens
+		statsByTime.Inference.EpochPocBlockHeight = inference.EpochGroupId
+		statsByTime.Inference.ActualConstInCoins = inference.ActualCost
```
patimen: ActualCostInCoins

inference-chain/x/inference/keeper/developer_stats_aggregation.go
```diff
@@ -0,0 +1,436 @@
...
+		prevEpochId uint64
+	)
+	if val := byTimeStore.Get(timeKey); val != nil {
+		k.LogInfo("record found by time key", types.Stat, "inference_id", inference.InferenceId, "developer", inference.RequestedBy)
+		k.cdc.MustUnmarshal(val, &statsByTime)
+		prevEpochId = statsByTime.EpochId
+
+		statsByTime.EpochId = currentEpochId
+		statsByTime.Inference.Status = inference.Status
+		statsByTime.Inference.AiTokensUsed = tokens
```
patimen: TotalTokenCount - but are we sure we don't want completion and input separate?

inference-chain/x/inference/keeper/msg_server_revalidate_inference.go
```diff
@@ -41,6 +41,11 @@ func (k msgServer) RevalidateInference(goCtx context.Context, msg *types.MsgReva
 
 	k.LogInfo("Saving inference", types.Validation, "inferenceId", inference.InferenceId, "status", inference.Status, "authority", inference.ProposalDetails.PolicyAddress)
 	k.SetInference(ctx, inference)
-
+	err := k.SetDeveloperStats(ctx, inference)
```
patimen: We do this a LOT.... maybe a single method that they all call that would log error/info as well as set?

inference-chain/scripts/init-docker.sh
```diff
@@ -62,6 +62,24 @@ SNAPSHOT_INTERVAL="${SNAPSHOT_INTERVAL:-10}"
 SNAPSHOT_KEEP_RECENT="${SNAPSHOT_KEEP_RECENT:-5}"
 TRUSTED_BLOCK_PERIOD="${TRUSTED_BLOCK_PERIOD:-2}"
 
+update_configs_for_explorer() {
+  if [ "${WITH_EXPLORER:-}" = true ]; then
+    echo "Updating configs for enable explorer..."
+    sed -i 's/^enable *= *false/enable = true/' "$STATE_DIR/config/app.toml"
```
patimen: patch-toml here as well.

inference-chain/proto/inference/inference/participant.proto
```diff
@@ -37,4 +37,3 @@ message CurrentEpochStats {
   uint64 validated_inferences = 6;
   uint64 invalidated_inferences = 7;
 }
-
```
patimen: Can we back out this change?

patimen (approved):
Let's break developer_stats_aggregation.go into a more CRUD based approach, where each Key we are querying by gets it's own file, like we see in inference.go or participant.go. Ignite usually does this for us.

inference-chain/x/inference/keeper/developer_stats_aggregation.go
0xBECEDA: We want to have total

inference-chain/proto/inference/inference/query.proto
```diff
@@ -583,6 +628,88 @@ message ParticipantCurrentStats {
   int32  reputation     = 3;
 }
 
+message ParticipantFullStats {
+  string account_address = 1;
+  string validator_operator_addres = 2;
```
GLiberman: Participant is in fact validator - I think operator_address is shorter and better?

GLiberman: Fixed

PR #186, opened 2025-07-01
Use epochs for dev stats

DimaOrekhovPS:


inference-chain/proto/inference/inference/inference_validation_details.proto
```diff
@@ -4,12 +4,13 @@ package inference.inference;
...
-  uint64 epoch_id = 1;
+  uint64 epoch_id = 1; // poc start block height
   string inference_id = 2;
   string executor_id = 3;
   int32 executor_reputation = 4;
   uint64 traffic_basis = 5;
   uint64 executor_power = 6;
   string model = 7;
   uint64 total_power = 8;
+  uint64 epoch_index = 9;
```
DimaOrekhovPS: Think of an alternative: `epoch_id_index`?

inference-chain/x/inference/keeper/developer_stats_aggregation.go
```diff
@@ -16,38 +16,41 @@ type StatsSummary struct {
...
 	}
 
 	inferenceStats := types.InferenceStats{
-		EpochPocBlockHeight: inference.EpochGroupId,
-		InferenceId:         inference.InferenceId,
-		Status:              inference.Status,
-		TotalTokenCount:     tokens,
-		Model:               inference.Model,
-		ActualCostInCoins:   inference.ActualCost,
+		EpochId:           inference.EpochId,
```
DimaOrekhovPS, line 43: EpochId can be removed now, we already store it in the enclosing entity `DeveloperStatsByTime `

inference-chain/x/inference/keeper/developer_stats_aggregation.go
```diff
@@ -16,38 +16,41 @@ type StatsSummary struct {
...
-	k.LogInfo("SetDeveloperStats: got stat", types.Stat, "inference_id", inference.InferenceId, "inference_status", inference.Status.String(), "developer", inference.RequestedBy, "poc_block_height", inference.EpochGroupId)
-	epoch, err := k.GetCurrentEpochGroup(ctx)
-	if err != nil {
-		return err
+	k.LogInfo("SetDeveloperStats: got stat", types.Stat,
+		"inference_id", inference.InferenceId,
+		"inference_status", inference.Status.String(),
+		"developer", inference.RequestedBy,
+		"poc_block_height", inference.EpochGroupId)
+	effectiveEpoch, found := k.GetEffectiveEpoch(ctx)
```
DimaOrekhovPS: First check if `inference.EpochId` is present here?

inference-chain/proto/inference/inference/inference_validation_details.proto
DimaOrekhovPS: Do we even need to add it now?

PR #199, opened 2025-07-11
[Require chain restart?]: OperatorAddress from account key

gmorgachev:


inference-chain/x/inference/keeper/query_get_participant_current_stats.go
```diff
@@ -79,3 +88,57 @@ func (k Keeper) GetParticipantsFullStats(ctx context.Context, _ *types.QueryPart
...
+	for _, validator := range validators {
+		if validator.OperatorAddress == accountBasedValAddr.String() {
+			k.LogDebug("Found validator using account-derived address", types.Participants,
+				"participant", participantAddr, "validatorAddr", accountBasedValAddr.String())
+			return accountBasedValAddr.String(), nil
+		}
+	}
+
+	// Second try: consensus-derived validator address (runtime case)
+	if validatorKey != "" {
```
DimaOrekhovPS: Looks like a fallback for handling the old incorrect addressing? or not
didn't get this one

DimaOrekhovPS (approved):

inference-chain/x/inference/keeper/query_get_participant_current_stats.go
gmorgachev: it's actually what we have in main now. hotfix and proposal were done at the same time

PR #205, opened 2025-07-15
Cosmos 0.53.3 upgrade (plus Replay attacks)

patimen:
# Major fix for exploits:
1. Signatures. Each inference participant signs.
    - Dev - Already signing, but now includes a `timestamp` and the address of the `transferAgent`. This signature becomes the InferenceId (preventing duplication at the chain level), the timestamp allows TA and EA to keep short lived lists of recent signatures for immediate rejection (preventing GPU burn), and of course only the designated TA can serve the request.
    - TA - The TA also adds a signature, with the EA address added on.
    - EA - the EA adds it's own signature, though at the moment it is the same payload as the TA
 2. Immediate Dedup - Both TA and EA keep a list of recent signatures they have served, and check for timestamps that are either too far in the future or too far in the past. This prevents simply replay attacks from even hitting inference resources.
 3. On Chain validation - On-Chain, since InferenceId is now the Dev signature, replay attacks by the TA are prevented, though they could still be used to burn GPU (see long term plan for TA reputation). Also, ALL the signatures that are available (Dev,TA and EA) are verified at each step.

## Open Questions
- As mentioned, there are still several attacks, especially by rogue TAs, available.
- FinishInference now has to include more data, especially the promptPayload, in order to verify signatures on chain. We should consider signing a sha256 hash of the payload instead of the full payload.
- StartInference and FinishInference contain duplicate data, but we need to figure out what to do when/if the data from each doesn't match the other. Not entirely clear what things COULD be different, given the signatures.

# Cosmos 0.53.3 upgrade
1. Update lots of dependencies
2. Add `auth` module to preBlockers
3. A few changes in `testermint` structs
4. Test fixes to verify changes - add File serving to mock-server


inference-chain/x/inference/calculations/inference_state.go
```diff
@@ -59,8 +59,15 @@ func ProcessStartInference(
...
 	// Works if FinishInference came before
+	currentInference.RequestTimestamp = startMessage.RequestTimestamp
+	currentInference.TransferredBy = startMessage.Creator
+	currentInference.TransferSignature = startMessage.TransferSignature
 	currentInference.PromptHash = startMessage.PromptHash
 	currentInference.PromptPayload = startMessage.PromptPayload
+	currentInference.OriginalPrompt = startMessage.OriginalPrompt
+	if currentInference.PromptTokenCount == 0 {
+		currentInference.PromptTokenCount = startMessage.PromptTokenCount
+	}
```
gmorgachev, line 70: I don't think we should charge with `startMessage.PromptTokenCount`, i deleted it on purpose. It's raw estimation for escrow, much bigger then final cost. If there is no Usage from executor - it's problem of executor
Let's fix it before merge if there is no new info that i'm missin

inference-chain/proto/inference/inference/inference.proto
```diff
@@ -45,5 +45,10 @@ message Inference {
   string node_version = 24;
   uint64 epoch_id = 25;
   uint64 epoch_poc_start_block_height = 26;
+  string transferred_by = 27;
+  int64 request_timestamp = 28;
+  string transfer_signature = 29;
+  string execution_signature = 30;
+  string original_prompt = 31;
```
gmorgachev, line 52: We're storing `original_prompt` 3 times, am i right?:
- Inference
- MsgStartInference
- MsgFinishInference

It'a actually a real problem already now - we were at the limit of data transfer capacity at testnet before these changes. We need to get back to it

inference-chain/x/inference/types/params.go
```diff
@@ -63,6 +63,8 @@ func DefaultParams() Params {
 			MinValidationTrafficCutoff:  100,
 			MissPercentageCutoff:        DecimalFromFloat(0.01),
 			MissRequestsPenalty:         DecimalFromFloat(1.0),
+			TimestampExpiration:         60, // 10 seconds for timestamp expiration
+			TimestampAdvance:            30, // 10 seconds for timestamp advance
```
gmorgachev: Comments are not clear

inference-chain/app/upgrades/v1_14/upgrades.go
```diff
@@ -0,0 +1,41 @@
...
+			params.ValidationParams.TimestampExpiration = 60
+			err := k.SetParams(ctx, params)
+			if err != nil {
+				k.LogError("Failed to set params during upgrade: %v", types.Upgrades, "error", err)
+			}
+
+			// For some reason, the capability module doesn't have a version set, but it DOES exist, causing
+			// the `InitGenesis` to panic.
+			if _, ok := vm["capability"]; !ok {
+				vm["capability"] = mm.Modules["capability"].(module.HasConsensusVersion).ConsensusVersion()
```
gmorgachev, line 36: Did you have a chance to discuss with Gabriel that it's ok?

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -346,19 +625,36 @@ func (s *Server) sendInferenceTransaction(inferenceId string, response completio
...
-		PromptPayload:        promptPayload,
-		ResponseHash:         responseHash,
-		ResponsePayload:      string(bodyBytes),
-		PromptTokenCount:     usage.PromptTokens,
-		CompletionTokenCount: usage.CompletionTokens,
-		Model:                model,
-		Id:                   id,
-	}
-
 	if s.recorder != nil {
```
gmorgachev, line 629: should we record err in opposite case?

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -3,31 +3,152 @@ package public
...
+		timestampExpiration := validationParams.TimestampExpiration
+
+		// Use default value if parameter is not set
+		if timestampExpiration == 0 {
+			timestampExpiration = 10 // Default 10 seconds
+		}
+
+		// Use twice the timestamp_expiration value (converted to blocks)
+		// Assuming average block time of 5 seconds
+		expirationBlocks = (timestampExpiration * 2) / 5
```
gmorgachev: I would get coeff bigger then 2 just in case. As in ideal word blocks are faster then 5 (might be twice?)

gmorgachev (approved):
1. I'm really worried about amount of data
2. I did stress tests many time. Some time saw insufficient funds many times but it might be same reason with bugs fixed in my PR => next steps would be testing together
3. About performance - not clear the impact on short request. On long results are close

inference-chain/proto/inference/inference/inference.proto
patimen: The best option for this is to store a hash and modify our signing so it generates the key on the hash, not a has of the hash. It could be a pretty quick follow up, I have an idea how to do it.

decentralized-api/internal/server/public/post_chat_handler.go
patimen: That's a multiple of 2, with a minimum of 4 blocks? It seems enough.

gmorgachev: Is block length stable ~ 5s? I'm not sure if it can be like 2.5 for some period and then we would clean auth keys before timestamp expiration. I would add a gap just in case

PR #214, opened 2025-07-21
Inference val details epoch

DimaOrekhovPS:
@gmorgachev I want to include this in the chain-restart upgrade


inference-chain/x/inference/keeper/msg_server_claim_rewards_test.go
```diff
@@ -196,14 +196,6 @@ func TestMsgServer_ClaimRewards_ValidationLogic(t *testing.T) {
 	k, ms, ctx, mocks := setupKeeperWithMocks(t)
 	sdkCtx := sdk.UnwrapSDKContext(ctx)
 
-	mockCreator := NewMockAccount(testutil.Creator)
-	mockExecutor1 := NewMockAccount("executor1")
-	mockExecutor2 := NewMockAccount("executor2")
-	// Setup participants
-	MustAddParticipant(t, ms, ctx, *mockCreator)
```
DimaOrekhovPS, line 209: @patimen I actually noticed that these were silently failing, but it didn't affect the test. Decided to delete so setup only contains what's necessary

PR #216, opened 2025-07-22
Issue 335 unordered txs

0xBECEDA:
This pull request introduces updates to the decentralized API, focusing on improving transaction management. It introduces 3 different transactions sending strategies. 

1. **SendTransactionAsyncNoRetry**
Broadcast tx, do not retry if broadcasting failed. 

2. **SendTransactionSyncNoRetry**
Broadcast tx and sync. wait for tx result for 15 seconds. Do not retry to send tx.

3. **SendTransactionAsycWithRetry**
Broadcast tx, if error is not critical or there is no error at all, put to queue on disk. If tx wouldn't be found in blockchain then, retry to send tx. Keep trying till tx wouldn't be included in block or would get critical error.

Sending txs is manages by new module TxManager. 
Screenshot from 2025-07-23 19-50-25


decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -124,347 +136,187 @@ type CosmosMessageClient interface {
...
-		logging.Error("Failed to send transaction", types.Messages, "error", err, "result", result)
-		return nil, err
-	}
+	transaction.Creator = icc.address
+	msg := &inference.MsgCreateTrainingTaskResponse{}
 
-	transactionAppliedResult, err := icc.Client.WaitForTx(icc.Context, result.TxHash)
-	if err != nil {
-		logging.Error("Failed to wait for transaction", types.Messages, "error", err, "result", transactionAppliedResult)
+	if err := icc.SendTransactionSyncNoRetry(transaction, msg); err != nil {
```
patimen, line 349: Great abstraction.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+	nc *nats.Conn,
+	address string) (*manager, error) {
+	js, err := nc.JetStream()
+	if err != nil {
+		return nil, err
+	}
+
+	types.RegisterInterfaces(client.Context().InterfaceRegistry)
+
+	m := &manager{
```
patimen, line 75: It's not immediately clear what `sendTxs` and `observeTxs` are doing in `StartTxManager`

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+
+		err := m.putOnRetry(id, "", timeout, rawTx, false)
+		if err != nil {
+			logging.Error("tx failed to broadcast, failed to put in queue", types.Messages, "tx_id", id, "broadcast_err", broadcastErr, "resend_err", err)
+		}
+		return nil, ErrTxFailedToBroadcastAndPutOnRetry
+	}
+	if err := m.putOnRetry(id, resp.TxHash, timeout, rawTx, true); err != nil {
+		logging.Error("tx broadcasted, but failed to put in queue", types.Messages, "tx_id", id, "err", err)
```
patimen: English trivia: The past tense of "broadcast" is actually "broadcast". :)

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		logging.Error("Failed to wait for transaction", types.Messages, "tx_id", id, "tx_hash", resp.TxHash, "error", err)
+		return nil, err
+	}
+	return result, nil
+}
+
+func (m *manager) SignBytes(seed []byte) ([]byte, error) {
+	name := m.account.Name
+	// Kind of guessing here, not sure if this is the right way to sign bytes, will need to test
```
patimen: I think we can safely say this is the correct way now.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+
+type manager struct {
+	ctx              context.Context
+	client           *cosmosclient.Client
+	account          *cosmosaccount.Account
+	txFactory        *tx.Factory
+	accountRetriever client.AccountRetriever
+	address          string
+	defaultTimeout   time.Duration
+	nc               *nats.Conn
```
patimen: Let's go ahead and use natsConnection and natsJetStream for these names, not nc and js.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+		Timeout: timeout,
+	})
+	if err != nil {
+		return err
+	}
+	_, err = m.js.Publish(server.TxsToObserveStream, b)
+	return err
+}
+
+func (m *manager) sendTxs() error {
```
patimen, line 237: So, if I read this right, this sets up the Subscription from NATS to handle when new Txs show up in the queue? I think we could name this better, and maybe add some function comments.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+			logging.Error("error pushing to observe queue", types.Messages, "id", tx.TxInfo.Id, "err", err)
+			msg.NakWithDelay(defaultSenderNackDelay)
+		} else {
+			msg.Ack()
+		}
+	}, nats.Durable(txSenderConsumer), nats.ManualAck())
+	return err
+}
+
+func (m *manager) observeTxs() error {
```
patimen, line 286: Same as for sendTxs

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+		logging.Debug("SendTxs: got tx", types.Messages, "id", tx.TxInfo.Id)
+
+		rawTx, err := m.unpackTx(tx.TxInfo.RawTx)
+		if err != nil {
+			logging.Error("error unpacking raw tx", types.Messages, "id", tx.TxInfo.Id, "err", err)
+			msg.Term() // malformed, drop it
+			return
+		}
+
+		if !tx.Sent {
```
patimen, line 257: How would something get in the toSendStream and already have tx.Sent to be true?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+			msg.Ack()
+		}
+	}, nats.Durable(txSenderConsumer), nats.ManualAck())
+	return err
+}
+
+func (m *manager) observeTxs() error {
+	logging.Info("Tx manager: observeTxs txs: run in background", types.Messages)
+	_, err := m.js.Subscribe(server.TxsToObserveStream, func(msg *nats.Msg) {
+		var tx txInfo
```
patimen, line 289: Given the importance of the logic in this method (it defines a big part of our queue flow), I think we could compose this a little more? As in, look at some ways of making the logic a little clearer, maybe move some of the error handling around?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+
+	}, nats.Durable(txObserverConsumer), nats.ManualAck())
+	return err
+}
+
+func (m *manager) GetClientContext() client.Context {
+	return m.client.Context()
+}
+
+func (m *manager) checkTxStatus(hash string) (bool, error) {
```
patimen, line 348: More of a `checkTxOnChain`? Are we checking the status of the Tx at all?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -0,0 +1,472 @@
...
+		return nil, NewTransactionErrorFromResult(transactionAppliedResult)
+	}
+	return transactionAppliedResult, nil
+}
+
+func (m *manager) BankBalances(ctx context.Context, address string) ([]sdk.Coin, error) {
+	return m.client.BankBalances(ctx, address, nil)
+}
+
+func (m *manager) broadcastMessage(id string, rawTx sdk.Msg) (*sdk.TxResponse, time.Time, error) {
```
patimen, line 389: Maybe make it a *time.Time instead? the `time.Time{}` parts are just placeholders? Maybe that's a Go idiom I'm just not used to.

patimen (approved):
Great job on the extraction of the tx_manager from the client.
I would say we might want to extract out the NATS implementation out of tx_manager and into an interface, both for clarity on for flexibility.
There's some nits/comments here and there, but this is a really clear implementation, great work.

decentralized-api/scripts/init-docker.sh
```diff
@@ -42,6 +42,9 @@ echo "init for cosmovisor"
 mkdir -p /root/.dapi
 mkdir -p /root/.dapi/data
 
+echo "init for nats"
+mkdir -p /root/.nats
+
```
gmorgachev: Let's put it inside /root/.dapi

testermint/src/main/kotlin/DockerGroup.kt
```diff
@@ -96,13 +97,17 @@ data class DockerGroup(
             put("DAPI_API__PUBLIC_SERVER_PORT", "9000")
             put("DAPI_API__ML_SERVER_PORT", "9100")
             put("DAPI_API__ADMIN_SERVER_PORT", "9200")
+            put("DAPI_NATS__HOST", "0.0.0.0")
+            put("DAPI_NATS__PORT", "4222")
+            put("DAPI_NATS__STORAGE_DIR", "/root/nats")
```
gmorgachev: Why another path?

local-test-net/docker-compose-base.yml
```diff
@@ -39,7 +41,9 @@ services:
       - DAPI_API__ML_GRPC_SERVER_PORT=9300
       - DAPI_API__TEST_MODE=true
       - NODE_CONFIG_PATH=/root/node_config.json
-      - DEBUG=false
+      - DAPI_NATS__HOST=0.0.0.0
+      - DAPI_NATS__PORT=4222
```
gmorgachev: Inside, full, not local deploy, can we set this values by default? Don't want to make configuration more complex

gmorgachev (approved):
looks solid. if all integration tests + stress tests passed and can be applied by upgrade - let's merge

decentralized-api/cosmosclient/tx_manager/tx_manager.go
0xBECEDA: 1. Tx was resent but we couldn't put it to 'observe queue' because of some reason. We try to move it later, but we do not need to send it again till observer wouldn't find this tx missed. Otherwise we eill send same tx just twice. 

2. We first try to broadcast tx and then put it into queue. 'Entry point' for whole resend algorithm is 'send queue', because then we have simply flow.  If tx was broadcast successfully, same as in example above, we just do not want to send it twice.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
0xBECEDA: We check if tx was included in block at all. We do not care if result is success or fail.

local-test-net/docker-compose-base.yml
0xBECEDA: Yes, we can

PR #226, opened 2025-07-24
Simple Pruning

patimen:
### Pruning System for Inference-Ignite Blockchain

This PR adds a pruning system to reduce storage usage by removing stale data while preserving essential analytics and integrity.

#### What’s Pruned

* **Inference Records**: Prompt and response payloads are removed after a configurable number of epochs (`inference_pruning_epoch_threshold`). Only finalized or expired inferences are eligible.
* **PoC Data**: `PoCBatch` and `PoCValidation` entries are removed after `poc_data_pruning_epoch_threshold` epochs.

#### Execution Timing

* Runs during the PoC phase of each epoch via `EndBlock`.
* Skips recent or incomplete records.

#### Implementation

* Adds new keeper methods: `PruneInferences` and `PrunePoCData`.
* Introduces two config params:

  * `inference_pruning_epoch_threshold` (in `EpochParams`)
  * `poc_data_pruning_epoch_threshold` (in `PocParams`)
* Uses epoch-based thresholds to maintain consistency across nodes.

#### Governance

* Both thresholds are adjustable via on-chain governance.
* Retention tradeoffs should consider analytics needs, auditability, and storage cost.

#### Testing
- Added two queries, on for the number of PocBatch records and the other for PocValidation records in order to verify correct pruning.
- Added unit tests and testermint tests for functionality


inference-chain/cmd/inferenced/main_test.go
```diff
@@ -0,0 +1,16 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestMainFunc(t *testing.T) {
```
0xBECEDA, line 8: Do we really need this test? For me it looks like it tests common panic-recover flow. Am I wrong?

inference-chain/x/inference/keeper/pruning.go
```diff
@@ -0,0 +1,215 @@
...
+import (
+	"context"
+	"cosmossdk.io/store/prefix"
+	"github.com/cosmos/cosmos-sdk/runtime"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// PruneInferences removes old inference records based on threshold and status
+func (k Keeper) PruneInferences(ctx context.Context, currentEpochIndex uint64, pruningThreshold uint64) error {
+	// Get all inferences
```
0xBECEDA: Minor, but I would avoid commenting code too much.

inference-chain/x/inference/keeper/pruning.go
```diff
@@ -0,0 +1,215 @@
...
+func (k Keeper) PrunePoCData(ctx context.Context, currentEpochIndex uint64, pruningThreshold uint64) error {
+	_, found := k.GetEpoch(ctx, currentEpochIndex)
+	if !found {
+		k.LogError("Failed to get current epoch", types.Pruning, "epoch_index", currentEpochIndex)
+		return types.ErrEffectiveEpochNotFound
+	}
+
+	// Calculate the maximum number of epochs to check (5 times the pruning threshold)
+	// This limits how far back we look to avoid performance issues on deep chains
+	maxEpochsToCheck := pruningThreshold * 5
```
0xBECEDA: I would make number as constant to easily find in code in future

inference-chain/x/inference/keeper/query_count_po_c_batches_at_height.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (k Keeper) CountPoCbatchesAtHeight(goCtx context.Context, req *types.QueryCountPoCbatchesAtHeightRequest) (*types.QueryCountPoCbatchesAtHeightResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO: Process the query
+	_ = ctx
```
0xBECEDA: this looks strange to me

inference-chain/x/inference/keeper/query_count_po_c_validations_at_height.go
```diff
@@ -0,0 +1,29 @@
...
+
+func (k Keeper) CountPoCvalidationsAtHeight(goCtx context.Context, req *types.QueryCountPoCvalidationsAtHeightRequest) (*types.QueryCountPoCvalidationsAtHeightResponse, error) {
+	if req == nil {
+		return nil, status.Error(codes.InvalidArgument, "invalid request")
+	}
+
+	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// TODO: Process the query
+	_ = ctx
```
0xBECEDA: same comment as above

inference-chain/x/inference/types/params.go
```diff
@@ -42,14 +42,15 @@ func DefaultGenesisOnlyParams() GenesisOnlyParams {
...
-			PocValidationDuration:     6,
+			EpochLength:                    40,
+			EpochMultiplier:                1,
+			EpochShift:                     0,
+			DefaultUnitOfComputePrice:      100,
+			PocStageDuration:               10,
+			PocExchangeDuration:            2,
+			PocValidationDelay:             2,
+			PocValidationDuration:          6,
+			InferencePruningEpochThreshold: 2, // Number of epochs after which inferences can be pruned
```
0xBECEDA, line 55: for me this default value looks too small, because in dashboard we show statistic data for 7 epochs.

0xBECEDA (approved):
Generally looks good to me, left some comments.

inference-chain/cmd/inferenced/main_test.go
patimen: This does indeed serve a purpose. I had an error with `ignite` at one point that caused a panic in main that would have been captured by this test cost me some time.

inference-chain/x/inference/keeper/pruning.go
patimen: I've got to tell my AI agent to cut it with the excessive comments so I don't need to edit after the fact. :)

inference-chain/x/inference/keeper/query_count_po_c_batches_at_height.go
patimen: It's leftover `ignite` boilerplate.

inference-chain/x/inference/types/params.go
patimen: The statistical data is preserved, this won't impact it.

inference-chain/x/inference/keeper/pruning.go
patimen: Resolved this. It's an interesting balance... I find it useful to see what the LLM _thinks_ it's doing in code, in case it gets it wrong. That way I know if it is mistaken in implementation vs understanding.
But it's super annoying after the fact.

PR #235, opened 2025-07-30
Standardize transaction logging across all modules

patimen:
## Summary
Create a new module, bookkeeper, that encapsulates a simple wrapper around `bank` that logs transactions in a double-entry accounting style. (and a "simple" style more suitable for examining in test logs).
Introduce the idea of sub-account transactions, to keep track of balance changes on things like coins-owed, collateral unbonding, etc.
Sub-accounts are essentially modeled as claims-registries. They are **tracking only** and are not an actual cosmos account.
### Example
 For example, rewards vesting flow is as follows:
- Mint Rewards: This shows as moving from `supply` into `inference`.
- Settled Sub-account: The reward amount for an account (let's say for `bob`) is "moved" from `bob-settled` to `inference-settled`, showing the `inference-settled` owes bob that amount (when they claim).
- Claim comes in. Three things happen (just rewards, let's ignore work):
1. Actual account movement: `inference` moves the reward amount to `streamvesting`
2. Vesting sub-account: `bob-vesting` moves the reward amount to `streamvesting-vesting` (to mark the amount owed to `bob`
3. Claim Settled: `inference-settled` moves the reward amount to `bob-settled`, zeroing out the `bob-settled` amount.

- Later, when a vesting portion happens:
1. Actual account movement: `streamvesting` moves the vested portion to `bob`
2. Vesting sub-account: `streamvesting-vesting` moves the vested portion to `bob-vesting` (reducing the bob-vesting amount)

- In the event of a forfeited claim:
1. Actual account movement: `inference` moves the reward amount to `supply` (burning)
2. Settled sub-account: `inference-settled` moves the reward amount to `inference-bob` (the account has been settled)

### Future Possible Work
1. Currently `bookkeeper` only logs. It could store a proper on the chain, or in a database.
2. `bookkeeper` should emit events
3. `bookkeeper` could model balances for sub-accounts, allowing a more detailed audit and status for these accounts. This _might_ even alleviate the need to have entries like vesting_owed, though things like vesting period would still need to be stored.
4. Settings like log level and whether to use double entry, simple entry or both should be configurable by the node (not by Params, which would require votes).

### DIFF
--- a/proposals/bookkeeping/bookkeeping.md
+++ b/proposals/bookkeeping/bookkeeping.md
@@ -1,0 +1,26 @@
+## Summary
+Create a new module, bookkeeper, that encapsulates a simple wrapper around `bank` that logs transactions in a double-entry accounting style. (and a "simple" style more suitable for examining in test logs).
+Introduce the idea of sub-account transactions, to keep track of balance changes on things like coins-owed, collateral unbonding, etc.
+Sub-accounts are essentially modeled as claims-registries. They are **tracking only** and are not an actual cosmos account.
+### Example
+For example, rewards vesting flow is as follows:
+- Mint Rewards: This shows as moving from `supply` into `inference`.
+- Settled Sub-account: The reward amount for an account (let's say for `bob`) is "moved" from `bob-settled` to `inference-settled`, showing the `inference-settled` owes bob that amount (when they claim).
+- Claim comes in. Three things happen (just rewards, let's ignore work):
+1. Actual account movement: `inference` moves the reward amount to `streamvesting`
+2. Vesting sub-account: `bob-vesting` moves the reward amount to `streamvesting-vesting` (to mark the amount owed to `bob`
+3. Claim Settled: `inference-settled` moves the reward amount to `bob-settled`, zeroing out the `bob-settled` amount.
+
+- Later, when a vesting portion happens:
+1. Actual account movement: `streamvesting` moves the vested portion to `bob`
+2. Vesting sub-account: `streamvesting-vesting` moves the vested portion to `bob-vesting` (reducing the bob-vesting amount)
+
+- In the event of a forfeited claim:
+1. Actual account movement: `inference` moves the reward amount to `supply` (burning)
+2. Settled sub-account: `inference-settled` moves the reward amount to `inference-bob` (the account has been settled)
+
+### Future Possible Work
+1. Currently `bookkeeper` only logs. It could store a proper on the chain, or in a database.
+2. `bookkeeper` should emit events
+3. `bookkeeper` could model balances for sub-accounts, allowing a more detailed audit and status for these accounts. This _might_ even alleviate the need to have entries like vesting_owed, though things like vesting period would still need to be stored.
+4. Settings like log level and whether to use double entry, simple entry or both should be configurable by the node (not by Params, which would require votes).
--- a/proposals/bookkeeping/flows.md
+++ b/proposals/bookkeeping/flows.md
@@ -1,0 +1,116 @@
+# Flows as of Tokenomics v2
+Rough chart of funding flows as of Tokenomics v2.
+
+In all charts, the accounts with underscores are "subaccounts", the others are full Cosmos accounts.
+## Inferences
+
+```mermaid
+sequenceDiagram
+    participant inference as inference
+    participant dev as dev
+    participant exec as exec
+    participant exec_balance as exec_balance
+    participant exec_settled as exec_settled
+    participant exec_vesting as exec_vesting
+    participant vesting as vesting
+    participant supply as supply
+
+    note over inference: Module cosmos<br>account
+    note over dev: dev cosmos<br>account
+    note over exec: executor cosmos<br>account
+    note over exec_balance: executor balance<br>owed
+    note over exec_settled: executor amount<br>awaiting claim
+    note over exec_vesting: amount held for<br>vesting for exec
+    note over vesting: Module cosmos<br>account
+    note over supply: Burns/Mints come from<br>here
+
+    %% Inference Starts
+    rect rgb(220, 255, 255)
+    Note over inference,dev: Inference Starts
+    dev->>inference: escrow
+    end
+
+    %% Inference Finishes
+    rect rgb(220, 255, 255)
+    Note over inference,dev: Inference Finishes
+    inference-->>dev: refund
+    inference-->>exec_balance: work
+    end
+
+    %% Inference Expires
+    rect rgb(255, 255, 220)
+    Note over inference,dev: Inference Expires
+    inference-->>dev: full refund
+    end
+
+    %% Inference Invalidated
+    rect rgb(255, 220, 220)
+    Note over inference,dev: Inference Invalidated
+    exec_balance-->>inference: removed
+    inference-->>dev: refunded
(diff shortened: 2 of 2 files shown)


inference-chain/app/app_config.go
```diff
@@ -101,8 +105,9 @@ var (
 		consensustypes.ModuleName,
 		circuittypes.ModuleName,
 		// chain modules
-		collateralmoduletypes.ModuleName,
+		bookkeepermoduletypes.ModuleName,
 		inferencemoduletypes.ModuleName,
+		collateralmoduletypes.ModuleName,
```
libermans, line 110: I put collateral before inference, as inferencemodule uses collateral module?

inference-chain/x/collateral/keeper/keeper.go
```diff
@@ -28,8 +28,8 @@ type (
 		// should be the x/gov module account.
 		authority string
 
-		bankKeeper       types.BankKeeper
-		bankEscrowKeeper types.BankEscrowKeeper
+		bankKeeper            types.BankKeeper
```
libermans: should we call it bankViewKeeper?

inference-chain/x/collateral/keeper/msg_server_deposit_collateral.go
```diff
@@ -39,6 +39,8 @@ func (k msgServer) DepositCollateral(goCtx context.Context, msg *types.MsgDeposi
 		currentCollateral = msg.Amount
 	}
 
+	k.bookkeepingBankKeeper.LogSubAccountTransaction(goCtx, types.ModuleName, msg.Participant, types.SubAccountCollateral, msg.Amount, "collateral deposit")
```
libermans, line 42: Should the sub account be just an optional property for SendCoinsFromAccountToModule? So that the logic of logging stays in the bookkeeper?

inference-chain/x/inference/keeper/collateral_integration_test.go
```diff
@@ -67,17 +68,17 @@ func setupRealKeepers(t testing.TB) (sdk.Context, keeper.Keeper, collateralKeepe
 		runtime.NewKVStoreService(collateralStoreKey),
 		keepertest.PrintlnLogger{},
 		authority.String(),
-		nil,        // bank keeper
-		bankKeeper, // bank escrow keeper
+		nil,                  // bank keeper
+		bookkepingBankKeeper, // bank escrow keeper
```
libermans: should we keep escrow in the comment?

inference-chain/x/inference/keeper/payment_handler.go
```diff
@@ -67,41 +62,27 @@ func (k *Keeper) PayParticipantFromModule(ctx context.Context, address string, a
...
-		err = k.BankKeeper.SendCoinsFromModuleToModule(ctx, moduleName, "streamvesting", vestingAmount)
-		if err != nil {
-			k.LogError("Error transferring coins to streamvesting module", types.Payments, "error", err, "amount", vestingAmount)
-			return err
-		}
-
-		// Then, add to vesting schedule with specified vesting period
-		err = k.GetStreamVestingKeeper().AddVestedRewards(ctx, address, vestingAmount, vestingEpochs)
+		// Vesting keeper should move funds and create vesting schedule
+		err = k.GetStreamVestingKeeper().AddVestedRewards(ctx, address, types.ModuleName, vestingAmount, vestingEpochs, memo+"_vested")
```
libermans, line 66: Isn’t it the right approach to allow only the modules themselves to transfer their funds to other modules?

inference-chain/x/inference/keeper/payment_handler.go
```diff
@@ -67,41 +62,27 @@ func (k *Keeper) PayParticipantFromModule(ctx context.Context, address string, a
...
 	if burnCoins <= 0 {
 		k.LogInfo("No coins to burn", types.Payments, "coins", burnCoins)
 		return nil
 	}
 	k.LogInfo("Burning coins", types.Payments, "coins", burnCoins)
-	err := k.BankKeeper.BurnCoins(ctx, types.ModuleName, types.GetCoins(burnCoins))
+	err := k.BankKeeper.BurnCoins(ctx, types.ModuleName, types.GetCoins(burnCoins), memo)
 	if err == nil {
-		k.LogTransaction("supply", types.ModuleName, burnCoins, memo)
 		k.AddTokenomicsData(ctx, &types.TokenomicsData{TotalBurned: uint64(burnCoins)})
```
libermans, line 86: We burn coins in collateral account as well, which we won't register in TokenomicsData now. Should we move TokenomicsData to bookkeeper as well?

inference-chain/x/inference/keeper/streamvesting_integration_test.go
```diff
@@ -322,34 +310,3 @@ func TestVestingIntegration_ParameterValidation(t *testing.T) {
 	require.Equal(t, uint64(180), retrievedParams.TokenomicsParams.RewardVestingPeriod)
 	require.Equal(t, uint64(365), retrievedParams.TokenomicsParams.TopMinerVestingPeriod)
 }
-
-func TestVestingIntegration_ErrorHandling(t *testing.T) {
```
libermans: Is it a redundant test?

inference-chain/app/app_config.go
patimen: I'm pretty sure the order doesn't actually matter, the dependency injection figures it out.

inference-chain/x/inference/keeper/payment_handler.go
patimen: I'm open to either, but it seemed odd that the streamvestingkeeper had no guarantee that the vested funds had been transferred. One would expect "AddVestedRewards" to move the funds, intuitively.
EIther way, I'm going to leave it this way for now. A refactor would actually be quite time consuming, as the unit test mocks are heavily impacted.

inference-chain/x/inference/keeper/payment_handler.go
patimen: I think having bookkeeper track overall transaction totals somehow makes sense, but would require some thought and consideration. I'd put it in the future work?

inference-chain/x/collateral/keeper/msg_server_deposit_collateral.go
patimen: That will make sense when/if we implement actual sub-accounts. Actual accounts with totals that can be queried and audited. Then this becomes an "operational" activity. 
For now, I wanted to distinguish these from the SendCoins activities, in that ALL this does is logging, vs SendCoins which actually does move coins in accounts.

inference-chain/x/collateral/keeper/keeper.go
patimen: Yeah, to distinguish from the "doing stuff" bankKeeper. Done.

inference-chain/x/inference/keeper/collateral_integration_test.go
patimen: refactored

inference-chain/x/inference/keeper/streamvesting_integration_test.go
patimen: Half of it was, I restored the other half

PR #244, opened 2025-08-02
Key Management v0

gmorgachev:
# New architecture for node operations

This PR replaces the single-key approach with separate Account Key (cold) and ML Operational Key (warm) for better security.

### New CLI Commands
- `inferenced register-new-participant` - Register new nodes with seed node API
- `inferenced tx inference grant-ml-ops-permissions` - Grant 18 AI operation permissions via authz

### Enhanced Decentralized-API Initialization
- Modified `decentralized-api/scripts/init-docker.sh` to support externally provided ACCOUNT_PUBKEY (separate from the existing private )
- Added CREATE_KEY=true option for test environments with automatic key creation (single key still used in testermint)
- Added `decentralized-api/apiconfig/accounts.go` - ApiAccount struct with dual key
- Added `inference-chain/x/inference/permissions.go` - Granting permissions
- `inference-chain/x/inference/module/module.go` - Grant permissions CLI command
- Added `inference-chain/cmd/inferenced/cmd/register_participant_command.go` - New participant registration CLI

### Signature Verification for Grantees

API service can now verify seed signatures and inference request signatures by fetching grantees:
```go
grantees, err := queryClient.GranteesByMessageType(ctx.Request().Context(), &types.QueryGranteesByMessageType{
    GranterAddress: accountAddress,
    MessageTypeUrl: "/inference.inference.MsgStartInference",
})
```

**Current Limitations:**
1. All grantees' public keys are checked 
2. If permission is revoked during PoC execution => unknown behavior  
3. Key rotation procedure is in TODO (not yet implemented)

### Current State

- **Testermint**: Uses test keyring with single key as both Account Key and ML Operational Key
- **Production**: External Account Key creation planned for future releases

Maintains backwards compatibility with existing single-key setups.

### Documentation

- **Overview**: `proposals/keys/README.md` - Key management architecture and implementation guide
- **Task Tracking**: `proposals/keys/todo.md` - Development progress and completed tasks


decentralized-api/apiconfig/config_manager.go
```diff
@@ -230,7 +230,25 @@ func readConfig(provider koanf.Provider) (Config, error) {
 		log.Fatalf("error unmarshalling config: %v", err)
 	}
 	if keyName, found := os.LookupEnv("KEY_NAME"); found {
-		config.ChainNode.AccountName = keyName
+		config.ChainNode.SignerKeyName = keyName
+		log.Printf("Loaded KEY_NAME: %+v", keyName)
+	}
+
+	if accountPubKey, found := os.LookupEnv("ACCOUNT_PUBKEY"); found {
```
DimaOrekhovPS, line 237: You decided to not use `DAPI_` because they require double underscores?

decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -2,38 +2,45 @@ package cosmosclient
...
-	Account   *cosmosaccount.Account
-	Address   string
-	Context   context.Context
-	TxFactory *tx.Factory
+	Client     *cosmosclient.Client
+	ApiAccount *apiconfig.ApiAccount
+	Address    string
+	Context    context.Context
+	TxFactory  *tx.Factory
+	NodeConfig apiconfig.ChainNodeConfig
```
DimaOrekhovPS: Hm, I think we never save reference to a config, but use `configManager`, because we assume config values can be updated and the manager will always fetch the most up to date data.

Not critical, can be kept as is since we don't expect the address to change without a restart

decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -281,20 +340,33 @@ func (icc *InferenceCosmosClient) BridgeExchange(transaction *types.MsgBridgeExc
...
+func (icc *InferenceCosmosClient) BroadcastMessage(ctx context.Context, msg sdk.Msg) (*sdk.TxResponse, error) {
+	factory, err := icc.getFactory()
 	if err != nil {
 		return nil, err
 	}
-	unsignedTx, err := factory.BuildUnsignedTx(msg)
+
+	var finalMsg sdk.Msg = msg
+	if !icc.ApiAccount.IsSignerTheMainAccount() {
+		granteeAddress, err := icc.ApiAccount.SignerAddress()
```
DimaOrekhovPS, line 353: So basically now nodes work with some keys that only have a subset of permissions of the original account?

decentralized-api/scripts/init-docker.sh
```diff
@@ -15,6 +15,48 @@ if [ -z "${KEY_NAME-}" ]; then
...
+
+# If ACCOUNT_PUBKEY is not provided but CREATE_KEY=false, try to extract from existing key
+if [ -z "${ACCOUNT_PUBKEY-}" ]; then
+  echo "WARNING: ACCOUNT_PUBKEY not provided, attempting to extract from existing key: $KEY_NAME"
+  echo "WARNING: For production, explicitly provide ACCOUNT_PUBKEY or set CREATE_KEY=true"
+  sleep 20
+
+  export KEYRING_BACKEND="test"
+  # Check if the key exists
+  if inferenced keys show "$KEY_NAME" --keyring-backend $KEYRING_BACKEND --keyring-dir /root/.inference >/dev/null 2>&1; then
```
DimaOrekhovPS, line 44: So technically this makes it backward compatible? we can still run the old way (tests and such). With time we'll update the tests to use this system

decentralized-api/Dockerfile
```diff
@@ -51,13 +51,23 @@ RUN --mount=type=cache,id=go-build-cache3,target=/root/.cache/go-build \
     go build -mod=readonly -tags muslc -ldflags "$LDFLAGS" \
     -o ./build/dapi \
     && patchelf --set-rpath '$ORIGIN:$ORIGIN/current:/root/.dapi/cosmovisor/current' ./build/dapi
+
+# TO FIX: temporary to have both binaries in api image
```
DimaOrekhovPS, line 55: Why? Because init-docker for API now tries to get the key?

inference-chain/cmd/inferenced/cmd/register_participant_command.go
```diff
@@ -0,0 +1,202 @@
...
+	pubKeyBytes, err := base64.StdEncoding.DecodeString(pubKeyBase64)
+	if err != nil {
+		return "", fmt.Errorf("failed to decode public key: %w", err)
+	}
+
+	pubKey := &secp256k1.PubKey{Key: pubKeyBytes}
+	return sdk.AccAddress(pubKey.Address()).String(), nil
+}
+
+func RegisterNewParticipantCommand() *cobra.Command {
```
DimaOrekhovPS, line 43: Will it replace the old command for participant registration or is it something different?

Before each node was automatically registering itself on app startup

DimaOrekhovPS (approved):

decentralized-api/apiconfig/config_manager.go
gmorgachev: it's not saved into config yaml

decentralized-api/cosmosclient/cosmosclient.go
gmorgachev: Yes

decentralized-api/scripts/init-docker.sh
gmorgachev: Yes

decentralized-api/Dockerfile
gmorgachev: First - yes, second - in onboarding we create warm key inside the `api` node and there is no reason to share this keys with `node` container too

inference-chain/cmd/inferenced/cmd/register_participant_command.go
gmorgachev: We registered participant with `func registerJoiningParticipant(`, not with cli command

If you're talking about `submit-new-unfunded-participant` - it has to be signed by existing account and node executed by seed node. RegisterNewParticipantCommand - just sending POST request and wait for result => transaction signed by seed node

PR #250, opened 2025-08-04
Verify participants by joining

0xBECEDA:
This PR implements verification of the active participants set starting from a given epoch (by default, the current epoch) until the specified app hash (or genesis app hash) is reached.

The verification mechanism is triggered if an app hash is provided and DAPI has detected a gap between blocks, as happens during synchronization with snapshots.

To support this mechanism, the following were added:

1. On-chain storage of block, validator signature proofs and merkle proofs for blocks where an active participant set was created. The block proof includes the block’s app_hash_hex and information about the validators who signed it (without the actual signatures). The validators proof includes the block_id with hashes and detailed signature information. Merkle tree proof proofs that given active participants set was in state in given block. The verification mechanism itself is implemented [here](https://github.com/gonka-ai/gonka-utils/pull/1)

2. Adding active participants for epoch 0 (without fully creating the epoch). Without it verification wouldn't be full. 

3. A debug endpoint to retrieve block info (block_proof + raw cosmos block).

### DIFF
--- a/inference-chain/x/inference/keeper/dynamic_pricing_test.go
+++ b/inference-chain/x/inference/keeper/dynamic_pricing_test.go
@@ -580,8 +580,7 @@ func TestDynamicPricingWithRealStats(t *testing.T) {
 		t.Logf("Querying stats from %d to %d (window: %d seconds)", timeWindowStart, timeWindowEnd, windowDuration)
 
 		// This is the same call UpdateDynamicPricing() makes
-		statsMap, err := k.GetSummaryByModelAndTime(goCtx, timeWindowStart, timeWindowEnd)
-		assert.NoError(t, err)
+		statsMap := k.GetSummaryByModelAndTime(goCtx, timeWindowStart, timeWindowEnd)
 		if len(statsMap) == 0 {
 			t.Logf("GetSummaryByModelAndTime returned no stats")
 			// If stats system isn't fully set up, fall back to testing the calculation logic


decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -48,9 +48,8 @@ type EventListener struct {
 	phaseTracker        *chainphase.ChainPhaseTracker
 	dispatcher          *OnNewBlockDispatcher
 	cancelFunc          context.CancelFunc
-
-	eventHandlers []EventHandler
-
+	eventHandlers       []EventHandler
+	// earliestCollectedBlockHeight int64
```
patimen: Remove?

decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -346,6 +346,116 @@ func (el *EventListener) handleBLSEvents(event *chainevents.JSONRPCResponse, wor
 	}
 }
 
+// TODO setup retry retrieving genesis block till success
```
patimen: These are in the `new_block_dispatcher.go`, right? Can we remove the commented code now?

inference-chain/x/inference/keeper/activeparticipants.go
```diff
@@ -41,3 +41,31 @@ func (k Keeper) SetActiveParticipants(ctx context.Context, participants types.Ac
 	b := k.cdc.MustMarshal(&participants)
 	store.Set(key, b)
 }
+
+func (k Keeper) SetActiveParticipantsProof(ctx context.Context, proof types.ProofOps, blockHeight uint64) {
```
patimen: We should use the new `collections` API for any new data we are adding.
There is documentation for it in `./proposals/cleanup_1/collections_quickstart.md`

inference-chain/x/inference/module/module.go
```diff
@@ -26,6 +29,7 @@ import (
 	"github.com/productscience/inference/x/inference/epochgroup"
 	"github.com/shopspring/decimal"
 	"github.com/spf13/cobra"
+	"strings"
```
patimen: I don't think this is ordered correctly... "strings" should be at the top of the imports in the standard library block.

inference-chain/x/inference/keeper/block_proof.go
```diff
@@ -0,0 +1,75 @@
...
+
+func pendingProofKey(height uint64) []byte {
+	var key []byte
+	key = append(key, []byte(pendingProofKeyPrefix)...)
+	key = append(key, sdk.Uint64ToBigEndian(height)...)
+	return key
+}
+
+func (k Keeper) SetBlockProof(ctx context.Context, proof types.BlockProof) error {
+	store := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
```
patimen: Again, this should use Collections, not the direct KVStore interactions.

patimen (changes requested):
We need two changes:
1. I have checked in a bunch of comments that start with `// PANIC:` These are all places inside `BeginBlock` that _could_ panic. Most of them are very unlikely. However, ANY panic during BeginBlock will take down the chain and must be avoided (I'm working on fixing this for EndBlock in another PR).
2. We have switched over to using `collections` in our chain vs the older style. There's still a few lingering, but I want new code to use the new pattern. One exception applies: For now, `active_participants` need to be the old style because we need an easy way to share the key to look up the merkle proofs. I believe most of the new data here could use `collections`.

PR #254, opened 2025-08-05
Early Network Protection Through Power Distribution Limits

libermans:
# Early Network Protection Through Power Distribution Limits

## Summary
Implements dual-layer protection system to prevent centralization attacks and ensure network stability during the critical bootstrap period of the Gonka network.

## Key Changes
- **Universal Power Capping**: Applies mathematical optimization to limit any participant's power to 30% maximum across all epoch calculations
- **Genesis Validator Enhancement**: Provides temporary developer veto authority (34% power) during network immaturity phases
- **Automatic Transition**: Protection systems automatically deactivate when network reaches 10M total power threshold

## Protection Mechanisms
1. **Power Concentration Prevention**: Sorting-based algorithm caps individual power at 30% while preserving total network power
2. **Early Network Security**: First genesis validator receives enhanced staking power (52% of others' total) for governance veto capability
3. **Network Maturity Detection**: Automatic protection removal when economic security thresholds are met

## Implementation Areas
- Enhanced genesis parameters in protobuf definitions and default configurations
- Power capping integration with epoch power calculations after weight computation
- Genesis enhancement integration with staking power modifications before validator updates
- Comprehensive test coverage for mathematical algorithms and integration workflows

## Benefits
- Prevents 67% supermajority attacks through mathematical power distribution limits
- Provides developer oversight during vulnerable network phases while preserving long-term decentralization
- Maintains network liveness and fault tolerance by preventing single points of failure
- Offers configurable protection levels (conservative, moderate, aggressive) for different deployment scenarios

This implementation addresses critical security vulnerabilities during network bootstrap while ensuring smooth transition to full decentralization as the network matures.


inference-chain/x/inference/module/genesis_enhancement.go
```diff
@@ -0,0 +1,169 @@
...
+	calculatedTotal := int64(0)
+	for _, result := range enhanced {
+		if result.Power < 0 {
+			return fmt.Errorf("negative power detected for validator %s: %d", result.OperatorAddress, result.Power)
+		}
+		calculatedTotal += result.Power
+	}
+
+	// Verify total power matches
+	if calculatedTotal != enhancedTotalPower {
```
gmorgachev: why are we doing that? we're checking exactly the same sum we did during enhancement

inference-chain/x/inference/module/power_capping.go
```diff
@@ -0,0 +1,238 @@
...
+			if denominator <= 0 {
+				cap = currentPower
+				break
+			}
+			cap = int64(numerator / denominator)
+			break
+		}
+
+		// Update running sum for next iteration
+		sumPrev += currentPower
```
gmorgachev: i have a concern about making this operations in float. can it lead to different cap? but not sure if problem might happen in reality

gmorgachev: Especially about ` float64(currentPower) > maxPercentageFloat * float64(weightedTotal)`

inference-chain/x/inference/module/module.go
```diff
@@ -593,6 +600,62 @@ func (am AppModule) moveUpcomingToEffectiveGroup(ctx context.Context, blockHeigh
...
+	result := ApplyGenesisEnhancement(ctx, am.keeper, computeResults)
+
+	// Log enhancement application results
+	originalTotal := int64(0)
+	for _, cr := range computeResults {
+		originalTotal += cr.Power
+	}
+
+	if result.WasEnhanced {
+		firstGenesisValidatorKey := am.keeper.GetFirstGenesisValidatorAddress(ctx)
```
gmorgachev: we should support 3 first validators

gmorgachev (changes requested):
I checked that during the executor sample and inference validation we use capped weight

inference-chain/x/inference/module/module.go
libermans: Added support of multiple validators

inference-chain/x/inference/module/power_capping.go
libermans: changed to Decimal

gmorgachev (approved):

PR #259, opened 2025-08-08
Make testermint work with warm/cold keys

patimen:
Overall: genesis works on a single `test` key, joins work with `file` keys, plus warm/cold
1. Add some env variables to the `join` compose file
2. Add the ability to inject passwords to the docker CLI (for file key operations)
3. Change around the init so that we launch genesis, then for each pair, we launch the chain-node, use that to setup keys/file keyring, then use that to set env variables for the api before launching
4. Update the entire API to use explicit warm or cold keys.
5. Switch from using API to send messages to sending messages straight using the CLI


testermint/src/main/kotlin/ApplicationCLI.kt
```diff
@@ -303,20 +350,55 @@ data class ApplicationCLI(
     }
 
     // New function that allows using TypeToken for proper deserialization of generic types
-    private fun <T> execAndParseWithType(typeToken: TypeToken<T>, args: List<String>): T {
+    private fun <T> execAndParseWithType(typeToken: TypeToken<T>, args: List<String>, password:String? = null): T {
         val argsWithJson = (listOf(config.execName) + args + "--output" + "json")
         Logger.debug("Executing command: {}", argsWithJson.joinToString(" "))
-        val response = exec(argsWithJson)
+        val response = exec(argsWithJson, password + "\n")
```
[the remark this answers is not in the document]
patimen: Good catch!

local-test-net/docker-compose.join.yml
```diff
@@ -10,4 +10,8 @@ services:
   api:
     environment:
       - DAPI_CHAIN_NODE__SEED_API_URL=${SEED_API_URL}
-      - CREATE_KEY=${CREATE_KEY:-true}
\ No newline at end of file
+      - CREATE_KEY=${CREATE_KEY:-true}
```
gmorgachev, line 13: probably we should have `false` everywhere

testermint/src/main/kotlin/DockerGroup.kt
```diff
@@ -435,6 +554,7 @@ class Consumer(val name: String, val pair: LocalInferencePair, val address: Stri
                 listOf()
             )
             cli.createContainer(doNotStartChain = true)
+            // PRTODO: This needs to use the file? Or override the test
```
gmorgachev, line 557: Is it temporary developer? If yes, it's probably not importnat which keyring is used

gmorgachev (approved):
From high level review - looks correct

PR #262, opened 2025-08-11
Liquidity Pool & Wrapped Token Update

GLiberman:


decentralized-api/internal/server/admin/bridge_handlers.go
```diff
@@ -0,0 +1,57 @@
...
+	"strings"
+
+	pserver "decentralized-api/internal/server/public"
+
+	"github.com/labstack/echo/v4"
+)
+
+// postBridgeBlock handles POST requests to submit finalized blocks with optional receipts
+func (s *Server) postBridgeBlock(c echo.Context) error {
+	// Debug: Log raw request body
```
DimaOrekhovPS, line 15: ?

inference-chain/proto/inference/inference/liquidity_pool.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+package inference.inference;
+
+option go_package = "github.com/productscience/inference/x/inference/types";
+
+message LiquidityPool {
```
DimaOrekhovPS, line 6: Here we don't need chain id?

inference-chain/x/inference/keeper/bridge.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Key prefixes for bridge storage
+const (
+	BridgeAddressKeyPrefix = "bridge_address:"
+)
+
+// Bridge address management functions
+
+// generateBridgeAddressKey creates a unique key for bridge addresses
+func (k Keeper) generateBridgeAddressKey(ctx context.Context, chainId string) string {
```
DimaOrekhovPS, line 20: minor thing, but the word `generate` feels like there's some randomness happening inside or something is being created. While it seems to be a deterministic key. So it's more like `get`

Anyway, I don't think we should spend time renaming it

inference-chain/x/inference/keeper/bridge.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+
+	"cosmossdk.io/store/prefix"
+	"github.com/cosmos/cosmos-sdk/runtime"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// Key prefixes for bridge storage
+const (
+	BridgeAddressKeyPrefix = "bridge_address:"
```
DimaOrekhovPS: I'm not sure what our current consensus on how keys should look like
Are we using `:` as a key separator anywhere else?

inference-chain/x/inference/keeper/msg_server_bridge_exchange.go
```diff
@@ -66,88 +51,128 @@ func (k msgServer) BridgeExchange(goCtx context.Context, msg *types.MsgBridgeExc
...
 
-	isActive := false
-	for _, participant := range activeParticipants.Participants {
-		// Get the account associated with this participant
-		participantAddr, err := sdk.AccAddressFromBech32(participant.Index)
+	// Check if this transaction has already been processed
+	existingTx, found := k.GetBridgeTransaction(ctx, msg.OriginChain, msg.BlockNumber, msg.ReceiptIndex)
+	if found {
+		// Get the epoch group for the existing transaction using pocStartBlockHeight
+		epochGroup, err := k.GetEpochGroup(goCtx, existingTx.PocStartBlockHeight, "")
```
DimaOrekhovPS: The alternative approach here would be to save `EpochId` in `existingTx`. And then give `epochId` retrieve the list of validators

inference-chain/x/inference/keeper/msg_server_register_liquidity_pool.go
```diff
@@ -0,0 +1,139 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// RegisterLiquidityPool handles the registration of a singleton liquidity pool.
+// This operation instantiates a new contract and registers it atomically.
+// Only the authorized governance account can perform this registration.
+func (k msgServer) RegisterLiquidityPool(goCtx context.Context, msg *types.MsgRegisterLiquidityPool) (*types.MsgRegisterLiquidityPoolResponse, error) {
+	// Validate authority explicitly and consistently
+	expectedAuthority := k.GetAuthority()
```
DimaOrekhovPS, line 22: Does it mean that the chain can collectively vote and change the liquidity pool?

DimaOrekhovPS (approved):

PR #309, opened 2025-08-29
Fix possible panics for negative coins

### LINKED ISSUE #313: [P0] Negative balance panic

patimen:
### Motivation
Panics in our EndBlock Go code will cause a consensus failure, halting the change and requiring a difficult chain reboot.

Even the most unlikely panic needs to be prevented. This PR focuses on possible panics that come from situations where there are negative coin balances at various points. While these negative balances should not happen, there are some edge cases where they might, and a consensus failure if far too risky to ignore.

Additionally, we use `uint64` in places inside our code, which could present dangerous roll-over errors when the value might go negative. We should start removing such code. 

### Solution

`sdk.NewInt64Coin` WILL panic if it gets a negative value. 

- Add error checking for every instance where we use NewInt64Coin (minus a few in Genesis we can ignore) 
- Add a Safe method for logging subaccount transactions to reduce the (now) needed boilerplate 
- Move from using uint64 in our interior methods. In the future, we need to strictly use uint64 at the edges (messages, apis) only. The risk of 0 - 1 = uint64.Max is real, and Go idioms avoid uints in general.

Of course, in most cases a negative balance is a serious problem that needs to be addressed, but causing a consensus failure needs to be impossible.

### DIFF
--- a/inference-chain/x/inference/types/settle_amount.go
+++ b/inference-chain/x/inference/types/settle_amount.go
@@ -1,5 +1,11 @@
 package types
 
-func (sa *SettleAmount) GetTotalCoins() uint64 {
-	return sa.RewardCoins + sa.WorkCoins
+import "math"
+
+func (sa *SettleAmount) GetTotalCoins() int64 {
+	sum := sa.RewardCoins + sa.WorkCoins
+	if sum > math.MaxInt64 {
+		return math.MaxInt64
+	}
+	return int64(sum)
 }
--- a/decentralized-api/cosmosclient/governance.go
+++ b/decentralized-api/cosmosclient/governance.go
@@ -16,9 +16,11 @@ type ProposalData struct {
 }
 
 func SubmitProposal(cosmosClient CosmosMessageClient, msg sdk.Msg, proposalData *ProposalData) error {
+	// cannot return err with fixed amount
+	coins, _ := types.GetCoins(100000000)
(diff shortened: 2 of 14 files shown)


inference-chain/x/inference/types/settle_amount.go
```diff
@@ -1,5 +1,11 @@
 package types
 
-func (sa *SettleAmount) GetTotalCoins() uint64 {
-	return sa.RewardCoins + sa.WorkCoins
+import "math"
+
+func (sa *SettleAmount) GetTotalCoins() int64 {
+	sum := sa.RewardCoins + sa.WorkCoins
+	if sum > math.MaxInt64 {
```
DimaOrekhovPS, line 7: what about this copilot comment?

DimaOrekhovPS (approved):
I didn't get all the type changes (uint64 > int64)
I guess we're still using uint64 at the storage level, but want to migrate to int64 at the code level?

inference-chain/x/inference/types/settle_amount.go
patimen: Discussed it with Gleb... the thing is that if we are dealing with numbers that could overflow uint64 or even int64, we're dealing with at least 9 times more that the total gonka ever intended to be created.
Plus, we have addition all over the place... shall we add checks everywhere? 
Even if the answer is yes, this is an arbitrary and confusing place to start.

PR #345, opened 2025-09-05
Fixes for validations, especially late (post-settle)

patimen:
1. Invalid payloads for an Inference will fail validation.
2. Validators with no nodes for the model will not submit a validation.
3. validations that are submitted after the settle will not get a share of the work, but will still count for inferences validated.
4. invalidations voted on and submitted after the settle will not be refunded.
5. Mock server for tests can now return an error at a validation attempt, allowing us to delay validations until after settle to test the above changes.
6. Support for direct submission of MsgValidation and MsgClaimRewards in tests
7. Support for retrieving and serializing app-config.yaml from API nodes in tests. Main purpose is to get Seed values for MsgClaimRewards.
8. Enabling some validation tests now. Some remain unstable


inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -52,6 +53,13 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 		return nil, types.ErrParticipantCannotValidateOwnInference
 	}
 
+	for _, validator := range inference.ValidatedBy {
```
DimaOrekhovPS, line 56: Great, it solves some of the questions I had in the missed validations PR

inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -146,6 +133,85 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 	return &types.MsgValidationResponse{}, nil
 }
 
+func (k msgServer) inferenceIsBeforeClaimsSet(ctx sdk.Context, inference types.Inference, group *epochgroup.EpochGroup) (bool, string) {
```
DimaOrekhovPS, line 136: I guess the most precise way to determine is to check every participant involved in this inference and see whether they claimed their rewards or not, but it is overcomplicated and unneeded. So this approach should work as well

DimaOrekhovPS (approved):

gmorgachev (approved):

PR #353, opened 2025-09-12
Missed validations recovery system

### LINKED ISSUE #567: Missed validations recovery system

patimen:
## Problem

During an epoch, a participant may legitimately miss validating some inferences due to network instability, hardware changes, or other temporary issues. Currently, once accounts are settled at epoch transition, missed validations cannot be recovered. This leads to:

* Gaps in inference validation coverage.
* Potentially lower reputation and compute credit for participants.
* Inconsistent incentives between those who missed validations for legitimate reasons and those who didn’t.
* A risk that some invalid inferences remain undetected if they were missed by validators.

---

## Solution

Introduce a recovery mechanism that allows participants to “catch up” on missed validations **after account settlement but before claiming rewards**. The key aspects are:

1. **Automatic Detection of Missed Validations**

   * Each participant node queries for any inferences it should have validated but did not.
   * These missed inferences are filtered against the models available to the participant.

2. **Recovery Validations Before Claim**

   * Any missed validations are executed and submitted prior to submitting the claim transaction.
   * This ensures participants remain consistent with the validation process, even if late.
   
 2. **Validations Retries**
  
  - Validation failures will now be retried several times before giving up, in an attempt to reduce the need for missed validations work.

3. **Modified Treatment of Recovery Validations**

   * Participants receive **validation credit** (reputation / proof of compute) but **no shared coin for work**, since payment has already been settled. They do still receive earned Bitcoin style rewards.
   * If a late validation identifies an invalid inference:

     - Invalidation verification and voting still happens on the network
     * If invalid, the inference submitter’s reputation is still penalized.
     * However, the requester does **not** receive a refund, since funds have already been distributed and cannot be clawed back. We may revisit this in the future for finding a way to claw back funds from the invalid executor.

4. **System Safeguards**

   * Ensure that no negative coin balances during settlement can cause a panic, preventing consensus-breaking chain halts. (While this PR will not cause such situations, testing revealed this possible error which needed to be addressed for chain stability)

---

## Implementation

* **Validation Logic Refactor**

  * Extract reusable validation logic into `shouldValidateInference()`.
* **Missed Validation Detection**

  * Implement `DetectMissedValidations()` to identify per-epoch gaps in validation.
* **Execution of Recovery Validations**

  * Add `ExecuteRecoveryValidations()` with parallel goroutine execution for efficiency.
* **Integration with Epoch Transition**

  * Trigger recovery validations at `IsSetNewValidatorsStage`, before claim submission.
  * Run background processing to minimize delays.
* **Infrastructure Reuse & Logging**

  * Use existing validation infrastructure for consistency.
  * Add logging for monitoring and debugging missed validation recovery.


decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -344,8 +350,34 @@ func (d *OnNewBlockDispatcher) handlePhaseTransitions(epochState chainphase.Epoc
...
+
+		// Get the previous epoch seed for validation recovery
+		previousSeed := d.configManager.GetPreviousSeed()
+
+		// Calculate previous epoch index
+		expectedPreviousEpochIndex := epochContext.EpochIndex - 1
+
+		// Verify the seed is from the correct epoch
+		if previousSeed.EpochIndex != expectedPreviousEpochIndex {
+			logging.Warn("Previous seed epoch mismatch for recovery", types.Validation,
```
DimaOrekhovPS, line 363: Shouldn't we skip claiming then?

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -422,6 +454,69 @@ func getCommandForPhase(phaseInfo chainphase.EpochState) (broker.Command, *chan
...
+	}
+
+	// Check for genesis epoch
+	if previousEpochIndex == 0 && previousSeed.EpochIndex == 0 {
+		logging.Info("Missed validation recovery skipped: genesis epoch", types.Validation, "previousEpochIndex", previousEpochIndex)
+		return
+	}
+
+	// Check if seed is valid
+	if previousSeed.Seed == 0 {
```
DimaOrekhovPS, line 474: We create seed with `newSeed := rand.Int63()`, the docs of `rand.Int63` say that it return a non-negative number, this making `0` a valid value.

We should either tweak `createNewSeed` so the seed is never 0 or change how we check here

decentralized-api/internal/event_listener/new_block_dispatcher.go
DimaOrekhovPS: ok, I see that it happens downstream, but shouldn't we then remove this check? And only track it and log downstream

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -422,6 +454,69 @@ func getCommandForPhase(phaseInfo chainphase.EpochState) (broker.Command, *chan
...
+	}
+
+	logging.Info("Found missed validations, executing recovery", types.Validation,
+		"previousEpochIndex", previousEpochIndex,
+		"missedCount", len(missedInferences))
+
+	// Execute recovery validations
+	d.validator.ExecuteRecoveryValidations(missedInferences)
+
+	time.Sleep(4 * time.Minute)
```
DimaOrekhovPS: I guess we're ok with this hack for now

decentralized-api/internal/server/admin/validation_recovery_handler.go
```diff
@@ -0,0 +1,130 @@
...
+		"missedCount", missedCount)
+
+	// Execute recovery validations
+	if missedCount > 0 {
+		s.validator.ExecuteRecoveryValidations(missedInferences)
+		logging.Info("Manual recovery validations completed", types.Validation,
+			"epochIndex", epochIndex,
+			"recoveredCount", missedCount)
+
+		time.Sleep(4 * time.Minute)
```
DimaOrekhovPS: We should document that the http timeout for this request should be 4mins+

decentralized-api/internal/server/admin/validation_recovery_handler.go
```diff
@@ -0,0 +1,130 @@
...
+	Success           bool   `json:"success"`
+	Message           string `json:"message"`
+	EpochIndex        uint64 `json:"epoch_index"`
+	Seed              int64  `json:"seed"`
+	MissedValidations int    `json:"missed_validations"`
+	AlreadyClaimed    bool   `json:"already_claimed"`
+	ClaimExecuted     bool   `json:"claim_executed"`
+}
+
+func (s *Server) postClaimRewardRecover(ctx echo.Context) error {
```
DimaOrekhovPS, line 29: Miners should be careful with this endpoint.

We prevent double claims with `alreadyClaimed`, but I'm worried that parallel multiple invocations of this endpoint could lead to duplicate validation tasks, which in turn may lead to multiple validation messages?

Is it possible we start multiple votings for the same inference or smth like that 
@patimen what do you think?

DimaOrekhovPS: I see some guardrails in `DetectMissedValidations`, so probably we should be fine

DimaOrekhovPS (changes requested):
Let's resolve comments and understand what needs to be addressed and what doesn't or can be postponed

decentralized-api/internal/event_listener/new_block_dispatcher.go
gmorgachev: That's just check if config is invalid, no?

decentralized-api/internal/event_listener/new_block_dispatcher.go
patimen: Yeah, this bothered me too. It's not blocking anything (in a separate goroutine). But still fragile.

decentralized-api/internal/event_listener/new_block_dispatcher.go
patimen: We should fix the seed generation to:
1. Use a crypto random generator (not math/rand)
2. Exclude 0.

gmorgachev: tha tmake sense but let's exclude from this PR

decentralized-api/internal/server/admin/validation_recovery_handler.go
DimaOrekhovPS: I see some of my concerns are addressed in PR#345

gmorgachev (approved):

PR #356, opened 2025-09-15
Certik Audit fixes (straightforward changes)

### LINKED ISSUE #342: [P0] Security: Minor
- [x] GOC-19 | Count Participant With Zero Balance Is Missing Balance Check
- [x] GOC-20 | Go Package Dependency Issues
- [x] GOC-22 | Inconsistent Epoch ID Handling In `GetPreviousEpochMLNodesWithInferenceAllocation`
- [x] GOC-29 | New Participants Initialized As ACTIVE Instead Of RAMPING
- [x] GOC-24 | Misleading Function Name In `storeMLNodeInfo`
- [x] GOC-25 | Misleading Log In Function `SetComputeValidators`
- [x] GOC-32 | Missing Error Handling For `GetPreservedNodesByParticipant`
- [x] GOC-33 | Println Usage In Production Code
- [x] GOC-34 | Var Can Be Declared As Constant
- [x] GOC-35 | Unhandled Error In Function `submitValidationProposals`
- [x] GOC-37 | Unused Variable
- [x] GOC-38 | Incorrect Comments
- [x] GOC-06 | Redundant SetParticipant In `handleExpiredInference`
- [x] GOC-36 | Redundant Maps And Double-Pass Iteration In `SetComputeValidators`
- [x] GOI-15 | Potential Division By Zero Leads To Panic
- [x] GOI-20 | Incorrect Remainder Distribution In Legacy Weight Allocation
- [x] GOI-05 | Stale Object Used For Completion Check In `FinishInference`
- [x] GOI-19 | Missing `nil` Check For `GetPubKey()` Could Lead To Panic
- [x] GOI-29 | Interface Mismatch In `RunMembershipService` Vs `RunManager`
- [x] GOI-07 | Constants Declaration Should Be Grouped At The Top Of The File
- [x] GOI-08 | Inconsistent Constant Declaration Pattern
- [x] GOI-31 | Unused Errors
- [x] GOI-33 | Duplicated Constant `DefaultMaxTokens`
- [x] GOI-34 | Incorrect Timeout Logging Value
- [x] GOI-35 | Redundant Boolean Comparison
- [x] GOI-09 | In `distributeLegacyWeight `, for each hardware node the code linearly scans newMLNodes to find an existing node with matching `NodeId `. This is O(H*N) and can degrade when many nodes exist.
- [x] GOI-10 | For each governance model, the code scans all `originalMLNodes` and `callsslices.Contains(supportedModelsByNode[mlNode.NodeId], model.Id) `to decide assignment. This is O(MNK) where K `ismodels` per node.

Ignore the following:

- [ ] GOC-10 | Subsidy Calculation Mints Total Payout Instead Of Reward Causing Systematic Over-Mint: Irrelevant given the switch to Bitcoin style rewards
- [ ] GOI-12 | Cryptographic Signature Collision In Sign/Verify Functions: we acknowledge that it's a good point but it's not a vulnerability. Going to fix it in the future
- [ ] GOC-09 | Reputation-Based Validation Reduction Undermines Byzantine Fault Tolerance: we're not lowering P to zero, only to 1% => we still have regular check (like every couple seconds under high load)

patimen:
These are mostly isolated or straightforward fixes for the Certik Audit. ALL issues addressed are in their own commit to this PR so they can be reviewed correctly.


inference-chain/x/inference/module/model_assignment.go
```diff
@@ -260,88 +259,99 @@ func (ma *ModelAssigner) distributeLegacyWeight(originalMLNodes []*types.MLNodeI
...
 		nodeId := hwNode.LocalId
 		distributedWeight := weightPerNode
-		if remainderCounter < remainderWeight {
-			distributedWeight++ // Give +1 to first remainderWeight nodes
-		}
 
-		if distributedWeight <= 0 {
+		if distributedWeight <= 0 && !remainder.RemainderLeft() {
 			continue
 		}
```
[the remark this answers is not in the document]
patimen: I'm pretty sure this is just inaccurate. this will continue (aka, not produce a MLNodeInfo) if there is BOTH nothing to distribute AND no remainders to distribute, which is what we want.

inference-chain/x/inference/keeper/msg_server_submit_new_participant.go
gmorgachev, line 1: Why default status is ramping?

decentralized-api/apiconfig/config_manager.go
```diff
@@ -214,7 +214,7 @@ func (cm *ConfigManager) SetLastUsedVersion(version string) error {
 func (cm *ConfigManager) ShouldRefreshClients() bool {
 	currentVersion := cm.GetCurrentNodeVersion()
 	lastUsedVersion := cm.GetLastUsedVersion()
-	return currentVersion != lastUsedVersion
+	return currentVersion != lastUsedVersion && lastUsedVersion != ""
```
gmorgachev: Why do we changing that?

gmorgachev (approved):

inference-chain/x/inference/keeper/msg_server_submit_new_participant.go
patimen: Because that's where you start... we recalculate it as you add history and get to ACTIVE. I think it is currently merely descriptive (not operative), but if we have it it should be accurate (as pointed out by Certik).

PR #358, opened 2025-09-16
Major Certik Audit fixes

### LINKED ISSUE #341: [P0] Security: Major
- [x] GOC-12. Unchecked PubKey–Address Binding Allows Account Squatting
- [x] GOC-13. Usage Of Must Marshaling Methods In `EndBlock ExecutedCode`
- [x] GOC-14. Unbounded Pruning At Epoch Boundaries Can Stall `EndBlock`
- [x] GOC-17. Unrestricted Status Manipulation In Validation GRPC Method
- [x] GOI-01. Lack Of Freshness Check Allows Signature Replay
- [x] GOI-24. Usage Of Must Marshaling Methods In `EndBlock ExecutedCode`

patimen:
These are fewer but more impactful than the "minor" audit fixes, and warrant more careful review and testing.
1. **Pubkey must match address** - make sure that the pubkey submitted for new unfunded participant matches the address given. (GOC-12)
2. **Panic avoidance** - We should NEVER panic during EndBlock. We may want to have more people look at the code tree from EndBlock to make sure we've caught ALL possible panics to avoid a consensus failure.
3. **Validate Timestamp for inferences on chain** - the chain itself needs to verify that timestamps are valid (not too old and not in the future) for inference signatures to prevent replay attacks. Note that this is _mostly_ handled by the InferenceId dedupe, but after pruning really old signatures could be reused. This check focuses on solving that problem by confirming the timestamp is not far in the future, so it allows for significantly older timestamps than the checks done by the API nodes.
4. **Pruning Rewrite** - Pruning was crude and could lead to out of memory issues or chain slow downs. This rewrite instead prunes a fixed amount each block until all pruning is done. It also introduces a generic `Pruner` type to handle the logic for different types of entities to prune, which can be reused for future pruning. Full details are in `proposals/pruning_v2`
5. **MsgValidation restrictions** - Invalidations are extremely costly to the network, and had zero restrictions. This implements a series of restrictions on the number of invalidations based on participant reputation and weight, as well as dropping the payload from the validation message.


inference-chain/proto/inference/inference/params.proto
```diff
@@ -187,4 +189,14 @@ message BandwidthLimitsParams {
...
+
+  // The total max simultaneous invalidations across the chain
+  uint64 invalidations_limit = 4;
+
+  // The seconds to look back to determine inference rate for invalidations
+  uint64 invalidations_sample_period = 5;
+
+  // The curve for the tanh function for determining how quickly traffic expands the
+  // max invalidations allowed
+  uint64 invalidations_limit_curve = 6;
```
DimaOrekhovPS, line 201: Is it somehow then converted to a float or smth? I just imagine a parameter of a function to be a real number

inference-chain/x/inference/calculations/maximum_invalidations.html
```diff
@@ -0,0 +1,183 @@
+<!DOCTYPE html>
```
DimaOrekhovPS, line 1: a UI tool to explore the function behavior?

inference-chain/proto/inference/inference/params.proto
DimaOrekhovPS: Found the actual formula, makes sense now

DimaOrekhovPS (approved):

inference-chain/x/inference/calculations/maximum_invalidations.html
patimen: Yeah, I created it in ChatGPT... it was very helpful in deciding defaults and understanding the algorithm.

PR #372, opened 2025-09-23
Empty nodeId

gmorgachev:
### Multiple Fixes and Security Improvements

The PR introduces multiple fixes and security improvements:

#### Missed `node_id` to PoCBatch (`inference-chain/x/inference/keeper/msg_server_submit_poc_batch.go`)

`node_id` is used to detect which node produced the nonce's batch

#### Remove legacy weight distribution from batches without `node_id`

`inference-chain/x/inference/module/model_assignment.go` distributed before batches without `node_id` between another nodes. As now all MLNodes returns `node_id` => that it not needed anymore

#### Remove MLNodes without HardwareNodes or models supported by Governance

`unassignedMLNodes` from `inference-chain/x/inference/module/model_assignment.go` are not counted in total weight anymore
Total weight of participant is recomputed after all filtering 

#### Statistical Validation for Missed inference and validation

Binom test `inference-chain/x/inference/calculations/stats.go` is now used for:   
- not pay reward if statistically signigicant > 10% of requests are missed (`inference-chain/x/inference/keeper/accountsettle.go`) 
- not pay claim if statistically signigicant > 10% of validations are missed (`inference-chain/x/inference/keeper/msg_server_claim_rewards.go`) instead of hard check for all validations

#### Set Participant status to `ACTIVE` for all ActiveParticipant when switch epoch

`inference-chain/x/inference/module/module.go:moveUpcomingToEffectiveGroup`

#### Fix counter for successful inferences

`inference-chain/x/inference/keeper/msg_server_start_inference.go`  
`inference-chain/x/inference/keeper/msg_server_finish_inference.go`  

#### Logging

### DIFF
--- a/inference-chain/x/inference/keeper/accountsettle.go
+++ b/inference-chain/x/inference/keeper/accountsettle.go
@@ -5,7 +5,9 @@ import (
 	"math"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/productscience/inference/x/inference/calculations"
 	"github.com/productscience/inference/x/inference/types"
+
 	"github.com/shopspring/decimal"
 )
 
@@ -85,6 +87,32 @@ func (k *Keeper) GetSettleParameters(ctx context.Context) *SettleParameters {
 		StageDecrease:            params.TokenomicsParams.SubsidyReductionAmount.ToFloat32(),
 		TotalSubsidySupply:       normalizedTotalSuply.Amount.Int64(),
 	}
+}
+
+func CheckAndPunishForDowntimeForParticipants(participants []types.Participant, rewards map[string]uint64) {
+	for _, participant := range participants {
+		rewards[participant.Address] = CheckAndPunishForDowntimeForParticipant(participant, rewards[participant.Address])
+	}
+}
+
+func CheckAndPunishForDowntimeForParticipant(participant types.Participant, reward uint64) uint64 {
+	totalRequests := participant.CurrentEpochStats.InferenceCount + participant.CurrentEpochStats.MissedRequests
+	missedRequests := participant.CurrentEpochStats.MissedRequests
+	return CheckAndPunishForDowntime(totalRequests, missedRequests, reward)
+}
+
+func CheckAndPunishForDowntime(total, missed, reward uint64) uint64 {
+	if total == 0 {
+		return reward
+	}
+	passed, err := calculations.MissedStatTest(int(missed), int(total))
+	if err != nil {
+		return reward
+	}
+	if !passed {
+		return 0
+	}
+	return reward
 }
 
 func (k *Keeper) SettleAccounts(ctx context.Context, currentEpochIndex uint64, previousEpochIndex uint64) error {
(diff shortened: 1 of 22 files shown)


inference-chain/x/inference/keeper/accountsettle.go
```diff
@@ -260,7 +288,8 @@ func getWorkTotals(participants []types.Participant) (int64, int64) {
 	for _, p := range participants {
 		// Do not count invalid participants work as "work", since it should not be part of the distributions
 		if p.CoinBalance > 0 && p.Status != types.ParticipantStatus_INVALID {
-			totalWork += p.CoinBalance
+			newCoinBalance := CheckAndPunishForDowntimeForParticipant(p, uint64(p.CoinBalance))
```
DimaOrekhovPS: Don't we want to zero out bitcoin reward as well in this case?

inference-chain/x/inference/keeper/bitcoin_rewards.go
```diff
@@ -239,6 +239,8 @@ func CalculateParticipantBitcoinRewards(participants []types.Participant, epochG
 		totalPoCWeight += pocWeight
 	}
 
+	CheckAndPunishForDowntimeForParticipants(participants, participantWeights)
```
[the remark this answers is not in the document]
DimaOrekhovPS: Is it relevant?

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -155,38 +155,44 @@ func (k msgServer) validateClaim(ctx sdk.Context, msg *types.MsgClaimRewards, se
...
+			missed++
 		}
 	}
 
-	return false, nil
+	passed, err := calculations.MissedStatTest(missed, total)
+	if err != nil {
+		return false, err
+	}
+	return !passed, nil
```
DimaOrekhovPS, line 198: A bit tricky to understand here

`passed` means "Yes, the participant missed the critical amount of activities" or vice versa?

And then we invert this result when returning from `checkMissedValidationsSignificance`. Would be nice to rename the variable and the function (e.g. `check...` > `hasSignificantMissedValidations`)

inference-chain/x/inference/keeper/accountsettle.go
gmorgachev: yes. but it's done in bitcoin_reward.go, no?

DimaOrekhovPS (approved):

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -126,6 +131,17 @@ func (k msgServer) handleInferenceCompleted(ctx sdk.Context, existingInference *
...
 	)
+
+	executedBy := existingInference.ExecutedBy
+	executor, found := k.GetParticipant(ctx, executedBy)
+	if !found {
+		k.LogError("handleInferenceCompleted: executor not found", types.Inferences, "executed_by", executedBy)
+	} else {
+		executor.CurrentEpochStats.InferenceCount++
+		executor.LastInferenceTime = existingInference.EndBlockTimestamp
+		k.SetParticipant(ctx, executor)
```
DimaOrekhovPS, line 142: So we were missing this before and the whole stats didn't work?

DimaOrekhovPS: We also increment `infereceCount` variable in `processInferencePayments`

PR #386, opened 2025-10-07
Fix tx timeout

### LINKED ISSUE #582: Fix tx timeout

0xBECEDA:
### Context:
We use unordered transactions. Each transaction has a TTL—the time window within which the transaction must be executed. Some transactions are send with retry: if transaction is expired, we send it again till find its result (negative or positive) on-chain. 

### Problem:
Before we were using timestamp from node's machine to create TTL and to verify if transaction TTL is expired. If chain slows down or even stops,  blockchain time is slowing down or “stopped” too. 

However, node clocks continue to advance. In this case, transaction TTLs often end up more than 10 minutes ahead of the current block time, which leads to their automatic rejection and endless re-sending. 

### Current solution:
- use latest block time to create and verify transactions TTL
- track if chain is slowing down or stops and do no send transacitons-with-retry in that case
- track ammount of attempts of retry (max 100)


decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -144,20 +159,31 @@ func (m *manager) Status(ctx context.Context) (*ctypes.ResultStatus, error) {
 func (m *manager) SendTransactionAsyncWithRetry(rawTx sdk.Msg) (*sdk.TxResponse, error) {
 	id := uuid.New().String()
 	logging.Debug("SendTransactionAsyncWithRetry: sending tx", types.Messages, "tx_id", id)
+
+	if halt, err := m.isChanHalt(); err != nil || halt {
+		logging.Error("chain is slowing down or couldn't fetch actual chain status", types.Messages, "latest_block_timestamp", m.blockTimeTracker.latestBlockTime.Load().(time.Time))
+
+		if err := m.putOnRetry(id, "", time.Time{}, rawTx, 0, false); err != nil {
```
gmorgachev, line 164: When will it retry in that case?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -342,9 +389,10 @@ func (m *manager) observeTxs() error {
 		}
 
 		if errors.Is(err, ErrTxNotFound) {
-			if time.Now().After(tx.Timeout) {
-				logging.Debug("tx expired", types.Messages, "tx_id", tx.Id, "tx_hash", tx.TxHash)
-				if err := m.putOnRetry(tx.Id, "", time.Time{}, rawTx, false); err != nil {
+			if m.blockTimeTracker.latestBlockTime.Load().(time.Time).After(tx.Timeout) {
```
gmorgachev, line 391: should we add some additional gap in this check? for example timeout + 10 sec. just in case if we're missing smth else

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -509,3 +565,36 @@ func (m *manager) getSignedBytes(ctx context.Context, id string, unsignedTx clie
 	}
 	return txBytes, timestamp, nil
 }
+
+func (m *manager) isChanHalt() (bool, error) {
```
patimen: `isChainHalt()`
Unless you mean "Is channel halt?"

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -99,8 +114,13 @@ func StartTxManager(
 		defaultTimeout:   defaultTimeout,
 		natsConnection:   natsConnection,
 		natsJetStream:    js,
+		blockTimeTracker: &blockTimeTracker{
+			latestBlockTime: ts,
+			maxBlockTimeout: 10 * time.Second,
+		},
 	}
 
+	m.isChanHalt()
```
patimen: What's the point of calling this here? Is there some side effect?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -204,6 +231,11 @@ func (m *manager) putOnRetry(
 		"sent", sent,
 	)
 
+	if attempts >= maxAttempts {
+		logging.Info("tx reached max attempts", types.Messages, "tx_id", id)
```
patimen: We should probably bump the log up to Warn

decentralized-api/cosmosclient/tx_manager/tx_manager.go
patimen: Actually, best to rename this "checkForChainHalt", which should at least give a hint that it's updating state. Or even "updateChainHalt". 
This is probably a Java thing... methods starting with "is" are nearly always idempotent wrappers around booleans, so this seems really awkward to me.

patimen (approved):
Maybe we can get a document/diagram of the queue/retry system?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
0xBECEDA: sendTxs will check every time when nats would deliver message if chain is alive - if yes, it would try to send it.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
0xBECEDA: We could, but I think it would be more confusing: we already have timeout and here we extend it +10 sec again.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
0xBECEDA: Shadow of previous implementation. 
Fixed

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -304,6 +344,10 @@ func (m *manager) sendTxs() error {
 func (m *manager) observeTxs() error {
 	logging.Info("Tx manager: observeTxs txs: run in background", types.Messages)
 	_, err := m.natsJetStream.Subscribe(server.TxsToObserveStream, func(msg *nats.Msg) {
+		if halt, err := m.updateChainHalt(); err != nil || halt {
+			logging.Error("chain is slowing down or couldn't fetch actual chain status", types.Messages, "latest_block_timestamp", m.blockTimeTracker.latestBlockTime.Load().(time.Time))
```
DimaOrekhovPS, line 348: So it just warns us, but doesn't do anything?

PR #390, opened 2025-10-09
Config management improvements

### LINKED ISSUE #315: [P0] Move config to DB (like seed, etc)

DimaOrekhovPS:


decentralized-api/broker/node_admin_commands.go
```diff
@@ -109,6 +109,88 @@ func (c RegisterNode) Execute(b *Broker) {
...
+	// Capture worker for client refresh
+	worker, _ := b.nodeWorkGroup.GetWorker(c.Node.Id)
+	b.mu.Unlock()
+
+	if worker != nil {
+		// Refresh client immediately so it picks up new URLs
+		oldVersion := b.configManager.GetLastUsedVersion()
+		newVersion := b.configManager.GetCurrentNodeVersion()
+		worker.RefreshClientImmediate(oldVersion, newVersion)
+	}
```
gmorgachev: Why to do this start/stop every time? It might be needed if IP / PORT is updated, which is easy to check. But version can't be updated by this endpoint as version is on chain 

Also it might be needed if models is the same but args changed 

But i think most probable scenario is just replacing one model with another. In that case we should not refresh client

DimaOrekhovPS: Deleted it

Do you think we need to keep `RefreshClientImmediate` if IP/PORT changes? There's a status check below that will kind of start the reconciliation process if needed

patimen (approved):
Approved after in person review.

PR #397, opened 2025-10-17
Experimental refactor of tx_manager

patimen:
Break up tx_manager into major components:
1. ChainTracker - tracks chain height and time
2. Broadcaster - signs and broadcasts messages
3. TransactionQueue - manages the queue for send/observing

Got rid of some unecessary code and flattened the heirarchy


patimen:
Right idea, but we would need to retry given current implementations.

PR #400, opened 2025-10-20
Verify particiapnts chain by joining

### LINKED ISSUE #330: [P1] Security MerkleTree Proofs; Merge participant validation till block0; Need to add signature check at recording

0xBECEDA:
This PR implements verification of the active participants set starting from a given epoch (by default, the current epoch) until the specified app hash (or genesis app hash) is reached.

The verification mechanism is triggered if an app hash is provided and DAPI has detected a gap between blocks, as happens during synchronization with snapshots.

To support this mechanism, the following were added:

1. On-chain storage of block, validator signature proofs and merkle proofs for blocks where an active participant set was created. The block proof includes the block’s app_hash_hex and information about the validators who signed it (without the actual signatures). The validators proof includes the block_id with hashes and detailed signature information. Merkle tree proof proofs that given active participants set was in state in given block. The verification mechanism itself is implemented [here](https://github.com/gonka-ai/gonka-utils/pull/1)

2. Adding active participants for epoch 0 (without fully creating the epoch). Without it verification wouldn't be full. 

3. A debug endpoint to retrieve block info (block_proof + raw cosmos block).


decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -391,6 +399,12 @@ func (icc *InferenceCosmosClient) SendTransactionAsyncNoRetry(msg sdk.Msg) (*sdk
 	return icc.manager.SendTransactionAsyncNoRetry(msg)
 }
 
+func (icc *InferenceCosmosClient) SubmitActiveParticipantsMissingProofs(tx *types.MsgSubmitActiveParticipantsProofData) error {
+	tx.Creator = icc.Address
+	_, err := icc.manager.SendTransactionAsyncNoRetry(tx)
```
patimen, line 404: If this fails, we can just re-run the catch-up?

decentralized-api/internal/event_listener/chainevents/block.go
```diff
@@ -0,0 +1,92 @@
...
+	Attributes []Attribute `json:"attributes"`
+}
+
+type Header struct {
+	AppHash            string        `json:"app_hash"`
+	ChainId            string        `json:"chain_id"`
+	ConsensusHash      string        `json:"consensus_hash"`
+	DataHash           string        `json:"data_hash"`
+	EvidenceHash       string        `json:"evidence_hash"`
+	Height             string        `json:"height"`
```
patimen: I think we can do better here for these "IntsAsStrings" that we are seeing from Cosmos. Currently we parse it in what appears to be at least 4 places, each with error handling and some code clutter. Might I suggest we add a new custom type and give it an unmarhsal?:
```
type StringInt64 int64

func (s *StringInt64) UnmarshalJSON(data []byte) error {
    // Remove quotes if present
    var str string
    if err := json.Unmarshal(data, &str); err == nil {
        // Parse string to int64
        v, err := strconv.ParseInt(str, 10, 64)
        if err != nil {
            return err
        }
        *s = StringInt64(v)
        return nil
    }
    // Fallback: try to unmarshal directly as int64
    var v int64
    if err := json.Unmarshal(data, &v); err != nil {
        return err
    }
    *s = StringInt64(v)
    return nil
}
```

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -244,19 +276,93 @@ func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo ch
...
+		if err != nil {
+			panic(err)
+		}
+
+		d.lastVerifiedAppHashHex = data.BlockProof.AppHashHex
+		logging.Info("verify participants successfully", types.Stages, "new_app_hash", d.lastVerifiedAppHashHex)
+	}
+}
+
+// Creates proof for genesis block. Is called only ones and only on genesis node
```
patimen: ones=>once

inference-chain/proto/inference/inference/query.proto
```diff
@@ -373,7 +374,21 @@ service Query {
...
+
+  rpc GetBlockProofByHeight (QueryBlockProofRequest) returns (QueryBlockProofResponse) {
+    option (google.api.http).get = "/productscience/inference/inference/block_proof/{proof_height}";
+  }
+
+  rpc GetParticipantsProofByHeight (QueryGetParticipantsProofRequest) returns (QueryGetParticipantsProofResponse) {
+    option (google.api.http).get = "/productscience/inference/inference/validators_proof/{proof_height}";
+  }
+
+  rpc IfProofPending(QueryIsProofPendingRequest) returns (QueryIsProofPendingResponse){
```
patimen: IsProofPending?

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -244,19 +276,93 @@ func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo ch
...
+// verifyParticipantsChain verifies participants from current epoch to genesis epoch
+// (or to latest known verified epoch) backwards using proof stored on-chain
+// runs if verification is on AND if blocks coming with gaps (1.. 4... 8.. etc)
+// It panics if verification failed.
+func (d *OnNewBlockDispatcher) verifyParticipantsChain(ctx context.Context, curHeight int64, knownHeight int64) {
+	logging.Debug("verify participants", types.ParticipantsVerification, "known_height", knownHeight, "current_height", curHeight, "last_verified_app_hash", d.lastVerifiedAppHashHex)
+
+	// verification is on AND there is gap between latest processed block and current block
+	if d.lastVerifiedAppHashHex != "" && knownHeight != curHeight-1 {
+		logging.Info("verify participants: start", types.ParticipantsVerification, "lastVerifiedAppHashHex", d.lastVerifiedAppHashHex)
```
patimen, line 294: Thanks for making a new logging tag for this.

inference-chain/app/upgrades.go
```diff
@@ -4,9 +4,8 @@ package app
 
 import (
 	"context"
-	"fmt"
-
 	upgradetypes "cosmossdk.io/x/upgrade/types"
+	"fmt"
```
patimen, line 8: "fmt" should not be moved down here... the canonical Go format is built ins at the top of the `import`... can you check your linter/format settings?

inference-chain/app/upgrades/v0_2_5/upgrades.go
```diff
@@ -0,0 +1,180 @@
...
+	return func(ctx context.Context, plan upgradetypes.Plan, vm module.VersionMap) (module.VersionMap, error) {
+
+		k.LogInfo(fmt.Sprintf("%s - Starting Participants verification upgrade", UpgradeName), types.Upgrades)
+
+		for moduleName, version := range vm {
+			fmt.Printf("Module: %s, Version: %d\n", moduleName, version)
+		}
+		fmt.Printf("OrderMigrations: %v\n", mm.OrderMigrations)
+
+		configurator.RegisterMigration(types.ModuleName, 4, func(ctx sdk.Context) error {
```
patimen: If 4 the right number here? Aren't we at a higher version?

inference-chain/app/upgrades/v0_2_5/upgrades.go
```diff
@@ -0,0 +1,180 @@
...
+					continue
+				}
+
+				participantAddr, err := sdk.AccAddressFromBech32(participant.Address)
+				if err != nil {
+					k.LogError("error getting participant address from string", types.Upgrades, "err", err, "participant", participant.Address)
+					continue
+				}
+
+				granteesResp, err := k.GranteesByMessageType(ctx, &types.QueryGranteesByMessageTypeRequest{
```
patimen, line 148: I don't quite get what's going on here, can you add some comments?

inference-chain/proto/inference/inference/block_proof.proto
```diff
@@ -0,0 +1,80 @@
...
+message CommitInfo {
+  string validator_address = 1;
+  string validator_pub_key = 2;
+}
+
+message BlockProof {
+  int64 created_at_block_height = 1; // H
+  string app_hash_hex           = 2; // Header(H).AppHash (from Header(H+1))
+  int64 total_voting_power      = 3;
+  uint64 epoch_id = 4;
```
patimen: epoch_index (0 based index for each Epoch)

inference-chain/proto/inference/inference/tx.proto
```diff
@@ -333,7 +337,7 @@ message MsgSetBarrierResponse {
 message MsgJoinTrainingStatus {
   option (cosmos.msg.v1.signer) = "creator";
   string              creator = 1;
-  JoinTrainingRequest req = 2;
+  JoinTrainingRequest req     = 2;
```
patimen: Any chance we could revert this and all the above formatting changes? Makes the PR harder to understand.

inference-chain/x/inference/types/keys.go
```diff
@@ -44,6 +44,10 @@ var (
 	PruningStatePrefix               = collections.NewPrefix(21)
 	InferencesToPrunePrefix          = collections.NewPrefix(22)
 	ActiveInvalidationsPrefix        = collections.NewPrefix(23)
+	ActiveParticipantsMerkleProofPrefix = collections.NewPrefix(24)
```
patimen, line 47: Formatting: These should all line up

inference-chain/x/inference/keeper/validators_proof.go
```diff
@@ -0,0 +1,50 @@
...
+	"context"
+	"fmt"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	externalutils "github.com/gonka-ai/gonka-utils/go/utils"
+	"github.com/productscience/common"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+func (k Keeper) SetValidatorsProof(ctx context.Context, proof types.ValidatorsProof) error {
+	h := uint64(proof.BlockHeight)
```
patimen: Can we just call this `height`?

patimen (approved):
Do we need more validation/limits for MsgSubmitParticipantsProof and MsgSubmitActiveParticipantsProofData? From a DDOS kind of perspective? Should it be restricted to ActiveParticipants themselves or do you feel the validation of the proofs is sufficient?

We need `testermint` tests for some of this, especially the things that will be ongoing processes (generation of proofs at each Epoch) that cannot fail.

One thing to call out for future work:
We need to standardize NewBlock handling better. I think we track it in 3 places? Not the place to solve it, though.

common/convert_types.go
```diff
@@ -0,0 +1,66 @@
...
+func ToCryptoProofOps(src *types.ProofOps) *cryptotypes.ProofOps {
+	if src == nil {
+		return nil
+	}
+
+	proofOps := &cryptotypes.ProofOps{
+		Ops: make([]cryptotypes.ProofOp, 0),
+	}
+	for _, op := range src.Ops {
+		proofOps.Ops = append(proofOps.Ops, cryptotypes.ProofOp(op))
```
DimaOrekhovPS, line 18: Why does calling `cryptotypes.ProofOp` work for converting here?

common/convert_types.go
```diff
@@ -0,0 +1,66 @@
...
+		CreatedAtBlockHeight: src.CreatedAtBlockHeight,
+		AppHashHex:           src.AppHashHex,
+		TotalVotingPower:     src.TotalVotingPower,
+		Commits:              make([]*contracts.CommitInfo, len(src.Commits)),
+	}
+	for i, c := range src.Commits {
+		out.Commits[i] = &contracts.CommitInfo{
+			ValidatorAddress: c.ValidatorAddress,
+			ValidatorPubKey:  c.ValidatorPubKey,
+			// VotingPower:      c.VotingPower,
```
DimaOrekhovPS: Why commented out?

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -25,6 +27,7 @@ import (
 // Minimal interface for query operations needed by the dispatcher
 type ChainStateClient interface {
 	EpochInfo(ctx context.Context, req *types.QueryEpochInfoRequest, opts ...grpc.CallOption) (*types.QueryEpochInfoResponse, error)
+	GetCurrentEpoch(ctx context.Context, req *types.QueryGetCurrentEpochRequest, opts ...grpc.CallOption) (*types.QueryGetCurrentEpochResponse, error)
```
DimaOrekhovPS, line 30: This query is deprecated, could you please use `QueryEpochInfoRequest` or you can also use  `phaseTracker`, it should have this info and it's already a dependency of `new_block_dispatcher`

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -100,18 +107,21 @@ func NewOnNewBlockDispatcher(
...
+		nodePocOrchestrator:    nodePocOrchestrator,
+		queryClient:            queryClient,
+		phaseTracker:           phaseTracker,
+		reconciliationConfig:   reconciliationConfig,
+		getStatusFunc:          getStatusFunc,
+		setHeightFunc:          setHeightFunc,
+		randomSeedManager:      randomSeedManager,
+		configManager:          configManager,
+		validator:              validator,
+		lastVerifiedAppHashHex: configManager.GetGenesisAppHash(),
```
DimaOrekhovPS, line 123: Every time we restart the decentralized API container we'll do the verification from the genesis?

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -149,25 +160,43 @@ func NewOnNewBlockDispatcherFromCosmosClient(
...
-func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo chainphase.BlockInfo) error {
+func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo chainevents.FinalizedBlock, knownHeight int64) error {
 	logging.Debug("Processing new block", types.Stages,
-		"height", blockInfo.Height,
-		"hash", blockInfo.Hash)
+		"height", blockInfo.Block.Header.Height,
+		"hash", blockInfo.BlockId.Hash)
+
+	// 1. Check if active participants proof for genesis epoch is created
+	if err := d.collectGenesisBlockProof(); err != nil {
```
DimaOrekhovPS, line 174: Do you think we can do all verification-related activities in a separate routine? So we don't accidentally block or slow down this processing loop

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -244,19 +276,93 @@ func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo ch
...
+
+		d.lastVerifiedAppHashHex = data.BlockProof.AppHashHex
+		logging.Info("verify participants successfully", types.Stages, "new_app_hash", d.lastVerifiedAppHashHex)
+	}
+}
+
+// Creates proof for genesis block. Is called only ones and only on genesis node
+// It is needed, because dapi container starts later, than blockchain container and dapi misses 1-2 first blocks by subscription
+func (el *OnNewBlockDispatcher) collectGenesisBlockProof() error {
+	if !el.configManager.GetChainNodeConfig().IsGenesis || el.isGenesisBlockProcessed {
```
DimaOrekhovPS, line 329: Do I understand correctly that this code may go unexecuted in the real chain if a genesis code is no longer online? Theoretical question, since the genesis node is still online for Gonka

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -244,19 +276,93 @@ func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo ch
...
+
+		data, err := utils.QueryActiveParticipants(rpcClient, d.transactionRecorder.NewInferenceQueryClient())(ctx, "current")
+		if err != nil {
+			logging.Error("Failed to get current participants data", types.ParticipantsVerification, "error", err)
+			return
+		}
+
+		err = externalutils.VerifyParticipants(ctx, d.lastVerifiedAppHashHex, utils.QueryActiveParticipants(rpcClient, d.transactionRecorder.NewInferenceQueryClient()), "current")
+		if err != nil {
+			panic(err)
```
DimaOrekhovPS, line 318: Are we absolutely sure that panicking is the only way here?
If we failed to verify then it means the chain is invalid?

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -347,6 +453,7 @@ func (d *OnNewBlockDispatcher) handlePhaseTransitions(epochState chainphase.Epoc
 		go func() {
 			d.randomSeedManager.ChangeCurrentSeed()
 		}()
+		d.collectBlockProofs(finalizedBlock)
```
DimaOrekhovPS, line 452: Why are we calling it in both `ProcessNewBlock` and `handlePhaseTransition`? Isn't it redundant?

decentralized-api/internal/server/public/get_block_handler.go
```diff
@@ -0,0 +1,70 @@
...
+	cosmos_client "decentralized-api/cosmosclient"
+	"decentralized-api/logging"
+	coretypes "github.com/cometbft/cometbft/rpc/core/types"
+	"github.com/labstack/echo/v4"
+	"github.com/productscience/inference/x/inference/types"
+	"net/http"
+	"strconv"
+)
+
+type TemporaryResponse struct {
```
DimaOrekhovPS, line 14: Why is it called `Temporary`?

inference-chain/app/upgrades/v0_2_5/upgrades.go
0xBECEDA: Yes, we are, thanks for notice

common/convert_types.go
0xBECEDA: I think because our ProofOp protobuf message fully repeats tendermit type (including fileds name)
`message ProofOp {
  string type = 1;
  bytes  key  = 2;
  bytes  data = 3;
}
`

I copypasted it becase including proto types from tendermint into project was crazy (have no idea why it had be such big problem).

common/convert_types.go
0xBECEDA: Didn't use it anywhere, forgot to delete

decentralized-api/internal/event_listener/new_block_dispatcher.go
0xBECEDA: Yes

decentralized-api/internal/event_listener/new_block_dispatcher.go
0xBECEDA: We could, but as I got @libermans concept, we shouldn't process next block, till we verify participants chain.  And in general case verification should be pretty fast:
- dapi and inference node are on same machine, so rpc requests execute fast
- we do not verify whole chain of participants every time: we start from current epoch till latest "verified". So normally we will verify 1-2 epochs back and only on container start we verify everything 
- collectGenesisBlockProof executes  only on genesis node and only once.

decentralized-api/internal/event_listener/new_block_dispatcher.go
0xBECEDA: Yes, but it is not problem for us since we populate missing proofs using by new API request

decentralized-api/internal/event_listener/new_block_dispatcher.go
0xBECEDA: It was David's wish and yes, it means there is some incorrect data

decentralized-api/internal/event_listener/new_block_dispatcher.go
0xBECEDA: Yes, you are right, will fix

decentralized-api/internal/server/public/get_block_handler.go
0xBECEDA: Because it is debug endpoint, so after upgrade I can remove it

common/convert_types.go
DimaOrekhovPS: But why is it ok that to not populate it?

inference-chain/app/upgrades/v0_2_5/upgrades.go
DimaOrekhovPS: Yes, why do we need some grantees here?

inference-chain/app/upgrades.go
```diff
@@ -75,4 +76,7 @@ func (app *App) registerMigrations() {
 	app.Configurator().RegisterMigration(stakingtypes.ModuleName, 5, func(ctx sdk.Context) error {
 		return nil
 	})
+	app.Configurator().RegisterMigration(stakingtypes.ModuleName, 6, func(ctx sdk.Context) error {
```
DimaOrekhovPS: Is it migration not for the inference module, but for the staking module?

DimaOrekhovPS: Why do we keep them here and why are they all empty @patimen @gmorgachev ?

DimaOrekhovPS: Maybe let's add a comment

inference-chain/proto/inference/inference/block_proof.proto
DimaOrekhovPS: I think we use it interchangeably, id and index

DimaOrekhovPS (approved):

decentralized-api/internal/event_listener/new_block_dispatcher.go
gmorgachev: How long does it take? Is it single request to node to get all data or per each previous epoch?
We have restarts quite often

inference-chain/app/upgrades.go
gmorgachev: Migration has to be registered per each increment of `ConsensusVersion` per each module. If no real state update needed, we can leave it empty (that was the case for previous upgrades)

In that brnach, we actually have additional Migration registered in `inference-chain/app/upgrades/v0_2_5/upgrades.go` from 4 to 5 for `inference` module, which was used earlier

In my understanding, we should have only one new migration 7 -> 8 for `inference` module. `staking` hasn't changed, am i missing smth?

inference-chain/go.mod
```diff
@@ -2,8 +2,6 @@ module github.com/productscience/inference
 
 go 1.23.2
 
-toolchain go1.24.2
```
gmorgachev, line 5: why to remove?

common/go.mod
```diff
@@ -0,0 +1,164 @@
+module github.com/productscience/common
+
+go 1.23.2
+
+require (
+	github.com/cometbft/cometbft v0.38.17
+	github.com/gonka-ai/gonka-utils/go v0.0.0-20250917223316-7583c27b562a
```
gmorgachev, line 7: should we set versions before merge? as it's separate repo

inference-chain/app/upgrades.go
DimaOrekhovPS: Yes, I'm also not sure why we use `stakingtypes.ModuleName` and not inference module name

common/go.mod
0xBECEDA: Yes

PR #403, opened 2025-10-22
Gl/wrapped token merge

gmorgachev:


inference-chain/x/bls/keeper/msg_server_group_validation.go
```diff
@@ -28,41 +28,48 @@ func (k Keeper) LogDebug(msg string, keyVals ...interface{}) {
...
-		return nil, fmt.Errorf("previous epoch %d not found", previousEpochId)
+		// Emit a searchable event and continue using current epoch data as fallback
+		ms.Keeper.LogWarn("Previous epoch not found - using current epoch for validation", "previous_epoch_index", previousEpochId, "new_epoch_index", msg.NewEpochIndex)
+		ctx.EventManager().EmitTypedEvent(&types.EventGroupKeyValidationFailed{
+			NewEpochIndex: msg.NewEpochIndex,
+			Reason:        fmt.Sprintf("previous_epoch_missing_fallback:%d", previousEpochId),
+		})
+
+		previousEpochBLSData = newEpochBLSData
+		previousEpochId = msg.NewEpochIndex
```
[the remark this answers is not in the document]
GLiberman: we actually need that so bls signature start to make next signatures because its stopped doing so because there was error with participants with zero slots

decentralized-api/internal/bls/dealer.go
```diff
@@ -38,20 +38,20 @@ import (
 // ProcessKeyGenerationInitiated handles the EventKeyGenerationInitiated event
 func (bm *BlsManager) ProcessKeyGenerationInitiated(event *chainevents.JSONRPCResponse) error {
 	// Extract event data from chain event (typed event from EmitTypedEvent)
-	epochIDs, ok := event.Result.Events["inference.bls.EventKeyGenerationInitiated.epoch_id"]
-	if !ok || len(epochIDs) == 0 {
-		return fmt.Errorf("epoch_id not found in key generation initiated event")
+	epochIndexes, ok := event.Result.Events["inference.bls.EventKeyGenerationInitiated.epoch_index"]
```
0xBECEDA, line 41: @DimaOrekhovPS introduced epoch id to fix out naming problem. I would suggest use epoch id everywhere, not epoch index

decentralized-api/internal/bls/manager.go
```diff
@@ -29,7 +29,7 @@ type BlsManager struct {
 
 // VerificationResult holds the results of DKG verification for an epoch
```
0xBECEDA, line 30: What is DKG?

decentralized-api/internal/bls/manager.go
```diff
@@ -59,37 +59,37 @@ func (vc *VerificationCache) Store(result *VerificationResult) {
 	}
 
 	// Store the new result
-	vc.results[result.EpochID] = result
+	vc.results[result.EpochIndex] = result
```
0xBECEDA, line 62: Are we sure we call this func in one thread only? 
Just in case: maps in Go are not thread-safe.

bridge/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Extract Geth binary from prebuilt container
+FROM ghcr.io/product-science/bridge-geth:latest AS geth-source
```
gmorgachev, line 2: What is that container?

bridge/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Extract Prysm binaries from prebuilt container  
+FROM ghcr.io/product-science/bridge-prysm:latest AS prysm-source
```
gmorgachev, line 5: What is that container?

bridge/Makefile
```diff
@@ -0,0 +1,41 @@
...
+	@echo "--> pushing bridge docker image"
+	@echo "pushing to GitHub Container Registry"
+	@docker push ghcr.io/product-science/bridge:$(VERSION)
+	@if [ "$(SET_LATEST)" = "1" ]; then \
+		echo "Setting latest tag..."; \
+		docker tag ghcr.io/product-science/bridge:$(VERSION) ghcr.io/product-science/bridge:latest; \
+		docker push ghcr.io/product-science/bridge:latest; \
+	fi
+
+release: build-docker
```
gmorgachev, line 39: Please add in high level build docker and releae

decentralized-api/Dockerfile
```diff
@@ -24,15 +24,15 @@ RUN apk add --no-cache make gcc musl-dev git patchelf
 
 # Use the musl libc version of wasmvm and create a symlink for the linker
 RUN if [ "$GOARCH" = "arm64" ]; then \
-        wget -O /lib/libwasmvm_muslc.aarch64.a https://github.com/CosmWasm/wasmvm/releases/download/v1.5.9/libwasmvm_muslc.aarch64.a \
-        && echo "a43cb22bf85e89bea45c9af04a229bc92f70ddd8216fee7db21d349d7579cff6 /lib/libwasmvm_muslc.aarch64.a" | sha256sum -c \
+        wget -O /lib/libwasmvm_muslc.aarch64.a https://github.com/CosmWasm/wasmvm/releases/download/v2.2.4/libwasmvm_muslc.aarch64.a \
+        && echo "27fb13821dbc519119f4f98c30a42cb32429b111b0fdc883686c34a41777488f /lib/libwasmvm_muslc.aarch64.a" | sha256sum -c \
```
gmorgachev, line 28: @GLiberman have you had a chance to check if it's embedded? Or we need to add them in migration?

That files are then copied to final image:
> COPY --from=builder /lib/libwasmvm_muslc.*.a /lib/

deploy/join/docker-compose.yml
```diff
@@ -71,6 +71,35 @@ services:
...
+    container_name: bridge
+    image: ghcr.io/product-science/bridge-ethereum:0.1.22.alphabridge
+    restart: unless-stopped
+    environment:
+      - GETH_DATA_DIR=/data/geth
+      - PRYSM_DATA_DIR=/data/prysm
+      - JWT_SECRET_PATH=/data/jwt/jwt.hex
+      - BRIDGE_API_BASE=http://api
+      - BRIDGE_POSTBLOCK=:9200/admin/v1/bridge/block
+      - BRIDGE_GETADDRESSES=:9000/v1/bridge/addresses
```
gmorgachev, line 84: Could we just explicitly set URLs? I'm worried that we're using too many different ways to compose URL in different parts of codebase

deploy/join/docker-compose.yml
```diff
@@ -71,6 +71,35 @@ services:
...
+      - BEACON_STATE_URL=https://beaconstate.ethstaker.cc/
+      - GETH_HTTP_PORT=8545
+      - GETH_AUTHRPC_PORT=8551
+      - GETH_P2P_PORT=30303
+      - GETH_DISCOVERY_PORT=30303
+      - PRYSM_P2P_TCP_PORT=13000
+      - PRYSM_P2P_UDP_PORT=12000
+      - PERSISTENT_DB_DIR=/persistent-db
```
gmorgachev, line 92: Should init docker has default value for that? Then end variable in docker compose might be optional

genesis/genesis-draft.json
gmorgachev, line 1: please don't change this file. it's read only

genesis/genesis.json
gmorgachev, line 1: please don't change this file, it's read only

decentralized-api/internal/server/admin/bls_handlers.go
```diff
@@ -42,10 +42,10 @@ func (f FlexibleUint64) ToUint64() uint64 {
 }
 
 type RequestThresholdSignatureDto struct {
-	CurrentEpochId FlexibleUint64 `json:"current_epoch_id"`
-	ChainId        []byte         `json:"chain_id"`
-	RequestId      []byte         `json:"request_id"`
-	Data           [][]byte       `json:"data"`
+	CurrentEpochIndex FlexibleUint64 `json:"current_epoch_index"`
```
0xBECEDA, line 45: I would definitely keep epoch id here since in another endpoints we have epoch_id and from user perspective it might be not clear what is epoch index

decentralized-api/internal/server/admin/bridge_handlers.go
```diff
@@ -0,0 +1,57 @@
...
+	// Debug: Log raw request body
+	rawBody := c.Request().Body
+	bodyBytes, err := io.ReadAll(rawBody)
+	if err != nil {
+		slog.Error("Failed to read request body", "error", err)
+		return c.JSON(400, map[string]string{"error": "Failed to read request body"})
+	}
+
+	// Log the raw JSON for debugging
+	slog.Info("Received raw request body", "body", string(bodyBytes))
```
0xBECEDA, line 24: we have logging package for that in dapi, please use it instead of slog

decentralized-api/internal/server/admin/bridge_handlers.go
```diff
@@ -0,0 +1,57 @@
...
+
+	// Reset the body for binding
+	c.Request().Body = io.NopCloser(strings.NewReader(string(bodyBytes)))
+
+	var blockData pserver.BridgeBlock
+	if err := c.Bind(&blockData); err != nil {
+		slog.Error("Failed to decode block data", "error", err)
+		return c.JSON(400, map[string]string{"error": "Invalid request body: " + err.Error()})
```
0xBECEDA, line 32: please use errors.New() instead of map or echo.NewHTTPError

decentralized-api/internal/server/admin/bridge_handlers.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	// Validate required fields
+	if blockData.BlockNumber == "" || blockData.ReceiptsRoot == "" || blockData.OriginChain == "" {
+		return c.JSON(400, map[string]string{"error": "Required fields missing: blockNumber, receiptsRoot, originChain"})
```
0xBECEDA, line 37: same here

decentralized-api/internal/server/admin/bridge_handlers.go
```diff
@@ -0,0 +1,57 @@
...
+		"originChain", blockData.OriginChain,
+		"receiptsRoot", blockData.ReceiptsRoot,
+		"receiptsCount", len(blockData.Receipts))
+
+	// Add the block to the shared queue
+	blockNumber := s.blockQueue.AddBlock(blockData)
+
+	// Return success response
+	return c.JSON(200, map[string]interface{}{
+		"status":        "success",
```
0xBECEDA, line 51: please create a response object instead of using map

decentralized-api/internal/server/public/bridge_handlers.go
```diff
@@ -335,3 +336,37 @@ func (s *Server) getBridgeStatus(c echo.Context) error {
 
 	return c.JSON(200, response)
 }
+
+// getBridgeAddresses returns bridge addresses for a specific chain by name
+func (s *Server) getBridgeAddresses(c echo.Context) error {
```
0xBECEDA, line 341: same comment about using maps in responses as in endpoint before

inference-chain/contracts/auto_vote.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/sh
+
```
0xBECEDA, line 2: why we need this script?

inference-chain/x/bls/module/autocli.go
```diff
@@ -19,11 +19,11 @@ func (am AppModule) AutoCLIOptions() *autocliv1.ModuleOptions {
 				},
 				{
 					RpcMethod: "EpochBLSData",
-					Use:       "epoch-data [epoch_id]",
+					Use:       "epoch-data [epoch_index]",
```
0xBECEDA, line 22: would keep epoch id here because of same reason as in API endpoints

inference-chain/x/inference/keeper/bridge.go
```diff
@@ -0,0 +1,109 @@
...
+// Key prefixes for bridge storage
+const (
+	BridgeAddressKeyPrefix = "BridgeAddress/"
+)
+
+// Bridge address management functions
+
+// generateBridgeAddressKey creates a unique key for bridge addresses
+func (k Keeper) generateBridgeAddressKey(ctx context.Context, chainId string) string {
+	return fmt.Sprintf("%s_%d", chainId, len(k.GetBridgeContractAddressesByChain(ctx, chainId))+1)
```
0xBECEDA, line 21: why +1 here?

inference-chain/x/inference/keeper/bridge_transaction.go
```diff
@@ -43,17 +30,108 @@ func (k Keeper) SetBridgeTransaction(ctx context.Context, tx *types.BridgeTransa
...
+
+	for ; iterator.Valid(); iterator.Next() {
+		key := string(iterator.Key())
+
+		// Parse key format: chainId_blockNumber_contentHash
+		// Check if this transaction is for the specified chain and is old enough
+		if len(key) > len(chainId)+1 && key[:len(chainId)+1] == chainId+"_" {
+			// Extract block number from key
+			remaining := key[len(chainId)+1:]
+			// Find the next underscore to get the block number
```
0xBECEDA, line 114: could we use strings.Split() here? I think it would be easier

inference-chain/x/inference/keeper/bridge_utils.go
```diff
@@ -0,0 +1,103 @@
...
+func ethereumAddressToBytes(address string) []byte {
+	// Remove 0x prefix if present
+	addr := address
+	if len(addr) >= 2 && addr[:2] == "0x" {
+		addr = addr[2:]
+	}
+
+	// Convert hex string to 20 bytes
+	addrBytes := make([]byte, 20)
+	for i := 0; i < 40 && i < len(addr); i += 2 {
```
0xBECEDA, line 24: I think we could use  https://pkg.go.dev/encoding/hex here

inference-chain/x/inference/keeper/bridge_wrapped_token.go
```diff
@@ -0,0 +1,907 @@
...
+	}
+
+	return nil
+}
+
+// GetTokenMetadata retrieves token metadata from chain state
+func (k Keeper) GetTokenMetadata(ctx sdk.Context, externalChain, externalContract string) (TokenMetadata, bool) {
+	store := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
+	normalizedContract := strings.ToLower(externalContract)
+	key := []byte(TokenMetadataKeyPrefix + externalChain + "/" + normalizedContract)
```
0xBECEDA, line 220: how about create function for key creation? since you create key in same way  a few methods

inference-chain/x/inference/keeper/bridge_wrapped_token.go
```diff
@@ -0,0 +1,907 @@
...
+	// - Other chains: various formats
+
+	// For Ethereum-style addresses, validate hex format
+	if strings.HasPrefix(strings.ToLower(address), "0x") {
+		hexPart := address[2:] // Remove 0x prefix
+		if len(hexPart) != 40 {
+			return false
+		}
+		// Check if it's valid hex
+		for _, r := range hexPart {
```
0xBECEDA, line 365: may be regexp?

decentralized-api/Dockerfile
GLiberman: I double checked - we used to have an dynamic libs but now we have static which embeds directly into executable. Unless we plan to rebuild inside the container we dont need these files

PR #407, opened 2025-10-28
Invalid and Inactive Participant improvements

### LINKED ISSUE #381: [P0] How to change `inference_url`
1. Change `inference_url`. Probably, it should happen immediately and propagate everywhere.
2. Vefigy `inference_url`. Let's think on how can it be verified, at least asynchronousl,y whena  node with that URL is already running
Example: `api` container has a new endpoint /v1/verify, which returns:
```
{
    "requester_address": "gonka...",
    "timestamps": <timestamp in last Xmin>,
    "signature": <singature of timestamps by this node's warm key>
}
```
The signature should not be refreshed more than once within X minutes.
Such an endpoint should be enough to have voting for claiming the wrong address. Ideally, every `api` node should verify all `inference_url` once in an epoch automatically and initiate this voting, but it's hard to estimate it for now, it might be okay to leave it manual at the moment.
3. Add a check that a new participant can't be created if there is the same URL across active participants (are all?), and also a participant can't be edited to set the existing URL.

### LINKED ISSUE #394: [P0] Invalid participants in the `ActiveParticipant` list
1. Proper removal (Check that we also jail => no voting power)
2. Removing for inactivity

- Goal: to remove inactive/invalid participants faster, not send inference requests to them, not allow them to be TAs / Validators, and remove their voting power
- We need to check for missed inference not only once in an epoch, but more often. E.g., once in X blocks (X = 500?)
- if `missed_stat_signigicant([start_of_epoch, current_heigh])` => remove and jail
- if `invalid_stat_significant([start_of_epoch, current_heigh])` => remove and jail
- For invalid - we should check how it intercepts with the current check, but we check only for sequential invalidations.
- introduce additional `min_n_samples = 100` to make this regular check less strong
-  open question: we now allow being jailed but an active participant, there is some slashing for that, but the participant still receives most of the reward. It might be okay to leave it that way, but let's think if removing from active participants is better in that case

**All the details in that task are sketches, not exact solutions, and should be criticized accordingly.**

### LINKED ISSUE #405: [P0] Removing participants for inactivity
Removing for inactivity

- Goal: to remove inactive/invalid participants faster, not send inference requests to them, not allow them to be TAs / Validators, and remove their voting power
- We need to check for missed inference not only once in an epoch, but more often. E.g., once in X blocks (X = 500?)
- if `missed_stat_signigicant([start_of_epoch, current_heigh])` => remove and jail
- if `invalid_stat_significant([start_of_epoch, current_heigh])` => remove and jail
- For invalid - we should check how it intercepts with the current check, but we check only for sequential invalidations.
- introduce additional `min_n_samples = 100` to make this regular check less strong
-  open question: we now allow being jailed but an active participant, there is some slashing for that, but the participant still receives most of the reward. It might be okay to leave it that way, but let's think if removing from active participants is better in that case

**All the details in that task are sketches, not exact solutions, and should be criticized accordingly.**

patimen:
### The Problem
Previous implementation had several key oversights when a participant was invalidated because of invalid inferences or downtime:
1. ActiveParticipants still included them, and gave no indication they had been excluded.
2. They were not removed from all EpochGroup sub-groups.

Additionally, the logic around removal was scattered and hard to understand, which is partly the reason these had not been addressed earlier.

#### ExcludedParticipants list
We add `ExcludedParticipants` to the state. It is keyed by epoch index and participant id. It includes the specific height of the exclusion and the reason for it. At the moment, exclusion is purely for invalidation because of too many invalid inferences. Other reasons are planned in the near future, thus the open-ended nature of the list.

When a participant is initially marked as invalidated for an epoch, they are added to the list.

#### ActiveParticipants response includes ExcludedParticipants
The response now includes the list of excluded participants, which clients will use to filter the Active list. The actual Active list remains the same, as it is important to have a cryptographically traceable ActiveParticipants list and this would become impractical if it also included arbitrary changes during an Epoch. (see related changes in this release).

#### Collateral has a hard dedupe for slashing
Slashing cannot rely on application logic to ensure that a single participant is slashed for a given reason only once per epoch. Given the dire nature of the punishment, we introduce a list in x/collateral that will ensure this at the lowest level, making application logic impossible to result in double punishment.

### The Problem
1. Status update was not uniformly calculated when states changed, thus it would be easy to forget to call the right methods to check status. 
2. The algorithm for invalid status was naive and likely to have both false positives and false negatives, especially during the early phase of an Epoch
3. Inactivity was detected and punished only at the end of an epoch during Claims, leaving inactive participants on the network

#### Unified status calculation
- Status is now calculated alongside `SetParticipant`. Additionally, if a significant state change is detected(from ACTIVE to INVALID for instance), then the procedures for removing the participant are executed.
- Status calculations, for both Invalid and Inactive, have been standardized to use an SPRT algorithm. (see sprt.go and sprt.html in `calculations`.
- Once a participant is moved to INVALID or INACTIVE, it will not be moved back until the next epoch.
- Many governance parameters added to address this.

#### Participant status reset with new participant
A participant could "re-add" themselves and erase their current epoch stats for the epoch. This could both inadverantly punish good behavior by erasing it or allow masking of bad behavior.

#### Ramping status deprecated
Early plans for using ramping never materialized. For now, this status is superfluous and only adds complication and confusion.


inference-chain/x/collateral/keeper/keeper.go
```diff
@@ -437,22 +444,52 @@ func (k Keeper) Slash(ctx context.Context, participantAddress sdk.AccAddress, sl
...
 	return totalSlashedAmount, nil
 }
+
+func (k Keeper) checkIfAlreadySlashed(ctx context.Context, participantAddress sdk.AccAddress, reason string) (*collections.Triple[uint64, sdk.AccAddress, string], error) {
+	if reason == "" {
+		return nil, nil
+	}
+	// get current epoch; default to 0 if not set yet
+	epoch := uint64(0)
+	if v, err := k.CurrentEpoch.Get(ctx); err == nil {
```
0xBECEDA, line 482: what if err != nil ?

inference-chain/x/inference/calculations/status.go
```diff
@@ -0,0 +1,106 @@
...
+
+const (
+	// ConsecutiveFailures indicates the participant has too many consecutive failures
+	ConsecutiveFailures ParticipantStatusReason = "consecutive_failures"
+	// Ramping indicates the participant is in ramp-up phase
+	Ramping ParticipantStatusReason = "ramping"
+	// StatisticalInvalidations indicates the participant has statistically significant invalidations
+	StatisticalInvalidations ParticipantStatusReason = "statistical_invalidations"
+	// NoReason indicates no specific reason for the status
+	NoReason ParticipantStatusReason = ""
```
0xBECEDA: May be NoSpecificReason ?

inference-chain/x/inference/calculations/status.go
```diff
@@ -0,0 +1,106 @@
...
+	zScore := (observedFailureRate - expectedFailureRate) / stdDev
+
+	return zScore
+}
+
+// MeasurementsNeeded calculates the number of measurements required
+// for a single failure to be within one standard deviation of the expected distribution
+func MeasurementsNeeded(p float64, max uint64) uint64 {
+	if p <= 0 || p >= 1 {
+		panic("Probability p must be between 0 and 1, exclusive")
```
0xBECEDA: do we really want panic here?

inference-chain/x/inference/calculations/status.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	// This value is derived from solving the inequality: |1 - np| <= sqrt(np(1 - p))
+	// Which leads to the quadratic inequality: y^2 - 3y + 1 >= 0, where y = np
+	// The solution to this inequality is np >= (3 + sqrt(5)) / 2
+	requiredValue := (3 + math.Sqrt(5)) / 2
```
0xBECEDA: how did you get this formula?

inference-chain/x/inference/calculations/status.go
```diff
@@ -0,0 +1,106 @@
...
+	return needed
+}
+
+// probabilityOfConsecutiveFailures returns P(F^N|G) = x^N
+func probabilityOfConsecutiveFailures(expectedFailureRate float64, consecutiveFailures int64) float64 {
+	if expectedFailureRate < 0 || expectedFailureRate > 1 {
+		panic("expectedFailureRate must be between 0 and 1")
+	}
+	if consecutiveFailures < 0 {
+		panic("consecutiveFailures must be non-negative")
```
0xBECEDA: same question about panic

inference-chain/x/inference/keeper/participant_status.go
```diff
@@ -0,0 +1,99 @@
...
+	// 4) Remove from current-epoch EpochGroup memberships
+	return k.removeFromEpochGroups(ctx, participant, reason)
+}
+
+// This is way messier than you'd expect...
+func multiply(completed uint32, preserve *types.Decimal) uint32 {
+	if preserve == nil {
+		return completed
+	}
+	pd := preserve.ToDecimal()
```
0xBECEDA, line 97: why wouldn't we round preserve to uint32 here, if we round final result in any case? Do I miss smth?

inference-chain/x/collateral/keeper/keeper.go
DimaOrekhovPS: yea, probably we should return it from the func then?

inference-chain/x/collateral/module/hooks.go
```diff
@@ -81,6 +80,7 @@ func (h StakingHooks) BeforeValidatorSlashed(ctx context.Context, valAddr sdk.Va
 		"fraction", fraction.String(),
 	)
 
-	_, err := h.k.Slash(sdkCtx, accAddr, fraction)
+	// Tendermint driven slashing is not limited per epoch, so pass in a blank reason
+	_, err := h.k.Slash(sdkCtx, accAddr, fraction, "")
```
DimaOrekhovPS, line 84: but if we always pass the same string, won't we hit "already slashed"?

inference-chain/x/inference/epochgroup/epoch_group.go
```diff
@@ -334,13 +334,24 @@ func (eg *EpochGroup) updateMember(ctx context.Context, address string, weight i
 	return err
 }
 
-func (eg *EpochGroup) UpdateMember(ctx context.Context, previousVersion *types.Participant, currentVersion *types.Participant) error {
```
DimaOrekhovPS, line 337: Why was UpdateMember deleted? Now we're always supposed to do Remove+Add?

inference-chain/x/collateral/module/hooks.go
patimen: The slashing portion ignores blank values and doesn't check for duplicates.

inference-chain/proto/inference/inference/excluded_participant.proto
```diff
@@ -0,0 +1,16 @@
...
+
+option go_package = "github.com/productscience/inference/x/inference/types";
+
+// ExcludedParticipant represents a participant excluded for a given epoch.
+// Upgrade-safety: New, additive message. Epoch-scoped; no migration required.
+message ExcludedParticipant {
+  string address = 1;           // Bech32 address of the participant
+  uint64 epoch_index = 2;       // Epoch index for which the participant is excluded
+  string reason = 3;            // Free-form reason string (e.g., "invalidated")
+  uint64 effective_height = 4;  // Block height when exclusion took effect
```
0xBECEDA: may be exclusion_block_height? effective block height is not really obvious, I think

inference-chain/x/collateral/keeper/keeper.go
```diff
@@ -431,28 +438,58 @@ func (k Keeper) Slash(ctx context.Context, participantAddress sdk.AccAddress, sl
 
 	// 3. Burn the total slashed amount from the module account
 	if !totalSlashedAmount.IsZero() {
-		err := k.bookkeepingBankKeeper.BurnCoins(sdkCtx, types.ModuleName, sdk.NewCoins(totalSlashedAmount), "collateral slashed")
+		err := k.bookkeepingBankKeeper.BurnCoins(sdkCtx, types.ModuleName, sdk.NewCoins(totalSlashedAmount), "collateral slashed:"+reason)
```
0xBECEDA: minor: "collateral slashed: "+reason instaed of "collateral slashed:"+reason
need extra space, would be easier to read

0xBECEDA (approved):
Generally looks good to me, added couple minor comments.

inference-chain/proto/inference/inference/excluded_participant.proto
DimaOrekhovPS: I don't have a strong opinion here, but `exclusion_block_height` may be a bit more clear

inference-chain/proto/inference/inference/participant.proto
```diff
@@ -8,7 +9,7 @@ enum ParticipantStatus {
   ACTIVE = 1;
   INACTIVE = 2;
   INVALID = 3;
-  RAMPING = 4;
+  RAMPING = 4 [deprecated = true];
```
DimaOrekhovPS, line 12: Deprecating it with this update? Lately I saw a message on discord that someone got stuck in the RAMPING status initially, maybe they just didn't get any inference to work their way into a different status

inference-chain/x/collateral/keeper/keeper.go
DimaOrekhovPS: or we should do no space approach, e.g. `collateral_slashed:some_reason`

inference-chain/x/collateral/keeper/keeper.go
DimaOrekhovPS: @patimen how does it work now?

inference-chain/x/inference/calculations/sprt.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	rFail := p1.Div(p0)
+	logFail, err := rFail.Ln(prec)
+	if err != nil {
+		return nil, fmt.Errorf("ln(P1/P0): %w", err)
+	}
+
+	rPass := one.Sub(p1).Div(one.Sub(p0))
+	logPass, err := rPass.Ln(prec)
```
DimaOrekhovPS, line 45: shouldn't we also validate `prec`?

inference-chain/x/inference/calculations/status.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	// Once INVALID or INACTIVE, this can only be reset deliberately (at epoch start)
+	if newValue.Status == types.ParticipantStatus_INVALID || newValue.Status == types.ParticipantStatus_INACTIVE {
+		return newValue.Status, AlreadySet, newStats
+	}
+
+	// If we have consecutive failures with a likelihood of less than 1 in a million times, we're assuming bad
+	falsePositiveRate := validationParameters.FalsePositiveRate.ToDecimal()
+	if probabilityOfConsecutiveFailures(falsePositiveRate, newValue.ConsecutiveInvalidInferences).LessThan(decimal.NewFromFloat(0.000001)) {
```
DimaOrekhovPS: Should we extract 0.000001 into a const?

inference-chain/x/inference/keeper/msg_server_submit_new_participant.go
```diff
@@ -9,12 +9,25 @@ import (
 
 func (k msgServer) SubmitNewParticipant(goCtx context.Context, msg *types.MsgSubmitNewParticipant) (*types.MsgSubmitNewParticipantResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
+
+	// Check if participant already exists. If it does, restrict updates to only
+	// ValidatorKey, WorkerKey, and Url as per requirements.
+	if existing, found := k.GetParticipant(ctx, msg.GetCreator()); found {
```
DimaOrekhovPS, line 15: 👍

inference-chain/x/inference/keeper/participant_status.go
DimaOrekhovPS: Because we want to do all computations in Decimal and convert only once in the end?

inference-chain/x/inference/module/module.go
```diff
@@ -640,6 +642,11 @@ func (am AppModule) moveUpcomingToEffectiveGroup(ctx context.Context, blockHeigh
 		am.LogError("Unable to clear active invalidations", types.EpochGroup, "error", err.Error())
 	}
 
+	// Clear ExcludedParticipants for the epoch that just ended to keep scope to current epoch only
+	err = am.keeper.ExcludedParticipantsMap.Clear(ctx, collections.NewPrefixedPairRange[uint64, sdk.AccAddress](previousEpochIndex))
```
DimaOrekhovPS, line 646: Maybe we should keep them for history? A few epochs maybe? For ease of debugging

DimaOrekhovPS (approved):

inference-chain/proto/inference/inference/participant.proto
patimen: I don't see how it serves any real purpose. It doesn't have any impact and just complicates the logic.

inference-chain/x/collateral/keeper/keeper.go
patimen: Get will only really return an error if we have yet to set the CurrentEpoch... I can't think of a better default behavior than to set the slash to zero, for the purposes of this change. That way we won't get a duplicate if we're in some how-did-this-happen no current epoch state.
I'd point out too that currentepoch IS set in the actual chain.

inference-chain/x/inference/calculations/status.go
patimen: This was replaced in the new version. It was derived from the equation for $|1 - np| \leq \sqrt{np(1 - p)}$ which gives us the probability a single failure is within one standard deviation give the failure rate.

inference-chain/x/inference/calculations/status.go
patimen: I just replaced these with returning Zero... they're both impossible in any event.

inference-chain/x/inference/keeper/participant_status.go
patimen: To use the shopspring library, we must convert all numbers to their Decimal representation.

inference-chain/x/inference/epochgroup/epoch_group.go
patimen: We don't currently have a scenario where we merely edit a member.

inference-chain/x/collateral/keeper/keeper.go
patimen: I'm going with the no_space approach.

inference-chain/x/inference/calculations/sprt.go
patimen: I'm going to follow this PR up with an exhaistive take on validating params in params.go, so that we can be sure we'll never have invalid values.

inference-chain/x/inference/module/module.go
patimen: We should have pretty clear logs for when these events happen. Let's get this merged.

PR #408, opened 2025-10-28
Node status check fixes

### LINKED ISSUE #326: [P2] Improve onboarding experience
Improve onboarding experience: 

- [ ] Clearer logging when node is launched and waiting for PoC
- [ ] remove errors which doesn’t mean errors
- [ ] automatic testing that everything will work when PoC starts (models can be deployed, all endpoints are accessible); 
- [ ] Clean up logs to avoid confusing ERROR messages;

Description of the proposal: 
[https://github.com/gonka-ai/gonka/blob/a2a15267ea4aa55288fc873f4e5e68bc69366447/proposals/onboarding-clarity-v1/README.md](https://www.google.com/url?q=https://github.com/gonka-ai/gonka/blob/a2a15267ea4aa55288fc873f4e5e68bc69366447/proposals/onboarding-clarity-v1/README.md&sa=D&source=docs&ust=1756918029145609&usg=AOvVaw3j8xsw5yvYWpDeCcVHBhZY)

### LINKED ISSUE #336: [P1] Retry for inference + don’t send inference to inactive nodes
- [x] Identify causes why inference requests are sent to inactive or unresponsive (“dead MLNode”) MLNodes.

DimaOrekhovPS:


decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -414,13 +418,23 @@ func (s *Server) handleExecutorRequest(ctx echo.Context, request *ChatRequest, w
...
+			return nil, fmt.Errorf("%w: %v", errInferencePost, postErr)
+		}
+		return resp, nil
 	})
 	if err != nil {
+		if errors.Is(err, errInferencePost) {
+			logging.Warn("Transport error posting to inference worker; triggering manual node status query", types.Inferences,
+				"inferenceId", inferenceId, "error", err)
+			// Trigger a manual status query to refresh node health
+			s.nodeBroker.TriggerStatusQuery()
```
gmorgachev: What if 1000 such query triggered in paralle? Does broker have logic to have only single one?

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -414,13 +418,23 @@ func (s *Server) handleExecutorRequest(ctx echo.Context, request *ChatRequest, w
 		if err != nil {
 			return nil, err
 		}
-		return http.Post(
+		resp, postErr := http.Post(
```
gmorgachev, line 423: @DimaOrekhovPS 
1. could you add logic to retry Lock if error during post? befiore proxy. let's say 3 attempts. imporant to have another node locked if > 1
2. could you add clear error message to distinguish when node can't be locked (not healthy nodes for model available) to response?

DimaOrekhovPS: 1. ok
2. We already have `ErrNoNodesAvailable = errors.New("no nodes available for inference")` defined in `broker.go` and it's returned when there are no nodes to lock. Doesn't it work as intended?

decentralized-api/internal/server/public/post_chat_handler.go
DimaOrekhovPS: Only one max can be queued at once, but I've also just added a 2 second cooldown between subsequent status query runs

PR #417, opened 2025-11-04
[P2] [DRAFT]: Websocket

### LINKED ISSUE #560: [P2] Finalizing the WebSocket (merge with `call_Back`, new vLLM will require python side implementation)

gmorgachev:


tcharchian:
Hey @x0152 @akup, it would be great if the two of you could align on the next steps for this task and make the key decisions together. If you are able to move it forward independently, this task could be considered for inclusion in v0.2.12. But overall, this is more of a nice-to-have than a must-have.

patimen:
Agree with low priority... I also have concerns about this approach in general, we may want to revisit. The PR tries to do more than it really needs to, it seems we should try to move entirely to the new solution as opposed to keeping the old HTTP method around, ai-review has found some substantive issues (Go client shutdown can hang, message handling blocks the read loop, idle connections may reconnect after 60s, and the Python side may block FastAPI’s async loop with multiprocessing queue reads)...

PR #418, opened 2025-11-04
Models: gemma, deepseek, qwen235, qwen30

### LINKED ISSUE #361: [P0] Threshold + Params for big model. Part 2.
1. Threshold + Params for big model; 
2. Scripts to compute, re-check for existing; 
3. New Models 

- [ ] `gpt-oss-120b`
- [x] `DeepSeek-R1-0528`
- [x] `gemma-3-27b-it`
- [x] `Qwen3-30B-A3B-Instruct-2507`
- [ ] `gpt-oss-20b`
- [x] `Qwen3-235B`

4. Instruction to do it
5. Inference Validation finetuning;
6. Fine-tuning Qwen 235

tamazgadaev:
- Updates to the threshold estimation workflow
- Proposals for the new models


islavutin:
Hi @tamazgadaev — Really glad to see the scope of supported models growing! I think that is crucial aspect of Gonka AI adoption. 

I’m new to the project, so apologies if I’m missing context! I have a few questions about the distance normalization implementation:

### Question 1: The `max(100, length)` normalization

I noticed in [`distance2()` at line 278](https://github.com/islavutin/gonka/blob/main/mlnode/packages/benchmarks/src/validation/utils.py#L278) you're using:

```python
total_dist = (total_dist + 1.0) / (max(100, len(inf_result.results))*len(inf_result.results[0].logprobs) + 1.0)
```
I'm curious about the choice to normalize by max(100, length) rather than just the actual length. Is this to protect against noisy short responses? Was there empirical testing that led to choosing 100 as the threshold?

### Question 2: Calibration data distribution
What was the length distribution of responses in your calibration dataset? I'm wondering if the threshold might behave differently for very short responses (like 10 tokens) vs longer ones (500+ tokens) if the calibration data was mostly in a specific range.

### Question 3: Length-dependent thresholds?
Have you considered using adaptive thresholds based on response length? Something like:

```python
if length < 50:
    threshold = 0.05  # More lenient for short
elif length < 200:
    threshold = 0.042
else:
    threshold = 0.035  # Stricter for long
```
Would love to hear your thoughts on whether this might improve accuracy!

###  Question 4: Adversarial testing
Did you get a chance to test any adversarial scenarios? I'm thinking of cases like:

Attackers generating intentionally verbose responses to game the normalization
Prompt engineering to hit a "sweet spot" in the distance calculation
Would be interesting to know how robust this is against those edge cases!

Great stuff! Keep it rolling

tamazgadaev:
Hey @islavutin,
Thank you for the attention to the PR, it is so precious to have people who care and review this kind of work! 
Let me answer your questions

Q1: Yes, it is for the protection from the noisy short answers. This method is quite rough, but it is enough to be able to validate short answers and not get a lot of False Positives.

Q2: You can take a look at the notebooks in https://github.com/gonka-ai/gonka/tree/tg/qwen_models/proposals/models/models_proposal . You will see distance-length graphs there and they will tell you almost everything about the distribution and basically answer your questions.

Q3: This is a promising idea! I think I would first try to look into distance function itself and choose something more theoretically meaningful in terms of statistics of sequences, because the current one is more of a heuristic. If the new distances depend too much on the length, it makes sense to use adaptive threshold. I am more inclined to follow "the simpler, the better" and don't increase the number of moving parts in the system, especially, when in general everything works.

Q4: given that the longer sequence, the easier it is to detect fraud (generally), the adversary would be to cut the response to  produce 2-3 word only, but in this case we won't see the <end-of-seqence> or its occurence will largely mismatch its probability in top logprobs. We explicitly check for it. If shortness of the answer is a true characteristic of the prompt (like answers to test), then different one-word answers for different models will be detectable and in case the words same, we are okay with the usage of smaller models for this specific case:).

I'd alse mention, that the core obstacle for QWEN models is that their token distributions are mostly one-hot encoded (one token is given some logprob, evrything else is -999 or similar).

I can take a closer look into this distributions, if you want! All the raw data is saved, links are available, and analysis notebooks can be a good quickstart or your own analysis. Everything is available here https://github.com/gonka-ai/gonka/blob/tg/qwen_models/proposals/models/models_proposal/README.md and in notebooks.

PR #446, opened 2025-11-19
Jf/token based authentication

### LINKED ISSUE #412: [P2] MLNode Token-Based Authentication and FQDN Support
## Goal / Problem

Current MLNode registration in the API service requires three static parameters:
- Static IP address
- PoC port (management API, default 8080)
- Inference port (vLLM inference API, default 5000)

Both ports serve the same MLNode container through nginx proxy for version management (see `deploy/join/docker-compose.mlnode.yml` and `deploy/join/nginx.conf`).

Problems:
- Some cloud providers (e.g., Aliyun EAS) assign new IPs on container recreation, making static IP registration impractical
- Managing two ports adds operational complexity
- Cloud providers often assign stable FQDNs with token-based authentication (e.g., `http://<some-id>.ap-southeast-1.pai-eas.aliyuncs.com/api/predict/eas02/`) that remain consistent across deployments
- Current registration doesn't support using authentication tokens that managed services provide for access control

**Note:** Segment fields (InferenceSegment, PoCSegment) are legacy parameters that are always empty in current deployments.

## Proposal

Support additional registration method using full URLs:

1. Use single port (8080) since both endpoints proxy to the same container   
2. Allow registration using stable baseURLs with authentication tokens instead of IP/ports

The system will support both registration methods, allowing users to choose between IP/port configuration or baseURL-based registration

### Single-Port Operation

Current state (see `deploy/join/nginx.conf` for nginx setup and `mlnode/packages/api/src/api/proxy.py` for internal routing logic):

Management API (port 8080) supports:
- `http://<host>:<poc_port>/api/v1/*` - management API endpoints
- `http://<host>:<poc_port>/v1/*` - proxies to vLLM endpoints
- `http://<host>:<poc_port>/readyz` - ready for inference
- `http://<host>:<poc_port>/health` - whole service health, *not only inference*

Inference API (port 5000, backward compatible) supports:
- `http://<host>:<inference_port>/v1/*` - proxies to vLLM endpoints  
- `http://<host>:<inference_port>/health` - vLLM health check (proxied from vLLM backend)

`<poc_port>/health` checks whole service health while `<inference_port>/health` checks only vLLM backend health. API node currently checks MLNode health via `client.InferenceHealth()` at `http://<host>:<inference_port>/health`. New API binary must support both old MLNodes (port 5000) and new single-port configuration.

### Solution
Use registration method to determine which health endpoint to check:
- Legacy registration (Host/Port/Segment): Check `http://<host>:<inference_port>/health`
- New registration (baseURL): Check `<baseURL>/readyz` (management API readiness endpoint on port 8080)

### FQDN and Token Authentication

Current structure (`decentralized-api/apiconfig/config.go`):
```go
type InferenceNodeConfig struct {
    Host             string
    InferenceSegment string
    InferencePort    int
    PoCSegment       string
    PoCPort          int
    // ... other fields
}
```

Proposed structure for `InferenceNodeConfig` and `broker.Node`:
```go
type InferenceNodeConfig struct {
    // Existing fields (preserved for backward compatibility)
    Host             string
    InferenceSegment string  // Legacy, always empty
    InferencePort    int
    PoCSegment       string  // Legacy, always empty
    PoCPort          int
    
    // New optional fields (SQLite only, not stored on-chain)
    BaseURL          string  // Optional: full URL to MLNode (e.g., "http://service.provider.com/path/")
    AuthToken        string  // Optional: bearer token for authentication
    
    // ... other fields
}

type Node struct {
    // Existing fields
    Host             string
    InferenceSegment string
    InferencePort    int
    PoCSegment       string
    PoCPort          int
    
    // New optional fields
    BaseURL          string
    AuthToken        string
    
    // ... other fields
}
```

URL construction uses baseURL when present, otherwise falls back to `http://<host>:<port>/<segment>`. Version insertion for rolling upgrades works identically for both approaches: `<baseURL>/<version>/<path>` or `http://<host>:<port>/<version>/<path>`.

baseURL and AuthToken are stored in local SQLite database only, not on-chain. This allows each API node to configure its own MLNode access methods independently.

Required changes:

1. Add `base_url` and `auth_token` columns to SQLite schema with empty defaults for automatic migration
2. Update URL construction methods:
   - `broker.Node.InferenceUrlWithVersion()` and `broker.Node.PoCUrlWithVersion()` in `broker/broker.go` (main methods used for all inference and management calls including `/v1/chat/completions`)
   - Helper functions in `mlnode_background_manager.go` 
   - Helper functions in `setup_report.go` for consistency
3. Add `Authorization: Bearer <token>` header to all MLNode requests when AuthToken is set
4. Validate registration: require either (Host+Ports) OR baseURL, not both. baseURL must be valid HTTP(S) URL. AuthToken is always optional.

### Testing

1. Covered by unit tests and they pass:
- local `make local-build`
- [CICD](https://github.com/gonka-ai/gonka/actions/workflows/verify.yml)
2. Existing testermint tests pass:
- local `make build-docker && ./local-test-net/stop.sh &&  make run-tests`
- [Recommended] [CICD](https://github.com/gonka-ai/gonka/actions/workflows/integration.yml)
3. New node joins testnet and works with new MLNode registration

### Backward Compatibility

- Existing nodes using Host/Port configuration are unaffected
- baseURL and AuthToken are local SQLite configuration, not stored on-chain
- No migration needed - old and new registration methods coexist

Pegasus-starry:


decentralized-api/apiconfig/sqlite_store.go
```diff
@@ -104,8 +106,62 @@ CREATE TABLE IF NOT EXISTS seed_info (
...
+	if _, err := db.ExecContext(ctx, stmt); err != nil {
+		return err
+	}
+
+	// Migrate existing tables: add base_url and auth_token columns if they don't exist
+	return migrateInferenceNodesTable(ctx, db)
+}
+
+// migrateInferenceNodesTable adds base_url and auth_token columns to existing inference_nodes table
+func migrateInferenceNodesTable(ctx context.Context, db *sql.DB) error {
```
DimaOrekhovPS: I'd be ideal if you could write a unit test to test the migration. I see it as:

1. Keep old `EnsureSchema` and rename it to `EnsureSchemaV1`
2. Add `EnsureSchemaV2` that has new changes.
3. In the test initialize 2 empty sqlite DBs and run `EnsureSchemaV1` + `migrateInferenceNodesTable` on the first one. On the second one run `EnsureSchemaV2`
4. Test success criteria: both schemas are identical

See `config_manager_migration_test.go`, it has examples of creating test db for unit tests

I think it's really important to test our migrations since if it fails then it's a showstopper.

decentralized-api/broker/broker.go
```diff
@@ -155,6 +157,14 @@ func (n *Node) InferenceUrl() string {
 }
 
 func (n *Node) InferenceUrlWithVersion(version string) string {
+	// If BaseURL is provided, build on top of it
+	if n.BaseURL != "" {
```
DimaOrekhovPS: Please add `TrimSpace` to every string comparison

decentralized-api/broker/broker.go
```diff
@@ -166,6 +176,14 @@ func (n *Node) PoCUrl() string {
 }
 
 func (n *Node) PoCUrlWithVersion(version string) string {
```
DimaOrekhovPS, line 188: Please add unit tests for both `PoCUrlWithVersion` and `InferenceUrlWithVersion`. You can see how they're already tested in `broker_test.go`

We need to test `BaseUrl` scenario you're adding here

decentralized-api/internal/modelmanager/mlnode_background_manager.go
```diff
@@ -207,18 +208,30 @@ func getInferenceUrlWithVersion(node apiconfig.InferenceNodeConfig, version stri
 }
 
 func getPoCUrl(node apiconfig.InferenceNodeConfig) string {
```
DimaOrekhovPS: I see that we have duplicating logic in `broker.go` and here. Ideally we should refactor:

1. Create reusable function like:
```
func getMlNodeUrl(pathElements MlNodePathElements) { ...}

type MlNodePathElements struct {
    Host string
    Port int
    BaseUrl string
    Version string
    Segment string
 }
```

2. Then this common function can be used in both of the files

3. Ideally we should use tools from Go stdlib for joining path, e.g. `url.JoinPath`

decentralized-api/internal/server/admin/setup_report.go
```diff
@@ -784,10 +784,16 @@ func getPoCUrlWithVersion(node apiconfig.InferenceNodeConfig, version string) st
 }
 
 func getPoCUrl(node apiconfig.InferenceNodeConfig) string {
+	if node.BaseURL != "" {
```
DimaOrekhovPS: Same thing about avoiding dupilcating logic by introducing a common function

DimaOrekhovPS (changes requested):
The changes look on point, but they need to be tested. I suggested a few unit tests you can write.

🔴 There are a few spots where we might be missing adding auth token to requests:
- [ ] Node completion requests, look at `post_chat_handler.go`. There are 2 `http.Post` calls we make to the 
- [ ] Inference validation requests, also look for `http.Post` in `inference_validation.go`

Please move them to `MlNodeClient` interface so we don't forget about them in the future.

DimaOrekhovPS:
Please check-in this file into this branch:
https://github.com/gonka-ai/gonka/tree/gm/proposals/proposals/mlnode-token-auth

So it shows up in the diff as well

decentralized-api/apiconfig/sqlite_store.go
Pegasus-starry: done， will commit later

Pegasus-starry: https://github.com/gonka-ai/gonka/pull/446/commits/4a01b0d475bd0270ae1852d76737577d72689f36

decentralized-api/broker/broker.go
Pegasus-starry: https://github.com/gonka-ai/gonka/pull/446/commits/75d4361f16747f662dec5551ba8cbb63fe6a8463

decentralized-api/broker/broker.go
Pegasus-starry: https://github.com/gonka-ai/gonka/pull/446/commits/75d4361f16747f662dec5551ba8cbb63fe6a8463

decentralized-api/internal/modelmanager/mlnode_background_manager.go
Pegasus-starry: https://github.com/gonka-ai/gonka/pull/446/commits/75d4361f16747f662dec5551ba8cbb63fe6a8463

decentralized-api/internal/server/admin/setup_report.go
Pegasus-starry: https://github.com/gonka-ai/gonka/pull/446/commits/75d4361f16747f662dec5551ba8cbb63fe6a8463

Pegasus-starry:
> Please check-in this file into this branch: https://github.com/gonka-ai/gonka/tree/gm/proposals/proposals/mlnode-token-auth
> 
> So it shows up in the diff as well

Pegasus-starry:
> Please write a testermint test that does the following:
> 
> 1. Creates a setup with mlnodes that use auth tokens
> 2. Wait for first PoC and make sure participants got their weight
> 3. When it's Inference phase after, do some inferences.
> 4. Wait for next poc to end and check that:
> 
> * Inferences succeeded
> * Inferences validated
> * Rewards were received
> 
> You can modify `testermint/mock-server` by adding an endpoint to it to enable authentication by token. For general examples of testermint tests I recommend looking at `ParticipantPowerTests.kt` (test PoC weights are being assigned correctly), `ValidationTests.kt`, `InferenceAccountingTests.kt`

added: https://github.com/gonka-ai/gonka/pull/446/commits/d12182a1d529b6ace26f3e1b50ac75a983a5fecd

decentralized-api/broker/broker_test.go
```diff
@@ -412,6 +412,79 @@ func TestNodeShouldBeOperationalTest(t *testing.T) {
 	require.False(t, ShouldBeOperational(adminState, 12, types.InferencePhase))
 }
 
+func TestGetMlNodeUrl(t *testing.T) {
+	tests := []struct {
+		name     string
+		elements MlNodePathElements
+		expected string
+	}{
+		{
```
DimaOrekhovPS, line 489: Could add BaseUrl + empty segment scenario. Usually segment is empty, so it's something we expect

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -420,11 +415,14 @@ func (s *Server) handleExecutorRequest(ctx echo.Context, request *ChatRequest, w
...
-		resp, postErr := http.Post(
-			completionsUrl,
-			request.Request.Header.Get("Content-Type"),
-			bytes.NewReader(modifiedRequestBody.NewBody),
-		)
+		var requestBody map[string]interface{}
+		if err := json.Unmarshal(modifiedRequestBody.NewBody, &requestBody); err != nil {
+			return nil, broker.NewApplicationActionError(fmt.Errorf("failed to unmarshal request body: %w", err))
+		}
+		resp, postErr := utils.SendPostJsonRequestWithAuth(context.Background(), http.DefaultClient, completionsUrl, requestBody, node.AuthToken)
```
DimaOrekhovPS, line 422: Can we actually make them methods of `MlNodeClient`?

testermint/src/test/kotlin/AuthTokenFlowTests.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    @Test
+    fun `full flow with auth tokens`() {
+        // 1. Setup with auth tokens
+        val (cluster, genesis) = initCluster(reboot = true)
+        val authToken = "test-auth-token-${UUID.randomUUID()}"
+        
+        // Configure MLNodes to verify auth token
+        cluster.allPairs.forEach { pair ->
+            pair.mock?.setRequestValidator { request ->
```
DimaOrekhovPS, line 21: Please add some check that makes sure that requests without the token fail

decentralized-api/mlnodeclient/client.go
```diff
@@ -282,12 +286,23 @@ func (api *Client) GetPowStatus(ctx context.Context) (*PowStatusResponse, error)
 }
 
 func (api *Client) InferenceHealth(ctx context.Context) (bool, error) {
-	requestURL, err := url.JoinPath(api.inferenceUrl, "/health")
+	var requestURL string
```
DimaOrekhovPS, line 289: Are you sure we need to handle it in a special way? Why?

I think every call to an mlnode should follow the same scheme and include a version

Pegasus-starry: according to https://github.com/gonka-ai/gonka/tree/gm/proposals/proposals/mlnode-token-auth,   two different way is suggested to implement:

Use registration method to determine which health endpoint to check:
Legacy registration (Host/Port/Segment): Check http://<host>:<inference_port>/health
New registration (baseURL): Check <baseURL>/readyz (management API readiness endpoint on port 8080)

decentralized-api/broker/broker_test.go
Pegasus-starry: added:https://github.com/gonka-ai/gonka/pull/446/commits/197a0e4a0655acbe4fc161349a71669acd13f86c

testermint/src/test/kotlin/AuthTokenFlowTests.kt
Pegasus-starry: "Validate registration: require either (Host+Ports) OR baseURL, not both. baseURL must be valid HTTP(S) URL. AuthToken is always optional."  in the proposal https://github.com/gonka-ai/gonka/tree/gm/proposals/proposals/mlnode-token-auth,  AuthToken is designed to always optional,  is it need to modify?  update to ensure AuthToken must be set when baseURL exist?

decentralized-api/internal/server/public/post_chat_handler.go
Pegasus-starry: MlNodeClient is preferred to handle function. And utils handle this. I think it's reasonable here now

testermint/src/test/kotlin/AuthTokenFlowTests.kt
Pegasus-starry: added:https://github.com/gonka-ai/gonka/pull/446/commits/91363b01565a5b25720e9f4cc180bbc1aa618549

decentralized-api/mlnodeclient/client.go
Pegasus-starry: added : https://github.com/gonka-ai/gonka/pull/446/commits/e9c3ab99872f87eafc833a4b113568146ca1062e

DimaOrekhovPS (approved):

DimaOrekhovPS:
The latest version: https://github.com/gonka-ai/gonka/pull/717

PR #452, opened 2025-11-22
feat: add configurable port and disable option for backward compatibility server

malmstrem:
Add support for BACKWARD_COMPAT_PORT and DISABLE_BACKWARD_COMPAT environment variables to fix port conflicts when running multiple ML nodes with hostNetwork: true in Kubernetes.


patimen:
We've addressed this in different ways since the PR was introduced.

PR #459, opened 2025-11-28
Upgrade v0.2.6

### LINKED ISSUE #387: Bug: fatal error: too many concurrent timer firings
node container restarted due to:
```
fatal error: too many concurrent timer firings

runtime stack:
runtime.throw({0x5422134?, 0x7f3d16444260?})
	/usr/local/go/src/runtime/panic.go:1067 +0x48 fp=0x7f3d164441f8 sp=0x7f3d164441c8 pc=0x4c80e8
runtime.(*timer).unlockAndRun(0xc0027fe250, 0x473246?)
	/usr/local/go/src/runtime/time.go:1076 +0x28e fp=0x7f3d16444260 sp=0x7f3d164441f8 pc=0x4b030e
runtime.(*timers).run(0xc000288290, 0x7e04cdbdb5274)
	/usr/local/go/src/runtime/time.go:1008 +0xf0 fp=0x7f3d16444288 sp=0x7f3d16444260 pc=0x4b0030
runtime.(*timers).check(0xc000288290, 0x0?)
	/usr/local/go/src/runtime/time.go:942 +0x13d fp=0x7f3d164442d0 sp=0x7f3d16444288 pc=0x4afe5d
runtime.stealWork(0x941cbc0?)
	/usr/local/go/src/runtime/proc.go:3687 +0x1f3 fp=0x7f3d16444340 sp=0x7f3d164442d0 pc=0x498433
runtime.findRunnable()
	/usr/local/go/src/runtime/proc.go:3364 +0x405 fp=0x7f3d164444b8 sp=0x7f3d16444340 pc=0x497485
runtime.schedule()
	/usr/local/go/src/runtime/proc.go:3995 +0xb1 fp=0x7f3d164444f0 sp=0x7f3d164444b8 pc=0x498eb1
runtime.park_m(0xc00042ec40)
	/usr/local/go/src/runtime/proc.go:4102 +0x1eb fp=0x7f3d16444548 sp=0x7f3d164444f0 pc=0x4992cb
runtime.mcall()
	/usr/local/go/src/runtime/asm_amd64.s:459 +0x4e fp=0x7f3d164445
```

[node-failure-oct8.log](https://github.com/user-attachments/files/22790665/node-failure-oct8.log)

### LINKED ISSUE #464: Difficulty with PoC, defined by `RTarget`
There is a difficulty with PoC, defined by `RTarget` in the repo. It essentially defines the percentage of "correct" nonces from all nonces => how many nonces participants has to check to find the correct one. 

Let's say we:
- increase complexity 
- add coefficient which transforn new weight to ~old weight (just to maintain same numbers in dashboard)

Please figure out how we're doing that. The open question - how we check which nodes were preserved, which are not, and which we preserved for > 1 epochs
=> to understand which weight to transform and which not 

There is some simple and elegant solution for that, e.g., use this coefficient at transforming len(nonces) -> weight
=> already transformed weight will be recorded in further preserved nodes

### LINKED ISSUE #470: Move payload off chain

gmorgachev:
# Upgrade Proposal: v0.2.6

This document outlines the proposed changes for on-chain software upgrade v0.2.6. The `Changes` section details the major modifications, and the `Upgrade Plan` section describes the process for applying these changes.

## Upgrade Plan

This PR updates the code for the `api` and `node` services. The PR modifies the container versions in `deploy/join/docker-compose.yml`.

The binary versions will be updated via an on-chain upgrade proposal. For more information on the upgrade process, refer to [`/docs/upgrades.md`](https://github.com/gonka-ai/gonka/blob/gm/dev-0.2.6/docs/upgrades.md).

Existing hosts are **not** required to upgrade their `api` and `node` containers. The updated container versions are intended for new hosts who join after the on-chain upgrade is complete.

## Proposed Process

2. Once the PR is approved by a majority, a `v0.2.6` release will be created from this branch, and an on-chain upgrade proposal for this version will be submitted.

Creating the release from this branch (instead of `main`) minimizes the time that the `/deploy/join/` directory on the `main` branch contains container versions that do not match the on-chain binary versions, ensuring a smoother onboarding experience for new hosts.

## Optional: PostgreSQL for Payload Storage

Off-chain payloads use file-based storage by default, which is suitable for small nodes. For larger deployments, payloads can optionally be stored in PostgreSQL. The database is defined in a separate `docker-compose.postgres.yml` file.

It's recommended to deploy PostgreSQL on a separate machine or at least point its volume to a separate disk.

**Environment variables for PostgreSQL container** (`docker-compose.postgres.yml`):
- `POSTGRES_PASSWORD` (required)
- `POSTGRES_USER` (default: `payloads`)
- `POSTGRES_DB` (default: `payloads`)

**Environment variables for API to connect to PostgreSQL** (`docker-compose.yml`):
- `PGHOST` - PostgreSQL host address (if not set, file storage is used)
- `PGPASSWORD` - PostgreSQL password
- `PGPORT` (default: `5432`)
- `PGUSER` (default: `payloads`)
- `PGDATABASE` (default: `payloads`)

Start after upgrade:
```
git pull
source config.env && docker compose -f docker-compose.postgres.yml up -d
```

## Testing

The on-chain upgrade from version `v0.2.5` to `v0.2.6` has been successfully deployed and verified on the testnet.

## Migration

The on-chain migration logic is defined in [`upgrades.go`](https://github.com/gonka-ai/gonka/blob/gm/dev-0.2.6/inference-chain/app/upgrades/v0_2_6/upgrades.go).

Migration sets new PoC parameters (see "PoC Parameters On-Chain" in Changes section).

## Changes

---

### Off-Chain Payloads

Commit: [477fb6e81](https://github.com/gonka-ai/gonka/commit/477fb6e81)

Moves inference prompts and response payloads off-chain. Only hashes are stored on-chain.

**Motivation:** Block size limit (22MB) and payload sizes (up to MBs for long responses) constrain throughput below compute capacity. Moving payloads off-chain reduces transaction size to ~500 bytes, removing bandwidth as a bottleneck.

Details: [proposals/offchain-payloads/README.md](https://github.com/gonka-ai/gonka/blob/gm/dev-0.2.6/proposals/offchain-payloads/README.md)

---

### Transaction Batching

Commit: [288b37732](https://github.com/gonka-ai/gonka/commit/288b37732)

Batching for StartInference/FinishInference transactions.

---

### PoC Parameters On-Chain

Commits: [86ebd4d65](https://github.com/gonka-ai/gonka/commit/86ebd4d65), [806b01616](https://github.com/gonka-ai/gonka/commit/806b01616), [f41e05142](https://github.com/gonka-ai/gonka/commit/f41e05142)

PoC parameters moved to on-chain governance. RTarget changed to 1.398077, increasing PoC difficulty ~2.5 times. Weights scaled by 2.5x to maintain absolute values.

---

### BLS DealerShared Recovery

Commit: [5b22aafd5](https://github.com/gonka-ai/gonka/commit/5b22aafd5)

Enables BLS secret recovery when container restarts.

---

### Nodes Always Available

Commit: [093c2e36a](https://github.com/gonka-ai/gonka/commit/093c2e36a)

Nodes remain available for inference even when disabled for next epoch.

---

### NATS Storage Fix

Commit: [ae938d357](https://github.com/gonka-ai/gonka/commit/ae938d357)

NATS messages retained for 24 hours instead of forever. Resolves issue with large `.dapi/.nats` directory.

---

### Force Reward Recovery

Commit: [fe02ed509](https://github.com/gonka-ai/gonka/commit/fe02ed509)

Enables reward recovery even when stored seed is missing or corrupted. Seed can now be regenerated deterministically from epoch number, allowing recovery for any epoch.

---

### Epoch Performance Query

Commit: [08ad82a7b](https://github.com/gonka-ai/gonka/commit/08ad82a7b)

Adds `EpochPerformanceSummaryAll` query endpoint.

---

### GPU Distribution Fix

Commit: [fae7d20a5](https://github.com/gonka-ai/gonka/commit/fae7d20a5)

Limits PoC batch queue per GPU to prevent one GPU from accumulating all batches. Ensures even distribution across multiple GPUs.


decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -462,6 +596,57 @@ func (m *manager) BankBalances(ctx context.Context, address string) ([]sdk.Coin,
...
 
+func (m *manager) GetJetStream() nats.JetStreamContext {
+	return m.natsJetStream
+}
+
+func (m *manager) BroadcastMessages(id string, msgs ...sdk.Msg) (*sdk.TxResponse, time.Time, error) {
+	if len(msgs) == 0 {
+		return nil, time.Time{}, nil
+	}
+	if len(msgs) == 1 {
```
patimen, line 607: Why not simplify and just use only the multi-message version?

decentralized-api/broker/broker.go
```diff
@@ -1143,9 +1152,19 @@ func (b *Broker) queryCurrentPoCParams(epochPoCStartHeight int64) (*pocParams, e
 		logging.Error("Failed to query PoC start block hash", types.Nodes, "height", epochPoCStartHeight, "error", err)
 		return nil, err
 	}
+
+	paramsResp, err := b.chainBridge.GetParams()
+	var modelParams *types.PoCModelParams
+	if err != nil {
+		logging.Warn("Failed to query chain params, will use default model params", types.Nodes, "error", err)
```
0xBECEDA, line 1167: May be Error level?

decentralized-api/broker/broker.go
```diff
@@ -1346,16 +1368,36 @@ func (b *Broker) UpdateNodeWithEpochData(epochState *chainphase.EpochState) erro
...
+	for nodeId := range b.nodes {
+		if !nodesInEpoch[nodeId] {
+			nodeIds = append(nodeIds, nodeId)
+		}
+	}
+	b.mu.RUnlock()
+
+	for _, nodeId := range nodeIds {
+		if err := b.populateNodeWithGovernanceModels(nodeId); err != nil {
+			logging.Warn("Failed to populate governance models for node not in epoch", types.Nodes, "node_id", nodeId, "error", err)
```
0xBECEDA, line 1405: May be Error level also?

decentralized-api/broker/broker.go
```diff
@@ -1383,6 +1425,125 @@ func (b *Broker) UpdateNodeEpochData(mlNodes []*types.MLNodeInfo, modelId string
 	}
 }
 
+// PopulateSingleNodeEpochData populates epoch data for a specific node.
+// If the node is found in current epoch data, it uses that data.
+// If not found (new node not yet assigned), it populates with governance model snapshots.
+func (b *Broker) PopulateSingleNodeEpochData(nodeId string) error {
+	if b.phaseTracker == nil {
```
0xBECEDA, line 1440: I think in our case phase tracker can't be nil in any case, otherwise the whole node wouldn't work

0xBECEDA, line 1440: how phaseTracker might be nil?

decentralized-api/cosmosclient/tx_manager/batch_consumer.go
```diff
@@ -0,0 +1,229 @@
...
+	var shouldFlush bool
+	c.startMu.Lock()
+	if len(c.startBatch) == 0 {
+		c.startCreatedAt = time.Now()
+	}
+	c.startBatch = append(c.startBatch, pendingMsg{msg: sdkMsg, natsMsg: msg})
+	shouldFlush = len(c.startBatch) >= c.config.FlushSize
+	c.startMu.Unlock()
+
+	if shouldFlush {
```
0xBECEDA, line 106: According to the [NATS documentation,](https://docs.nats.io/nats-concepts/jetstream/consumers?utm_source=chatgpt.com#ackpolicy) if a message is processed for too long without an Ack/Nack, NATS may redeliver it before the previous delivery has been fully processed.

Have we verified this?

decentralized-api/cosmosclient/tx_manager/batch_consumer.go
```diff
@@ -0,0 +1,229 @@
...
+		msgs[i] = p.msg
+	}
+
+	logging.Info("Broadcasting batch", types.Messages, "type", batchType, "count", len(msgs))
+
+	err := c.txManager.SendBatchAsyncWithRetry(msgs)
+	if err != nil {
+		logging.Error("Failed to hand off batch to TxManager", types.Messages, "type", batchType, "error", err)
+		for _, p := range batch {
+			p.natsMsg.Nak()
```
0xBECEDA: don't we need Ack here instead of Nack?

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -515,6 +700,18 @@ func (m *manager) unpackTx(bz []byte) (sdk.Msg, error) {
 	return rawTx, nil
 }
 
+func (m *manager) unpackBatch(rawBatch [][]byte) ([]sdk.Msg, error) {
+	msgs := make([]sdk.Msg, len(rawBatch))
+	for i, bz := range rawBatch {
+		msg, err := m.unpackTx(bz)
+		if err != nil {
```
0xBECEDA, line 707: why we return here instead of skip invalid msg and keep sending other?

decentralized-api/internal/bls/manager.go
```diff
@@ -42,26 +43,26 @@ type VerificationResult struct {
 
 // VerificationCache manages verification results for multiple epochs
 type VerificationCache struct {
-	results map[uint64]*VerificationResult // epochID -> VerificationResult
+	sync.RWMutex
+	results map[uint64]*VerificationResult
```
0xBECEDA, line 47: why we do not use sync.Map here?
It is map with mutex already

decentralized-api/internal/epoch_group_cache.go
```diff
@@ -9,18 +9,30 @@ import (
...
+
 type EpochGroupDataCache struct {
 	mu sync.RWMutex
 
+	// Legacy single-epoch cache for GetCurrentEpochGroupData
 	cachedEpochIndex uint64
 	cachedGroupData  *types.EpochGroupData
 
+	// Multi-epoch cache for GetEpochGroupData (max 2 epochs)
+	epochCache map[uint64]*cachedEpochData
```
0xBECEDA, line 27: same question about sync.Map here

decentralized-api/internal/server/admin/payload_handlers.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"log/slog"
+	"net/http"
+	"strconv"
+
+	"github.com/labstack/echo/v4"
+)
+
+// StorePayloadRequest is the request body for storing payloads
+// Note: EpochId is string because Kotlin's cosmosJson serializes Long as string
```
0xBECEDA: I think we need to explain here, why we talking about Kotlin here. It is not obvious for all that out testing framework is written on Kotlin

decentralized-api/internal/server/admin/payload_handlers.go
```diff
@@ -0,0 +1,69 @@
...
+	EpochId     uint64 `json:"epoch_id"`
+}
+
+// storePayload handles POST requests to store payloads directly to PayloadStorage.
+// This endpoint is used by testermint to store payloads when using InferenceTestHelper
+// which bypasses the normal REST API flow.
+func (s *Server) storePayload(c echo.Context) error {
+	inferenceId := c.QueryParam("inference_id")
+	if inferenceId == "" {
+		return c.JSON(http.StatusBadRequest, map[string]string{"error": "inference_id required"})
```
0xBECEDA, line 31: let's use here echo.NewHTTPError here and always

decentralized-api/payloadstorage/cached_storage.go
```diff
@@ -0,0 +1,104 @@
...
+	go c.cleanupLoop()
+	return c
+}
+
+func (c *CachedStorage) Store(ctx context.Context, inferenceId string, epochId uint64, promptPayload, responsePayload string) error {
+	return c.storage.Store(ctx, inferenceId, epochId, promptPayload, responsePayload)
+}
+
+func (c *CachedStorage) Retrieve(ctx context.Context, inferenceId string, epochId uint64) (string, string, error) {
+	c.mu.RLock()
```
0xBECEDA: I think we could use sync.Map here too

decentralized-api/payloadstorage/file_storage.go
```diff
@@ -0,0 +1,108 @@
...
+func filenameToInferenceId(filename string) (string, error) {
+	decoded, err := hex.DecodeString(filename)
+	if err != nil {
+		return "", fmt.Errorf("decode hex filename: %w", err)
+	}
+	return string(decoded), nil
+}
+
+// Atomic write: temp file + rename
+func (f *FileStorage) Store(ctx context.Context, inferenceId string, epochId uint64, promptPayload, responsePayload string) error {
```
0xBECEDA: What happens if two parallel calls to Store() end up generating the same tempPath (for example, when inferenceId and epochId are identical)?
Could this cause a race when writing the .tmp file, potentially resulting in a failed rename or one process’s data being lost?

iamoeco (approved):

decentralized-api/payloadstorage/postgres_storage.go
```diff
@@ -0,0 +1,165 @@
...
+
+const createTableSQL = `
+CREATE TABLE IF NOT EXISTS inferences (
+    epoch_id BIGINT NOT NULL,
+    inference_id TEXT NOT NULL,
+    prompt_payload BYTEA,
+    response_payload BYTEA,
+    prompt_hash TEXT,
+    response_hash TEXT,
+    created_at TIMESTAMP DEFAULT NOW(),
```
blizko, line 25: Consider adding an extra BTREE like index for `created_at`. It is highly likely to be used for a range scan.
In case prompt and response hashes are used for direct search - might consider adding an extra HASH index

deploy/join/docker-compose.postgres.yml
```diff
@@ -0,0 +1,46 @@
...
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:?POSTGRES_PASSWORD is required}
+      POSTGRES_DB: ${POSTGRES_DB:-payloads}
+    volumes:
+      - ./postgres:/var/lib/postgresql
+    ports:
+      - "127.0.0.1:5432:5432"
+    command:
+      - "postgres"
+      - "-c"
+      - "min_wal_size=4GB"
```
blizko: Consider moving values to .env

deploy/join/docker-compose.postgres.yml
```diff
@@ -0,0 +1,46 @@
...
+      - "max_wal_size=16GB"
+      - "-c"
+      - "checkpoint_timeout=15min"
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER:-payloads} -d ${POSTGRES_DB:-payloads}"]
+      interval: 10s
+      timeout: 5s
+      retries: 5
+
+  db-backup:
```
blizko: Would DB backup be considered required for non-genesis nodes?
If not - the retention for 6 months seems too big

inference-chain/x/inference/module/confirmation_poc.go
```diff
@@ -411,6 +414,11 @@ func (am AppModule) updateConfirmationWeights(ctx context.Context, event *types.
 	am.LogInfo("updateConfirmationWeights: Confirmation weights", types.PoC,
 		"confirmationWeights", confirmationWeights)
 
+	notPreservedWeights, err := am.GetNotPreservedTotalWeightByParticipant(ctx, event.EpochIndex)
+	if err != nil {
```
Pegasus-starry: Is TimeslotAllocation[1] == false indicates a PoC miner vs. an inference server?

decentralized-api/cosmosclient/tx_manager/batch_consumer.go
gmorgachev: Good call, adding automatic extending till in memory

decentralized-api/payloadstorage/file_storage.go
gmorgachev: But inferenceId are unique, no? Not unique id would be overriden on chain too

decentralized-api/payloadstorage/postgres_storage.go
gmorgachev: Why do you think it's needed?
I expect this DB to be used only for direct access by index (epochId, inferenceId). Once payload is required for validation

deploy/join/docker-compose.postgres.yml
gmorgachev: Good catch, that's a copy-paste issue. DB table is partitioned tables per epoch, parts which are older then 2 days should be pruned (i also found a bug that pruning was disabled by default)
So, it definitely doesn't make sense  to store backup longer then 48 hours

decentralized-api/payloadstorage/postgres_storage.go
blizko: It depends on the pattern of further DB searches.
If it is expected to be only `(epochId, inferenceId)` it is fine.
It is a general question whether (for example network explorers) will eventually ask for specific data that might lead to index miss

blizko (approved):

gmorgachev:
> I have a general offchain prompt flow question, not the in-code implementation.
> 
> Here is the attack flow:
> 
> 1. User computes `original_prompt_hash`, signs it with their key, sends request to transfer agent REST API
> 2. Transfer agent verifies user signature, assigns executor
> 3. Transfer agent creates `prompt_payload` (adds seed, logprobs), computes `prompt_hash`
> 4. Transfer agent DON'T SEND ANYTHING TO EXECUTOR
> 
> So the prompt payload and hash are onchain and will be validated, but executor didn't execute anything, and will be striked for not serving request. How the new suggested protocol handle this attack? It wasn't covered in [Security analysis](https://github.com/gonka-ai/gonka/blob/gm/dev-0.2.6/proposals/offchain-payloads/README.md)
> 
> What happens if transfer agent takes user request and even don't send to blockchain but says to user everything is fine with your request? How protocol can strike the transfer agent? What should user do if he is sending prompts to malware transfer agent?
> 
> And finally, how transfer agent selects the executor? What blockchain data is used for executor selection? What if TA sends all requests to one executor to make a kind of DDoS?

Hi @akup  ! Thanks for the careful review of the payload transfer logic. The attacks you described are definitely valid, but they relate to the general asynchronous request processing system rather than the off-chain payload functionality itself. That is why they were not covered in the Security Analysis.

Currently, the Transfer Agent (TA) samples executors based on participant weight distribution. You can see the implementation in `query_get_random_executor.go` (https://github.com/gonka-ai/gonka/blob/main/inference-chain/x/inference/keeper/query_get_random_executor.go). The correct distribution is currently not strictly enforced, and this enforcement should be the foundation for addressing the problems you described

One possible solution we are considering is to gather miss-rate data for the entire Host x Host matrix (all TAs vs Executors) and implementing a retry mechanism on the TA side. The TA will record a `StartInference` transaction on-chain with both the attempted sampled executors and those who actually responded. This allows us to use stat-tests to validate behavior
E.g.: If a TA reports that executors are not responding while others reach them fine, the TA is likely at fault and will be punished.
If multiple TAs cannot reach an executor, that executor is likely experiencing downtime and will be penalized

There is still some space for behavior where a single Executor doesn't respond to a specific TA, but it financially doesn't make sense as such an Executor would miss their share of Work Coins

Additionally, after Epoch 180, all Hosts will have Collateral, which is being slashed in case of any fraudulent behavior. That will make punishment more effective

As for the case where a TA pretends to process a request but doesn't send it to the chain: this approach has no economic impact on the user since they are not charged. It implies users should prefer TAs with higher reputation. For the case where a TA produces a pseudo-valid response, it would make sense to have the response signed by the TA's private key automatically. Then it can be used as an artifact to claim fraudulent behavior, but the TA can still simply not respond

I think it's quite important to incentivize TAs with a small share of each inference transferred. Then they will be motivated to respond correctly

Btw, if you would want to contribute in any of these improvements, it would be awesome and would really  fit the Bounty Program. Based on your question you has required skills for such contribution :)

PR #469, opened 2025-12-04
[P1?] Certik Fixes

### LINKED ISSUE #468: Certik (finalization)

patimen:
# Certik Audit Fixes
Several major areas, and some smaller ones as well:
## Unified Permissions
Addressing both Certik issues and a general disorder around message permissions, we unify the permissions mostly in `permissions.go`. The `CheckPermissions` method is responsible for verifying the Signers have permissions to send the message, and every message sent goes through this, minus a few that need authority from EpochGroups.

Note that the permissions are actually specified twice: Once when CheckPermissions is called, so that it is clear in every message what permissions are needed. Second, in `permissions.go` itself, so that all permissions across the chain can be seen and reasoned about at once. Tests were added for ALL message permissions and verified to be effective using a python script.

### Top Miner
The "Top miner" logic has been deprecated in favor of our Bitcoin style rewards
### Work based rewards
Again, in favor of our bitcoin style rewards, the original code that issued rewards based on the amount of work done has been removed.
### Deprecated parameters
All parameters related to old features have been marked as `deprecated` for now.

## Removal of panics and "Must"
Several places were using panics in the event of error, presenting a risk to consensus if called during EndBlock or other crucial code paths. These have been switched to returning `error`, even when there is an apparent zero chance of an error happening (famous last words).

## Additional Validations
In addition to permissions, range checks, string size checks and array limits have been added for even more Messages to protect against possible DoS or inadvertent edge conditions.

## Minor Additions
- Add instructions.md files for Github Copilot to help improve AI code reviews
- Optimize retrieving epoch group data with `GetEpochGroupDataForEpoch`
- File naming standardization: msg_server precursor for ALL message processing.
- model_assignment.go now uses a separate Logging Category.


inference-chain/proto/inference/inference/params.proto
```diff
@@ -43,14 +43,14 @@ message GenesisOnlyParams {
 
 message TokenomicsParams {
   option (gogoproto.equal) = true;
-  Decimal subsidy_reduction_interval = 1;
-  Decimal subsidy_reduction_amount = 2;
-  Decimal current_subsidy_percentage = 3;
-  Decimal top_reward_allowed_failure = 4;
-  int64 top_miner_poc_qualification = 5;
+  Decimal subsidy_reduction_interval = 1 [deprecated = true];
```
DimaOrekhovPS: Why are we deprecating?

inference-chain/x/inference/keeper/mlnode_version.go
```diff
@@ -9,11 +9,15 @@ import (
...
 
 // SetMLNodeVersion set mlNodeVersion in the store
-func (k Keeper) SetMLNodeVersion(ctx context.Context, mlNodeVersion types.MLNodeVersion) {
+func (k Keeper) SetMLNodeVersion(ctx context.Context, mlNodeVersion types.MLNodeVersion) error {
 	storeAdapter := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
 	store := prefix.NewStore(storeAdapter, types.KeyPrefix(types.MLNodeVersionKey))
-	b := k.cdc.MustMarshal(&mlNodeVersion)
+	b, err := k.cdc.Marshal(&mlNodeVersion)
+	if err != nil {
+		return err
```
DimaOrekhovPS: So all this errors are now just propagated to message/query handler callers?

0xBECEDA (approved):

inference-chain/x/inference/module/module.go
```diff
@@ -282,9 +282,18 @@ func (am AppModule) EndBlock(ctx context.Context) error {
...
+				err = am.keeper.SetMLNodeVersion(ctx, types.MLNodeVersion{
 					CurrentVersion: pu.NodeVersion,
 				})
+				if err != nil {
+					am.LogError("Unable to update MLNode version", types.Upgrades, "error", err.Error())
+				}
+			}
+
+			// Track last upgrade height
+			err = am.keeper.SetLastUpgradeHeight(ctx, blockHeight)
```
jacky6block, line 294: Repeatedly call SetLastUpgradeHeight, line 300

PR #473, opened 2025-12-05
Change RTarget and scale all weight by 2.5

### LINKED ISSUE #584: Change RTarget and scale all weight by 2.5

DimaOrekhovPS:
Changing `RTarget` to increase the difficulty of PoC challenge in approximately 2.5 times, meaning participants are expected to generate 2.5 nonces less in the allotted time. To keep absolute values of PoC weights we scale all the weight by 2.5

### DIFF
--- a/inference-chain/x/inference/module/chainvalidation.go
+++ b/inference-chain/x/inference/module/chainvalidation.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log/slog"
+	"math"
 	"sort"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -676,6 +679,7 @@ func calculateParticipantWeight(batches []types.PoCBatch) ([]nodeWeight, int64)
 			}
 		}
 
+		weight = int64(math.Round(WeightScaleFactor * float64(weight)))
```
gmorgachev: i don't think we should use float
Decimal -> then to int

PR #482, opened 2025-12-10
feat: script to sync data

ChungKiet:


patimen:
Closing this as stale. The local/testnet setup has moved on since this was opened, and if we want something along these lines it should be reintroduced against the current flow rather than revived from this branch.

PR #494, opened 2025-12-14
Bugfix/validation recovery improvements

aleks1k:
This PR improves the validation recovery system by:
1. **Fixing validation detection logic**: Uses `InferenceValidationDetailsAll` instead of `InferenceAll` to avoid pruning issues
2. **Adding new recovery options**: 
   - `forceValidateAll` - bypass shouldValidate check
   - `useBlockchainLogic` - use blockchain's exact logic for determining missed validations
   - `maxValidations` - limit number of validations sent per request
3. **Code quality improvements**: Fixed indentation issues and removed unused variables
4. **Safety improvements**: Added nil checks and early returns for better error handling


decentralized-api/cosmosclient/tx_manager/tx_manager.go
DimaOrekhovPS, line 558: Why does the limit needs to be increased?

decentralized-api/internal/validation/inference_validation.go
```diff
@@ -163,100 +158,328 @@ func (s *InferenceValidator) getCurrentSupportedModels() (map[string]bool, error
...
-	queryClient types.QueryClient,
-) ([]types.Inference, error) {
-	// Filter inferences by epoch
-	var epochInferences []types.Inference
-	for _, inf := range inferencePage {
-		if inf.EpochId == epochIndex {
-			epochInferences = append(epochInferences, inf)
+// DetectMissedValidations identifies which validations were missed for a specific epoch
+// Returns a list of inference objects that the current participant should have validated but didn't
+func (s *InferenceValidator) DetectMissedValidations(epochIndex uint64, seed int64, forceValidateAll bool) ([]types.Inference, error) {
```
DimaOrekhovPS, line 163: Why are you modifying this function and then also introducing a separate `DetectMissedValidationsUsingBlockchainLogic`? 

It just seems that this function `DetectMissedValidations` now also uses blockchain logic (use `InferenceValidationDetails` instead of `Inference`), or am I missing something here?

DimaOrekhovPS (commented):
Do we really need `forceValidateAll`? Is it for people who accidentally lost both their seed and previous warm key?

PR #503, opened 2025-12-22
Upgrade v0.2.7

gmorgachev:


decentralized-api/cosmosclient/tx_manager/errors.go
```diff
@@ -18,6 +20,98 @@ var (
...
+	// HTTP gateway errors
+	"post failed",
+	"bad gateway",
+	"service unavailable",
+	"gateway timeout",
+	// RPC errors
+	"rpc error",
+	"aborted",
+	// OS resource exhaustion
+	"too many open files",
```
blizko, line 58: Debatable. That is not always a transient failure, but might indicate that the OS simply is not configured properly to serve the load.

decentralized-api/cosmosclient/tx_manager/errors.go
```diff
@@ -18,6 +20,98 @@ var (
...
+	// Sequence errors (safety net for non-unordered scenarios)
+	"account sequence mismatch",
+	"incorrect account sequence",
+
+	"unordered transaction has a timeout_timestamp that has already passed",
+	"unordered tx ttl exceeds",
+}
+
+// isRetryableRawLog checks if the raw log contains any retryable error patterns
+func isRetryableRawLog(rawLog string) bool {
```
blizko, line 70: Analysis of logs by using textual search should be the last resort, since it is resource intensive. If possible - only strict status codes should be used.

decentralized-api/internal/nats/server/server.go
```diff
@@ -12,10 +12,12 @@ import (
...
-	TxsBatchFinishStream = "txs_batch_finish"
+	TxsToSendStream             = "txs_to_send"
+	TxsToObserveStream          = "txs_to_observe"
+	TxsBatchStartStream         = "txs_batch_start"
+	TxsBatchFinishStream        = "txs_batch_finish"
+	TxsBatchPocBatchStream      = "txs_batch_poc_batch"
+	TxsBatchPocValidationStream = "txs_batch_poc_validation"
 
 	storageDir    = "/root/.dapi/.nats"
 	defaultMaxAge = 24 * 60 * 60 // 24 hours
```
blizko, line 23: Consider lowering the max age, since majority of messages seem to become irrelevant in 5 minutes?

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -460,6 +544,24 @@ func (s *Server) handleExecutorRequest(ctx echo.Context, request *ChatRequest, w
 	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
blizko, line 544: Note extra caution for Response code 5xx.
In certain cases that indicates a problem with the inference engine that is not transient, but permanent (e.g. manipulation with model behaviour)

decentralized-api/internal/validation/inference_validation.go
```diff
@@ -865,12 +902,30 @@ func (s *InferenceValidator) validateWithPayloads(inference types.Inference, inf
...
 
 	respBodyBytes, err := io.ReadAll(resp.Body)
 	if err != nil {
 		return nil, err
 	}
 
+	// If the validator's inference node rejects the payload (400/422), treat validation as passed.
+	// This can happen when the original inference could not be executed due to upstream payload rejection,
+	// and validators on older versions may still attempt re-execution.
+	if resp.StatusCode == http.StatusBadRequest || resp.StatusCode == http.StatusUnprocessableEntity {
```
blizko, line 915: Extra node - whilst vLLM does not send by default 429 (Too many requests), that might be the case with other engines.
Double edged sword - a validator might be already out of capacity because of requests, or on the other hand - manipulating with the engine to not serve requests.

inference-chain/app/upgrades/v0_2_7/upgrades.go
```diff
@@ -0,0 +1,3404 @@
...
+
+	// Participant access gating:
+	// - Block NEW participant registrations until this height (registration opens at this height).
+	// - Blocklisted accounts cannot participate in PoC submissions.
+	// - Allowlist is disabled by default; can be enabled via governance later.
+	// NOTE: 2,222,222 is roughly ~2 weeks after the upgrade at typical block times.
+	params.ParticipantAccessParams = &types.ParticipantAccessParams{
+		NewParticipantRegistrationStartHeight: 2_222_222,
+		BlockedParticipantAddresses:           []string{"gonka1blockedxxxxxxxxxxxxxxxxxxxxxx"}, // placeholder; update before the upgrade
+		UseParticipantAllowlist:               false,                                           // disabled by default
```
blizko, line 3347: Note that allowlist is not activated in the PR, however it was announced

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -16,6 +16,11 @@ func (k msgServer) StartInference(goCtx context.Context, msg *types.MsgStartInfe
 	var ctx sdk.Context = sdk.UnwrapSDKContext(goCtx)
 	k.LogInfo("StartInference", types.Inferences, "inferenceId", msg.InferenceId, "creator", msg.Creator, "requestedBy", msg.RequestedBy, "model", msg.Model)
 
+	// Developer access gating: before the cutoff height, only allowlisted developers may request inferences.
+	if k.IsDeveloperAccessRestricted(ctx, ctx.BlockHeight()) && !k.IsAllowedDeveloper(ctx, msg.RequestedBy) {
```
blizko, line 20: What happens if a participant removes this check in his binary? In this case the TA will pass the request to chain? Or this is the check on the executor node?

inference-chain/app/upgrades/v0_2_7/upgrades.go
```diff
@@ -0,0 +1,3404 @@
...
+
+	return k.SetParams(ctx, params)
+}
+
+func distributeBountyRewards(ctx context.Context, k keeper.Keeper, distrKeeper distrkeeper.Keeper) error {
+	sections := []struct {
+		name     string
+		bounties []BountyReward
+	}{
+		{"epoch_117", epoch117Rewards},
```
blizko, line 3359: As noted by the community -  these rewards will have immediate vesting that would impact the current market volume.

decentralized-api/internal/nats/server/server.go
gmorgachev: Now each transaction has it's own deadline in blocks. But you're right, it should be lowered a lot too (it's configurable from api-config btw)

PR #505, opened 2025-12-23
Security Fixes for v0.2.7

ouicate:
This PR addresses five security vulnerabilities:

### 1. SSRF and Request-Hang DoS via Participant-Controlled InferenceUrl
- **Problem:** Participants could set `InferenceUrl` to internal IPs (SSRF) or slow endpoints (DoS)
- **Fix:** Adds SSRF protection to participant URL validation, rejecting private/internal IP ranges (localhost, RFC1918, link-local, AWS metadata endpoint). Also adds 30-second HTTP client timeout to prevent request-hang DoS from slow executor URLs.
- Commit: 85e6af1

### 2. PoC Validation Overwrite Enabling Vote Flipping  
- **Problem:** Validators could re-submit validations, overwriting their vote after seeing others
- **Fix:** Prevents vote flipping by rejecting duplicate PoC validations. Adds `HasPoCValidation()` check before storing to enforce first-write-wins semantics, preventing validators from changing their vote after seeing others' votes.
- Commit: 897ef58

### 3. MsgSubmitPocBatch Lacking Size/Count Bounds
- **Problem:** Arbitrarily large batches could cause state bloat and block-space DoS
- **Fix:** Adds maximum size limits to prevent state bloat and block-space DoS from arbitrarily large PoC batches.
- Commit: 5f29222

### 4. Ineffective PoC Exclusion for Inference-Serving Nodes
- **Problem:** [getInferenceServingNodeIds()](/inference-chain/x/inference/module/chainvalidation.go:743:0-783:1) always returned empty set due to using parent epoch group
- **Fix:** Fixes PoC exclusion filter that was always returning empty set. Rewrites `getInferenceServingNodeIds()` to use `GetActiveParticipants()` which properly stores MlNodes data, instead of parent epoch group which returns nil for MlNodes.
- Commit: 5fc4db4

### 5. Epoch-Based Authorization Bypass + Signature Replay Attack
- **Problem:** Attackers could bypass auth by providing epoch where they ARE active; signatures didn't bind epochId
- **Fix:** Fixes two related vulnerabilities: (1) Authorization bypass where attackers could provide an epoch header where they ARE active to access payloads from a different epoch - now queries inference's actual epoch first. (2) Signature replay attack - signatures now include epochId to prevent replay with different epoch headers.
- Commit: 8853af8


decentralized-api/internal/server/public/payload_handlers.go
```diff
@@ -175,10 +182,13 @@ func (s *Server) verifyActiveParticipant(ctx echo.Context, address string, epoch
...
-// Validator signs: inferenceId + timestamp + validatorAddress
-func validatePayloadRequestSignature(inferenceId string, timestamp int64, validatorAddress string, pubkeys []string, signature string) error {
+// Validator signs: inferenceId + timestamp + validatorAddress + epochId
+// EpochId binding prevents replay attacks within epoch windows
+func validatePayloadRequestSignature(inferenceId string, timestamp int64, validatorAddress string, epochId uint64, pubkeys []string, signature string) error {
+	// Include epochId in payload to bind signature to specific epoch
+	payloadWithEpoch := inferenceId + strconv.FormatUint(epochId, 10)
 	components := calculations.SignatureComponents{
-		Payload:         inferenceId,
+		Payload:         payloadWithEpoch,
```
patimen: Simply appending the epoch to the payload is obtuse and hard to observe. Instead, let's add Epoch to the SignatureComponent and do the appending in the same place we do all signature byte creation.

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -39,6 +39,10 @@ const (
 	BothContexts = TransferContext | ExecutorContext
 )
 
+// HTTP client with timeout for transfer agent requests to executor nodes
+// Prevents request-hang DoS attacks from malicious executor URLs
+var executorHttpClient = utils.NewHttpClient(30 * time.Second)
```
patimen: Inference requests can be much longer than 30 seconds. If we are going to protect against request-hangs, it will need to be far more sophisticated than the simple timeout provided here.

inference-chain/x/inference/keeper/msg_server_submit_poc_validation.go
```diff
@@ -52,6 +52,20 @@ func (k msgServer) SubmitPocValidation(goCtx context.Context, msg *types.MsgSubm
 			return nil, sdkerrors.Wrap(types.ErrPocTooLate, "Confirmation PoC validation window closed")
 		}
 
+		// Check if validation already exists (prevent vote flipping)
```
patimen: This is a good implementation, but we now have even more code duplication than we did before (with the branches for confirmation PoC vs sprint PoC). Can we dedupe the logic a bit, maybe include some more of the surrounding code?

inference-chain/x/inference/keeper/poc_batch.go
```diff
@@ -50,6 +50,15 @@ func (k Keeper) GetPoCBatchesCountByStage(ctx context.Context, pocStageStartBloc
 	return count, nil
 }
 
+// HasPoCValidation checks if a validation already exists for the given key combination.
+// Returns true if a validation exists, false otherwise.
+func (k Keeper) HasPoCValidation(ctx context.Context, pocStageStartBlockHeight int64, participantAddress string, validatorAddress string) (bool, error) {
+	pAddr := sdk.MustAccAddressFromBech32(participantAddress)
```
patimen: Let's use AccAddressFromBech32 here instead of Must, then return an error. We don't want to have potential panics lying around waiting for someone to call HasPoCValidation from EndBlock or similar consensus code.

inference-chain/x/inference/utils/signature_and_url_validation.go
```diff
@@ -43,3 +45,89 @@ func ValidateURL(fieldName, raw string) error {
...
+	// Parse as IP and check for private ranges
+	ip := net.ParseIP(host)
+	if ip != nil {
+		if isPrivateIP(ip) {
+			return errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "%s cannot point to private IP address", fieldName)
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
patimen: I think, while this is a good first step, it's mostly defense-in-depth. SSRF is a good weakness to highlight, but ultimately the checks need to be on the client/TA side, as we need to involve DNS and, additionally, ValidateBasic can be bypassed by malicious agents.

inference-chain/x/inference/types/message_submit_poc_batch.go
```diff
@@ -11,6 +11,16 @@ import (
 
 var _ sdk.Msg = &MsgSubmitPocBatch{}
 
+// Maximum limits for PoC batch to prevent state bloat and block-space DoS
+const (
+	// MaxPocBatchNonces is the maximum number of nonces allowed per batch
+	MaxPocBatchNonces = 10000
```
patimen, line 17: Let's double check this against the numbers we are seeing on-chain from our biggest miners.

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -741,7 +741,8 @@ func calculateValidationOutcome(currentValidatorsSet map[string]int64, validatio
 	}
 }
 
-// getInferenceServingNodeIds returns a set of node IDs that have POC_SLOT = true in the current epoch
+// getInferenceServingNodeIds returns a set of node IDs that have POC_SLOT = true in the current epoch.
+// Uses ActiveParticipants which properly stores MlNodes (unlike parent epoch groups which have nil MlNodes).
```
patimen, line 790: This is a good insight. Parent epoch groups (with modelId = "") will not have any MLNode information, as can be seen in `EpochGroup.getMLNodeInfo`.
Since we are looking for global exclusions, I believe your approach here should work.

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -795,25 +796,54 @@ func (am AppModule) getInferenceServingNodeIds(ctx context.Context, upcomingEpoc
...
-	// Find all nodes with POC_SLOT = true
-	for _, validationWeight := range currentEpochGroup.GroupData.ValidationWeights {
-		for _, mlNode := range validationWeight.MlNodes {
-			if len(mlNode.TimeslotAllocation) > 1 && mlNode.TimeslotAllocation[1] { // POC_SLOT = true
-				inferenceServingNodeIds[mlNode.NodeId] = true
-				am.LogInfo("getInferenceServingNodeIds: Found inference-serving node", types.PoC,
-					"nodeId", mlNode.NodeId,
-					"participantAddress", validationWeight.MemberAddress)
+	// Get active participants for the effective epoch - this properly stores MlNodes
+	activeParticipants, found := am.keeper.GetActiveParticipants(ctx, effectiveEpochIndex)
```
DimaOrekhovPS, line 807: We need to account for excluded participants here. Active participants are effectively a snapshot of the epoch group at the time of epoch start.

During the course of the epoch some might be excluded

@patimen Is it reliable to use excluded participants here? Or should we filter using the main epoch group?

patimen: Excluded participants can still participate in PoC's, right? They participate in the next PoC and get back on the network, exclusion isn't for multiple Epochs.

DimaOrekhovPS: Yes, since the purpose of this function is to get inference serving nodes for filtration only it should be ok. I agree, we can keel it as is

PR #507, opened 2025-12-24
Fix a bug in claim recovery

### LINKED ISSUE #511: Bug in claim recovery

patimen:
The query GetInferenceValidationParameters returned only a single ValidatorPower, despite checking both the previous and current epochs for the inferences. This meant that the claim recovery process used the validator power for the CURRENT epoch, NOT the previous. This meant that if power went down between epochs, fewer validations would be required and they would NOT be recovered, leading to missed ClaimRewards.

We fix this by returning the ValidatorPower for all Epochs that were retrieved. Note that we specifically return them as a list of Power/Epoch pairs. Then we use the Power for the epoch of the recovered Inference.

Note that this change should be not impact consensus, allowing impacted users to switch immediately.


decentralized-api/internal/validation/inference_validation.go
```diff
@@ -334,8 +334,12 @@ func (s *InferenceValidator) DetectMissedValidations(epochIndex uint64, seed int
 						logging.Error("Failed to get validator power", types.ValidationRecovery, "error", err)
 						return nil, fmt.Errorf("failed to get validator power: %w", err)
 					}
-					validatorPower = powerResp.ValidatorPower
-					validatorPowerFetched = true
+					for _, power := range powerResp.ValidatorPowers {
```
akup, line 339: If we always claim the previous epoch inferences, why we need validatorPowers including previous and current epoch?
It is enough to have previous epoch validator power.

patimen: I agree this is true... for now. But the code here is written not to know that (for sure)... I think it is cleaner to use the EpochId for the Inference, because that is the only data point that matters and that we can control.

PR #514, opened 2025-12-30
Improve TXManager reliability, add block-based deadlines, don't send invalid TXs

### LINKED ISSUE #590: Improve TXManager reliability, add block-based deadlines, don't send invalid TXs

gmorgachev:
## feat(tx_manager, validation): improve tx reliability and add block-based deadlines

Overhaul transaction management to prevent wasteful retries, add deadline expiration, and fix validation edge cases for inactive participants.

### Changes

**tx_manager: transaction deadlines**
- Add `DeadlineBlock` field to track when transactions become meaningless
- Configure per-message-type limits: PoC ~240 blocks, inference ~150 blocks
- Drop expired transactions before processing instead of retrying for hours
- Auto-calculate deadline from message type, allow caller override

**tx_manager: error classification**
- Implement three-way response classification (Observe/Retry/Fail)
- Allow-list approach: only transient errors (mempool full, timeout, sequence mismatch) retry
- Business errors (duplicate validation, POC too late, etc.) fail immediately
- Before: 100 retries over 12+ hours; After: instant failure for invalid txs

**tx_manager: retry improvements**
- Replace `NumDelivered` with payload-based `Attempts` tracking
- Add `RequeueTime` field to prevent tight retry loops
- Add jittered delays (±30%) to prevent thundering herd
- Increase max attempts to 100 for transient errors

**validation: skip inactive participants**
- Add `isActiveInCurrentEpoch()` check before recovery validations
- Skip confirmation PoC for non-active participants
- Prevents `ErrParticipantNotFound` errors

**poc: security fix**
- Use fresh block hash for nonce sampling instead of PoC start hash
- Prevents attacker from precomputing sampled nonces

### Files Changed

- `tx_manager/tx_manager.go` - core retry/deadline logic
- `tx_manager/errors.go` - response classification
- `tx_manager/tx_deadline_config.go` - per-message deadlines
- `validation/inference_validation.go` - epoch activity checks
- `event_listener/new_block_dispatcher.go` - PoC skip logic
- `poc/node_orchestrator.go` - fresh hash for sampling

### DIFF
--- a/decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
+++ b/decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
@@ -1,7 +1,7 @@
 package tx_manager
 
 // defaultMaxBlocks is the default deadline in blocks (500 blocks ≈ 40 minutes at 5s/block)
-const defaultMaxBlocks int64 = 500
+const defaultMaxBlocks int64 = 30
 
 // deadlineByMsgType defines message-specific deadlines based on chain epoch parameters
 var deadlineByMsgType = map[string]int64{
--- a/decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
+++ b/decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
@@ -1,0 +1,22 @@
+package tx_manager
+
+// defaultMaxBlocks is the default deadline in blocks (500 blocks ≈ 40 minutes at 5s/block)
+const defaultMaxBlocks int64 = 500
+
+// deadlineByMsgType defines message-specific deadlines based on chain epoch parameters
+var deadlineByMsgType = map[string]int64{
+	"/inference.inference.MsgSubmitPocBatch":      240,
+	"/inference.inference.MsgSubmitPocValidation": 240,
+	"/inference.inference.MsgFinishInference":     150,
+	"/inference.inference.MsgValidation":          150,
+	"/inference.inference.MsgStartInference":      150,
+}
+
+// getMaxBlocksForType returns the maximum blocks a transaction type can wait before expiring.
+// Returns defaultMaxBlocks if the message type is not in the configuration.
+func getMaxBlocksForType(msgType string) int64 {
+	if blocks, ok := deadlineByMsgType[msgType]; ok {
+		return blocks
+	}
+	return defaultMaxBlocks
+}
--- a/decentralized-api/cosmosclient/tx_manager/tx_manager.go
+++ b/decentralized-api/cosmosclient/tx_manager/tx_manager.go
@@ -48,6 +48,8 @@ const (
 
 	hashHeader = "TX_HASH"
 	idHeader   = "TX_ID"
+
+	maxBlockTimeDrift = 120 * time.Second
 )
 
 type TxManager interface {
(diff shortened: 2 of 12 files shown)


decentralized-api/internal/validation/inference_validation.go
```diff
@@ -645,6 +653,21 @@ func (s *InferenceValidator) isEpochStale(inferenceEpochId uint64) bool {
 	return epochState.LatestEpoch.EpochIndex >= inferenceEpochId+2
 }
 
+func (s *InferenceValidator) isActiveInCurrentEpoch() bool {
+	queryClient := s.recorder.NewInferenceQueryClient()
```
akup, line 657: We do not need to send a new query on every isActiveInCurrentEpoch check, as participants are updated every epoch so it's better to use the cache that is already implemented at [new_block_dispatcher.go](https://github.com/gonka-ai/gonka/pull/514/files#diff-74b84dd74d5619994aee6fe4c7f109ad10e14a85a57935decdec082d52fe8524)

decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
```diff
@@ -0,0 +1,22 @@
+package tx_manager
+
+// defaultMaxBlocks is the default deadline in blocks (500 blocks ≈ 40 minutes at 5s/block)
+const defaultMaxBlocks int64 = 30
+
+// deadlineByMsgType defines message-specific deadlines based on chain epoch parameters
+var deadlineByMsgType = map[string]int64{
+	"/inference.inference.MsgSubmitPocBatch":      240,
+	"/inference.inference.MsgSubmitPocValidation": 240,
+	"/inference.inference.MsgFinishInference":     150,
```
akup, line 10: MsgFinishInference and MsgStartInference deadlines are not in sync with timestampExpirationNs and timestampAdvanceNs parameters.

150 blocks is around 20-25 minutes. There are extra-seconds at inference-chain 60*60=1 hour that makes acceptable FinishInference during this hour.

And MsgStartInference should be sent before expiring, and expiration is regulated by timestampExpirationNs that is set to 1 minute.
If we are retrying to send MsgStartInference, it should be resigned with new request timestamp, but not just resent from the queue.

akup: Removed some from comment below as inference-chain can process MsgStartInference and MsgFinishInference out of order.

decentralized-api/internal/validation/inference_validation.go
patimen: ActiveParticipants are updated every Epoch, but EpochGroupData is updated more often. For instance, when a participant is marked as Invalid, their weight is set to 0 in the EpochGroupData.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -178,29 +187,65 @@ func (m *manager) SendTransactionAsyncWithRetry(rawTx sdk.Msg) (*sdk.TxResponse,
...
+		deadlineBlock = deadlineBlockOpt[0]
+	} else {
+		var minBlocks int64 = defaultMaxBlocks
+		for _, msg := range msgs {
+			if blocks := getMaxBlocksForType(sdk.MsgTypeURL(msg)); blocks < minBlocks {
+				minBlocks = blocks
+			}
+		}
+		deadlineBlock = m.getLatestBlockHeight() + minBlocks
+	}
```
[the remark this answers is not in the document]
patimen: The messages should be all the same type.

PR #525, opened 2026-01-06
fix: merge all generated batches before sending to dapi

### LINKED ISSUE #325: [P0] No inference failures on the edge of PoC; Check for missed validation during epoch

icydark:
## Problem
When MLNode retries sending ProofBatch requests to DAPI after failures, it was sending each batch individually. When retries succeed, multiple batches are sent almost simultaneously, causing unnecessary network overhead.

## Solution
Merge all batches in `generated_not_sent` list into a single batch before sending.

### Changes

**File**: `mlnode/packages/pow/src/pow/service/sender.py`

**Modified `_send_generated()` method**:
- Previous: Iterated through `generated_not_sent` list and sent each batch individually. Failed batches will be added into `generated_not_sent` list and sent during next call individually.
- Now: Merges all batches in `generated_not_sent` using `ProofBatch.merge()` and sends a single merged batch

## Potential Limitation

**Large merged batches**: If many batches accumulate, the merged batch could be large.

## Reviewers
@tamazgadaev @gmorgachev


tcharchian:
@icydark 
Actually, this PR is on the outdated code, the previous version of PoC
Once we switched to PoC v2, we couldn’t merge it

PR #526, opened 2026-01-06
Membership for correct epoch for Validation requests

patimen:
Especially for validations for the Previous Epoch (common with retries), membership MUST be determined by the active participants THEN, not in the current epoch, so we use the inference EpochId


patimen:
Git history is all scrambled, this was a simple change. Created a new branch and PR.

PR #533, opened 2026-01-07
feat: add connection pool for query clients

x0152:
Implements [TODO] from proposals/mlnode-upgrade/README.md: "Connection Pool with Auto-Healing"

Problem: single gRPC connection is fragile one EOF error breaks all 30+ blockchain query systems.

Solution:

Connection pool with configurable size
Round-robin load balancing
Health checks every 30s with auto-reconnect
Graceful degradation (fallback to single connection)
17 unit tests
Config:
chain_node:
connection_pool_size: 5

Disabled by default (pool_size = 0)


decentralized-api/cosmosclient/connection_pool.go
```diff
@@ -0,0 +1,183 @@
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	for i := 0; i < n; i++ {
+		if newClients[i] != nil {
+			oldClient := p.clients[i]
+			p.clients[i] = newClients[i]
+			p.healthy[i].Store(true)
+			if oldClient != nil && oldClient.RPC != nil {
+				oldClient.RPC.Stop()
```
patimen: I’m a bit concerned about how client shutdown is currently handled. Calling RPC.Stop() only stops the Tendermint RPC client, but the cosmosclient.Client likely involves other background processes — potentially including gRPC connections — that aren’t being stopped here.

In Go, the idiomatic way to manage long-lived clients or processes is via context.Context. If a client accepts a context, that context should be used to control its lifecycle. Relying on partial shutdown methods (like Stop()) risks leaking resources, especially under conditions like frequent client replacement.

Recommendation:
Derive a new cancelable context (context.WithCancel(poolCtx)) for each client. Use this context when creating the client, and cancel it when the client is replaced or shut down. This approach allows:
	•	Individual clients to be canceled safely
	•	All clients to be canceled at once by canceling the shared pool context
	•	Better resource cleanup for any background goroutines or dialers tied to the context

It's subtle but I think the right way to go here.

patimen: Additiionally, since we're locked in this context, could this cause contention/async issues?

decentralized-api/cosmosclient/connection_pool.go
```diff
@@ -0,0 +1,183 @@
...
+		select {
+		case <-p.ctx.Done():
+			return
+		case <-ticker.C:
+			p.check()
+		}
+	}
+}
+
+func (p *ConnectionPool) check() {
```
patimen, line 146: I think we're ok for now, but this seems like a possible issue if we ever crank up the number of pools. A lot of synchronous check work, all failing (in the event of a slowdown or failure)... Perhaps a comment or warning in ConnectionPoolSize is set beyond, say, 10?

patimen (changes requested):
We really do need to use Context for stop/cancel of Clients in the pool to be sure we won't have resource leaks.

decentralized-api/cosmosclient/connection_pool.go
x0152: Good point. Added per client cancelable contexts and cancel on replace/close to fully stop background work. Thanks

x0152: Fixed
RPC.Stop() and cancel() are now executed outside p.mu

decentralized-api/cosmosclient/connection_pool.go
x0152: Parallel checks now are capped (by semaphore) and large pools log a warning. What do you think?

decentralized-api/cosmosclient/connection_pool.go
```diff
@@ -0,0 +1,270 @@
...
+		case <-p.ctx.Done():
+			return
+		case <-time.After(p.jitterDuration(p.interval)):
+			p.check()
+		}
+	}
+}
+
+func (p *ConnectionPool) jitterDuration(base time.Duration) time.Duration {
+	factor := 1 + ((p.rand.Float64()*2 - 1) * DefaultHealthJitter)
```
patimen, line 141: I think we're ok for now, but I wanted to call out that math/rand is _not thread safe_. We have only one health loop, so it isn't an issue yet.

decentralized-api/cosmosclient/connection_pool_test.go
```diff
@@ -0,0 +1,347 @@
...
+		cancel:  cancel,
+	})
+
+	// Each call should cycle through clients
+	got1, _ := p.Get()
+	got2, _ := p.Get()
+	got3, _ := p.Get()
+	got4, _ := p.Get()
+
+	if got1 == got2 && got2 == got3 {
```
patimen: This doesn't seem like a very complete test... shouldn't all 3 gets be distinct? More like:
`got1 == got2 || got2 == got3 || got1 == got3`?

decentralized-api/cosmosclient/connection_pool.go
```diff
@@ -0,0 +1,270 @@
...
+		return false
+	}
+	ctx, cancel := context.WithTimeout(p.ctx, p.timeout)
+	defer cancel()
+	_, err := c.RPC.Status(ctx)
+	if err != nil {
+		logging.Warn("connection pool: ping failed", types.System, "error", err)
+	}
+	return err == nil
+}
```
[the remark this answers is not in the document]
patimen: @x0152 This seems legit? If we only get the RPC status, we aren't actually detecting the issue we're trying to fix.

decentralized-api/cosmosclient/connection_pool.go
```diff
@@ -0,0 +1,270 @@
...
+		} else {
+			p.healthy[i].Store(false)
+			unhealthy++
+		}
+	}
+	p.mu.Unlock()
+	logging.Info("connection pool: health check result", types.System,
+		"replaced", replaced, "healthy", healthy, "unhealthy", unhealthy, "total", len(pingResults))
+	for _, cancel := range toCancel {
+		cancel()
```
patimen: All this canceling and stopping is good... but should it happen immediately? What if we have pending queries? All of them will be canceled as well, even if they were going to succeed.
Do we want to instead fire this off on a delay? 
I know we're getting pretty intricate here, but this is a very important part of our system, we want to be precise.

decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -159,7 +168,7 @@ func NewInferenceCosmosClient(ctx context.Context, addressPrefix string, config
 		}
 	}()
 
-	mn, err := tx_manager.StartTxManager(ctx, &cosmoclient, apiAccount, time.Second*60, natsConn, accAddress, config.GetHeight)
+	mn, err := tx_manager.StartTxManager(ctx, cosmoclient, apiAccount, time.Second*60, natsConn, accAddress, config.GetHeight)
```
patimen: We're passing in the base cosmos client here, right? Won't that mean that the TxManager doesn't use the connection pool?

patimen (changes requested):
I'm seeing a few more issues we're going to need to resolve.
This is a really important change, I appreciate the time you're taking with me to get it 100% right.

decentralized-api/cosmosclient/connection_pool.go
x0152: Valid point. It's easy to fix, so I added locking

decentralized-api/cosmosclient/connection_pool_test.go
x0152: Fixed

decentralized-api/cosmosclient/connection_pool.go
x0152: Got your point. added a gRPC health check, switched from RPC.Status to RPC.Health and added delayed cancel handling as well

decentralized-api/cosmosclient/cosmosclient.go
x0152: Updated tx manager to use pool as well

PR #534, opened 2026-01-08
security: prevent SSRF via executor redirect

x0152:
Problem: `handleTransferRequest` uses `http.DefaultClient` which follows redirects automatically. A malicious executor could return 307 redirect to internal services.

Attack flow:

    User -> TA -> Malicious Executor
                        |
                  307 Redirect to http://localhost:9200/admin/v1/...
                        |
             TA follows redirect -> Internal Admin API
                        |
               Response goes back to attacker

Potential impact if exploitable:
- Steal private keys via GET /admin/v1/config
- Send transactions as TA via POST /admin/v1/tx/send
- Access internal endpoints (bridge, node management)
- Leak headers (Authorization, X-Transfer-Address)

Requirements for exploitation:
- Be a registered executor (pass PoC)
- Get selected for request
- Admin API reachable from TA network

Note: theoretical vulnerability from code review, not tested on live system.

Solution:
- Custom HTTP client with CheckRedirect returning http.ErrUseLastResponse
- Prevents following any redirects from executor

Tests:
- TestDefaultClientFollowsRedirects - shows vulnerable behavior
- TestNoRedirectClient - verifies fix works

### DIFF
--- a/decentralized-api/internal/poc/node_orchestrator.go
+++ b/decentralized-api/internal/poc/node_orchestrator.go
@@ -7,13 +7,17 @@ import (
 	cosmos_client "decentralized-api/cosmosclient"
 	"decentralized-api/logging"
 	"decentralized-api/mlnodeclient"
+	"errors"
+	"time"
 
 	"github.com/productscience/inference/x/inference/types"
 )
 
 const (
-	POC_VALIDATE_BATCH_RETRIES     = 5
-	POC_VALIDATE_SAMPLES_PER_BATCH = 200
+	POC_VALIDATE_BATCH_RETRIES         = 5
+	POC_VALIDATE_SAMPLES_PER_BATCH     = 200
+	POC_VALIDATE_GET_NODES_RETRIES     = 30
+	POC_VALIDATE_GET_NODES_RETRY_DELAY = 5 * time.Second
 )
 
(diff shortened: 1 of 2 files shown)


gmorgachev:
@patimen @DimaOrekhovPS

decentralized-api/internal/server/public/ssrf_test.go
```diff
@@ -0,0 +1,68 @@
...
+		http.Redirect(w, r, target.URL, http.StatusTemporaryRedirect)
+	}))
+	defer redirector.Close()
+
+	req, _ := http.NewRequest(http.MethodPost, redirector.URL+"/v1/chat/completions", bytes.NewReader([]byte(`{}`)))
+	req.Header.Set(utils.XInferenceIdHeader, "test-id")
+	req.Header.Set(utils.AuthorizationHeader, "Bearer key")
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{
```
DimaOrekhovPS: I think it might be good to extract client creation in a separate function and then use it both in `post_chat_handler.go` and `ssrf_test.go`

This way we can be sure we're testing the same client that's used in production

x0152: Makes sense, done

PR #536, opened 2026-01-08
perf: optimize participants endpoint with single balance query

x0152:
Problem: /v1/participants makes N separate gRPC calls for balances - one per participant. With 5000 participants this means 5000 queries, slow response (~15s) and high load on chain node.

Solution:

- New ParticipantsWithBalances query in blockchain
- Single call returns all participants with their balances
- Uses IterateAllBalances for efficient data fetch

Local benchmark (5000 participants):
- Before: 310ms
- After: 0.6ms
- Speedup: ~500x

Changes:
- inference-chain: new query endpoint
- decentralized-api: use new query in getAllParticipants

### DIFF
--- a/inference-chain/app/upgrades/v0_2_8/upgrades.go
+++ b/inference-chain/app/upgrades/v0_2_8/upgrades.go
@@ -2,10 +2,12 @@ package v0_2_8
 
 import (
 	"context"
+	"fmt"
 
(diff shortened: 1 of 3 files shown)


gmorgachev:
great improvement! it also should spend much less resources this way

@patimen @DimaOrekhovPS please take a look

DimaOrekhovPS (approved):

PR #537, opened 2026-01-09
[P2] security: optional signature verification for MLnode PoC endpoints

### LINKED ISSUE #330: [P1] Security MerkleTree Proofs; Merge participant validation till block0; Need to add signature check at recording

x0152:
MLnode api endpoints are currently unauthenticated, allowing an attacker who can reach the mlnode port to hijack POC by sending their own callback_url

Adds request signing using the existing signerAccount key. Network Node signs requests with `X-Signature` header, mlnode verifies using SIGNER_PUBKEY environment variable. For standard deployment where mlnode is on the same machine, it works automatically since SIGNER_PUBKEY defaults to ACCOUNT_PUBKEY

For remote mlnode setups, you need to set SIGNER_PUBKEY manually. If SIGNER_PUBKEY is not set, verification is skipped for backward compatibility

Get public key on network node:
```
source config.env
inferenced keys show $KEY_NAME --pubkey --keyring-backend $KEYRING_BACKEND | jq -r '.key'
```

Start mlnode with signature verification:
```
export SIGNER_PUBKEY=<public_key_from_above>
docker-compose -f docker-compose.mlnode.yml up -d
```


akup (changes requested):
It is uncomplete patch, as there are following functions in MLNodeClient:

```go
// Training operations
	StartTraining(ctx context.Context, taskId uint64, participant string, nodeId string, masterNodeAddr string, rank int, worldSize int) error
	GetTrainingStatus(ctx context.Context) error

	// Node state operations
	Stop(ctx context.Context, num *int) error
	NodeState(ctx context.Context) (*StateResponse, error)

	// PoC operations
	GetPowStatus(ctx context.Context, num *int) (*PowStatusResponse, error)
	InitGenerate(ctx context.Context, dto InitDto, num *int) error
	InitValidate(ctx context.Context, dto InitDto, num *int) error
	ValidateBatch(ctx context.Context, batch ProofBatch, num *int) error

	// Inference operations
	InferenceHealth(ctx context.Context) (bool, error)
	InferenceUp(ctx context.Context, model string, args []string) error

	// GPU operations
	GetGPUDevices(ctx context.Context) (*GPUDevicesResponse, error)
	GetGPUDriver(ctx context.Context) (*DriverInfo, error)

	// Model management operations
	CheckModelStatus(ctx context.Context, model Model) (*ModelStatusResponse, error)
	DownloadModel(ctx context.Context, model Model) (*DownloadStartResponse, error)
	DeleteModel(ctx context.Context, model Model) (*DeleteResponse, error)
	ListModels(ctx context.Context) (*ModelListResponse, error)
	GetDiskSpace(ctx context.Context) (*DiskSpaceInfo, error)
```

Stop and InferenceUp are also critical to protect, as they are changing node state.
Other are get commands, but it could be more complete, to sign all the MLNode requests.

mlnode/packages/pow/src/pow/service/auth.py
```diff
@@ -0,0 +1,21 @@
+import os
+import base64
+import hashlib
+
+SIGNER_PUBKEY = os.environ.get("SIGNER_PUBKEY", "")
+
+def verify_signature(body: bytes, signature_b64: str) -> bool:
+    if not SIGNER_PUBKEY:
+        return True
```
IgnatovFedor: In `check_signature` from `routes.py` it's already checked `if not SIGNER_PUBKEY:`, so the same check inside `verify_signature` is redundant in the current usage pattern.

mlnode/packages/pow/src/pow/service/routes.py
```diff
@@ -36,23 +43,16 @@ async def init_generate(
     request: Request,
     init_request: PowInitRequestUrl
 ):
+    await check_signature(request)
```
IgnatovFedor: I'd use FastAPI [dependency / security dependency](https://fastapi.tiangolo.com/reference/dependencies) instead of manually calling await check_signature(request) in each handler.

For example:
```python
from fastapi.security import APIKeyHeader
signature_header = APIKeyHeader(name="X-Signature", auto_error=False)
async def check_signature(
    request: Request,
    signature: str | None = Security(signature_header),
):
    ...
@app.post("/pow/init")
async def init(
    _auth = Security(check_signature),
    ...
```
This keeps the auth logic centralized and makes it harder to forget on new endpoints.
Totally fine to keep the current approach.

x0152: Good point! Thanks!

tcharchian:
Hey @x0152 @akup! It would be great if you could sync on the next steps for this pull request and make the needed decisions together. If you can move it forward on your own, it could be included in v0.2.12. But overall, this is a nice-to-have rather than something critical.

akup:
I think it is useful PR, but it should be aligned with https://github.com/gonka-ai/gonka/pull/717
And we should merge it with new MLNode and https://github.com/gonka-ai/gonka/pull/417

The artifacts that are sent to callback URL also should be signed, as the attacker can send bad artifacts if the URL is not protected, and node will not pass validation on PoC.

I think we should first update main repo with new MLNode PoC apis (now PoC completely lives at https://github.com/gonka-ai/vllm repo). Then look and compare approaches with [Jf/token auth](https://github.com/gonka-ai/gonka/pull/717) and finally take into account with [websocket](https://github.com/gonka-ai/gonka/pull/417) that should go next

x0152:
@akup If the poc api update is coming, let's postpone #537 and #417 until after that. #717 can go in this upgrade - it's optional, shouldn't be affected by PoC API changes and enables cloud-hosted MLNode setups

#537 and #417 for security and stability should go on top of the new api once it is validated on mainnet, to avoid regressions

what do you think about that?

akup:
@x0152 yes https://github.com/gonka-ai/gonka/pull/717 can go in this release, but I want to align vision that best option of jf/token and request signatures go to main.
And hornestly, haven't time to dig it yet

But it seams that really it is better to align signatures in other PRs to jf/tokens that are present at https://github.com/gonka-ai/gonka/pull/717

x0152:
Agree, let's merge #717 in this release and postpone #537/ #417 to the next one

akup:
@x0152 I've added small PR that cleans the mlnode. Take a look: https://github.com/gonka-ai/gonka/pull/994

I think this one can be merged next

patimen:
@akup , @x0152 - Do we want to push on this one? You seemed ready to push it forward.

akup:
I think that certification of communication between devshard (or dapi) and mlnode should be certified. Because there are always some attacks on open ports and this is negative to adoption and newcomers.
I think it could be scheduled to 0.2.13

x0152:
Agreed - let's include this in v0.2.13

tcharchian:
Hi @x0152, are you ready to include this PR in the next upgrade?

x0152:
https://github.com/gonka-ai/gonka/pull/1329

PR #538, opened 2026-01-09
Update docker-compose.mlnode.yml and docker-compose.yml

rufatpro:
Ports 5050 and 8080 are restricted to local access only. Update docker-compose.mlnode.yml


Sunsec2027 (approved):

Sunsec2027 (approved):

Sunsec2027 (approved):

Sunsec2027 (approved):

patimen:
Closing this as stale. The concern here is valid, but addressing it as a narrow compose-only change is not enough by itself. This should be handled as part of the broader MLNode hardening/auth work in #412 rather than revived piecemeal from this PR.

PR #541, opened 2026-01-10
PoC validation, retry getting nodes

### LINKED ISSUE #325: [P0] No inference failures on the edge of PoC; Check for missed validation during epoch

DimaOrekhovPS:


decentralized-api/internal/poc/node_orchestrator.go
```diff
@@ -265,6 +265,31 @@ func (o *NodePoCOrchestratorImpl) ValidateReceivedBatches(pocStageStartBlockHeig
...
+		}
+
+		if attempt == POC_VALIDATE_GET_NODES_RETRIES-1 {
+			break
+		}
+		time.Sleep(POC_VALIDATE_GET_NODES_RETRY_DELAY)
+	}
+
+	logging.Error("ValidateReceivedBatches. Failed to get nodes after all retry attempts", types.PoC, "pocStageStartBlockHeight", pocStageStartBlockHeight, "numAttempts", POC_VALIDATE_GET_NODES_RETRIES)
+	return nil, nil
```
patimen: I'll suggest we should return an error here, rather than the check above for number of nodes (line 164). This will work, but it seems a bit obtuse and it works only because `len(nil) == 0`

patimen (approved):
One comment, we should probably fix that.

PR #542, opened 2026-01-10
perf: gRPC caching

x0152:
Add grpc query caching with per block invalidation. Even with ~6s block time, it reduces chain node load by deduplicating queries within a block. 

Cache overhead is ~200ns vs ~3ms for gRPC calls.


akup (changes requested):
ttl approach seams bad to me, and cache should be on block basis.
When we process new block, we should clear the cache.

When we are using ttl approach it is possible that cache will return same value when we are in a next block. This can introduce a lot of possible errors and shouldn't be even tryed.

while per-block cache is safe and is what we are really want from cache, as data is not changing before the next block.

x0152:
> ttl approach seams bad to me, and cache should be on block basis. When we process new block, we should clear the cache.
> 
> When we are using ttl approach it is possible that cache will return same value when we are in a next block. This can introduce a lot of possible errors and shouldn't be even tryed.
> 
> while per-block cache is safe and is what we are really want from cache, as data is not changing before the next block.

You're right
Was focused on dashboard API latency. Switched to per block invalidation. 

Removed http caching for now - that's a separate topic

decentralized-api/cosmosclient/query_cache.go
```diff
@@ -0,0 +1,143 @@
...
+
+type CachingConn struct {
+	inner grpc.ClientConnInterface
+}
+
+func NewCachingConn(inner grpc.ClientConnInterface) *CachingConn {
+	return &CachingConn{inner: inner}
+}
+
+func (c *CachingConn) Invoke(ctx context.Context, method string, args any, reply any, opts ...grpc.CallOption) error {
```
patimen: Shouldn't the grpc.CallOptions be passed on to the CachedInvoke?

decentralized-api/cosmosclient/query_cache.go
```diff
@@ -0,0 +1,143 @@
...
+	grpcCache.Purge()
+}
+
+func cacheKey(method string, req proto.Message) (string, error) {
+	data, err := proto.Marshal(req)
+	if err != nil {
+		return "", err
+	}
+	hash := sha256.Sum256(data)
+	return method + ":" + hex.EncodeToString(hash[:16]), nil
```
patimen: Why the truncation?

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -275,6 +275,9 @@ func (d *OnNewBlockDispatcher) ProcessNewBlock(ctx context.Context, blockInfo ch
 		return nil
 	}
 
+	// Clear cache only when synced to avoid thrashing during catch-up
+	cosmosclient.ClearCache()
```
patimen: This is clearing the cache AFTER we have tried to get the networkInfo above... meaning that it will be info from the PREVIOUS block, not the new one.

patimen: Specifically, take a look at line 170 (queryNetworkInfo)

patimen (changes requested):
I'm really not sure about this PR.
There is some interaction between clients and caches here that could be a big problem. I called one out, I fear there may be more.
Generally speaking, we would want a cache as a gRPC Interceptor, so the cache is tied to a long lived connection vs a global cache that could come up and bite us some time, say, if we have multiple connections.
I do like the idea, and I can see how this could be very useful. But I think we need to maybe step back a bit and think harder about how to do this.

x0152:
@patimen Thanks for the review! 

The safest cache idea is to make it height‑scoped and explicit: pass the height in context, then cache only for that height. That way every request clearly says which block it belongs to, and the developer controls it. Example:
```go
ctx := context.WithValue(ctx, HeightKey, height) // or attach height to request metadata
resp, err := queryClient.Inference(ctx, req)     // cache uses ctx height
```

But I want to ask about the current solution - I rewrote it to a simpler and less buggy version that is easier to read, but it still uses a global height‑pinned cache. If we talk to a single node and single backend URL, what risks do you still see with a shared cache? I only see risks if we ever talk to different nodes or have a load balancer/proxy between the dapi and the node, which is not the case for us right now

What do you think?

akup:
@x0152 could you please share your @telegram to me? Please mail me to kuprin.alexander@gmail.com

x0152:
@akup hit me up on discord please and I'll share my tg there if necessary

decentralized-api/apiconfig/config_manager.go
akup, line 73: It is unrelated to this PR.
and was already merged at https://github.com/gonka-ai/gonka/pull/644/changes
Maybe it is better to remove this commit (b3fee3d) from this PR.

akup:
@x0152 
Please clean the commits of this PR

Only 3 commits are actually about gRPC caching:
3e41e57 — perf: add LRU cache for gRPC queries (the core feature)
4764bbc — fix: height-pinned query cache
09f1a8a — fix: cache by response block height

There are a lot of already merged unrelated commits at this PR and it is hard to review

x0152:
> @x0152 Please clean the commits of this PR
> 
> Only 3 commits are actually about gRPC caching: [3e41e57](https://github.com/gonka-ai/gonka/commit/3e41e5701e2dacc39552b05cd9ec60adcb1cce39) — perf: add LRU cache for gRPC queries (the core feature) [4764bbc](https://github.com/gonka-ai/gonka/commit/4764bbcae7aac023280972137a44cd17bf7209dc) — fix: height-pinned query cache [09f1a8a](https://github.com/gonka-ai/gonka/commit/09f1a8a43ae36b9574d33ebea55392637bccfdaa) — fix: cache by response block height
> 
> There are a lot of already merged unrelated commits at this PR and it is hard to review

Done: rebased and cleaned

decentralized-api/cosmosclient/query_cache.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	if err := c.inner.Invoke(pinnedCtx, method, args, reply, callOpts...); err != nil {
+		return err
+	}
+
+	responseHeight := heightFromIncomingMD(*responseHeader)
+	if responseHeight == 0 {
+		return nil
+	}
```
akup, line 91: We make gRPC call with pinned height at x-csomos-block-height request header.
Maybe it's better to return err if responseHeight != height (should be height or explicitHeight)

decentralized-api/cosmosclient/query_cache.go
```diff
@@ -0,0 +1,193 @@
...
+
+	if data, err := marshalProtoMessage(reply); err == nil {
+		cacheKey := buildCacheKey(method, args, responseHeight)
+		c.cache.mu.Lock()
+		if c.cache.height == responseHeight {
+			c.cache.entries[cacheKey] = data
+		}
```
akup, line 98: this means that, if we are making request with some other explicitHeight it will never be cached.
I think we need some method to cache it too.

akup:
What I would like to see in the new cachedQuery, is the way to ask for some certain height and cache it too.

Now I can do something like:

```go
// Pin query to a specific block height
height := int64(2827908)
ctx := metadata.AppendToOutgoingContext(context.Background(),
    grpctypes.GRPCBlockHeightHeader, strconv.FormatInt(height, 10))
// Now any gRPC query using this ctx will be executed at that height
resp, err := queryClient.Params(ctx, &types.QueryParamsRequest{})
```

Now it will not be cached.

First when the height is updated everything is cleared, so cache works only during this block.
Second it will not be cached because:

```go
if c.cache.height == responseHeight {
  c.cache.entries[cacheKey] = data
}
```

It would be nice if cache for heights are cleaned on timeout (from latest use of the key) but not cleaned every block.

I don't think it will take a lot of memory. Or maybe we can clear every block latest block, and keep all keys where height != latestHeight

Practical case is when we are often getting some value before epoch change (it has certain height) and we don't need to clear it every block.

akup:
And I think caches should be enabled by default, so `query_cache_enabled` should be true by default.

patimen:
Can we move on this now? It's gone through a good number of iterations and is looking close, and could definitely still help perf.

tcharchian:
@x0152 @akup hey, should this be included in v0.2.13?

akup:
I think it is important to have caching. @x0152 are you going to take it for 0.2.13 ?

x0152:
Yes, let's take this in v0.2.13

tcharchian:
@x0152 are you ready to open this PR and rebase it for the next upgrade?

x0152:
https://github.com/gonka-ai/gonka/pull/1272

PR #543, opened 2026-01-10
fix: data race conditions

x0152:
Fix data races detected by 'go test -race' and review comments


IgnatovFedor (changes requested):
@x0152, could you please rebase onto upgrade-v0.2.11 and force-push?

slandymani (approved):

IgnatovFedor (approved):

decentralized-api/broker/node_worker_commands.go
```diff
@@ -25,8 +25,10 @@ func (c StopNodeCommand) Execute(ctx context.Context, worker *NodeWorker) NodeRe
 	if ctx.Err() != nil {
 		result.Succeeded = false
 		result.Error = ctx.Err().Error()
+		worker.broker.mu.RLock()
```
DimaOrekhovPS: Let's not access this lock here, it feels like an architectural violation. This lock should be accessed only by broker itself and all-node-level commands

It's crucial to keep the number of access points to a lock low to avoid potential issues like deadlocks

I see that we can have a concurrency issue if we don't acquire a lock, but let's postpone solving it since it needs to be solved diferently

DimaOrekhovPS (changes requested):

decentralized-api/broker/node_worker_commands.go
x0152: Agree. I think we can fix it with an immutable snapshot from the broker
it's a small change. I'll revert if you still think we should postpone

DimaOrekhovPS: I've reviewed the snapshot approach, I think it still suffers from the problem that we now access the broker lock from node worker, something we've not been doing so far. 

I think the snapshot should be taken when we're creating the node command to be executed. Then it will be ok

DimaOrekhovPS (approved):

PR #544, opened 2026-01-10
inference: defense-in-depth against int overflow

ouicate:
**Problem:** 
The inference module computed escrow and actual cost using uint64 arithmetic and then cast the result to int64 without bounds checks. When the uint64 result exceeded math.MaxInt64, the cast could wrap negative and be treated as a “refund owed”, triggering module-account refunds. Additionally, StartInference reused a shared payment helper that would issue refunds when payments.EscrowAmount < 0, and it only logged refund errors (tx could still succeed). Token count fields were uint64 and unbounded, allowing extreme values that amplify overflow / DoS risk.

**Root Cause:** 
Escrow and cost math was performed unchecked in CalculateEscrow / CalculateCost, then converted into int64 for coin accounting. Because the Start/Finish signature bytes do not bind token counts, attackers could provide arbitrary token counts and force overflow conditions. The shared processInferencePayments() helper treated any negative escrow delta as refundable, even during StartInference.

**Fix:**
- Updated CalculateEscrow and CalculateCost to use checked arithmetic (math/bits) and return (int64, error); fail closed if the result overflows, exceeds MaxInt64, or token sums exceed range.
- Propagated these errors through the Start/Finish inference processing so overflows abort the transaction.
- Prevented module-account refunds during StartInference by adding an allowRefund flag to processInferencePayments; StartInference now fails closed on any negative escrow amount, while FinishInference keeps refund behavior intact.
- Added hard caps for max_tokens, prompt_token_count, and completion_token_count in both ValidateBasic and msg server handling, using a shared types.MaxAllowedTokens constant.

### DIFF
--- a/inference-chain/x/inference/calculations/inference_state.go
+++ b/inference-chain/x/inference/calculations/inference_state.go
@@ -1,6 +1,7 @@
 package calculations
 
 import (
+	"math"
 	"math/bits"
 
 	sdkerrors "cosmossdk.io/errors"
@@ -17,7 +18,7 @@ const (
 	PerTokenCost     = 1000 // Legacy fallback price
 )
 
-const maxInt64Uint64 = uint64(^uint64(0) >> 1)
+const maxInt64Uint64 = uint64(math.MaxInt64)
 
 type BlockContext struct {
 	BlockHeight    int64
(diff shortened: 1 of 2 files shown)


inference-chain/x/inference/calculations/inference_state.go
```diff
@@ -15,6 +17,8 @@ const (
 	PerTokenCost     = 1000 // Legacy fallback price
 )
 
+const maxInt64Uint64 = uint64(^uint64(0) >> 1)
```
patimen: Why not use math.MaxInt64?

PR #545, opened 2026-01-11
fix: minor bugs

### LINKED ISSUE #562: GetEpochModel in validation should use inference epoch
Follow-up to #553. Line 68 uses GetEpochModel (current epoch) instead of GetEpochModelForEpoch(ctx, inference.EpochId, inference.Model)

x0152:
There are three commits:
1. Division by zero rare edge case when a single participant runs a single model. The denominator becomes zero - panic and crashes the node
2. HTTP timeout - critical, potential DoS. if remote side accepts connection but never respond, goroutine blocks forever. Possible scenario: collect many connection and will crash node
3. CheckStatusRoutine create timer that triggers only once. After first tick, goroutine blocks forever waiting for a channel that will never respond
4. Validation model lookup uses wrong epoch. Causes incorrect validator rejection for previous epoch inferences (Fixes #562)

### DIFF
--- a/inference-chain/x/inference/calculations/should_validate.go
+++ b/inference-chain/x/inference/calculations/should_validate.go
@@ -4,7 +4,6 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
-	"log/slog"
 	"math"
 	"strconv"
 
@@ -22,11 +21,7 @@ func ShouldValidate(
 	debug bool,
 ) (bool, string) {
 	if totalPower <= executorPower {
-		slog.Error("ShouldValidate: totalPower <= executorPower",
-			"inferenceId", inferenceDetails.InferenceId,
-			"totalPower", totalPower,
-			"executorPower", executorPower)
-		return false, "ShouldValidate:false"
+		return false, "ShouldValidate:false totalPower <= executorPower"
 	}
 	// Creating with exponent vs dividing
 	executorReputation := decimal.New(int64(inferenceDetails.ExecutorReputation), -2)


x0152:
@0xMayoor You're right. However, this is not a single isolated issue. I don't see any transactional approach. So for this purpose the best idea make another PR and set all atomicity l fixes there.

I've removeв the 4th commit from this PR to keep it focused on minor bugs

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -397,7 +397,7 @@ func (s *Server) handleTransferRequest(ctx echo.Context, request *ChatRequest) e
 	req.Header.Set(utils.XPromptHashHeader, inferenceRequest.PromptHash)
 	req.Header.Set("Content-Type", request.Request.Header.Get("Content-Type"))
 
-	resp, err := NewNoRedirectClient().Do(req)
```
DimaOrekhovPS, line 400: Please take a look at how `NewNoRedirectClient` is constructed and make sure the new `s.httpClient` also doesn't allow redirects

inference-chain/x/inference/calculations/should_validate.go
```diff
@@ -20,6 +20,9 @@ func ShouldValidate(
 	validationParams *types.ValidationParams,
 	debug bool,
 ) (bool, string) {
+	if totalPower <= executorPower {
+		return false, "ShouldValidate:false"
```
DimaOrekhovPS: Let's log an error here, it's an abnormal situation

inference-chain/x/inference/keeper/epoch_models.go
```diff
@@ -8,14 +8,23 @@ import (
...
+	effectiveEpochIndex, found := k.GetEffectiveEpochIndex(ctx)
+	if !found {
+		return nil, types.ErrEffectiveEpochNotFound
+	}
+	return k.GetEpochModelForEpoch(ctx, effectiveEpochIndex, modelId)
+}
+
+// GetEpochModelForEpoch retrieves the model snapshot for a given model ID from a specific epoch.
+func (k Keeper) GetEpochModelForEpoch(ctx context.Context, epochId uint64, modelId string) (*types.Model, error) {
+	epochGroup, err := k.GetEpochGroup(ctx, epochId, "")
```
DimaOrekhovPS: We can just pass in `modelId` here instead of the empty string and get the `modelSubGroup` right away

DimaOrekhovPS (changes requested):
Please fix the http client and optimize getting the epoch group

decentralized-api/internal/server/public/post_chat_handler.go
x0152: Thanks! NoRedirectClient used

inference-chain/x/inference/calculations/should_validate.go
x0152: Logged as error

inference-chain/x/inference/keeper/epoch_models.go
x0152: Fixed. Direct modelId lookup. Thanks

DimaOrekhovPS (approved):

PR #551, opened 2026-01-13
fix(bls): reject duplicate slot indices in partial signatures

0xMayoor:
Slot indices aren't deduplicated before counting coverage. 
Someone can submit [0,0,0,...] fifty times and hit threshold with just one actual slot.

The aggregation still runs but Lagrange math breaks on duplicate x-coords - 
final signature fails verification.

Fix: reject submissions with repeated indices early.


x0152:
> > @0xMayoor Hi! I would add slot count validation here to prevent DOS. What do you think?
> > **threshold_signing.go:240**
> > ```go
> >  func (k Keeper) validateSlotOwnership(ctx sdk.Context, submitter string, slotIndices []uint32, epochBLSData *types.EpochBLSData) error {
> > +	// reject if slot count exceeds epoch total
> > +	if uint32(len(slotIndices)) > epochBLSData.ITotalSlots {
> > +		return fmt.Errorf("slot count %d exceeds epoch total %d", len(slotIndices), epochBLSData.ITotalSlots)
> > +	}
> > +
> >  	// Find submitter in epoch participants
> > 	var participantStartSlot, participantEndSlot uint32
> > 	found := false
> > ```
> 
> Appreciate it, but that's a separate validation. This PR is scoped to the duplicate slot issue I reported.
> 
> Also worth noting - validateSlotOwnership already constrains slots to the submitter's assigned range, so the epoch-total check would be redundant in practice. But if you still think it adds value, happy to review a separate PR!

@0xMayoor Got it, thanks! I missed the validateSlotOwnership check

patimen (approved):
Very simple and clear, thank you.

DimaOrekhovPS (approved):

PR #553, opened 2026-01-14
Membership for correct epoch for Validation requests

### LINKED ISSUE #566: Epoch and timestamp consistency across inference, validation, and claims

### LINKED ISSUE #574: Membership for correct epoch for Validation requests

patimen:
Especially for validations for the Previous Epoch (common with retries), membership MUST be determined by the active participants THEN, not in the current epoch.


DimaOrekhovPS (approved):
makes perfect sense

patimen:
> Found a similar pattern with GetEpochModel at msg_server_validation.go:68. It currently only uses ValidationThreshold (line 76), but under the hood it pulls from the current epoch via GetCurrentEpochGroup. Right now it's probably fine since thresholds rarely change, but if someone adds more model fields to validation logic later, they might not realize it
> 
Totally valid, can you create an issue for this? I don't want to add to v0.2.8 at this point.

PR #563, opened 2026-01-15
fix(inference): variable shadowing in direct payment path

0xMayoor:
Line 83 assigns to shadowed err, line 85 returns outer err (nil).

IssueRefund uses this path for all refunds. When SendCoins fails, user never gets their tokens back but system records show refund succeeded.

No error logged, no way to trace it!


patimen:
Yes, this would be a good fix. The question of what to do in an underfunded module situation may not be clear, but all calls to this do in fact check for underfunded messages and at least log something we could see later.

PR #609, opened 2026-01-19
BLS optimized

libermans:
BLS signature verification on chain is too computationally expansive, which makes every BLS signature message to execute 2s. After this change it is less than 10ms

Precompute slot public keys once. And use MultiExp for the optimization.
More BLST optimizations, including on api node side.

Thanks for this optimizations inspiration to prof. Dan Boneh & Ash Vardanian from Alex Petrov aka sysman

---

> [!NOTE]
> Introduces a blst-backed BLS pipeline with multi-scalar multiplication and precomputed per-slot public keys to remove per-request heavy math and slash latency.
> 
> - Add `blst` dependency and timing tests; implement blst variants for commitments, share verification, partial signing, signature aggregation, and final verification
> - Extend proto (`EpochBLSData`) with `slot_public_keys` (96-byte G2 per slot); upgrade handler migrates and precomputes keys; phase transitions now precompute and store them
> - Keeper: replace gnark paths with blst MSM for commitment eval, participant pubkey aggregation, partial verification, final verification, and G2 aggregation; add strict input validation and payload filtering
> - API: new blst decompress helpers (G1/G2) and handlers updated to use them; compute/compare results kept byte-identical
> - Dealer: compute G2 commitments via blst; numerous legacy gnark methods marked deprecated but retained for reference
> - Tests: add comprehensive timing/comparison tests; Kotlin e2e updated to expect epoch `SIGNED` post-validation
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b5a5d6ddfdc43725911aaa2490c6ff7f75d7f99d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


decentralized-api/internal/bls/verifier.go
```diff
@@ -444,6 +445,55 @@ func (bm *BlsManager) verifyShareAgainstCommitments(share *fr.Element, slotIndex
...
+// verifyShareAgainstCommitmentsBlst verifies a decrypted share against the dealer's polynomial commitments using blst.
+func (bm *BlsManager) verifyShareAgainstCommitmentsBlst(share *fr.Element, slotIndex uint32, commitments [][]byte) (bool, error) {
+	if len(commitments) == 0 {
+		return false, fmt.Errorf("no commitments provided")
+	}
+
+	// 1. Prepare points (commitments) in blst format
+	points := make([]*blst.P2Affine, len(commitments))
+	for j, cb := range commitments {
+		p := new(blst.P2Affine).Uncompress(cb)
```
x0152, line 465: Hi there!
Noticed one thing:
blst.Uncompress doesn't do subgroup check, it only validates that point is on curve. The docs say caller must call InG1()/InG2() after (https://github.com/supranational/blst#signature-verification) Otherwise someone could potentially send a point outside G1/G2 subgroup and it leads to unpredictable results

Quick fix would be to add !p.InG1()/p.InG2() check after each Uncompress, but not the most performant

x0152: InG1/InG2 accept infinity point. KeyValidate()/SigValidate(true) would be better for full validation

inference-chain/x/bls/keeper/bls_crypto.go
```diff
@@ -265,6 +578,224 @@ func (k Keeper) aggregateBLSPartialSignatures(partialSignatures []types.PartialS
...
+	}
+
+	res := aggregate.Bytes()
+	return res[:], nil
+}
+
+// aggregateG2PointsBlst aggregates a list of G2 points using blst.
+func (k Keeper) aggregateG2PointsBlst(points [][]byte) ([]byte, error) {
+	if len(points) == 0 {
+		return new(blst.P2).ToAffine().Compress(), nil
```
x0152, line 716: `return new(blst.P2).ToAffine().Compress(), nil`
Empty list -> returns infinity point (0xC0...) instead of error.

Maybe it should be `return nil, fmt.Errorf("cannot aggregate empty list")`

And in **msg_server_dealer.go:52** should be check
```go
if len(msg.Commitments) == 0 {
    return nil, fmt.Errorf("commitments cannot be empty")
}
```

inference-chain/x/bls/keeper/msg_server_group_validation.go
```diff
@@ -164,7 +162,7 @@ func (ms msgServer) SubmitGroupKeyValidationSignature(goCtx context.Context, msg
 	}
 
 	// Verify BLS partial signature against participant's computed individual public key
-	if !ms.verifyBLSPartialSignature(msg.PartialSignature, validationState.MessageHash, &previousEpochBLSData, filteredSlots) {
+	if !ms.verifyBLSPartialSignatureBlst(msg.PartialSignature, validationState.MessageHash, &previousEpochBLSData, filteredSlots) {
```
akup: According to https://github.com/gonka-ai/gonka/pull/610
msg.PartialSignature should be filtered with slots. Now duplicated slots are filtered, but partialSignature is not

inference-chain/app/upgrades/v0_2_8/upgrades.go
```diff
@@ -67,3 +77,42 @@ func burnExtraCommunityCoins(ctx context.Context, k *keeper.Keeper) error {
...
+			slotKeys, err := bk.PrecomputeSlotPublicKeysBlst(&epochData)
+			if err != nil {
+				return fmt.Errorf("failed to precompute slot keys for epoch %d: %w", epochID, err)
+			}
+			epochData.SlotPublicKeys = slotKeys
+			if err := bk.SetEpochBLSData(sdkCtx, epochData); err != nil {
+				return fmt.Errorf("failed to save migrated epoch %d data: %w", epochID, err)
+			}
+			bk.Logger().Info("Successfully precomputed slot public keys for epoch", "epoch_id", epochID)
+		}
```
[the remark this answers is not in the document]
libermans: This it only for migration from previous state which has no SlotPublicKey.

inference-chain/x/bls/keeper/phase_transitions.go
```diff
@@ -192,6 +191,13 @@ func (k Keeper) CompleteDKG(ctx sdk.Context, epochBLSData *types.EpochBLSData) e
 		// Store valid dealers in epoch data
 		epochBLSData.ValidDealers = validDealers
 
+		// Precompute per-slot public keys for faster validation later
+		slotPublicKeys, err := k.PrecomputeSlotPublicKeysBlst(epochBLSData)
+		if err != nil {
+			return fmt.Errorf("failed to precompute slot public keys for epoch %d: %w", epochBLSData.EpochId, err)
+		}
```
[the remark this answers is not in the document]
libermans: It was just precomputed with the right amount of slots. So we shouldn't have any check here.

inference-chain/x/bls/keeper/bls_crypto.go
```diff
@@ -19,69 +24,279 @@ func (k Keeper) computeParticipantPublicKey(epochBLSData *types.EpochBLSData, sl
...
+		// Parse precomputed slot public key with blst
+		if len(epochBLSData.SlotPublicKeys) <= int(slotIndex) {
+			k.Logger().Error("Precomputed slot public key missing", "slot", slotIndex)
+			return false
+		}
+		slotPubKey := new(blst.P2Affine).Uncompress(epochBLSData.SlotPublicKeys[slotIndex])
+		if slotPubKey == nil {
+			k.Logger().Error("Failed to unmarshal precomputed slot public key with blst", "slot", slotIndex)
+			return false
+		}
```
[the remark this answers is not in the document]
libermans: The current logging is solid.

inference-chain/x/bls/keeper/bls_crypto.go
```diff
@@ -19,69 +24,279 @@ func (k Keeper) computeParticipantPublicKey(epochBLSData *types.EpochBLSData, sl
...
+
+		var power fr.Element
+		power.SetOne()
+
+		scalars := make([]byte, len(totalCommitmentsAffine)*32)
+		for i := 0; i < len(totalCommitmentsAffine); i++ {
+			pBytes := power.Bytes()
+			// gnark-crypto uses big-endian, blst expects little-endian for scalars
+			for j := 0; j < 16; j++ {
+				pBytes[j], pBytes[31-j] = pBytes[31-j], pBytes[j]
```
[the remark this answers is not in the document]
libermans: It should be always 32 bytes

inference-chain/x/bls/keeper/msg_server_group_validation.go
libermans: No vulnerability. If SlotIndices are filtered but msg.PartialSignature isn’t, the submission fails validation in verifyBLSPartialSignatureBlst because:
sigCount != len(filteredSlots) (count mismatch), or the per-slot pairing checks fail if ordering/chunks don’t match.

But filtered partialSignatures in any case. Can you please check.

decentralized-api/internal/bls/verifier.go
libermans: Fixed. Can you please check left InG1/InG2 in two places. The rest KeyValidate()/SigValidate(true).

inference-chain/x/bls/keeper/bls_crypto.go
libermans: added commitments length check to two places, which should cover all cases. can you please check.

inference-chain/x/bls/keeper/threshold_signing.go
```diff
@@ -298,7 +298,7 @@ func (k Keeper) verifyPartialSignature(partialSignature []byte, messageHash []by
 	}
 
 	// Verify BLS partial signature against participant's computed individual public key
-	if !k.verifyBLSPartialSignature(partialSignature, messageHash, epochBLSData, slotIndices) {
+	if !k.verifyBLSPartialSignatureBlst(partialSignature, messageHash, epochBLSData, slotIndices) {
```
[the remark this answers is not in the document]
libermans: Add "deprecated" comments, will remove the functions later.

inference-chain/x/bls/keeper/msg_server_group_validation.go
akup: Valid

decentralized-api/internal/bls/verifier.go
x0152: looks good! Split make sense

inference-chain/x/bls/keeper/bls_crypto.go
x0152: Looks good. Maybe worth adding a comment in aggregateG2PointsBlst about returning infinity on empty list, but not in this PR

patimen (approved):

PR #610, opened 2026-01-19
fix: sync filtered signature with slots

x0152:
filter signatures alongside slots to fix length mismatch during verification
+add test for it


AlexeySamosadov (changes requested):
Good refactoring direction — extracting `FilterSlotsAndSignature` with a named constant and a test is a clear improvement. Two issues to address:

### [Must fix] Lost `LogWarn` for duplicate slot submissions

The original inline code logged duplicates:

```go
ms.Keeper.LogWarn("Ignoring duplicate slot submission", "slot_index", idx, "creator", msg.Creator)
```

The extracted function drops this entirely. Duplicate BLS slot submissions are an anomalous event worth tracking — could indicate a client bug or an attack attempt. Since the function doesn't have access to the logger, one option is to return the count of filtered duplicates so the caller can log:

```go
if skipped := len(msg.SlotIndices) - len(filteredSlots); skipped > 0 {
    ms.Keeper.LogWarn("Ignored duplicate slot submissions", "count", skipped, "creator", msg.Creator)
}
```

### [Must fix] No bounds check in exported function

`FilterSlotsAndSignature` slices `signature[sigStart:sigEnd]` without checking `sigEnd <= len(signature)`. The caller validates this beforehand, but since the function is exported, it can be called from anywhere. A mismatched input would cause a panic — which violates the "never panic in chain code" rule.

Options: (a) add a length guard and return an error, or (b) make the function unexported (`filterSlotsAndSignature`) since the test is in the same package anyway.

### [Nit] Unrelated blank line removals

Two blank lines removed (after unmarshal block, before "Add the partial signature") are unrelated to the fix.

### [Suggestion] Edge case tests

The test covers the happy path well. Consider adding: empty slots input, all-duplicates (result should be empty), and `nil` seen map.

x0152:
Closing as outdated

PR #614, opened 2026-01-21
[P2] feat: improve openai api compatibility and error handling

### LINKED ISSUE #351: BUG: Wrong error message for unsupported models in /chat/completions

### LINKED ISSUE #558: [P2] URLs with `/chat/completions` and `/completions` for Open Router
Provide a minimal but verifiable example that demonstrates Gonka’s inference capability via standard OpenAI-compatible endpoints, suitable for validation and review by openrouter.ai. We need to run a simple inference and provide publicly accessible URLs for the following endpoints:
- `/completions`
- `/chat/completions`

These endpoints are required for listing Gonka as an inference vendor on openrouter.ai.

Requirements
- The inference should be very simple (e.g. a basic prompt like asking about the weather).
- The URLs must return real inference results, not mock data.

Inference logs for the provided URLs must:
- Be readable
- Remain accessible for a reasonably long period of time
- Just a minimal, clear demonstration of how to query the API

**Must be openAI-compliant and return usage for both stream and non-stream**

+ include pricing models endpoint

Before implementation, please review OpenRouter documentation and any relevant provider integration requirements.
This will help ensure full compatibility and avoid iteration during validation.

x0152:
Improve OpenAI API compatibility and error handling:
- /v1/completions endpoint (single, streaming) that proxies to /v1/chat/completions
- /v1/models with OpenAI-compatible response format
- Proper HTTP error codes (404 model not found, 503 no executors, 429 capacity reached)
- Model validation before balance check and executor selection after capacity (fixes #351)
- Message content validation and request body size limit


kotelnikova (commented):
Please add openrouter slug "gonka"

x0152:
Hi @gmorgachev
Thanks for the detailed explanation. I think you're right that your approach is cleaner but let me explain my reasoning:

I created separate endpoints because openrouter might need specific fields (like openrouter.slug) or different auth that we don't want in our main inference api. Keeping them separate means changes to one won't affect the other. I don't think these extra fields would break openai like connections, but having separate endpoints gives us more flexibility. This way, if we need to extend or change inference api later, you don't need to worry about openrouter provider api

P.S. I'll rework it to match your approach as described in the issue 😅

tcharchian:
Hi @x0152, thank you for the work on this implementation. After review, the merge is temporarily paused. A broker within the community needs to be identified to act as an intermediary between OpenRouter and Gonka. Once that structure is defined, the required adjustments on the integration side will become clearer.

In particular, this will determine whether separate endpoints (for example, with /openrouter in the address) are necessary, or whether a different architectural approach would be more appropriate.

Further feedback and next steps will be shared once the broker setup is clarified.

Thank you for your patience.

cc: @kotelnikova @gmorgachev @mtvnastya

x0152:
Hi @tcharchian! Thanks for review.

Just to clarify: most of the work in this PR is done for OpenAI API compatibility (and better error handling) for /v1/chat/completions, /v1/completions, and /v1/models - these fixes are useful regardless of open router and I think we would need them anyway. OpenRouter only fields are small (e.g. openrouter.slug). I can remove them and rename the PR to "Improve OpenAI compatibility and error handling" to keep it generic

Then we can do OpenRouter integration later, when details are clear

Happy to adjust if you prefer

tcharchian:
thanks @x0152!
@akup Would be awesome to have a second pair of eyes on this and decide whether we can merge it without breaking anything.

akup:
It seams that OpenRouter expects model pricing in USD https://openrouter.ai/docs/guides/get-started/for-providers

And the Broker instance is needed that will be accounted in USD and is responsible to pay for inferences in Gonka for the chain.


@tcharchian your previous comment was

> A broker within the community needs to be identified to act as an intermediary between OpenRouter and Gonka. Once that structure is defined, the required adjustments on the integration side will become clearer.

Are there anything clear about the broker?

Current PR is just covering opening OpenAI compatible public endpoints, but it is far from what OpenRouter expects for Provider integration. One important point  from OpenRouter provider integration docs is [Auto Top Up or Invoicing](https://openrouter.ai/docs/guides/get-started/for-providers#2-auto-top-up-or-invoicing). And it is the Broker responsibility.
Another point is pricing in USD, and that models listing also provides pricing in USD. So there anyway is needed Broker integration and the Broker should provide gonka exchange rate as it will be payed in USD from OpenRouter and it is the Broker's business to have reliable exchange rate.

I can't understand how to review and approve the PR when there is no information and technical requirements from the Broker.

x0152:
Hi @akup,

What exactly is missing here besides payment? Payment was explicitly out of scope for the original issue (#558). I followed the OpenRouter provider docs for formats and error codes: https://openrouter.ai/docs/guides/get-started/for-providers

As for the full OpenRouter integration - from my point of view this should be done through a separate relay/broker service, not in the decentralized API

Anyway, OpenRouter integration is on hold for now. So I'd suggest we don't close this PR but just remove the OpenRouter-specific bits (like openrouter.slug) and rename it (e.g. "Improve OpenAI API compatibility and error handling"). Almost all the logic here is about OpenAI-like compatibility: error handling, /v1/completions, model validation, response formats. We'll need all of that regardless of OpenRouter

What do you think?

alexanderkuprin:
> What exactly is missing here besides payment?

I think that OpenRouter expects pricing in **models endpoint to be in USD**.
And this **strongly couples this logic to exchange rate, that is Broker's business**.

> As for the full OpenRouter integration - from my point of view this should be done through a separate relay/broker service, not in the decentralized API

Moreover it can't be done at decentralized-api, it will be actually centralized service that is accountant at OpenRouter and has payment relations with it.

> Almost all the logic here is about OpenAI-like compatibility: error handling, /v1/completions, model validation, response formats. We'll need all of that regardless of OpenRouter

For this part it is correct. And I have no concerns on this part.

There is another part for OpenRouter, that logs should be recorded and provided, but I still see no reason to start to code it, before Broker technical requirements are provided.

akup:
Accidentely used another account to reply

x0152:
Removed OpenRouter specific stuff (slug, pricing, provider metadata) from /v1/models as discussed. The rest of the pr is purely OpenAI compat + better error handling, no OpenRouter dependency left + renamed PR

akup:
> Removed OpenRouter specific stuff (slug, pricing, provider metadata) from /v1/models as discussed. The rest of the pr is purely OpenAI compat + better error handling, no OpenRouter dependency left + renamed PR

Now it seams to be pure OpenAI compatibility PR.

But current inference/validation API is going to be legacy code soon.
There is new `/v1/subnet/sessions/:id/chat/completions` endpoint, and this compatibility features should go there. But if `chat/completions` will become old path in a next version, maybe this PR should wait a little, not because it's bad, but because API is changing and subnet adoption is the primary goal.

@gmorgachev what will be the new paths for public apis, should it contain `sessionId`?

@tcharchian what is the destiny of the root issue? Are there some requirements or plan for OpenRouter integrations?

akup (approved):
Latest fixes make the PR clean and targeted to solve openai api compatibility

x0152:
add strict request validation according to openai api spec: https://platform.openai.com/docs/api-reference/chat/create

akup:
> add strict request validation according to openai api spec: https://platform.openai.com/docs/api-reference/chat/create

Maybe we need some high-level plan/proposal how to actually support openAI API?

Now it validates against it but actually does not support. Also are there some endpoint in standart that tells what features are supported?

But this is industry standart protocol, and actually it is good to have in mind it's features to understand big picture what could/should be supported in Gonka.

x0152:
> Maybe we need some high-level plan/proposal how to actually support openAI API?
> 
> Now it validates against it but actually does not support. Also are there some endpoint in standart that tells what features are supported?
> 
> But this is industry standart protocol, and actually it is good to have in mind it's features to understand big picture what could/should be supported in Gonka.

The idea was just to reject bad requests early at the TA before they hit vLLM. The current validation may be a bit stricter than vLLM, but it still stays compatible with openai connectors, so it should be okay

Maybe it would be better to move this into a separate pr

0xgonka (commented):
Auth pipeline is correctly shared with `/v1/chat/completions` — no bypass risk. Signature verification uses the original body hash. Message validation is thorough. Test coverage is strong. Executor query error differentiation (NotFound vs Unavailable → 404 vs 503) is well-designed.

**Medium:**

- `/v1/completions` forwards the original body to executor nodes at `/v1/completions` path. This creates a hard dependency on executor nodes natively supporting this endpoint. If any executor only supports `/v1/chat/completions`, requests will silently fail. Should either be documented as a requirement or guarded with a capability check.

**Minor:**

- Body size limits are inconsistent: `server.go:96` uses `BodyLimit("10M")` (10,000,000 bytes) while `MaxRequestBodySize` in `post_chat_handler.go:44` is `10 * 1024 * 1024` (10,485,760 bytes). Echo's stricter limit wins in practice, but the constants should match.
- `get_models_handler.go:47` — `Created` is hardcoded to `0`. OpenAI returns a Unix timestamp. Clients relying on this for sorting or cache invalidation will get wrong results.
- `completions_handlers.go:22` — original error from `readRequestBody` is discarded and replaced with a generic message. The underlying error (e.g., body too large) would be more informative.
- `query_get_random_executor.go:85` — pre-existing typo `"nill"` → should be `"nil"`.

**LGTM with the above addressed.**

0xgonka:
also merge conflicts

x0152:
> Auth pipeline is correctly shared with /v1/chat/completions — no bypass risk. Signature verification uses the original body hash. Message validation is thorough. Test coverage is strong. Executor query error differentiation (NotFound vs Unavailable → 404 vs 503) is well-designed.
> 
> Medium:
> 
> /v1/completions forwards the original body to executor nodes at /v1/completions path. This creates a hard dependency on executor nodes natively supporting this endpoint. If any executor only supports /v1/chat/completions, requests will silently fail. Should either be documented as a requirement or guarded with a capability check.
> Minor:
> 
> Body size limits are inconsistent: server.go:96 uses BodyLimit("10M") (10,000,000 bytes) while MaxRequestBodySize in post_chat_handler.go:44 is 10 * 1024 * 1024 (10,485,760 bytes). Echo's stricter limit wins in practice, but the constants should match.
> get_models_handler.go:47 — Created is hardcoded to 0. OpenAI returns a Unix timestamp. Clients relying on this for sorting or cache invalidation will get wrong results.
> completions_handlers.go:22 — original error from readRequestBody is discarded and replaced with a generic message. The underlying error (e.g., body too large) would be more informative.
> query_get_random_executor.go:85 — pre-existing typo "nill" → should be "nil".
> 

Fixed all comments. The only exception is exposing raw read errors - it's unsafe to leak internal io errors to clients. Instead, I mapped them to safe 400/413 responses

0xgonka (approved):

PR #618, opened 2026-01-22
PoC v2 & Offchain PoC data

gmorgachev:
# Proposal: PoC v2 & Offchain PoC data

This proposal describes the current status of the PoC v2 migration, which integrates the PoC procedure into vLLM and uses `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8` model.

The PoC v2 initiative addresses two key objectives:

1. Integrate the PoC procedure directly into vLLM, enabling an immediate switch from inference to PoC without offloading the model or loading a separate PoC model, so PoC can be triggered quickly with minimal phase-switch overhead and no dedicated setup.

2. Migrate artifact storage off-chain using Merkle commitments (only the root hash + count is recorded on-chain) to reduce on-chain data volume.

These changes maintain robustness against minimal model changes (e.g., quantization). In the current vLLM integration, inference and PoC can't be executed in the same forward pass, but PoC can be scheduled in the next forward pass.

Keeping PoC inside vLLM also makes it straightforward to integrate PoC into the chat completion path next. In such setup, inference and PoC can run in the same batch and the same forward pass.

## vLLM

The current version of PoC integrated with vLLM: https://github.com/gonka-ai/vllm/tree/gm/poc-layers

Instead of offloading the inference model and loading a randomly initialized model for PoC, the integrated approach applies block-dependent randomization to specific layers:

- Input: bypass token embeddings by feeding deterministic random `inputs_embeds` seeded by `(block_hash, public_key, nonce)`.
- Hidden layers: apply per-layer Householder transforms via forward hooks seeded by `(block_hash, layer_idx)`, active only during PoC forward context.
- Output: from the last-token hidden state, normalize, select `k_dim` indices per nonce, and apply a deterministic Haar-like rotation (Householder chain) seeded by `(block_hash, public_key, nonce)`.
- Artifact: output is a `k_dim` vector (default `k_dim=12`, FP16, base64-encoded). In our experiments, increasing `k_dim` beyond 12 did not improve fraud detection quality, while lower values led to degradation.

### Experiment

We compare honest outputs across different hardware generations (H100, B200, A100) and model variants (FP8 vs INT4-quantized) to measure expected divergence. We also compare against a modified model on the same hardware (closest case for fraud detection). Validation uses an L2-distance mismatch rule with a calibrated threshold and a statistical test.

Report: [report.md](./PoC_V2_Validation_Report.md)

## Merkle Tree (MMR, Off-chain Artifacts)

PoC v2 artifacts are larger than v1, and we need enough total nonces per participant to allow meaningful sampling without pushing large payloads on-chain. The approach is to keep artifacts off-chain and store only an on-chain commit: `(root_hash, count)` (MMR root + leaf count).

Validators:
- Query the chain for all participants’ latest commits for the stage (`root_hash`, `count`) and validate each participant against their commit.
- Sample `leaf_index` values in `[0, count)`
- Fetch `(artifact, proof)` from the participant API (`POST /v1/poc/proofs`) and verify against the committed `root_hash`.
- Run statistical validation on the verified artifacts and submit results in batch (one `MsgSubmitPocValidationsV2` can include validations for multiple participants at the same stage height).

```
┌────────────────────────────────────────────────────────────────────────┐
│ Generation Phase                                                       │
│   Participant generates artifacts → stores locally in MMR              │
│   Periodically submits MsgPoCV2StoreCommit(root_hash, count) to chain  │
└────────────────────────────────────────────────────────────────────────┘
         ↓
┌────────────────────────────────────────────────────────────────────────┐
│ After Generation Phase                                                 │
│   Submits MsgMLNodeWeightDistribution (weight per MLNode)              │
└────────────────────────────────────────────────────────────────────────┘
         ↓
┌────────────────────────────────────────────────────────────────────────┐
│ Validation Phase                                                       │
│   Validator queries chain for commits (root_hash, count)               │
│   Samples leaf_indices in [0, count) deterministically                 │
│   Requests artifacts + proofs from participant API (/v1/poc/proofs)    │
│   Verifies proofs → statistical validation → submits MsgSubmitPoc...V2 │
└────────────────────────────────────────────────────────────────────────┘
```

## Migration

PoC v2 rollout requires coordinated updates across:
- `mlnode`: new version with V2 endpoints + switching to `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`
- `api`
- `node` (chain)

`mlnode` can be upgraded independently (before or after the `api`/`node` upgrade), but must be upgraded before PoC v2 is enabled. It is backward compatible and can keep running PoC v1 while the network is still in v1 mode.

`api` and `node` are updated together via a single on-chain upgrade. That upgrade enables the end-to-end PoC v2 flow.

If issues arise during migration, `api` fixes can be deployed off-chain without requiring a governance vote, as long as they don't change protocol behavior.

Because of this coordination requirement, the migration keeps PoC v1 and PoC v2 available at the same time. During migration:

- **Regular PoC**: V1 (on-chain batches, affects weights)
- **Confirmation PoC**: V1 for all events, plus V2 **tracking** for the first sampled event per epoch (`event_sequence == 0`)

The V2 tracking confirmation event records coverage metrics on-chain for monitoring but does **not** affect weights/slashing. This allows us to measure mlnode V2 adoption before enabling full V2.

The V1/V2 switch is controlled by governance params:
- `poc_v2_enabled` (false = PoC v1, true = PoC v2)
- `confirmation_poc_v2_enabled` (enables migration mode when `poc_v2_enabled=false`)

Setting `poc_v2_enabled=true` requires no software upgrade - just a governance proposal. After this switch, V2 affects weights/slashing. The tracking-only mode only exists during migration.

## Status

Implementation complete. All testermint tests pass for PoC v1, v2, and migration mode.

### Completed

- PoC v2 integration with vLLM (`v0.9.1-poc-v2-blackwell`)
- Off-chain artifact storage with MMR proofs
- Migration mode: V1 enforcement + V2 tracking for confirmation PoC
- Grace epoch: dry-run mode when switching to full V2
- Governance-controlled V1/V2 switching via `poc_v2_enabled` and `confirmation_poc_v2_enabled`
- V1/V2 dispatch in chain and DAPI based on migration state
- `ListConfirmationPoCEvents` query for monitoring V2 adoption
- Validation retries with dynamic node count and backoff

### Pending

- [to define]: PoC phase length for PoC v2

See [migration-dual.md](./dev/migration-dual.md) for detailed migration design.


deploy/join/node-config.json
```diff
@@ -11,7 +11,8 @@
           "--quantization",
           "fp8",
           "--gpu-memory-utilization",
-          "0.9"
+          "0.8",
+          "--enable-poc"
```
blizko: `--enable-poc` needs to be in default or in governance parameters. Otherwise high chance that will be missed during configurations/re-configurations of nodes

decentralized-api/internal/server/public/server.go
```diff
@@ -108,6 +129,19 @@ func NewServer(
 	g.GET("restrictions/exemptions", s.getRestrictionsExemptions)
 	g.GET("restrictions/exemptions/:id/usage/:account", s.getRestrictionsExemptionUsage)
 
+	// PoC proofs endpoint with IP rate limiting (100 req/min per IP)
+	pocProofsRateLimiter := echomw.RateLimiter(echomw.NewRateLimiterMemoryStoreWithConfig(
+		echomw.RateLimiterMemoryStoreConfig{
```
blizko, line 134: Could this trigger an issue with proxy or NAT setups? In case multiple nodes are registered under the same IP?

decentralized-api/poc/proof_client.go
```diff
@@ -0,0 +1,262 @@
...
+		"signature":                    base64.StdEncoding.EncodeToString(signature),
+	}
+
+	bodyBytes, err := json.Marshal(requestBody)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal request: %w", err)
+	}
+
+	// Build URL
+	proofUrl, err := url.JoinPath(participantUrl, "v1/poc/proofs")
```
blizko, line 140: Is this consistent with the notation of v1 and v2 PoC?

decentralized-api/poc/validator.go
```diff
@@ -0,0 +1,655 @@
...
+
+// filterNodesForValidation returns nodes available for PoC validation.
+// - Accept nodes in POC status with any sub-status
+// - Accept nodes in INFERENCE status
+// - Exclude FAILED or administratively disabled nodes
+func filterNodesForValidation(nodes []broker.NodeResponse) []broker.NodeResponse {
+	filtered := make([]broker.NodeResponse, 0, len(nodes))
+	for _, node := range nodes {
+		// Exclude failed nodes
+		if node.State.CurrentStatus == types.HardwareNodeStatus_FAILED {
```
blizko, line 645: Assuming only applicable to v2 where inference is loaded. In v1 failed nodes might occur because of inference parameters selected

decentralized-api/poc/validator.go
```diff
@@ -0,0 +1,655 @@
...
+		}
+
+		// Accept nodes in POC status (any sub-status)
+		if node.State.CurrentStatus == types.HardwareNodeStatus_POC {
+			filtered = append(filtered, node)
+			continue
+		}
+
+		// Accept nodes in INFERENCE status
+		if node.State.CurrentStatus == types.HardwareNodeStatus_INFERENCE {
```
blizko, line 669: Ideally - requires a check between "RAMPING" and "READY" for v2 PoC, since might be sending unnecessary requests to a unready node

decentralized-api/poc/validator_v1.go
```diff
@@ -0,0 +1,467 @@
...
+	"decentralized-api/logging"
+	"decentralized-api/mlnodeclient"
+
+	"github.com/productscience/inference/x/inference/types"
+)
+
+const (
+	POC_VALIDATE_GET_NODES_RETRIES     = 30
+	POC_VALIDATE_GET_NODES_RETRY_DELAY = 5 * time.Second
+	POC_VALIDATE_BATCH_RETRIES         = 5
```
blizko, line 25: Note that for for a validator with 1 node it will retry on the same node, for validator with N nodes, with N>5 might fall into a subset.

decentralized-api/poc/validator_v1.go
```diff
@@ -0,0 +1,467 @@
...
+			BlockHash:   work.pocStartBlockHash,
+			BlockHeight: work.blockHeight,
+			Nonces:      sampledBatch.nonces,
+			Dist:        sampledBatch.dist,
+		}
+
+		// Send to ML node with retry
+		validationSucceeded := false
+		for attempt := 0; attempt < POC_VALIDATE_BATCH_RETRIES; attempt++ {
+			node := nodes[nodeCounter%len(nodes)]
```
blizko, line 279: In cases where len(nodes) > POC_VALIDATE_BATCH_RETRIES might be worth to extend the retry count? Thus ensuring higher probability of actually completing the validation?

inference-chain/x/inference/keeper/msg_server_poc_v2_commit.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return nil, sdkerrors.Wrap(types.ErrInvalidAddress, fmt.Sprintf("invalid creator address: %v", err))
+	}
+	pk := collections.Join(startBlockHeight, addr)
+	commit, err := k.PoCV2StoreCommits.Get(ctx, pk)
+	if err != nil {
+		return nil, sdkerrors.Wrap(types.ErrIllegalState, "no store commit found for this stage")
+	}
+
+	var sum uint32
```
x0152: Hi!
If I send weights [4294967196, 200] with commit.Count = 100, uint32 sum wraps to 100, so check passes and chain stores huge weight 4294967196. If I'm not mistaken, this can lead to outsized rewards

x0152: and It looks like distribution validation is missing

decentralized-api/poc/validator.go
```diff
@@ -0,0 +1,655 @@
...
+// sampleLeafIndices generates deterministic leaf indices for sampling.
+func sampleLeafIndices(validatorPubKey string, blockHash string, blockHeight int64, count uint32, sampleSize int) []uint32 {
+	if count == 0 {
+		return nil
+	}
+
+	n := int(count)
+	if sampleSize >= n {
+		// Return all indices
+		indices := make([]uint32, n)
```
x0152, line 478: If I commit Count = 4,294,967,295, this code allocates 18gb of ram for the uint32 slice, which can dos some validators

decentralized-api/poc/validator.go
```diff
@@ -0,0 +1,655 @@
...
+	failCount *int,
+	pendingCount *int,
+) {
+	nodeCounter := workerID // Start from different nodes per worker
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case work, ok := <-workChan:
```
x0152, line 298: Potential deadlock - workers block on empty workChan while main goroutine waits for them. Channel is only closed after wg.Wait()

quick fix:
one of the goroutines calls cancel() when *pendingCount <= 0
```go
// validator.go:236
v.worker(
    ctx,
    cancel,
    workerID,
    workChan,
    ...
)

// validator.go:273
func (v *OffChainValidator) worker(
    ctx context.Context,
    cancel context.CancelFunc, 
    workerID int,
    ...
)

// validator.go:348
if done {
    cancel()
    return
}
```

decentralized-api/broker/node_worker_commands.go
```diff
@@ -356,3 +223,96 @@ func (c *NoOpNodeCommand) Execute(ctx context.Context, worker *NodeWorker) NodeR
...
+
+	if ctx.Err() != nil {
+		result.Succeeded = false
+		result.Error = ctx.Err().Error()
+		result.FinalStatus = worker.node.State.CurrentStatus
+		result.FinalPocStatus = worker.node.State.PocCurrentStatus
+		return result
+	}
+
+	// No network call - just transition broker state.
```
DimaOrekhovPS, line 317: But what if the node is in inference/stopped state. 

The current implementation seem to assume that the node is always in PoC generating phase when this command is being called.

Shouldn't we do a status check here?

decentralized-api/poc/commit_worker.go
```diff
@@ -0,0 +1,315 @@
...
+		scaledSum += scaled
+	}
+
+	diff := int(targetCount) - int(scaledSum)
+	for i := 0; diff > 0; i++ {
+		weights[i%len(weights)].Weight++
+		diff--
+	}
+
+	return weights, nil
```
ouicate, line 310: **getWeightDistribution()** builds weights by iterating over a **map[string]uint32**. Since Go map iteration order is randomized, the remainder fix-up loop **(weights[i%len(weights)].Weight++)** applies leftover weights in a non-deterministic order. As a result, identical inputs can produce different weight allocations across runs, which may lead to fairness or incentive inconsistencies.

This can be fixed straightforwardly by sorting the keys (or otherwise fixing the iteration order) before building weights, ensuring the remainder distribution is stable and reproducible.

decentralized-api/poc/proof_client.go
```diff
@@ -0,0 +1,262 @@
...
+	ctx context.Context,
+	participantUrl string,
+	req ProofRequest,
+) ([]VerifiedArtifact, error) {
+	// Build request body
+	timestamp := time.Now().UnixNano()
+	validatorAddress := c.recorder.GetAccountAddress()
+	signerAddress := c.recorder.GetSignerAddress()
+
+	// Build signature payload
```
ouicate, line 100: `FetchAndVerifyProofs()` verifies each proof item returned, but does not enforce that the response covers **exactly** the requested `leaf_indices` (no full coverage check, no duplicate check). As a result, a participant API can omit “hard” leaves and return only easy/valid proofs; the verifier still accepts the response as long as the included proofs individually verify. This undermines the sampling challenge.

decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
```diff
@@ -1,15 +1,14 @@
 package tx_manager
 
-// defaultMaxBlocks is the default deadline in blocks (500 blocks ≈ 40 minutes at 5s/block)
+// defaultMaxBlocks is the default deadline in blocks (30 blocks ≈ 2.5 minutes at 5s/block)
 const defaultMaxBlocks int64 = 30
 
 // deadlineByMsgType defines message-specific deadlines based on chain epoch parameters
 var deadlineByMsgType = map[string]int64{
-	"/inference.inference.MsgSubmitPocBatch":      240,
```
DimaOrekhovPS, line 8: Why are we deleting old message types from the deadline config?

decentralized-api/internal/server/mlnode/server.go
```diff
@@ -4,18 +4,30 @@ import (
...
 )
 
 type Server struct {
-	e        *echo.Echo
-	recorder cosmos_client.CosmosMessageClient
-	broker   *broker.Broker
+	e             *echo.Echo
+	recorder      cosmos_client.CosmosMessageClient
+	broker        *broker.Broker
+	artifactStore *artifacts.ManagedArtifactStore // Optional: for off-chain PoC artifacts
```
DimaOrekhovPS: It's optional, but a participant won't be able to participate in PoC if they don't setup a store, right?

decentralized-api/poc/artifacts/managed_store.go
```diff
@@ -0,0 +1,264 @@
...
+	m.stores[pocStageStartHeight] = store
+	return store, nil
+}
+
+// GetStore returns the store for the given PoC stage, or an error if it doesn't exist.
+// Does not create new stores (for proof requests).
+func (m *ManagedArtifactStore) GetStore(pocStageStartHeight int64) (*ArtifactStore, error) {
+	m.mu.RLock()
+	store, ok := m.stores[pocStageStartHeight]
+	m.mu.RUnlock()
```
DimaOrekhovPS, line 74: Technically once we release the lock here the cleanup routine can proceed and start cleaning the store we just obtained

deploy/join/node-config.json
gmorgachev: Good call. I think it might be even better to rebuild mlnode/vLLM without this flag. So it's enabled by default (there is no overhead related to this flag, only during generation itself)

gmorgachev: Addressed in https://github.com/gonka-ai/vllm/tree/gm/poc-layers/

decentralized-api/internal/server/public/server.go
gmorgachev: Only single request for /poc/proofs/ expected from Participant A to Participant B per PoC round. It should be good if it's less then 100 `api` containers with identical api

But i think we can adjust this parameter at migration PoC v1 -> PoC v2 migration period if it'd cause issues

decentralized-api/poc/proof_client.go
gmorgachev: /v2/ is introduced for internal callback api. for public api is new endpoint in addition to another off-chain exchange endpoint. I think should be good...

decentralized-api/poc/validator.go
gmorgachev: It's at the moment between PoC phases, so if node is failed here, it can't validated PoC in any cases 

In general yes, here it's used for PoC v2

decentralized-api/poc/validator.go
gmorgachev: Yes, we probably need to double check time between `/inference/up` and until successful `/health`. Now it should be resolved by retry...

inference-chain/x/inference/keeper/msg_server_poc_v2_commit.go
gmorgachev: Hi. You're right, thanks. Adding fix

decentralized-api/poc/validator.go
gmorgachev: Thanks. Couple lines below it really created full array. Changing implementation to use lazy approach https://www.drmaciver.com/2018/01/lazy-fisher-yates-shuffling-for-precise-rejection-sampling/

decentralized-api/poc/validator.go
gmorgachev: Good call, thanks! Adding

decentralized-api/broker/node_worker_commands.go
gmorgachev: Status inference would be good, as PoC v2 works during the inference now 
Adding explicit check that status is inference or poc

decentralized-api/poc/proof_client.go
gmorgachev: Good call, thanks! Adding explicit check

decentralized-api/poc/commit_worker.go
gmorgachev: You're right, it might not be deterministic. The impact itself is low as it's only of potentially small leftover across MLNodes of the same participant. Still worth to fix for easier debugging  (adding fix)

decentralized-api/cosmosclient/tx_manager/tx_deadline_config.go
gmorgachev: forgot to get back, reverting

decentralized-api/internal/server/mlnode/server.go
gmorgachev: yes, it's not optional. if store is nil, it must be handled as error

decentralized-api/poc/artifacts/managed_store.go
gmorgachev: We prune only oldest store (if > 10). But de-facto write only in 1 latest and read only from one latest after write finished. Buffer is mostly just in case


Adding comment:
```
// ManagedArtifactStore wraps per-stage ArtifactStores with automatic pruning.
//
// Lifecycle:
//   - Short-lived stores, one per PoC stage (poc_stage_start_block_height)
//   - Writes go only to the newest store (via GetOrCreateStore)
//   - Reads happen from the newest store after writes complete
//   - Old stores are pruned when count exceeds retainCount (default 10)
//
// The large retention buffer ensures pruned stores are "cold" with no active use.
```

decentralized-api/poc/validator_v1.go
gmorgachev: made it longer

mlnode/packages/api/src/api/inference/pow_v2_routes.py
```diff
@@ -0,0 +1,249 @@
...
+        
+        # For queued requests, create composite request_id
+        if data.get("status") == "queued" and "request_id" in data:
+            data["request_id"] = f"{port}:{data['request_id']}"
+        
+        await _release_vllm_backend(port)
+        return data
+        
+    except HTTPException:
+        raise
```
jacky6block, line 215: When an `HTTPException` is caught and re-raised here, the backend connection counter doesn't get decremented

mlnode/packages/api/src/api/proxy.py
```diff
@@ -182,6 +219,20 @@ async def _health_check_vllm(interval: float = 2.0):
...
+            
+            # Poll PoC status for healthy backends
+            if ok:
+                try:
+                    r = await vllm_client.get(f"http://{VLLM_HOST}:{p}/api/v1/pow/status", timeout=2)
+                    if r.status_code == 200:
+                        data = r.json()
+                        poc_status_by_port[p] = data.get("status", "")
+                    else:
+                        poc_status_by_port[p] = ""
```
jacky6block, line 231: the 2-second timeout here might be a bit tight under high load. Also, silently swallowing exceptions could make debugging tricky later.

  Maybe something like:
```

except Exception as e:
      logger.debug(f"PoC status check failed for port {p}: {e}")
      poc_status_by_port[p] = ""
```
  
  And consider bumping the timeout to 5 seconds if you've seen any flakiness in testing.

mlnode/packages/api/src/api/inference/pow_v2_routes.py
```diff
@@ -0,0 +1,249 @@
...
+    block_height: int
+    public_key: str
+    node_id: int
+    node_count: int
+    nonces: List[int]
+    params: PoCParamsModel
+    batch_size: int = 32
+    wait: bool = False
+    url: Optional[str] = None
+    validation: Optional[ValidationModel] = None
```
jacky6block, line 70: It might be worth validating that nonces and validation.artifacts actually match at this layer. Right now mismatches only get caught deeper in vLLM, which gives less clear error messages.

  A Pydantic validator could catch this early:
```
  @model_validator(mode='after')
  def validate_nonces_match(self):
      if self.validation:
          artifact_nonces = {a.nonce for a in self.validation.artifacts}
          request_nonces = set(self.nonces)
          if artifact_nonces != request_nonces:
              raise ValueError(
                  f"nonces mismatch: request has {len(request_nonces)}, "
                  f"validation has {len(artifact_nonces)}"
              )
      return self
```

jacky6block (commented):
Hi, overall the mlnode PoC v2 implementation looks solid

  Left a few comments:
  - **One issue to address**: Connection counter leak in `/generate` when HTTPException is re-raised (should use `finally` block)
  - **Two suggestions**: Longer timeout for PoC status polling, and early validation of nonces/artifacts match

PR #623, opened 2026-01-23
Special allowance for Status endpoint

blizko:
This PR adds an explicit route for the status endpoint (bypassing the general RPC block).
This allows external observers to see the sync status of the node and thus evaluate the health of the network.
Overall load on this particular endpoint is negligible in comparison to other RPC calls

### DIFF
--- a/proxy/nginx.unified.conf.template
+++ b/proxy/nginx.unified.conf.template
@@ -124,6 +124,25 @@ http {
             proxy_read_timeout ${CHAIN_RPC_TRANSFER_TIMEOUT}s;
         }
 
+        # Status endpoint without limits
+        location /chain-rpc/status {
+            ${LIMIT_REQ_RULE_CHAIN_RPC}
+            ${CHAIN_RPC_STATUS}
+            proxy_pass http://chain_rpc_backend/status;
+            proxy_set_header Host $$host;
+            proxy_set_header X-Real-IP $$remote_addr;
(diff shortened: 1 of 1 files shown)


AlexeySamosadov (changes requested):
The idea of keeping `/status` accessible when Chain RPC is disabled makes sense, but the implementation has a couple of issues:

### [Must fix] Comment contradicts the code

The comment says `# Status endpoint without limits`, but the block includes `${LIMIT_REQ_RULE_CHAIN_RPC}` — so rate limiting IS applied. Either remove the limit (if "without limits" is the intent) or fix the comment to something like:

```nginx
# Status endpoint - accessible even when Chain RPC is disabled
```

### [Suggestion] Document why `${CHAIN_RPC_STATUS}` is absent

The key difference from the main `/chain-rpc/` block is the missing `${CHAIN_RPC_STATUS}` directive — this is what allows `/status` to work when `DISABLE_CHAIN_RPC=true`. This design choice is correct but non-obvious. A brief comment would help future readers understand it's intentional, not an omission.

### [Suggestion] `${STREAMING_CONFIG}` is unnecessary here

`/status` returns a simple JSON object — WebSocket upgrade headers and disabled buffering from `${STREAMING_CONFIG}` don't apply. Removing it keeps the config cleaner.

patimen:
Closing this as stale for now. The proxy config has moved on since this was opened, and this patch would need to be refreshed against the current layout and review comments. Feel free to open this again if you get a chance to address the requested changes.

PR #625, opened 2026-01-23
fix: avoid uint32 truncation for large validation weights

x0152:
switch ShouldValidate weights to int64 to prevent uint32 truncation from corrupting validation selection at scale

+guard the totalPower == executorPower division‑by‑zero case


AlexeySamosadov (commented):
## Review Notes

Clean PR, looks good overall.

### Positive

- Correct fix: `uint32` → `int64` prevents silent truncation for large validation weights
- `int64` over `uint64` is the right call here — the signed subtraction `totalPower - executorPower` needs a negative check, which wouldn't work with unsigned types
- Division-by-zero and negative-denom guards are good defense-in-depth
- Descriptive guard messages with actual values (`totalPower=%d, executorPower=%d, denom=%d`) — helpful for production debugging
- Call site cleanup in `msg_server_claim_rewards.go` removes unnecessary casts since underlying values are already `int64`
- Tests with actual assertions via `require`, checking both the boolean and the message string

### Minor suggestions

- Consider adding a test case with a value exceeding `uint32` max (e.g. `totalPower = 5_000_000_000`) — that's the original motivation for the fix and would serve as a regression test
- `validatorPower` could also theoretically be negative now with `int64` — a guard similar to the denom guards would catch data corruption early
- A brief comment explaining the `int64` choice (signed arithmetic needed) would help future reviewers wondering why not `uint64`

patimen:
This should be folded in to the larger effort for handling math overflows.

x0152:
> This should be folded into the larger effort for handling math overflows.

Agree - let's close this one and continue in #1101 as part of the larger math-overflow effort

P.S. weird, I didn’t see this pr in the pr list yesterday

PR #634, opened 2026-01-24
security: add request body size limit to prevent DoS

AlexeySamosadov:
## Summary

Add request body size limit to prevent memory exhaustion DoS attacks.

## Problem

The `readRequestBody` function reads the entire HTTP request body into memory without any size limit:

```go
func readRequestBody(r *http.Request) ([]byte, error) {
    var buf bytes.Buffer
    if _, err := io.Copy(&buf, r.Body); err != nil {  // NO SIZE LIMIT
        return nil, err
    }
    // ...
}
```

An attacker can send arbitrarily large request bodies to `/v1/chat/completions`, causing:
- Unbounded memory allocation
- OOM kill and node crash
- Network-wide DoS if multiple nodes are targeted

## Solution

- Add `MaxRequestBodySize` constant (10 MB)
- Use `http.MaxBytesReader` to enforce the limit
- Requests exceeding the limit are rejected with an error

## Changes

  - Add `MaxRequestBodySize = 10 * 1024 * 1024` constant
  - Wrap `r.Body` with `http.MaxBytesReader` in `readRequestBody`

## Test plan

- [x] Add unit test for normal-sized requests
- [x] Add unit test for oversized requests (should error)
- [x] Add unit test for exactly-at-limit requests
- [x] Add unit test for empty body
- [x] Add test verifying constant value

### DIFF
--- a/decentralized-api/internal/server/public/post_chat_handler.go
+++ b/decentralized-api/internal/server/public/post_chat_handler.go
@@ -61,8 +61,9 @@ var (
 	configManagerRef *apiconfig.ConfigManager
 )
 
-func NewNoRedirectClient() *http.Client {
+func NewNoRedirectClient(timeout time.Duration) *http.Client {
 	return &http.Client{
+		Timeout: timeout,
 		CheckRedirect: func(req *http.Request, via []*http.Request) error {
 			return http.ErrUseLastResponse
 		},
@@ -401,7 +402,7 @@ func (s *Server) handleTransferRequest(ctx echo.Context, request *ChatRequest) e
 	req.Header.Set(utils.XPromptHashHeader, inferenceRequest.PromptHash)
 	req.Header.Set("Content-Type", request.Request.Header.Get("Content-Type"))
 
-	resp, err := NewNoRedirectClient().Do(req)
+	resp, err := s.httpClient.Do(req)
 	if err != nil {
 		logging.Error("Failed to make http request to executor", types.Inferences, "error", err, "url", executor.Url)
 		return err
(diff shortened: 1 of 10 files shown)


decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -950,6 +954,9 @@ func readRequest(request *http.Request, transferAddress string) (*ChatRequest, e
 }
 
 func readRequestBody(r *http.Request) ([]byte, error) {
+	// Limit request body size to prevent memory exhaustion attacks
+	r.Body = http.MaxBytesReader(nil, r.Body, MaxRequestBodySize)
```
patimen, line 968: Maybe we should pass in the Writer instead of using nil?

decentralized-api/internal/server/public/post_chat_handler_test.go
```diff
@@ -164,3 +173,57 @@ func TestEmptyButParseableResponsePayload_EnforcedTokensEmptySlice(t *testing.T)
...
+	result, err := readRequestBody(req)
+	require.NoError(t, err)
+	require.Equal(t, body, result)
+}
+
+// TestReadRequestBody_ExceedsMaxSize tests that oversized requests are rejected
+func TestReadRequestBody_ExceedsMaxSize(t *testing.T) {
+	// Create a body larger than MaxRequestBodySize (10 MB)
+	oversizedBody := make([]byte, MaxRequestBodySize+1)
+	for i := range oversizedBody {
```
patimen, line 191: It's a nit, but you could use `bytes.Repeat([]byte("a"), MaxRequestBodySize+1)` instead of a for loop.

PR #635, opened 2026-01-24
security: add response body size limits to prevent DoS

AlexeySamosadov:
## Summary

Add response body size limits to prevent memory exhaustion DoS attacks from malicious executor nodes.

## Problem

Multiple locations use `io.ReadAll(resp.Body)` without size limits:

```go
// post_chat_handler.go:908
bodyBytes, err := io.ReadAll(resp.Body)  // NO SIZE LIMIT

// proxy.go:91
var bodyBytes, err = io.ReadAll(resp.Body)  // NO SIZE LIMIT
```

A malicious executor node could return extremely large responses, causing Transfer Agent nodes to allocate unbounded memory and crash from OOM.

## Solution

- Add `MaxResponseBodySize` constant (50 MB) for normal responses
- Add `MaxErrorResponseSize` constant (1 MB) for error messages
- Use `io.LimitReader` to enforce limits before `io.ReadAll`

## Changes

  - Add `MaxResponseBodySize` and `MaxErrorResponseSize` constants
  - Use `io.LimitReader` in `getInferenceErrorMessage`

- `decentralized-api/internal/server/public/proxy.go`
  - Use `io.LimitReader` in `proxyJsonResponse`

## Test plan

- [ ] Manual test with normal responses
- [ ] Verify responses over 50 MB are truncated
- [ ] Verify error responses over 1 MB are truncated

### DIFF
--- a/decentralized-api/internal/server/public/post_chat_handler.go
+++ b/decentralized-api/internal/server/public/post_chat_handler.go
@@ -37,6 +37,13 @@ const (
 	ExecutorContext AuthKeyContext = 2
 	// BothContexts indicates the AuthKey was used for both transfer and executor requests
 	BothContexts = TransferContext | ExecutorContext
+
+	// MaxResponseBodySize is the maximum allowed size for response bodies (50 MB)
+	// This prevents memory exhaustion attacks from malicious executor responses
+	MaxResponseBodySize = 50 * 1024 * 1024
+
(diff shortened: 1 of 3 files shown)


patimen:
Closing this as stale. The issue is valid, but this area is being refactored as part of the devshards/sharded direction, and it is not worth patching in this older code path separately.

PR #636, opened 2026-01-24
security: fix race condition in AuthKey check (TOCTOU)

AlexeySamosadov:
## Summary

Fix time-of-check to time-of-use (TOCTOU) race condition in `checkAndRecordAuthKey` that could allow replay attacks.

## Problem

The `checkAndRecordAuthKey` function had a race condition between checking if an AuthKey exists and recording it:

```go
func checkAndRecordAuthKey(...) bool {
    authKeysMutex.RLock()
    existingContext, exists := usedAuthKeys[authKey]  // CHECK
    authKeysMutex.RUnlock()
    // <-- RACE WINDOW: another goroutine can add the same key here

    authKeysMutex.Lock()
    usedAuthKeys[authKey] = context  // USE
    // ...
}
```

In the window between releasing `RLock` and acquiring `Lock`, another goroutine could record the same key, potentially allowing:
- Replay attacks with the same AuthKey
- Duplicate inference requests

## Solution

Use a single `Lock` for the entire check-and-record operation:

```go
func checkAndRecordAuthKey(...) bool {
    authKeysMutex.Lock()
    defer authKeysMutex.Unlock()

    existingContext, exists := usedAuthKeys[authKey]
    // ... rest of logic under same lock
}
```

## Changes

  - Replace `RLock` + `Lock` pattern with single `Lock`
  - Add explanatory comment about the fix

## Test plan

- [x] Add basic unit test for AuthKey recording
- [x] Add concurrent test with 100 goroutines
- [x] Verify exactly one goroutine records the key first
- [x] Verify all other goroutines see key as already used

### DIFF
--- a/decentralized-api/internal/server/public/post_chat_handler.go
+++ b/decentralized-api/internal/server/public/post_chat_handler.go
@@ -118,10 +118,15 @@ func emptyButParseableResponsePayload(inferenceId, model string, promptTokens ui
 
 // checkAndRecordAuthKey checks if an AuthKey has been used before and records it if not
 // Returns true if the key has been used before in the specified context, false otherwise
+//
+// This function uses a single Lock for the entire check-and-record operation to prevent
+// race conditions (TOCTOU). Previously, using RLock for check followed by Lock for write
+// created a window where another goroutine could record the same key.
 func checkAndRecordAuthKey(authKey string, currentBlockHeight int64, context AuthKeyContext) bool {
-	authKeysMutex.RLock()
+	authKeysMutex.Lock()
+	defer authKeysMutex.Unlock()
+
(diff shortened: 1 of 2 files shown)


patimen:
Closing this as stale.

The original concern was on the legacy `AuthKey` replay-check path in `decentralized-api/internal/server/public/post_chat_handler.go`. That locking issue is already fixed on the current baseline, so this PR is no longer the right vehicle.

Separately, current devshard session traffic uses the newer session-scoped `/v1/devshard/sessions/:id/chat/completions` path with different request authentication, so this specific TOCTOU concern does not apply there. We also do not have evidence that this was happening in practice.

Given the deprecation of the old path, this no longer looks worth carrying forward as an open PR.

PR #641, opened 2026-01-24
Add MsgTransferWithVesting for vesting transfers

### LINKED ISSUE #631: Add message to transfer amount with vesting
When the community distributes funds to miners, the transferred tokens should vest over a fixed 180-epoch period, rather than being fully available at the time of transfer.

AlexeySamosadov:
## Summary
Add a new message type `MsgTransferWithVesting` to the streamvesting module that allows transferring tokens with a vesting schedule.

### Features
- Transfer tokens from sender to recipient with vesting over N epochs
- Default vesting period of 180 epochs if not specified
- Tokens vest linearly with remainder going to first epoch
- Full validation of sender/recipient addresses and amount

### CLI Usage
```bash
inferenced tx streamvesting transfer-with-vesting [recipient] [amount] [vesting-epochs]
```

### Files changed
- `proto/inference/streamvesting/tx.proto` - Add message definition
- `x/streamvesting/types/tx.pb.go` - Generated protobuf code
- `x/streamvesting/types/codec.go` - Register new message
- `x/streamvesting/types/events.go` - Add event types
- `x/streamvesting/keeper/msg_transfer_with_vesting.go` - Handler implementation
- `x/streamvesting/keeper/msg_transfer_with_vesting_test.go` - Unit tests
- `x/streamvesting/module/autocli.go` - CLI configuration

## Test plan
- [x] Unit tests for validation (invalid addresses, zero amount)
- [x] Unit tests for valid transfer with custom epochs
- [x] Unit tests for default 180 epoch vesting
- [x] All existing streamvesting tests pass

Closes #631

### DIFF
--- a/proposals/governance-artifacts/update-v0.2.10/README.md
+++ b/proposals/governance-artifacts/update-v0.2.10/README.md
@@ -1,0 +1,89 @@
+# Upgrade Proposal: v0.2.10
+
+This document outlines the proposed changes for on-chain software upgrade v0.2.10. The `Changes` section details the major modifications, and the `Upgrade Plan` section describes the process for applying these changes.
+
+## Upgrade Plan
+
+This PR updates the code for the `api` and `node` services. The PR modifies the container versions in `deploy/join/docker-compose.yml`.
+
+The binary versions will be updated via an on-chain upgrade proposal. For more information on the upgrade process, refer to [`/docs/upgrades.md`](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.10/docs/upgrades.md).
+
+Existing hosts are **not** required to upgrade their `api` and `node` containers. The updated container versions are intended for new hosts who join after the on-chain upgrade is complete.
+
+## Proposed Process
+
+1. Active hosts review this proposal on GitHub.
+2. Once the PR is approved by a majority, a `v0.2.10` release will be created from this branch, and an on-chain upgrade proposal for this version will be submitted.
+3. If the on-chain proposal is approved, this PR will be merged immediately after the upgrade is executed on-chain.
+
+Creating the release from this branch (instead of `main`) minimizes the time that the `/deploy/join/` directory on the `main` branch contains container versions that do not match the on-chain binary versions, ensuring a smoother onboarding experience for new hosts.
+
+## Testing
+
+This branch includes upgrade-path changes and supporting testnet tooling. Reviewers are encouraged to request access to testnet environments to validate both node behavior and the on-chain upgrade process, or to replay the upgrade on private testnets.
+
+## Migration
+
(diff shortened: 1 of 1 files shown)


inference-chain/x/streamvesting/keeper/msg_transfer_with_vesting.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Implement aggregation logic for each coin denomination
+	for _, coin := range req.Amount {
+		// Divide amount by epochs
+		epochsInt := math.NewInt(int64(vestingEpochs))
+		amountPerEpoch := coin.Amount.Quo(epochsInt)
+		remainder := coin.Amount.Mod(epochsInt)
+
+		// Add the base amount to each epoch
+		for i := 0; i < int(vestingEpochs); i++ {
```
patimen: Same as above, avoid `int`

inference-chain/x/streamvesting/keeper/msg_transfer_with_vesting.go
```diff
@@ -0,0 +1,121 @@
...
+	schedule, found := k.GetVestingSchedule(ctx, req.Recipient)
+	if !found {
+		schedule = types.VestingSchedule{
+			ParticipantAddress: req.Recipient,
+			EpochAmounts:       []types.EpochCoins{},
+		}
+	}
+
+	// Extend the schedule if necessary
+	requiredLength := int(vestingEpochs)
```
patimen: This should convert explicitly to either int64 or int32, we should not be using `int`

inference-chain/x/streamvesting/keeper/msg_transfer_with_vesting.go
patimen: We need some kind of upper limit on vestingEpochs, or this could be given a gigantic number causing resource problems in this loop.

decentralized-api/apiconfig/config_test.go
```diff
@@ -37,6 +41,159 @@ func TestNodeVersion(t *testing.T) {
 	require.Equal(t, testManager.GetCurrentNodeVersion(), "v3.0.8")
 }
 
+func TestSetCurrentNodeVersion(t *testing.T) {
```
patimen, line 44: These don't seem related to the rest of the PR.

inference-chain/x/streamvesting/keeper/msg_transfer_with_vesting.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Validate amount
+	if req.Amount.IsZero() {
+		return nil, errorsmod.Wrap(sdkerrors.ErrInvalidCoins, "amount cannot be zero")
+	}
+
+	if !req.Amount.IsValid() {
+		return nil, errorsmod.Wrap(sdkerrors.ErrInvalidCoins, "invalid coins")
+	}
+
```
patimen, line 49: Right here we should check:
VestingEpochs upper limit.
Upper limit to number of values in req.Amount.
Without these two, an attacker could push in high numbers for both and we have N*M complexity below.

PR #642, opened 2026-01-26
[P0] fix: model specific weights

### LINKED ISSUE #629: [P0] Possible cause of missed inferences
We're distributing inference requests on the chain based on the total weight of the participant, not the weight of the participant's mlnode for a specific `model_id`. Seems like it's easy can be the cause of missed inferences (e.g. I have 100 nodes for `model_id1` and 3 nodes for `model_id2`, but I get the amount of requests based on the weight of 103 nodes for `model_id2`)

x0152:
fixes #629

### DIFF
--- a/inference-chain/x/inference/epochgroup/epoch_group.go
+++ b/inference-chain/x/inference/epochgroup/epoch_group.go
@@ -52,6 +52,35 @@ func NewEpochMemberFromActiveParticipant(p *types.ActiveParticipant, reputation 
 	}
 }
 
+func (m EpochMember) ForModel(modelId string) EpochMember {
+	// We're explicitly passing only this model to prevent further recursion
+	subMember := m
+	subMember.Models = []string{modelId}
+	subMember.MlNodes = m.modelMLNodes(modelId)
+
+	// Model subgroup weight = PoC node weight for this model's nodes
+	modelWeight := calculatePocParticipatingNodesWeight(subMember.MlNodes)
+	subMember.Weight = modelWeight
+	subMember.ConfirmationWeight = modelWeight
+
+	return subMember
+}
+
+func (m EpochMember) modelMLNodes(modelId string) []*types.ModelMLNodes {
+	modelIndex := -1
+	for i, model := range m.Models {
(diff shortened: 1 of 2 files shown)


AlexeySamosadov (commented):
Clean fix — the one-liner correctly recalculates the weight per model instead of carrying the parent's aggregate weight. Test demonstrates the issue well.

### [Question] Does `ConfirmationWeight` have the same problem?

`ConfirmationWeight` is initialized from all MlNodes in `NewEpochMemberFromActiveParticipant` but is not recalculated when the member is added to a model subgroup — same pattern as the `Weight` bug this PR fixes. If a participant has 100 nodes for model A and 3 for model B, model B's subgroup would get a `ConfirmationWeight` based on all 103 nodes. Worth checking if this needs the same treatment, or if `ConfirmationWeight` is intentionally kept at the aggregate level.

### [Suggestion] Add a two-model test case

The current test covers the fix well, but a test with two models (the exact scenario from issue #629 — e.g., 100 nodes for model A, 3 for model B) would make it even more convincing and prevent regressions.

AlexeySamosadov:
Makes sense — if `ConfirmationWeight` is only relevant at the main group level, then no action needed here. Thanks for clarifying.

x0152:
I decided to also make ConfirmationWeight model specific to prevent future bugs, refactored the subgroup member creation to be cleaner, and added one more test

tcharchian:
This PR is a subset of multi-model support (https://github.com/gonka-ai/gonka/issues/728)
We'll figure out when to review and merge this PR during work on multimodels https://github.com/gonka-ai/gonka/issues/728
@x0152 @0xgonka

x0152:
This was meant to be a quick fix for the specific issue described in #629 only. The capacity/serving weight stuff is a bit tricky in the current code so better to deal with it in a separate pr

Copilot is pointing to a different issue: in this flow subgroup weight is for validation/voting, not serving capacity. Including POC_SLOT=true here would change capacity behavior, so it should be handled in a separate pr

x0152:
Closing as outdated

PR #643, opened 2026-01-26
fix: guard nil participants in executor selection

### LINKED ISSUE #499: Chat Completions aren't working
```
base_url: http://192.241.240.19:8000/v1
Request signing is enabled through a custom HTTP client implementation.
Using Gonka address: gonka1twg5lhad6sc86yjqspanmp94md2dx0fhfppxus
INFO:httpx:HTTP Request: POST http://192.241.240.19:8000/v1/chat/completions "HTTP/1.1 500 Internal Server Error"
INFO:openai._base_client:Retrying request to /chat/completions in 0.400175 seconds
INFO:httpx:HTTP Request: POST http://192.241.240.19:8000/v1/chat/completions "HTTP/1.1 500 Internal Server Error"
INFO:openai._base_client:Retrying request to /chat/completions in 0.879420 seconds
INFO:httpx:HTTP Request: POST http://192.241.240.19:8000/v1/chat/completions "HTTP/1.1 500 Internal Server Error"
Traceback (most recent call last):
  File "/Users/bob/git/gonka/script.py", line 9, in <module>
    response = client.chat.completions.create(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bob/git/gonka/.venv/lib/python3.12/site-packages/openai/_utils/_utils.py", line 286, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bob/git/gonka/.venv/lib/python3.12/site-packages/openai/resources/chat/completions/completions.py", line 1192, in create
    return self._post(
           ^^^^^^^^^^^
  File "/Users/bob/git/gonka/.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1259, in post
    return cast(ResponseT, self.request(cast_to, opts, stream=stream, stream_cls=stream_cls))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bob/git/gonka/.venv/lib/python3.12/site-packages/openai/_base_client.py", line 1047, in request
    raise self._make_status_error_from_response(err.response) from None
openai.InternalServerError: Error code: 500 - {'error': 'rpc error: code = Unknown desc = runtime error: invalid memory address or nil pointer dereference: panic'}
```

script

```
import os
from gonka_openai import GonkaOpenAI

client = GonkaOpenAI(
    gonka_private_key=os.environ.get('GONKA_PRIVATE_KEY'),
    source_url="http://node3.gonka.ai:8000",
)

response = client.chat.completions.create(
    model="Qwen/Qwen3-235B-A22B-Instruct-2507-FP8",
    messages=[
        { "role": "user", "content": "Write a one-sentence bedtime story about a unicorn" }
    ]
)

print(response.choices[0].message.content)
```

x0152:
guard nil participants (maybe fixes #499)

### DIFF
--- a/decentralized-api/internal/server/public/post_chat_handler.go
+++ b/decentralized-api/internal/server/public/post_chat_handler.go
@@ -37,6 +37,10 @@ const (
 	ExecutorContext AuthKeyContext = 2
 	// BothContexts indicates the AuthKey was used for both transfer and executor requests
 	BothContexts = TransferContext | ExecutorContext
+
+	// MaxRequestBodySize is the maximum allowed size for request bodies (10 MB)
+	// This prevents memory exhaustion attacks from oversized requests
+	MaxRequestBodySize = 10 * 1024 * 1024
 )
 
 // Package-level variables for AuthKey reuse prevention
@@ -57,8 +61,9 @@ var (
 	configManagerRef *apiconfig.ConfigManager
 )
 
-func NewNoRedirectClient() *http.Client {
+func NewNoRedirectClient(timeout time.Duration) *http.Client {
 	return &http.Client{
+		Timeout: timeout,
 		CheckRedirect: func(req *http.Request, via []*http.Request) error {
 			return http.ErrUseLastResponse
 		},
@@ -397,7 +402,7 @@ func (s *Server) handleTransferRequest(ctx echo.Context, request *ChatRequest) e
 	req.Header.Set(utils.XPromptHashHeader, inferenceRequest.PromptHash)
 	req.Header.Set("Content-Type", request.Request.Header.Get("Content-Type"))
 
-	resp, err := NewNoRedirectClient().Do(req)
+	resp, err := s.httpClient.Do(req)
 	if err != nil {
 		logging.Error("Failed to make http request to executor", types.Inferences, "error", err, "url", executor.Url)
 		return err
@@ -469,7 +474,7 @@ func (s *Server) getPromptTokenCount(text string, model string) (int, error) {
 			return nil, broker.NewApplicationActionError(err)
 		}
 
-		resp, postErr := http.Post(
+		resp, postErr := s.httpClient.Post(
 			tokenizeUrl,
 			"application/json",
 			bytes.NewReader(jsonData),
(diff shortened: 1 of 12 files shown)


inference-chain/x/inference/epochgroup/random.go
```diff
@@ -87,9 +89,26 @@ func computeCumulativeArray(participants []*group.GroupMember) []int64 {
...
+		return 0
+	}
 	weight, err := strconv.Atoi(participant.Member.Weight)
 	if err != nil {
 		return 0
 	}
 	return int64(weight)
 }
+
+func sanitizeMembers(members []*group.GroupMember) []*group.GroupMember {
```
patimen, line 101: We might want to filter in place instead of creating a new slice? But unlikely to be an issue.

inference-chain/x/inference/epochgroup/random.go
```diff
@@ -43,11 +43,13 @@ func (eg *EpochGroup) GetRandomMember(
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+	activeParticipants = sanitizeMembers(activeParticipants)
 	if len(activeParticipants) == 0 {
 		return nil, status.Error(codes.Internal, "Active participants found, but length is 0")
 	}
 
 	filteredParticipants := filterFn(activeParticipants)
+	filteredParticipants = sanitizeMembers(filteredParticipants)
```
patimen: Why do we need to sanitize again?

x0152: Hi!
My thinking was that filterFn can be any filter and GetRandomMember is reusable, so the extra sanitizeMembers is a cheap safety net if future filter doesn’t guard against nil

x0152: But now I think it looks a bit strange, since it’s unlikely anyone will add logic like that inside the filter lambda so I’ll remove it. 
Thanks

inference-chain/x/inference/epochgroup/random.go
x0152: Just in case activeParticipants is being used by another goroutine at the same time
I think copying is safer and cleaner, and the overhead here is minimal

PR #644, opened 2026-01-26
fix: avoid rewriting config on every startup

### LINKED ISSUE #466: Stop rewriting static config at each app start

x0152:
fixes #466

### DIFF
--- a/decentralized-api/Dockerfile
+++ b/decentralized-api/Dockerfile
@@ -8,6 +8,7 @@ FROM golang:1.24.2-alpine3.20 AS builder
 ARG BUILD_FLAGS
 ARG GOOS
 ARG GOARCH
+ARG BLST_PORTABLE=0
 
 ENV GOOS=${GOOS} \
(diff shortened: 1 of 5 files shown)


decentralized-api/apiconfig/config_manager.go
```diff
@@ -70,17 +70,19 @@ func LoadConfigManagerWithPaths(configPath, sqlitePath, nodeConfigPath string) (
...
 		log.Printf("Error migrating dynamic data to DB: %+v", err)
 		return nil, err
 	}
 
-	if err = manager.Write(); err != nil {
-		log.Printf("Error writing config: %+v", err)
-		return nil, err
+	if migrated {
+		if err = manager.Write(); err != nil {
+			log.Printf("Error writing config: %+v", err)
```
patimen, line 81: Let's use our standard logging (logging.Info) instead of a println.

x0152: valid point, thanks - fixed it

PR #645, opened 2026-01-26
feat(inference): add punishment statistics query helper

0xMayoor:
Query helper that re-runs MissedStatTest on existing EpochPerformanceSummary data to determine if participant was punished. No proto changes needed.

#561

### DIFF
--- a/inference-chain/x/inference/keeper/query_punishment_stats.go
+++ b/inference-chain/x/inference/keeper/query_punishment_stats.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/productscience/inference/x/inference/calculations"
 	"github.com/productscience/inference/x/inference/types"
+	"github.com/shopspring/decimal"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
@@ -16,7 +17,7 @@ type PunishmentStats struct {
 	WasPunished    bool
 	MissedRequests uint64
 	TotalRequests  uint64
-	MissedRatio    float64
+	MissedRatio    decimal.Decimal
 	// RewardedCoins from EpochPerformanceSummary (0 if punished, >0 if not)
 	RewardedCoins uint64
 }
@@ -25,12 +26,17 @@ type PunishmentStats struct {
 // by re-running the same statistical test used during settlement.
 // This uses existing EpochPerformanceSummary data - no additional storage required.
 func (k Keeper) GetPunishmentStats(ctx context.Context, epochIndex uint64, participantId string) (*PunishmentStats, error) {
-	// Get the stored performance summary for this epoch/participant
 	summary, found := k.GetEpochPerformanceSummary(ctx, epochIndex, participantId)
 	if !found {
 		return nil, status.Error(codes.NotFound, "participant not found for epoch")
 	}
 
+	return k.ComputePunishmentStats(ctx, summary)
+}
+
+// ComputePunishmentStats computes punishment stats from a summary.
+// Exported so callers with existing summary data can avoid redundant lookups.
+func (k Keeper) ComputePunishmentStats(ctx context.Context, summary types.EpochPerformanceSummary) (*PunishmentStats, error) {
 	total := summary.InferenceCount + summary.MissedRequests
 
 	// Case 1: No work assigned - code returns full reward, not punishment
@@ -40,7 +46,7 @@ func (k Keeper) GetPunishmentStats(ctx context.Context, epochIndex uint64, parti
 			WasPunished:    false,
 			MissedRequests: 0,
 			TotalRequests:  0,
-			MissedRatio:    0,
+			MissedRatio:    decimal.Zero,
 			RewardedCoins:  summary.RewardedCoins,
 		}, nil
 	}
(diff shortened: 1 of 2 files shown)


inference-chain/x/inference/keeper/query_punishment_stats.go
```diff
@@ -0,0 +1,104 @@
...
+		MissedRequests: summary.MissedRequests,
+		TotalRequests:  total,
+		MissedRatio:    float64(summary.MissedRequests) / float64(total),
+		RewardedCoins:  summary.RewardedCoins,
+	}, nil
+}
+
+// GetPunishmentStatsForEpoch returns punishment stats for all participants in an epoch.
+func (k Keeper) GetPunishmentStatsForEpoch(ctx context.Context, epochIndex uint64) ([]PunishmentStats, error) {
+	all := k.GetAllEpochPerformanceSummary(ctx)
```
patimen: This could cause issues at scale.
At a minimum, you should walk directly over the EpochPerformanceSummaries and grab the data needed, instead of loading the entire data structure in memory and iterating over each one.
We likely need to add a secondary index to EpochPerformanceSummaries so we could query by epoch directly in the long term.
Ping me if you need pointers for how to walk through a Cosmos SDK collection. There are examples in the code (look for Walk(. )

inference-chain/x/inference/keeper/query_punishment_stats.go
```diff
@@ -0,0 +1,104 @@
...
+	// RewardedCoins from EpochPerformanceSummary (0 if punished, >0 if not)
+	RewardedCoins uint64
+}
+
+// GetPunishmentStats computes whether a participant was punished in a given epoch
+// by re-running the same statistical test used during settlement.
+// This uses existing EpochPerformanceSummary data - no additional storage required.
+func (k Keeper) GetPunishmentStats(ctx context.Context, epochIndex uint64, participantId string) (*PunishmentStats, error) {
+	// Get the stored performance summary for this epoch/participant
+	summary, found := k.GetEpochPerformanceSummary(ctx, epochIndex, participantId)
```
patimen, line 29: Don't we have the summaries from GetAllEpochPerformanceSummary?

inference-chain/x/inference/keeper/query_punishment_stats.go
```diff
@@ -0,0 +1,104 @@
...
+		p0.ToDecimal(),
+	)
+	if err != nil {
+		// Error in test = code returns full reward (not punishment)
+		// See accountsettle.go:57-58
+		return &PunishmentStats{
+			WasPunished:    false,
+			MissedRequests: summary.MissedRequests,
+			TotalRequests:  total,
+			MissedRatio:    float64(summary.MissedRequests) / float64(total),
```
patimen: It's a minor quibble, but maybe we should use shopspring/decimal here instead of relying on unreliable FP logic? This doesn't go directly into state, of course, but that might change.

inference-chain/x/inference/keeper/query_punishment_stats.go
0xMayoor: Looked into Walk(): 
From what i can tell since the key is (participant, epoch), querying by epoch alone would still need a full scan even with Walk? just wouldn't load everything into memory at once. 

You think we just drop the epoch-wide function for now and keep the single lookup?

inference-chain/x/inference/keeper/query_punishment_stats.go
0xMayoor: Sorry not sure what you mean, are you asking if we should reuse data from GetAll instead of the direct lookup? or something about how the two functions relate?

inference-chain/x/inference/keeper/query_punishment_stats.go
0xMayoor: fair, will switch to decimal

inference-chain/x/inference/keeper/query_punishment_stats.go
0xMayoor: Actually @patimen what are your thoughts on this approach:
EpochGroupData already stores ValidationWeights which has all participant addresses for each epoch. Could do:                                         

1. GetEpochGroupData(epochIndex,"") - direct key lookup
2. Extract participant ID's from ValidatioWeights
3. Batch lookup summaries via GetParticipantsEpochSummaries

Avoids scanning all historical summaries, just looks up the ones for that epoch's participants. 
Would that work or am I missing something?

patimen: One limit is that EpochGroupData will only have the participants left at the end of an epoch (none removed for bad behavior). Maybe ActiveParticipants instead (as it is immutable)?

inference-chain/x/inference/keeper/query_punishment_stats.go
patimen: Yes, I was thinking we already had the data and could just pass it in instead of another lookup.

inference-chain/x/inference/keeper/query_punishment_stats.go
0xMayoor: So @patimen  I could be wrong but here is something interesting:
I noticed RemoveMember updates the Cosmos group membership but doesn't touch ValidationWeights. 
Is that intentional or should it be synced?

ActiveParticipant might be indeed a better option either way but curious if I missed something?

inference-chain/x/inference/keeper/query_punishment_stats.go
0xMayoor: Agreed! Will refactor to pass it directly.

PR #668, opened 2026-01-29
Upgrade v0.2.9

patimen:
# Upgrade v0.2.9

Full PoC V2 activation with model consolidation and access controls.

## Summary

- PoC V2 fully enabled (tracking mode from v0.2.8 -> full enforcement)
- Network consolidated to single model: `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`
- Transfer Agent whitelist for request gating
- Guardian tiebreaker for undecided PoC V2 votes
- 24 suspicious participants removed from allowlist

## Chain Changes

**PoC V2 Activation**
- `PocV2Enabled=true`, `ConfirmationPocV2Enabled=true`
- `WeightScaleFactor=0.262`, `InferenceValidationCutoff=2`, `PocValidationDuration=480`
- All nodes' `POC_SLOT` allocations reset (both `ActiveParticipants` and `EpochGroupData`)
- First V2 epoch runs in grace mode, full enforcement begins the following epoch

**Guardian Tiebreaker**
- When neither valid nor invalid votes reach majority, guardians can break the tie
- Requires: no majority exists, guardians enabled, at least one guardian voted, all voting guardians agree

**Transfer Agent Whitelist**
- New `TransferAgentAccessParams.AllowedTransferAddresses` in params
- Validation in `StartInference` and `FinishInference` messages
- Empty whitelist = all TAs allowed; non-empty = only listed TAs

**Model Consolidation**
- All governance models except `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8` deleted during migration

**Participant Access Update**
- Registration and allowlist heights set to 2475000
- 24 addresses removed from allowlist (completed POC generation but did not vote at validation)

## API Changes

**Enforced Model Auto-Switch**
- Auto-switches nodes to Qwen235B with `--tensor-parallel-size 4 --max-model-len 240000` if model is not configured
- Three-layer enforcement: config load, state sync, runtime verification via vLLM `/v1/models`

**Transfer Agent Whitelist**
- Early enforcement in `/chat/completion` before expensive operations
- Cache synced from chain on every new block for O(1) lookups

## Bug Fixes

**[#674](https://github.com/gonka-ai/gonka/pull/674) Missed inferences fix**
- Don't punish for missed inferences of non-preserved nodes during PoC
- Don't punish if participant doesn't support the model

**[#678](https://github.com/gonka-ai/gonka/pull/678) CPoC downtime penalty redistribution**
- Penalty transferred to community pool instead of being lost

**Inference expiry during PoC/CPoC**
- New `InferenceExpiryContext` tracks latest PoC or CPoC range
- Inferences started during PoC/CPoC not punished for expiry

**Bitcoin rewards weight calculation**
- Full weights used for denominator (prevents redistribution of invalidated shares)
- CPoC reductions and invalidated shares go to governance pool


decentralized-api/broker/enforced_model.go
```diff
@@ -0,0 +1,49 @@
...
+	"os"
+	"strings"
+
+	"decentralized-api/apiconfig"
+)
+
+// Default enforced model for v0.2.9 upgrade
+const defaultEnforcedModelId = "Qwen/Qwen3-235B-A22B-Instruct-2507-FP8"
+
+var defaultEnforcedModelArgs = []string{"--tensor-parallel-size", "4", "--max-model-len", "240000"}
```
blizko, line 13: enforcing `max-model-len` in the local parameter set is incorrect. It is a governance parameter

x0152:
https://github.com/gonka-ai/gonka/pull/674#discussion_r2751148316

decentralized-api/broker/enforced_model.go
```diff
@@ -0,0 +1,49 @@
...
+func EnforceModel(node *apiconfig.InferenceNodeConfig) {
+	modelId, args := getEnforcedModel()
+	if modelId == "" {
+		return // enforcement disabled
+	}
+	if _, ok := node.Models[modelId]; ok {
+		return // node already has required model ID
+	}
+	node.Models = map[string]apiconfig.ModelConfig{
+		modelId: {Args: args},
```
x0152, line 47: Maybe this should be node.Models[modelId] = ...

decentralized-api/broker/enforced_model.go
```diff
@@ -0,0 +1,49 @@
...
+	"decentralized-api/apiconfig"
+)
+
+// Default enforced model for v0.2.9 upgrade
+const defaultEnforcedModelId = "Qwen/Qwen3-235B-A22B-Instruct-2507-FP8"
+
+var defaultEnforcedModelArgs = []string{"--tensor-parallel-size", "4", "--max-model-len", "240000"}
+
+// getEnforcedModel returns the enforced model ID and args from env vars or defaults.
+// Returns empty string if enforcement is disabled (ENFORCED_MODEL_ID="").
```
x0152, line 16: But defaultEnforcedModelId isn't empty string

decentralized-api/broker/enforced_model.go
gmorgachev: it's essentially the same once governance support only single model. the plan it to remove all enforcement at `api` side in 0.2.10

decentralized-api/broker/enforced_model.go
gmorgachev: agree, comment is a bit misleading. it returns empty enforced model only if disabled explicitly. that's also should be removed in next upgrade

gmorgachev (approved):

PR #673, opened 2026-01-30
PoC v2 Fixes/TA Agent Whitelist

patimen:
# Upgrade v0.2.9

Full PoC V2 activation with model consolidation and access controls.

## Summary

- PoC V2 fully enabled (tracking mode from v0.2.8 -> full enforcement)
- Network consolidated to single model: `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`
- Transfer Agent whitelist for request gating
- Guardian tiebreaker for undecided PoC V2 votes
- 24 suspicious participants removed from allowlist

## Chain Changes

**PoC V2 Activation**
- `PocV2Enabled=true`, `ConfirmationPocV2Enabled=true`
- `WeightScaleFactor=0.262`, `InferenceValidationCutoff=2`, `PocValidationDuration=480`
- All nodes' `POC_SLOT` allocations reset (both `ActiveParticipants` and `EpochGroupData`)
- First V2 epoch runs in grace mode, full enforcement begins the following epoch

**Guardian Tiebreaker**
- When neither valid nor invalid votes reach majority, guardians can break the tie
- Requires: no majority exists, guardians enabled, at least one guardian voted, all voting guardians agree

**Transfer Agent Whitelist**
- New `TransferAgentAccessParams.AllowedTransferAddresses` in params
- Validation in `StartInference` and `FinishInference` messages
- Empty whitelist = all TAs allowed; non-empty = only listed TAs

**Model Consolidation**
- All governance models except `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8` deleted during migration

**Suspicious Participant Removal**
- 24 addresses removed from allowlist (completed POC generation but did not vote at validation)

## API Changes

**Enforced Model Auto-Switch**
- Auto-switches nodes to qwen235B with `--tensor-parallel-size 4 --max-model-len 240000`
- Config enforcement overwrites node model config on load
- Runtime verification queries vLLM `/v1/models` endpoint, triggers redeploy on mismatch
- Disabled via `ENFORCED_MODEL_ID=none`

**Transfer Agent Whitelist**
- Early enforcement in `/chat/completion` before expensive operations
- Cache synced from chain on every new block for O(1) lookups

## Bug Fixes

**[#674](https://github.com/gonka-ai/gonka/pull/674) Missed inferences fix**
- Don't punish for missed inferences of non-preserved nodes during PoC
- Don't punish if participant doesn't support the model

**[#678](https://github.com/gonka-ai/gonka/pull/678) CPoC downtime penalty redistribution**
- Lower confirmation reward penalty transferred to community pool instead of being lost

**Inference expiry during PoC/CPoC**
- New `InferenceExpiryContext` tracks latest PoC or CPoC range
- Inferences started during PoC/CPoC not punished for expiry

**Bitcoin rewards weight calculation**
- Full weights used for denominator (prevents redistribution of invalidated participants' shares)
- CPoC reductions and invalidated shares go to governance pool

### DIFF
--- a/test-net-cloud/nebius/genesis-overrides.json
+++ b/test-net-cloud/nebius/genesis-overrides.json
@@ -183,6 +183,23 @@
             "value": "95",
             "exponent": -2
           }
+        },
+        {
+          "proposed_by": "genesis",
+          "id": "Qwen/Qwen3-235B-A22B-Instruct-2507-FP8",
+          "units_of_compute_per_token": "10000",
+          "hf_repo": "Qwen/Qwen3-235B-A22B-Instruct-2507-FP8",
+          "hf_commit": "c0eb82898e3da8fb6dd017e3e6698a5e37b3a3e6",
+          "model_args": [
+            "--max-model-len",
+            "240000"
+          ],
+          "v_ram": "320",
+          "throughput_per_nonce": "1500",
+          "validation_threshold": {
+            "value": "970917",
+            "exponent": -6
+          }
         }
       ],
       "genesis_only_params": {


decentralized-api/apiconfig/config_manager.go
```diff
@@ -292,6 +292,16 @@ func (cm *ConfigManager) GetBandwidthParams() BandwidthParamsCache {
 	return cm.currentConfig.BandwidthParams
 }
 
+func (cm *ConfigManager) SetTransferAgentAccessCache(cache TransferAgentAccessCache) {
+	cm.mutex.Lock()
+	defer cm.mutex.Unlock()
+	cm.currentConfig.TransferAgentAccessCache = cache
+}
+
+func (cm *ConfigManager) GetTransferAgentAccessCache() TransferAgentAccessCache {
```
patimen, line 301: We should have a read lock here to match our write lock

decentralized-api/broker/broker.go
```diff
@@ -995,10 +995,31 @@ func (b *Broker) reconcileIfSynced(triggerMsg string) {
...
+		return
+	}
+
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	for id, node := range b.nodes {
+		if _, ok := node.Node.Models[modelId]; ok {
+			continue
+		}
+		node.Node.Models = map[string]ModelArgs{
```
patimen: To be clear, we are replacing ALL models with only the enforced model? Not just making sure that models INCLUDE the enforced model?
BUT... if the model already exists in the list of models, it's left, along with all other models that might be there? This seems unintuitive.
If so, we likely should rename a few things... maybe this should be called `enforceSingleModelOnAllNodes`?

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -21,6 +21,13 @@ func (k msgServer) StartInference(goCtx context.Context, msg *types.MsgStartInfe
 		return failedStart(ctx, sdkerrors.Wrap(types.ErrDeveloperNotAllowlisted, msg.RequestedBy), msg), nil
 	}
 
+	// Transfer Agent access gating: only allowlisted TAs may submit StartInference.
+	if k.IsTransferAgentRestricted(ctx) && !k.IsAllowedTransferAgent(ctx, msg.Creator) {
```
patimen, line 25: Both of these calls call k.GetParams... I'm concerned about the performance of GetParams, since it serializes an entire gigantic params struct.
Maybe we could instead call getParams close to the top of the StartInference call and pass it in? We have other uses of getting params...
We can postpone this for now, but I think we do need to consider the perf of GetParams and start looking to optimize it soon.

inference-chain/proto/inference/inference/params.proto
```diff
@@ -312,3 +313,12 @@ message ParticipantAccessParams {
   // Set to 0 to disable cutoff (allowlist stays active as long as use_participant_allowlist=true).
   int64 participant_allowlist_until_block_height = 4;
 }
+
+// TransferAgentAccessParams controls which Transfer Agents are allowed to process inference requests.
+// If allowed_transfer_addresses is empty, all TAs are allowed (no restriction).
+message TransferAgentAccessParams {
```
patimen, line 319: We should add these new Params to the Testermint params in AppExport.kt, otherwise it could cause issues in governance tests when they try to serialize/deserialize parameter objects.

inference-chain/x/inference/keeper/params.go
```diff
@@ -233,3 +233,35 @@ func (k Keeper) IsPoCParticipantBlocked(ctx context.Context, address string) boo
...
+	p := k.GetTransferAgentAccessParams(ctx)
+	if p == nil || len(p.AllowedTransferAddresses) == 0 {
+		return true // no restriction
+	}
+	for _, a := range p.AllowedTransferAddresses {
+		if a == taAddress {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
gmorgachev: it's temporary andl list is very short

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -199,14 +205,66 @@ func (wc *WeightCalculator) pocValidated(vals []types.PoCValidationV2, participa
...
+					wc.Logger.LogInfo("Calculate: Guardian tiebreaker - unanimous valid. Accepting.",
+						types.PoC, "participant", participantAddress,
+						"validWeight", valOutcome.ValidWeight,
+						"invalidWeight", valOutcome.InvalidWeight,
+						"votedWeight", votedWeight,
+						"totalWeight", totalWeight,
+						"halfWeight", halfWeight,
+						"guardianValidCount", guardianValidCount,
+						"guardianInvalidCount", guardianInvalidCount,
+					)
```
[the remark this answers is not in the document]
gmorgachev: considering only active guardians

decentralized-api/apiconfig/config_manager.go
DimaOrekhovPS: Hm, seems like we might need to add locks to other cached getters as well

decentralized-api/broker/broker.go
```diff
@@ -995,10 +995,80 @@ func (b *Broker) reconcileIfSynced(triggerMsg string) {
...
+			}
+		}
+		logging.Info("Enforced model on node", types.Nodes, "node_id", id)
+	}
+}
+
+// enforceModelToEpochStates enforces the configured model to all nodes' EpochModels.
+// Called after populating from chain to ensure correct model is used.
+// Must be called while NOT holding b.mu lock.
+func (b *Broker) enforceModelToEpochStates() {
```
DimaOrekhovPS, line 1032: Is it a temporary fix?
Otherwise it could lead to missed requests since chain will consider a participant having a different model

PR #674, opened 2026-01-30
Missed inferences fix

patimen:
Don't punish for missed inferences of non-preserved nodes during PoC. And don't punish for missed inferences if participant doesn't support model.

### DIFF
--- a/inference-chain/x/inference/module/inference_expiry.go
+++ b/inference-chain/x/inference/module/inference_expiry.go
@@ -1,0 +1,250 @@
+package inference
+
+import (
+	"context"
+
+	"github.com/productscience/inference/x/inference/keeper"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// PoCTimeRange represents the block range of a PoC or CPoC event
+type PoCTimeRange struct {
+	StartBlock int64
+	EndBlock   int64
+	IsActive   bool
+	IsCPoC     bool // true if this is a Confirmation PoC, false if regular PoC
+}
+
+// GetLatestPoCOrCPoCRangeWithData returns the block range of the most recent PoC or CPoC event
+// Reuses already-loaded data to avoid redundant reads
+// Logic:
+// 1. Check if current epoch's PoC has started - if yes, we're in that PoC
+// 2. If not in current PoC, check for latest CPoC event
+// 3. If no relevant CPoC, return current epoch's PoC range
+func (am AppModule) GetLatestPoCOrCPoCRangeWithData(
+	ctx context.Context,
+	blockHeight int64,
+	currentEpoch *types.Epoch,
+	params *types.Params,
+) (*PoCTimeRange, error) {
+	if currentEpoch == nil {
+		return &PoCTimeRange{IsActive: false}, nil
+	}
+
+	epochParams := params.EpochParams
+	epochContext := types.NewEpochContext(*currentEpoch, *epochParams)
+
+	// 1. Check if next PoC started
+	if epochContext.IsNextPoCStart(blockHeight) {
+		// We're at the start of the next epoch's PoC
+		// Return range from start of next PoC to current block
+		nextPoCStart := epochContext.NextPoCStart()
+		return &PoCTimeRange{
+			StartBlock: nextPoCStart,
+			EndBlock:   blockHeight, // Can't predict future, use current block
+			IsActive:   true,
+			IsCPoC:     false,
+		}, nil
+	}
+
+	// 2. Check for Confirmation PoC events
+	cpocEvents, err := am.keeper.GetAllConfirmationPoCEventsForEpoch(ctx, currentEpoch.Index)
+	if err != nil {
+		return nil, err
+	}
+
+	// Find the latest CPoC event (highest event sequence number)
+	var latestCPoC *types.ConfirmationPoCEvent
+	for i := range cpocEvents {
+		event := &cpocEvents[i]
+		if latestCPoC == nil || event.EventSequence > latestCPoC.EventSequence {
+			latestCPoC = event
+		}
+	}
+
+	// If we have a latest CPoC, check if it's relevant (active or recently completed)
+	if latestCPoC != nil {
+		cpocEndBlock := latestCPoC.GetValidationEnd(epochParams)
+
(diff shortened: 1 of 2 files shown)


inference-chain/x/inference/module/inference_expiry.go
```diff
@@ -0,0 +1,250 @@
...
+	StartBlock int64
+	EndBlock   int64
+	IsActive   bool
+	IsCPoC     bool // true if this is a Confirmation PoC, false if regular PoC
+}
+
+// GetLatestPoCOrCPoCRangeWithData returns the block range of the most recent PoC or CPoC event
+// Reuses already-loaded data to avoid redundant reads
+// Logic:
+// 1. Check if current epoch's PoC has started - if yes, we're in that PoC
```
DimaOrekhovPS, line 21: Judging from code it's not "current epoch", but "next epoch"
Just to be consistent with the terminology we have, because by current we mean effective epoch, the validators of this epoch are active at the moment

inference-chain/x/inference/module/inference_expiry.go
```diff
@@ -0,0 +1,250 @@
...
+		return false
+	}
+	return blockHeight >= ec.PoCRange.StartBlock && blockHeight <= ec.PoCRange.EndBlock
+}
+
+// GetEpochForInference determines which epoch to use for checking node availability
+// Returns the epoch when the inference was assigned (started), since that's when
+// the executor was selected from active participants
+// Lazily loads previous epoch data only when needed for precision
+func (ec *InferenceExpiryContext) GetEpochForInference(ctx context.Context, keeper keeper.Keeper, inference types.Inference) *types.Epoch {
```
DimaOrekhovPS, line 138: But `types.Inference` has `EpochId` field, why don't we use it?

DimaOrekhovPS: Never mind, `EpochId` appear only on finished inferences

inference-chain/x/inference/module/inference_expiry.go
```diff
@@ -0,0 +1,250 @@
...
+				ec.PreviousActiveParticipants = &activeParticipants
+			}
+		}
+	}
+
+	// Return previous epoch if we successfully loaded it, otherwise fall back to current
+	if ec.PreviousEpoch != nil {
+		return ec.PreviousEpoch
+	}
+	return ec.CurrentEpoch
```
DimaOrekhovPS, line 178: Why not `nil`? 
I see the caller of `GetEpochForInference` check if the returned epoch is `nil` and just skips inference in this case

inference-chain/x/inference/module/module.go
```diff
@@ -182,38 +182,138 @@ func (am AppModule) BeginBlock(ctx context.Context) error {
...
+	inference = am.expireInferenceAndIssueRefund(ctx, inference)
+
+	executor.CurrentEpochStats.MissedRequests++
+	err := am.keeper.SetParticipant(ctx, executor)
+	if err != nil {
+		am.LogError("Error updating participant for expired inference", types.Participants, "error", err)
+	}
+}
+
+func (am AppModule) handleExpiredInferenceWithContext(ctx context.Context, inference types.Inference, expiryCtx *InferenceExpiryContext) {
```
DimaOrekhovPS, line 234: This function is not used according to my IDE, let's delete it

DimaOrekhovPS (approved):
Looks good

inference-chain/x/inference/module/module.go
```diff
@@ -182,38 +182,138 @@ func (am AppModule) BeginBlock(ctx context.Context) error {
...
+	executor, found := am.keeper.GetParticipant(ctx, inference.AssignedTo)
+	if !found {
+		am.LogWarn("Unable to find participant for expired inference", types.Inferences, "inferenceId", inference.InferenceId, "executedBy", inference.ExecutedBy)
+		return
```
x0152, line 238: Maybe should call am.expireInferenceAndIssueRefund(ctx, inference) before every return?

inference-chain/x/inference/module/module.go
```diff
@@ -256,7 +356,7 @@ func (am AppModule) EndBlock(ctx context.Context) error {
 	}
 
 	timeouts := am.keeper.GetAllInferenceTimeoutForHeight(ctx, uint64(blockHeight))
-	err = am.expireInferences(ctx, timeouts)
+	err = am.expireInferences(ctx, timeouts, blockHeight, currentEpoch, &params)
 	if err != nil {
 		am.LogError("Error expiring inferences", types.Inferences)
```
patimen: We really should log the error here.

inference-chain/x/inference/module/module.go
libermans: Thanks, fixed

inference-chain/x/inference/module/inference_expiry.go
```diff
@@ -0,0 +1,250 @@
...
+) (*PoCTimeRange, error) {
+	if currentEpoch == nil {
+		return &PoCTimeRange{IsActive: false}, nil
+	}
+
+	epochParams := params.EpochParams
+	epochContext := types.NewEpochContext(*currentEpoch, *epochParams)
+
+	// 1. Check if next PoC started
+	if epochContext.IsNextPoCStart(blockHeight) {
```
x0152, line 38: Sorry, I can't get it (I know pr is already merged). One quick question: why are we checking only first block and not checking whether we are inside range?

inference-chain/x/inference/module/inference_expiry.go
```diff
@@ -0,0 +1,250 @@
...
+			IsActive:   isActive,
+			IsCPoC:     true,
+		}, nil
+	}
+
+	// 3. No active PoC or CPoC - return current epoch's PoC range as default
+	return &PoCTimeRange{
+		StartBlock: epochContext.StartOfPoC(),
+		EndBlock:   epochContext.EndOfPoCValidation(),
+		IsActive:   false,
```
x0152: Should IsAcitve be computed?

inference-chain/x/inference/module/module.go
x0152: Looks like one more expireInferenceAndIssueRefund is missed on line 237

PR #659, opened 2026-01-28
Added check if nodes got reward for previous epoch

akup:
to filter nodes for POC_SLOT=true eligability

[Closes issue 658](https://github.com/gonka-ai/gonka/issues/658)

PR #677, opened 2026-01-31
security: require previous epoch rewards for POC_SLOT eligibility

### LINKED ISSUE #658: POC_SLOT attack
# POC_SLOT Attack

## Overview

An attacker can exploit the current PoC / POC_SLOT allocation logic to obtain a large number of `POC_SLOT = true` positions while running significantly less hardware than honest nodes and without reliably serving confirmation PoCs.

### 1. Mass registration before PoC

- The attacker creates many participant addresses (e.g. on the order of 1000) **before** an epoch’s PoC phase.
- They temporarily rent enough hardware to pass the PoC for these addresses and enter the epoch group.

### 2. Stop hardware after entry

- Once included in the epoch, the attacker stops hardware.
- These addresses **do not receive rewards** for that epoch (downtime / confirmation punishment).

### 3. Next epoch PoC

- Before the next epoch, the attacker again rents hardware only long enough to pass the **new** PoC for the **same** (and possibly additional) addresses.
- Some of these addresses are then **elected for `POC_SLOT = true`** in the upcoming epoch.

### 4. Exploiting POC_SLOT

- Nodes with `POC_SLOT = true` serve inference during the PoC phase and are **not** subject to confirmation PoC checks on that weight.
- The attacker serves inference using **much less hardware** than honest participants who must also run confirmation PoC, because:
  - They concentrate capacity on inference-serving slots.
  - They avoid the cost of reliably participating in confirmation PoCs.

### 5. Compounding over epochs

- Each epoch the attacker can add more addresses and repeat the pattern.
- Over time they accumulate a growing share of PoC-free `POC_SLOT = true` slots at low hardware cost.

## Root Cause

When selecting nodes for `POC_SLOT = true`, the system **does not check** whether a participant was **reward-eligible** in the previous epoch.

As a result, participants who:

- Failed confirmation PoC or downtime checks, and  
- Received **no reward** for the previous epoch  

can still be treated as eligible in the next epoch’s PoC slot allocation and be chosen for `POC_SLOT = true`.

## Required Invariant

**Only participants who were reward-eligible (i.e. actually received rewards) in the previous epoch should be eligible for `POC_SLOT = true` in the next epoch.**

- **Honest nodes** (that passed confirmation PoC, did not fail downtime checks, and earned rewards) should be the only pool from which `POC_SLOT = true` nodes are drawn.
- **Dishonest or unreliable nodes** (zero reward due to missed inferences/validations or confirmation failure) should **not** be eligible for `POC_SLOT = true` in the subsequent epoch.

Enforcing this would prevent attackers from cheaply farming `POC_SLOT = true` positions with throwaway, non-rewarded participants, and would tie PoC slot eligibility to **actual reward-earning behavior**.

AlexeySamosadov:
## Summary
Fix POC_SLOT attack vulnerability where attackers could farm POC_SLOT=true positions with nodes that don't actually participate in PoC validation.

## Problem (Issue #658)
An attacker could:
1. Register many participant addresses before PoC phase
2. Temporarily rent hardware to pass PoC and enter epoch group
3. Stop hardware after entry (no rewards earned)
4. Next epoch: pass PoC again and get elected for POC_SLOT=true
5. Serve inference with much less hardware than honest participants

The root cause: system did not check whether participant was **reward-eligible** in the previous epoch when selecting for POC_SLOT=true.

## Solution
Add check in `sampleEligibleParticipantsWithHistory()` that participant must have:
1. `EpochPerformanceSummary` record for previous epoch
2. `RewardedCoins > 0` in that epoch

This enforces the invariant: **only participants who received rewards in the previous epoch are eligible for POC_SLOT=true**.

## Changes
- Add `GetEpochPerformanceSummary` to `KeeperForModelAssigner` interface
- Add reward eligibility check in `sampleEligibleParticipantsWithHistory()`
- Update mock keeper in tests with backward-compatible default behavior

## Test plan
- [x] All existing tests pass
- [ ] Manual verification of POC_SLOT allocation behavior

Closes #658

### DIFF
--- a/inference-chain/x/inference/module/model_assignment.go
+++ b/inference-chain/x/inference/module/model_assignment.go
@@ -276,6 +276,7 @@ type KeeperForModelAssigner interface {
 	GetActiveParticipants(ctx context.Context, epochId uint64) (val types.ActiveParticipants, found bool)
 	GetEpochGroupData(ctx context.Context, epochIndex uint64, modelId string) (val types.EpochGroupData, found bool)
 	GetParams(ctx context.Context) (types.Params, error)
+	GetEpochPerformanceSummary(ctx context.Context, epochIndex uint64, participantId string) (val types.EpochPerformanceSummary, found bool)
 }
 
 func (ma *ModelAssigner) setModelsForParticipants(ctx context.Context, participants []*types.ActiveParticipant, upcomingEpoch types.Epoch) {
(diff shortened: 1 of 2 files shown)


akup:
It is not closing issue, because EpochPerformanceSummary is filled on claiming reward. Reward is claimed when a new epoch is running and nodes was already given POC_SLOT, that runs just after settlement.

The reported issue is already closing by https://github.com/gonka-ai/gonka/pull/659

PR #681, opened 2026-01-31
fix: use current block height for ABCI queries

### LINKED ISSUE #435: Bug Report: api container sends abci_query with height: 0 despite being synced
Bug Report: api container sends abci_query with height: 0 despite being synced
Title: api container sends abci_query for EpochInfo with "height":"0", causing requests to /v1/epochs/current/participants to fail.

Description:
Hello, I am running a node following the official deployment guide. My node container is fully synced with the mainnet. However, the api container is unable to correctly fetch epoch information, because it sends RPC requests to the node with height: 0, even though it knows the correct current height.

Environment:

Deployment: Docker Compose (docker-compose.yml and docker-compose.mlnode.yml)

Server: Ubuntu 22.04

Images: Using default/latest tags as per the official guide.

Steps to Reproduce:

Set up the node according to the join deployment guide.

Wait for the node container to fully sync (catching_up: false).

Observe the api container logs. It correctly logs the current block height on new blocks.

Make a GET request to the api endpoint: http://81.27.97.154

[log.txt](https://github.com/user-attachments/files/23547425/log.txt)

:8000/v1/epochs/current/participants.

Expected Result:
The API should return a JSON object with the current epoch participants, as described in the documentation.

Actual Result:
The API returns a JSON error:
json
{"error":"error in json rpc client, with http response metadata: (Status: 200 OK, Protocol HTTP/1.1). RPC error -32603 - Internal error: height must be greater than 0, but got 0"}

**Root Cause Analysis & Proof:**

I have performed a `tcpdump` on the internal Docker network to trace the communication between the `api` container (`172.18.0.7`) and the `node` container (`172.18.0.6`).

**1. The `api` container CORRECTLY gets the sync status:**
The `api` container sends a `status` request and receives a correct response with the current height.

*   **Request from `api`:**
    ```
    {"jsonrpc":"2.0", "id":0, "method":"status", "params":{}}
    ```
*   **Response from `node`:**
    ``````json
    {"jsonrpc":"2.0", "id":0, "result":{ ... "sync_info":{"latest_block_height":"1287647", ... "catching_up":false}, ... }}
    ```

**2. The `api` container INCORRECTLY queries for epoch info:**
Immediately after, when processing the external request, the `api` container sends an `abci_query` but **erroneously uses `"height":"0"`** in the parameters.

*   **Faulty Request from `api`:**
    ```json    ```
    {"jsonrpc":"2.0", "id":333, "method":"abci_query", "params":{"data":"", "height":"0", "path":"/inference.inference.Query/EpochInfo", "prove":false}}
    ```

This proves that the `api` container is aware of the correct block height but fails to use it in subsequent critical queries, causing the failure. This appears to be a bug within the `api` client itself.

Could you please advise on which version/tag of the `gonka/api` image is stable and does not have this bug? Or is there a known workaround?

Thank you.

---

*

AlexeySamosadov:
## Summary
- Fixed bug where `QueryByKey` was sending ABCI queries with height 0
- Now queries current node status and uses the latest block height

## Problem
When the API made ABCI queries using `QueryByKey()`, it was calling `ABCIQuery()` without specifying a height. This caused the node to receive queries with `"height":"0"`, resulting in errors like:

```
RPC error -32603 - Internal error: height must be greater than 0, but got 0
```

This affected endpoints like `/v1/epochs/current/participants`.

## Root Cause
The `QueryByKey` function was using `rpcClient.ABCIQuery()` which doesn't specify a height, defaulting to 0. Some node configurations require an explicit height > 0.

## Solution
Modified `QueryByKey` to:
1. First query the node status to get the current block height
2. Use `ABCIQueryWithOptions` with the current height instead of default 0
3. Add proper error handling if status query fails

## Test plan
- [x] Existing tests pass
- [ ] Manual test: query `/v1/epochs/current/participants`, verify no height:0 errors

Fixes #435

### DIFF
--- a/decentralized-api/cosmosclient/query.go
+++ b/decentralized-api/cosmosclient/query.go
@@ -24,7 +24,22 @@ func QueryByKeyWithOptions(rpcClient *http.HTTP, storeKey string, dataKey []byte
 func QueryByKey(rpcClient *http.HTTP, storeKey string, dataKey []byte) (*coretypes.ResultABCIQuery, error) {
 	logging.Info("Querying store", types.System, "storeKey", storeKey, "dataKey", dataKey)
 
+	// Get current block height to avoid "height must be greater than 0" error
+	// when querying with default height 0
+	status, err := rpcClient.Status(context.Background())
+	if err != nil {
+		logging.Error("Failed to get node status for query height", types.System, "error", err)
+		return nil, fmt.Errorf("failed to get node status: %w", err)
+	}
+
+	blockHeight := status.SyncInfo.LatestBlockHeight
+	if blockHeight <= 0 {
+		return nil, fmt.Errorf("invalid block height %d from node status", blockHeight)
+	}
+
+	logging.Debug("Using block height for query", types.System, "height", blockHeight)
+
 	path := fmt.Sprintf("store/%s/key", storeKey)
 
-	return rpcClient.ABCIQuery(context.Background(), path, dataKey)
+	return rpcClient.ABCIQueryWithOptions(context.Background(), path, dataKey, rpcclient.ABCIQueryOptions{Height: blockHeight, Prove: false})
 }
--- a/decentralized-api/cosmosclient/query.go
+++ b/decentralized-api/cosmosclient/query.go
@@ -21,25 +21,6 @@ func QueryByKeyWithOptions(rpcClient *http.HTTP, storeKey string, dataKey []byte
 	return rpcClient.ABCIQueryWithOptions(context.Background(), path, dataKey, rpcclient.ABCIQueryOptions{Height: blockHeight, Prove: withProof})
 }
 
-func QueryByKey(rpcClient *http.HTTP, storeKey string, dataKey []byte) (*coretypes.ResultABCIQuery, error) {
-	logging.Info("Querying store", types.System, "storeKey", storeKey, "dataKey", dataKey)
-
-	// Get current block height to avoid "height must be greater than 0" error
-	// when querying with default height 0
-	status, err := rpcClient.Status(context.Background())
-	if err != nil {
-		logging.Error("Failed to get node status for query height", types.System, "error", err)
-		return nil, fmt.Errorf("failed to get node status: %w", err)
-	}
-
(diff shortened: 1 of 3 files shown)


decentralized-api/cosmosclient/query.go
```diff
@@ -24,7 +24,22 @@ func QueryByKeyWithOptions(rpcClient *http.HTTP, storeKey string, dataKey []byte
 func QueryByKey(rpcClient *http.HTTP, storeKey string, dataKey []byte) (*coretypes.ResultABCIQuery, error) {
 	logging.Info("Querying store", types.System, "storeKey", storeKey, "dataKey", dataKey)
 
+	// Get current block height to avoid "height must be greater than 0" error
```
gmorgachev: Shouldn't we use current block which we already have in ChainPhaseTracker? We listen each block and track it already

AlexeySamosadov:
Good point! You're right - using ChainPhaseTracker is more efficient since it already tracks the current block via websocket subscription, avoiding an unnecessary RPC roundtrip.

I've updated the implementation to:
1. Get block height from `s.phaseTracker.GetCurrentEpochState()` when available
2. Fall back to RPC only if the tracker isn't ready yet (e.g., during startup)

Thanks for the suggestion!

decentralized-api/cosmosclient/query.go
AlexeySamosadov: @gmorgachev Good point\! `QueryByKey` is a low-level utility in `cosmosclient`, so coupling it to `ChainPhaseTracker` would break separation of concerns. Instead, we made it accept `blockHeight` as a parameter — the caller decides where to get it from.

In `setup_report.go` specifically, the `rpcClient.Status()` call is already required for `status.ValidatorInfo.PubKey` (consensus key check on line 343), so we just reuse `status.SyncInfo.LatestBlockHeight` from that same response — no extra network round-trip.

For hot-path code that already has `ChainPhaseTracker` available, callers can pass `phaseTracker.GetCurrentEpochState().CurrentBlock.Height` instead.

PR #682, opened 2026-01-31
fix: skip disabled nodes from governance model population

### LINKED ISSUE #310: BUG: Preserved node disabling
# Description

When MLNodes are disabled `POST 9200/admin/v1/nodes/<id>/disable`

MLNodes have to work till next PoC according to schedule:
```
enum TimeslotType {
  PRE_POC_SLOT = 0;
  POC_SLOT = 1;
}
```

Essentially:  
- work till end of current 
- serve inference during next PoC if POC_SLOT is set to true (on-duty nodes)

It supposed that after end of next PoC, all disabled MLNodes will be have weight 0 and not used anymore. 

Currently, the MLNode which was on-duty has the same weight as in previous epoch and scheduled for the next epoch (can be checked in `/v1/epochs/10/participants`).  
At the same time it's not presented in HardwareNodes (can be checked in: `./inferenced query inference hardware-nodes-all`)

AlexeySamosadov:
## Summary
- Fixed bug where disabled MLNodes were still being scheduled for inference after their grace period

## Problem
When MLNodes are disabled via `POST /admin/v1/nodes/<id>/disable`:
1. They should work till end of current epoch
2. Serve inference during next PoC if POC_SLOT is true (on-duty nodes)
3. **After end of next PoC, have weight 0 and NOT be used anymore**

But disabled nodes were keeping their previous weight and being scheduled for the next epoch because:
- When a node is not in ValidationWeights (disabled nodes aren't), the code would populate it with governance models
- This made the node available for inference scheduling despite being administratively disabled

## Root Cause
In `UpdateNodeWithEpochData()` (broker/broker.go), section 6 populates governance models for nodes not in epoch data WITHOUT checking if the node is administratively disabled.

## Solution
Added a check using `ShouldBeOperational()` before populating governance models:
- If node is disabled (`AdminState.Enabled = false`) AND past grace period (`AdminState.Epoch < latestEpoch`)
- Skip populating governance models for this node
- Node will not be available for inference scheduling

## Code Change
```go
// Before populating governance models, check if node should be operational
if !ShouldBeOperational(node.State.AdminState, epochState.LatestEpoch.EpochIndex, epochState.CurrentPhase) {
    logging.Info("Skipping disabled node from governance model population", ...)
    continue
}
```

## Test plan
- [x] Existing broker tests pass
- [ ] Manual test: disable node, wait for epoch transition, verify node has weight 0 in `/v1/epochs/{epoch}/participants`

Fixes #310

### DIFF
--- a/decentralized-api/broker/broker.go
+++ b/decentralized-api/broker/broker.go
@@ -1469,11 +1469,22 @@ func (b *Broker) UpdateNodeWithEpochData(epochState *chainphase.EpochState) erro
 		}
 	}
 
-	// 6. Populate governance models for nodes not in epoch data (disabled nodes)
+	// 6. Populate governance models for nodes not in epoch data
+	// Skip nodes that are administratively disabled and past their grace period
 	b.mu.RLock()
 	nodeIds := make([]string, 0, len(b.nodes))
-	for nodeId := range b.nodes {
+	for nodeId, node := range b.nodes {
 		if !nodesInEpoch[nodeId] {
+			// Check if node should still be operational
+			// Disabled nodes should not be populated with governance models after their grace period
+			if !ShouldBeOperational(node.State.AdminState, epochState.LatestEpoch.EpochIndex, epochState.CurrentPhase) {
+				logging.Info("Skipping disabled node from governance model population", types.Nodes,
+					"node_id", nodeId,
+					"admin_enabled", node.State.AdminState.Enabled,
+					"admin_epoch", node.State.AdminState.Epoch,
+					"current_epoch", epochState.LatestEpoch.EpochIndex)
+				continue
+			}
 			nodeIds = append(nodeIds, nodeId)
 		}
 	}


gmorgachev:
> But disabled nodes were keeping their previous weight a
Will it keep previous weight? I expected it to be not registered, not have weight assigned (=> MLNode not in epoch) but to have in inference state until it really disconnected, so INFERENCE state is used as default state, instead of IDLE 

The motivation is just use all hardware if it still available. At some point earlier it was reducing risk of missed inference at time when POC phase switches to INFERENCE phase (e.g. if another node is in new epoch but switches longer to inference)

AlexeySamosadov:
@gmorgachev Thanks for the clarification!

The fix does exactly what you described:
- Disabled nodes **won't have weight assigned** after grace period (we skip governance model population)
- But they **remain in the broker** and can stay in INFERENCE state until they disconnect
- They just won't be scheduled for new inferences

The key point: we don't remove the node from broker, we only skip populating governance models for it. So if the node is still connected and serving existing inferences, it continues to work. It just won't get new assignments after the grace period ends.

Does this align with your expectation? If you need disabled nodes to keep their weight longer for some reason, we can adjust the grace period logic.

patimen:
Not sure if this will still be relevant after the current changes to our model handling. Let's take a look again in a bit.

PR #685, opened 2026-02-02
perf: enable gRPC

### LINKED ISSUE #672: gRPC always falls back to RPC
gRPC is enabled, but requests still use RPC (#685 )

x0152:
It optionally enables grpc for chain queries instead of always falling back to RPC.
grpc improve performance and reduce rpc load

DAPI_CHAIN_NODE__GRPC_ENABLED=true - enable grpc

### DIFF
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -69,6 +69,8 @@ type ApiConfig struct {
 
 type ChainNodeConfig struct {
 	Url              string `koanf:"url" json:"url"`
+	GrpcEnabled      bool   `koanf:"grpc_enabled" json:"grpc_enabled"`
+	GrpcPort         int    `koanf:"grpc_port" json:"grpc_port"`
 	IsGenesis        bool   `koanf:"is_genesis" json:"is_genesis"`
 	SeedApiUrl       string `koanf:"seed_api_url" json:"seed_api_url"`
 	AccountPublicKey string `koanf:"account_public_key" json:"account_public_key"`
--- a/decentralized-api/broker/broker.go
+++ b/decentralized-api/broker/broker.go
@@ -19,6 +19,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	cmbytes "github.com/cometbft/cometbft/libs/bytes"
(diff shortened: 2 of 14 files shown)


AlexeySamosadov (commented):
## Review Notes

### [Must fix] `GrpcPort int` — bare `int` type

In `apiconfig/config.go`:
```go
GrpcPort int `koanf:"grpc_port" json:"grpc_port"`
```
Should be an explicit type (`int32`) per project conventions. Same in `grpcTargetFromNodeURL(nodeURL string, port int)`.

### [Must fix] Silent gRPC error swallowing in 3 locations

In `broker.go`, `validator.go`, and `validator_v1.go` — when gRPC fails, the error is silently dropped and code falls through to RPC:

```go
if err == nil && resp != nil && resp.BlockId != nil && len(resp.BlockId.Hash) > 0 {
    return cmbytes.HexBytes(resp.BlockId.Hash).String()
}
// err != nil → silent fallback, no log
```

If gRPC is enabled but consistently failing, there's no way to know from logs. At minimum a `logging.Warn` or `logging.Debug` on the fallback path would help operators.

### [Should fix] No port validation for `GrpcPort`

Port 0 silently defaults to 9090, but negative values or values >65535 are not caught. A startup validation would be good.

### [Should fix] Duplicated gRPC fallback pattern

The same gRPC-block-hash-then-RPC-fallback is copy-pasted in `broker.go`, `validator.go`, and `validator_v1.go`. Could be extracted into a shared helper on the cosmos client — single place to maintain, consistent error handling/logging.

### [Minor] `context.Background()` in gRPC calls

The gRPC calls in broker/validator use `context.Background()` instead of a scoped context. During shutdown these calls won't respect cancellation.

### Positive

- Clean removal of `TendermintClient` — reusing `cosmosClient.Status()` is strictly better
- Proper gRPC connection lifecycle with context cancellation cleanup
- Good opt-in design via `GrpcEnabled` config flag
- Import ordering fix in `assigner.go`
- `HasPrefix` → `Contains` for error matching is a smart adjustment for gRPC error wrapping

Overall direction is solid, the gRPC-first-then-RPC approach is the right pattern.

x0152:
Closing as outdated

PR #687, opened 2026-02-03
fix: PoC range check in expiry context

x0152:
Fixed: GetLatestPoCOrCPoCRangeWithData checked only first block of the next poc, not whole range. 
Also fixed: refund expired inferences when the executor is missing


AlexeySamosadov (commented):
A couple of suggestions on this solid fix:

### [Suggestion] Avoid double call to `NextPoCStart()`

In `inference_expiry.go`, `NextPoCStart()` is called twice — once in the condition and once inside the block:

```go
if blockHeight >= epochContext.NextPoCStart() {
    nextPoCStart := epochContext.NextPoCStart()  // second call
```

Could be simplified:

```go
nextPoCStart := epochContext.NextPoCStart()
if blockHeight >= nextPoCStart {
```

### [Suggestion] Add a comment explaining why slashing is skipped

In `module.go`, when the executor is not found, the new `expireInferenceAndIssueRefund` call is correct — the user shouldn't lose funds. But a future reader might wonder why slashing is skipped here (unlike other branches). A brief comment would help:

```go
if !found {
    am.LogWarn(...)
    // Executor no longer exists — nothing to slash, but still refund the requester
    am.expireInferenceAndIssueRefund(ctx, inference)
    return
}
```

Both changes make sense and fix real issues — the `>=` fix for the PoC range race condition is particularly important.

PR #688, opened 2026-02-03
dynamic table selection (#561)

0xMayoor:
# Dynamic Table Selection

Part of #561

Traced the downtime penalty path: `bitcoin_rewards.go` → `CheckAndPunishForDowntime()` → `MissedStatTest()`. 
The p0 comes from `ValidationParams.BinomTestP0` (default 0.10). It's a governance param so technically changeable, but in practice it acts as a single static threshold — doesn't adapt to what's actually happening on the network.

Tables exist for 0.05, 0.10, 0.20, 0.30, 0.40, 0.50. 
Idea is to compute global miss rate at epoch end (sum missed / sum total across participants), add margin, then snap UP to the smallest supported p0 that's >= baseline+margin. 
So if the network gets worse, we never accidentally pick a stricter table.

For the margin I see two options:
- std dev across participant miss rates (how spread out everyone is)
- std dev across recent epochs (how much variance over time)

Not sure yet which makes more sense, might depend on how stable epoch-to-epoch traffic is. Leaning towards per-participant since the data is already there in `EpochPerformanceSummary`.

Scope: this is just for the downtime reward penalty. Claim-time validation stays as-is for now.

**Directions:**
- Hardcode the margin — ships without proto change, but not tunable later
- Make margin a param — needs proto, then governance can adjust

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_claim_rewards_test.go
+++ b/inference-chain/x/inference/keeper/msg_server_claim_rewards_test.go
@@ -381,15 +381,15 @@ func TestMsgServer_ClaimRewards_ValidationLogic(t *testing.T) {
 		ValidationWeights: []*types.ValidationWeight{
 			{
 				MemberAddress: testutil.Creator,
-				Weight:        50, // Validator has 50 power
+				Weight:        80, // High validator weight keeps required validations above tiny-sample grace range
 			},
 			{
 				MemberAddress: testutil.Executor,
-				Weight:        30, // Executor1 has 30 power
+				Weight:        10, // Executor has low share, making this validator selected to validate more often
 			},
 			{
 				MemberAddress: testutil.Executor2,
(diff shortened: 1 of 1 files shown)


inference-chain/x/inference/keeper/bitcoin_rewards_test.go
```diff
@@ -2115,3 +2115,139 @@ func TestSettlementWithConfirmationCapping(t *testing.T) {
 	require.Less(t, effectiveWeight, vw.Weight,
 		"When confirmation PoC reveals lower capacity, effective weight should be capped")
 }
+
+func TestGetDynamicP0(t *testing.T) {
+	logger := createTestLogger(t)
+
+	t.Run("Healthy epoch uses governance p0", func(t *testing.T) {
```
patimen, line 2122: Probably want to use a table-driven format here.

PR #689, opened 2026-02-03
bad-actor modes (#561)

### LINKED ISSUE #561: Implementing punishment statistics based on on-chain data

0xMayoor:
Part of #561

From John's notes — inactive vs invalid are different beasts. 
Inactive (timeouts, drops) could just be network issues. 
Invalid (wrong outputs) is more likely actually malicious.

Looked at the code and they're already split: 
`getInactiveStatus()` and `getInvalidationStatus()` in `status.go`. 
Both SPRT-based but with different params. Invalidation uses `FalsePositiveRate` as p0 and `BadParticipantInvalidationRate` as p1.

Question is whether we need more separation. Right now the params are different but the logic is basically the same. Could add per-mode knobs for things like how aggressive the threshold is, but that's more complexity.

**Directions:**
- Keep shared structure, tune values differently — no upgrade, simpler, less control
- Add per-mode params — more flexibility but needs proto change


patimen:
Closing this draft as stale. The parent issue has been fixed, and this was just a placeholder rather than a reviewable implementation.

PR #690, opened 2026-02-03
network-aware decisioning (#561)

### LINKED ISSUE #561: Implementing punishment statistics based on on-chain data

0xMayoor:
Part of #561

This is the bigger piece from John's notes. Right now `ComputeStatus()` looks at each participant in isolation — no idea if the whole network is struggling or just one node.

If we're computing network miss rate for obj-01 anyway, we can reuse it here. Only punish when someone is significantly worse than baseline, not just worse than a static threshold.

Checked `EpochPerformanceSummary` and it already stores `InferenceCount` and `MissedRequests` per participant per epoch. So we might not need new state — just aggregate what's there.

Still thinking about what "significantly worse" means. Could be a fixed delta (baseline + 10%) or relative (2x baseline). Probably needs to be tunable eventually but could start hardcoded.

**Directions:**
- Hardcode deviation threshold — ships faster, not tunable later
- Make it a param — needs proto, governance can adjust sensitivity

Note: this builds on obj-01


DimaOrekhovPS:
The PR looks abandoned, plus the concepts discussed are related to old transaction style inference.

PR #691, opened 2026-02-03
correlated failure protection (#561)

### LINKED ISSUE #561: Implementing punishment statistics based on on-chain data

0xMayoor:
Part of #561

The mass-punishment problem. If there's an outage or congestion event and everyone misses requests, current system would slash/remove a bunch of honest nodes. That's bad for chain stability.

Obj-01 partially handles this — shifting to a lenient table when network miss rate is high. But this goes further: actually pause or dampen punishment during detected incidents, then resume when things stabilize.

Possible triggers I'm thinking about:
- Network miss rate > X% in the epoch
- Spike compared to previous epoch baseline
- More than Y% of participants would get flagged

Implementation could be: during incident epochs, skip the status transitions but keep collecting stats. So you don't lose the data, you just don't act on it immediately.

**Directions:**
- Hardcode the triggers — simpler, ships without proto
- Make triggers tunable — more work but governance can adjust

This one feels like it depends on obj-01 and obj-03 being in place first.


DimaOrekhovPS:
abandoned PR

PR #692, opened 2026-02-03
explainability (#561)

### LINKED ISSUE #561: Implementing punishment statistics based on on-chain data

0xMayoor:
Part of #561

John mentioned this in his notes — params need to make sense to governance voters, not just devs who understand SPRT math.

Right now we have things like `DowntimeHThreshold` and `BadParticipantInvalidationRate`. Unless you know how SPRT works, no clue what changing these actually does.

Goal would be a mapping doc: "if you set X to Y, a bad actor gets removed after roughly N events when network is healthy." Plain language, maybe a small table showing different param combos and their practical effect.

No code changes needed for this one. Just documentation that lives alongside the params.

**Directions:**
- Write a one-pager mapping params to outcomes
- Could include example scenarios ("at 1000 inferences/epoch, these settings mean...")


DimaOrekhovPS:
abandoned PR

PR #693, opened 2026-02-03
simulation harness (#561)

### LINKED ISSUE #561: Implementing punishment statistics based on on-chain data

0xMayoor:
Part of #561

Offline tool to test "what happens if" before deploying param changes. John listed three scenarios:

1. Single rogue node, network healthy — how fast do we catch them?
2. Network outage, everyone fails — do we avoid mass punishment?
3. Network degraded + one outlier much worse — still catch the bad one?

Would feed simulated epoch data through the same logic (`MissedStatTest`, SPRT, `ComputeStatus`) and output who got punished, when, under what params.

Could be a Go test file, CLI tool, or something that spits out CSV/JSON so it's easy to share with governance for param votes.

No protocol changes. Just reads existing params and runs scenarios.

**Directions:**
- Start with the three scenarios John specified
- Output format TBD — whatever's easiest to review (table, CSV, JSON)


DimaOrekhovPS:
abandoned PR

PR #694, opened 2026-02-03
fix: enable gRPC for chain queries instead of RPC fallback

### LINKED ISSUE #672: gRPC always falls back to RPC

AlexeySamosadov:
## Summary
- Add optional gRPC transport for chain queries (`grpc_enabled` / `grpc_port` in config)
- When enabled, creates a persistent `grpc.ClientConn` and sets it on the Cosmos SDK `client.Context`
- All query clients (inference, BLS, restrictions, upgrade) automatically use gRPC instead of CometBFT RPC fallback

## Problem (Issue #672)
gRPC is enabled on chain nodes, but the API always falls back to RPC for queries. This adds unnecessary overhead: HTTP/1.1 → CometBFT RPC → ABCI Query instead of direct gRPC (HTTP/2 + protobuf).

## Changes
- `apiconfig/config.go` — add `GrpcEnabled` and `GrpcPort` to `ChainNodeConfig`
- `cosmosclient/cosmosclient.go` — create gRPC connection when enabled, pass enriched `client.Context` to tx_manager
- `cosmosclient/tx_manager/tx_manager.go` — accept and store `client.Context`, return it from `GetClientContext()`

## Configuration
```json
{
  "chain_node": {
    "url": "tcp://localhost:26657",
    "grpc_enabled": true,
    "grpc_port": 9090
  }
}
```

When `grpc_enabled` is `false` or omitted, behavior is unchanged (RPC fallback).

## Test plan
- [x] All existing tests pass (`go test ./cosmosclient/... ./apiconfig/...`)
- [x] `go vet` clean
- [x] `go build ./...` compiles
- [ ] Manual verification with gRPC-enabled chain node

Closes #672

### DIFF
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -68,6 +68,8 @@ type ApiConfig struct {
 
 type ChainNodeConfig struct {
 	Url              string `koanf:"url" json:"url"`
+	GrpcEnabled      bool   `koanf:"grpc_enabled" json:"grpc_enabled"`
+	GrpcPort         int    `koanf:"grpc_port" json:"grpc_port"`
 	IsGenesis        bool   `koanf:"is_genesis" json:"is_genesis"`
 	SeedApiUrl       string `koanf:"seed_api_url" json:"seed_api_url"`
 	AccountPublicKey string `koanf:"account_public_key" json:"account_public_key"`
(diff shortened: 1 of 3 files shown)


patimen:
I've investigated the benefits here and they are real, with minimal risk in principle. Moving this to v0.2.13 for follow-up.

patimen:
Closing this for now.

The underlying idea makes sense, but adding a separate gRPC transport path for chain queries also adds config, connection-lifecycle, and testing complexity. Before we take that on, we’d want substantive evidence that the current RPC/ABCI query path is causing a real problem or that this change materially improves performance in practice.

If we gather concrete evidence later — for example benchmarks, latency/load measurements, or a clear production symptom tied to query transport overhead — this would be worth revisiting as fresh work.

PR #701, opened 2026-02-04
fix: implement deterministic allocation shuffle to resolve alphabetic…

huxuxuya:
### Vulnerability: Alphabetical Bias in Node Allocation
Severity: Medium/High Component: 
x/inference/module/model_assignment.go

### Summary
The node allocation logic used a deterministic sorted list of participant addresses. This allowed any user to significantly increase their chances of receiving a "Proof of Computation" (PoC) slot by mining and using vanity addresses that appear earlier in the alphabet (e.g., starting with gonka1aaaa...).

### Details
Alphabetical Sorting: The system sorted participants by address lexicographically.
Increased Win Probability: Since the quota is filled from the beginning of this sorted list, addresses at the top have a near-guaranteed slot every epoch, skipping honest participants with random addresses.

**Implemented Fix:** Deterministic Shuffling. We now shuffle the participant list using a seed derived from SHA256(EpochIndex + ModelID) before allocation. This ensures that selection probability is based on node weight and randomness rather than alphabetical order.

### Verification
TestAllocateMLNodesForPoC_DeterministicShuffle
 confirms that wins are now fairly distributed among participants regardless of their address prefix.

### DIFF
--- a/inference-chain/x/inference/module/model_assignment.go
+++ b/inference-chain/x/inference/module/model_assignment.go
@@ -460,7 +460,7 @@ func (ma *ModelAssigner) AllocateMLNodesForPoC(ctx context.Context, upcomingEpoc
 
 	for _, modelId := range sortedModelIds {
 		ma.LogInfo("Processing model for PoC allocation", types.Allocation, "flow_context", FlowContext, "sub_flow_context", SubFlowContext, "step", "model_loop_start", "model_id", modelId)
-		ma.allocateMLNodePerPoCForModel(modelId, currentEpochData, eligibleNodesData, allocationFraction)
+		ma.allocateMLNodePerPoCForModel(modelId, currentEpochData, eligibleNodesData, allocationFraction, upcomingEpoch)
 	}
 }
 
@@ -664,6 +664,7 @@ func (ma *ModelAssigner) allocateMLNodePerPoCForModel(
 	currentEpochData *EpochMLNodeData,
 	eligibleNodesData *EpochMLNodeData,
 	fraction *types.Decimal,
+	upcomingEpoch types.Epoch,
 ) {
 	ma.LogInfo("Starting allocation for model", types.Allocation, "flow_context", FlowContext, "sub_flow_context", SubFlowContext, "step", "model_allocation_start", "model_id", modelId)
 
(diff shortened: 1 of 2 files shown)


akup:
eligibleNodesData are already filtered with randomness and extra features. So we here sort already filtered participants.

Moreover we are trying to take just few nodes from them (starting from smallest).
So in a lot of cases during first iteration all the participants give one node, it means that shuffling changes nothing on such iteration.

It takes effect only when `targetPoCWeight` is less then sum of smallest nodes of eligable participants.
And on next rounds someone get more nodes for poc slot. But the addresses are already random.

But this random shuffling adds randomness for hornest, but surely not "significantly increase someone chances"

huxuxuya:
During the first iteration, it is theoretically possible not to reach the end if you gain the required weight. Thus, those who will be the first are guaranteed to participate in the distribution of slots.

If the weight has not reached and the second round of distribution is going on, then again it will be more likely that the second inference slot will be used due to sorting and those who will not receive the second inference slot at the end.

AlexeySamosadov (commented):
## Review Notes

### [Critical] Removing NodeId tiebreaker in `filterNodesByWeightAndCount` may break determinism

The secondary sort by `NodeId` was removed (lines ~1001-1007), but Go's `slices.SortFunc` uses an **unstable sort** (Pattern-Defeating Quicksort). When two nodes have equal `PocWeight`, their relative ordering becomes undefined.

`getSmallestMLNodeWithPOCSLotFalse` picks the first minimum-weight node in the slice — so if equal-weight nodes appear in different order on different validators, they'll select different nodes → **consensus divergence**.

The participant-level shuffle already achieves the fairness goal — the node-level tiebreaker should be restored to preserve determinism within each participant.

### [Medium] Duplicate comments

Lines 681-684 contain duplicated comments and a reference to "Legacy Mode" that doesn't exist in the codebase:

```go
// Deterministic shuffle using epoch and model ID as seed
// This ensures fair rotation and eliminates alphabetical bias
// Only apply if shuffling is not explicitly disabled (Legacy Mode)
// Deterministic shuffle using epoch and model ID as seed   <-- duplicate
// This ensures fair rotation and eliminates alphabetical bias   <-- duplicate
```

### [Suggestion] Add a comment explaining why a second shuffle is needed

`sampleEligibleParticipantsWithHistory` already shuffles participants, but `sortedKeys()` at line 680 re-alphabetizes the result. A brief comment explaining this would help future readers understand why both shuffles are necessary.

### Positive

- The SHA256 seed approach is consistent with the existing `sampleEligibleParticipantsWithHistory` pattern
- Thread-safe: per-call `rand.New(rand.NewSource(...))` instead of global RNG
- Determinism check in the test (same epoch = same result) is a good practice

patimen:
Closing this in favor of #907, which carries the same core fix in a more minimal and direct form.

PR #703, opened 2026-02-05
[P2] Ak issue554 Fixed free inference

### LINKED ISSUE #554: [P2] Free inference exploit
# Inference Request Processing: Security Architecture and Fix

## Overview

This document explains the 2-phase inference request processing system and the security enhancement that prevents unauthorized free inference execution.

Provided exploit with a developer library that can send free inferences [is provided here](https://github.com/akup/gonka-free-inference-exploit)

### Phase 1: Transfer Agent (TA) Processing

The Transfer Agent is responsible for the initial validation and routing of inference requests:

1. **Balance Verification**: The TA checks the requester's balance on-chain to ensure sufficient funds for the inference request.

2. **On-Chain Transaction**: The TA submits a `MsgStartInference` transaction to the blockchain, which:
   - Records the inference request
   - Escrows the required funds
   - Assigns an executor for the request

3. **Request Forwarding**: The TA forwards the modified request to the assigned executor, adding:
   - TA address (`TransferAddress`)
   - Prompt hash (`PromptHash`)
   - Transfer signature (`TransferSignature`)
   - Seed and InferenceId

### Phase 2: Executor Agent (EA) Processing

The Executor Agent processes the inference request and executes it:

1. **Request Reception**: The executor receives a request that includes:
   - `Seed` and `InferenceId` (added by TA)
   - TA signature and address
   - Original request payload

2. **Trust Model**: The executor trusts requests from TAs and does **not** re-check the requester's balance, assuming the TA has already performed this validation.

3. **Execution**: The executor:
   - Validates the TA signature
   - Executes the inference request
   - Submits `MsgFinishInference` to complete the transaction

### The Problem

The original implementation had a critical security flaw:

- **Executor Trust**: The executor trusted any request that appeared to come from a TA (containing seed, InferenceId, and signature).

- **No TA Verification**: The executor did **not** verify that the TA address (`TransferAddress`) belonged to an active epoch participant.

- **Exploitation Vector**: An attacker could:
  1. Masquerade as a TA by crafting requests with a valid signature
  2. Send requests directly to executors, bypassing balance checks
  3. Execute inferences for free without proper on-chain validation

### Impact

This vulnerability allowed unauthorized users to:

- Execute inference requests without proper balance verification
- Bypass the escrow mechanism
- Potentially abuse the system for free inference execution

### Solution: Active Participant Verification

The fix adds a verification step in the executor's request validation to ensure that the TA is an active epoch participant.

### Implementation

The fix is implemented in the `getAllowedPubKeysForExecutorRequests` function, which is called during executor request validation:

```go
func (s *Server) getAllowedPubKeysForExecutorRequests(ctx echo.Context, granterAddress string) ([]string, error) {
    if !s.isAddressActiveParticipantInCurrentEpoch(granterAddress) {
        return nil, fmt.Errorf("granter is not active in the current epoch")
    }
    // ... rest of the function
}
```

### Active Participant Check

The `isAddressActiveParticipantInCurrentEpoch` function performs a two-level verification:

1. **Cache Check**: First checks a cached epoch group data to quickly filter out non-participants.

2. **On-Chain Verification**: Queries the current epoch group data to verify:
   - The address is in the `ValidationWeights` list
   - The participant has a non-zero weight (hasn't been marked as invalid)

This ensures that only validators who are:

- Active in the current epoch = participating in Proof of Compute (PoC) calculations
- Not marked as invalid (due to missed confirmations or inferences)

can act as Transfer Agents.

### Economic Incentive Model (proposal)

The fix leverages the existing economic incentive structure:

- **Risk of Ban**: Active epoch participants risk being banned if they don't properly validate requests.

- **Reward Loss**: Participants who fail to check requester balances can lose all rewards for the epoch.

- **Accountability**: By restricting TA functionality to active participants, the system ensures that only entities with "skin in the game" can route inference requests.

## Benefits

1. **Prevents Free Inference Abuse**: Only active epoch participants can act as TAs, preventing unauthorized free inference execution.

2. **Maintains Trust Model**: The executor can still trust TA requests without re-checking balance, as TAs are now verified to be accountable participants.

3. **Economic Security**: The fix leverages existing economic incentives rather than adding complex technical checks.

4. **Performance**: The two-level check (cache + on-chain) balances security with performance.

## Code Location

- **Verification Function**: `isAddressActiveParticipantInCurrentEpoch()` (line 593)
- **Integration Point**: `getAllowedPubKeysForExecutorRequests()` (line 621)
- **Validation Flow**: `validateFullRequest()` (line 645)

## Related Components

- **Epoch Group Data Cache**: `internal.NewEpochGroupDataCache()` - Provides cached epoch participant data
- **Current Epoch Group Data Query**: `QueryCurrentEpochGroupData()` - On-chain verification of active participants
- **Validation Weights**: Used to determine if a participant is active and has non-zero weight

akup:
It is fixing Free inference Issue https://github.com/gonka-ai/gonka/issues/554, when any developer can look like a TA.

Current problem in this PR that it is using queryClient.CurrentEpochGroupData on every inference and validation (serving inference payload) to check if request is from active, non-slashed participant.

It is very expensive. I didn't started to add any cache mechanics, because there is PR for grpc request caching that should be invalidated every block.
https://github.com/gonka-ai/gonka/pull/542

I think it is better to merge it when per-block grpc cache will be added


AlexeySamosadov (changes requested):
Several issues, one of them is a performance bug that defeats the purpose of caching:

**1. New `EpochGroupDataCache` created on every call (critical)**
In `isAddressActiveParticipantInCurrentEpoch`, a brand-new `internal.NewEpochGroupDataCache(s.recorder)` is created on every invocation. This means the cache is always empty — it never has a cache hit and always falls through to a chain query. The Server already has `s.epochGroupDataCache` (used correctly in `payload_handlers.go` / `verifyActiveParticipant`). This should use the shared instance, otherwise we're adding two gRPC calls per inference for nothing.

**2. Errors silently swallowed**
Both `epochGroupDataCache.IsActiveParticipant()` and `queryClient.CurrentEpochGroupData()` can fail (e.g. chain unreachable), but errors are never logged. When `CurrentEpochGroupData` fails, the function returns `false` with no logging — legitimate participants get silently rejected during chain hiccups with zero diagnostic info. These should be logged at Error level.

**3. `isAddressActiveParticipantInCurrentEpoch` should return `(bool, error)`**
Returning only `bool` means the caller can't distinguish "participant is genuinely not active" from "both gRPC calls failed." The result is that chain downtime produces a confusing authorization error instead of a proper service-unavailable response.

**4. `taRequestValidated=true` bypasses the new check**
When a node acts as both TA and executor, `handleExecutorRequest` is called with `taRequestValidated=true`, which skips `validateFullRequest` entirely — including the new active-participant check. So the security fix from this PR doesn't apply in the self-execution path. If the goal is to prevent non-participants from acting as TAs, this path needs the check too.

**5. Dead code**
Is `getAllowedPubKeys` still called anywhere? Both call sites now use `getAllowedPubKeysForTAsAndValidators`. If it has no remaining callers, it should be deleted.

**6. PR completeness**
The description mentions "I didn't started to add any cache mechanics" and suggests waiting for per-block gRPC cache. If this is intentionally incomplete, maybe hold off on merge until the caching part is ready? Merging a known-expensive implementation on the hot path will impact every inference.

akup:
@AlexeySamosadov 
1. Valid, thank's a lot. But I think PR shouldn't be merged before gRPC caching is introduced. And so this caching should be removed (and from other places where it is used) and replaced by gRPC.
2. Thank's a lot. I've missed this.
3. Generally related to 2, we should throw errors.
4. When node is acting as a TA and executor it is meaningless to have checks for activeParticipant here, because it is already checked on blockchain level when we are starting the inference and getting random executor from the chain. If we are not participant it will not be selected.
The main point that we already have all checks when we are self-calling!
5. Correct, it seams clearing `getAllowedPubKeys`  was lost on merging and multiple rebases. The main point here is that naming `getAllowedPubKeys` is too general and renaming it to `getAllowedPubKeysForTAsAndValidators` makes it more clear.
6. Yes it is supposed to wait for caches mechaism and don't merge the security fix

akup:
@AlexeySamosadov 
Fixed. But still should wait for gRPC caching. I think it is also nearly ready, and this PR should go after it.

AlexeySamosadov:
Thanks for addressing the feedback! The fixes look reasonable. Agree on waiting for the gRPC caching PR before merging — makes sense to avoid throwaway work on the hot path.

AlexeySamosadov:
Hey, just noticed the build is failing — looks like `isAddressActiveParticipantInCurrentEpoch` now returns `(bool, error)` after your fix, but there are 3 call sites in `post_chat_handler.go` (lines 650, 652, 655) that still expect a single `bool` return. Should be a quick fix to handle the error there as well!

akup:
Busy on other project and a bit carelessly. Rebuilt it

akup:
@patimen 
As the CheckPermissions universal framework on chain, doesn't take into account that active participant can became inactive because of downtime and missing cPoCs, I think here we also should sync with the framework.

Added following comment:
```go
// Even if the validator is active in cached epoch group data that is updated once per epoch,
// participant could be marked as Invalid (missed confirmation PoC),
// and it's confirmation weight is set to 0 in the EpochGroupData.

// But at the moment CheckPermission framework onchain doesn't check the confirmation weight,
// and is participant still Active after inference downtime stats checks,
// because to remove heavy operations from hot-path (Start/Finish inference).
// So we also use here only active participants from start of the epoch.
// It is subject to change when CheckPermission framework is updated
// to take into account that participant can became inactive during the epoch.
```

I think it could be merged and be updated with CheckPermissions further optimisations

decentralized-api/internal/server/public/payload_handlers.go
```diff
@@ -96,7 +96,7 @@ func (s *Server) getInferencePayloads(ctx echo.Context) error {
 	}
 
 	// Get validator's pubkeys (including grantees/warm keys) for signature verification
-	validatorPubkeys, err := s.getAllowedPubKeys(ctx, validatorAddress)
+	validatorPubkeys, err := s.getAllowedPubKeysForTAsAndValidators(ctx, validatorAddress)
```
x0152, line 99: Shouldn't we pass the inference epochId here for payload validation?

akup: I think here we don't need to check for epochId.

The edge case is:
validator is asking for payload and validator was active in previous epoch, but the new epoch already began, and we are nearly selected new validators.
Validators set was changed, and sender validator isn't active anymore.
As it is not active it anyway can't validate the inference and we can skip his request.

But there is another case.
TA is active in previous epoch and it sent StartInference to chain.
New ActiveValidators are elected and executor receives the inference payload.
Executor checks that message is from someone not eligible to be TA (as he is not a validator anymore).
Executor misses this inference.

So at `post_chat_handler` we need to check also if sender was validator 1-2 blocks before.

What do you think?

x0152: I

> The edge case is:
> validator is asking for payload and validator was active in previous epoch, but the new epoch already began, and we are nearly selected new validators.
> Validators set was changed, and sender validator isn't active anymore.
> As it is not active it anyway can't validate the inference and we can skip his request.

#760 added PreviousActiveParticipantPermission recently, so chain now accepts validators from the previous epoch. I think we should use the same approach here for TA as well. Would this approach work for you?

ouicate:
I do not think this is ready to merge yet. There are several correctness / semantics issues and I also do not see tests for the new authorization boundary.

### 1. The code only checks epoch-set membership, not weight / confirmation status
`isAddressActiveParticipantInCurrentEpoch` delegates to `s.epochGroupDataCache.IsActiveParticipant(...)`, which is effectively an address-membership check against the selected epoch's cached group data.

That does not prove the signer is "non-slashed", currently eligible, or carrying non-zero effective confirmation weight. This distinction matters because `EpochGroupData.ValidationWeights[*].ConfirmationWeight` can change mid-epoch in `inference-chain/x/inference/module/confirmation_poc.go`.

So the issue is broader than just a missing `Weight` check specifically: the helper verifies cached membership in the selected epoch set, but not current eligibility semantics.

### 2. Executor-path inactive participant rejection is returned as a plain error, and payload-path mapping is also misleading
`getAllowedPubKeysForTAsAndValidators` returns `fmt.Errorf("granter is not active in the current epoch")` when the participant check fails.
In the executor path, `validateFullRequest` returns that plain error directly. The server-wide `TransparentErrorHandler` treats non-`*echo.HTTPError` values as `500` and returns the original message in JSON.

So in the executor path, an authorization failure here will currently be reported as an internal server error rather than `401/403`.

In the payload path, the same helper error is wrapped as `echo.NewHTTPError(http.StatusUnauthorized, "validator not found")`, which avoids `500` but still misclassifies an authorization failure as a lookup failure.

Worse, if the active-participant lookup fails due to a query error, the payload path still turns that failure into `401 "validator not found"` instead of surfacing a service error.

Return an explicit `echo.NewHTTPError(...)` for this path.

### 3. Request-path chain lookup uses `context.Background()` instead of the request context
`isAddressActiveParticipantInCurrentEpoch` calls the cache with `context.Background()`.

That drops request cancellation / deadline propagation. It is also a poor fit if query height pinning or request-scoped metadata is added later.
Thread the request context through this check instead of creating a background context.

### 4. `phaseTracker.GetCurrentEpochState()` is dereferenced without a nil check
The new helper reads `s.phaseTracker.GetCurrentEpochState().LatestEpoch.EpochIndex` directly.
`GetCurrentEpochState()` can return `nil` when `latestEpoch` or `currentEpochParams` is not populated yet.
There are already multiple call sites in the repo that guard against this case explicitly.

Handle the nil state here and return a clear service-unavailable style error instead of risking a panic.

### 5. The helper derives its epoch from `phaseTracker`'s `latestEpoch`, which is not guaranteed to be the effective epoch
`isAddressActiveParticipantInCurrentEpoch` uses `s.phaseTracker.GetCurrentEpochState().LatestEpoch.EpochIndex`.
But `phase_tracker.go` explicitly documents that `latestEpoch` is not necessarily the effective epoch and may be `effectiveEpoch + 1` around phase boundaries.

So this is not just an odd naming choice: near epoch transitions the helper may check participant membership against the wrong epoch source even before payload-specific `epochId` handling comes into play.

Use an explicit effective / inference epoch source here rather than implicitly relying on `LatestEpoch.EpochIndex`.

### 6. Payload validation already has an explicit inference epoch, but the shared helper adds a second, current-epoch-only rule
This PR changes payload validation to use `getAllowedPubKeysForTAsAndValidators(...)` as well.
The payload path already resolves and verifies the inference's actual `epochId` for authorization.
But `getAllowedPubKeysForTAsAndValidators(...)` then calls `isAddressActiveParticipantInCurrentEpoch`, which hardcodes a current-epoch check.

So this is a concrete policy mismatch, not just an abstract design concern: the payload flow now mixes inference-epoch authorization with current-epoch pubkey gating. The review discussion already raised epoch-boundary concerns here, and the latest thread points at `#760` as introducing previous-epoch acceptance on chain.

Make the epoch source explicit in each path, or mirror the chain permission semantics directly instead of layering a current-epoch-only helper on top of inference-epoch validation.

### 7. The original "CurrentEpochGroupData on every request" description is stale, but the broader merge concern remains
The PR thread originally describes this as calling `queryClient.CurrentEpochGroupData` on every inference / payload-validation request.
That no longer matches the latest diff: the helper now goes through `s.epochGroupDataCache.IsActiveParticipant(...)`, so the exact "raw current-epoch gRPC per request" concern is stale.

I still think the broader concern is valid though: this PR is adding another node-local authorization cache / rule in `decentralized-api` instead of lining up with the shared per-block gRPC caching work in `#542` and the broader permission work in `#760`.
So if the recommendation is still "wait before merge", I would frame it as broader caching / permission alignment, not as the latest diff still doing uncached `CurrentEpochGroupData` on every request.

### 8. Add tests for the new security boundary
I do not see tests covering the newly introduced authorization behavior.
At minimum I would want coverage for:
- **[direct executor spoof attempt]** non-active `TransferAddress`
- **[inactive participant error path]** ensure auth failure is not surfaced as `500`
- **[same-node TA/executor path]** `taRequestValidated=true`
- **[payload request path]** now using the same helper
- **[epoch-boundary behavior]** current epoch vs inference epoch / previous-epoch acceptance

gmorgachev:
postponing to 0.2.12+

akup:
As devshards is a primary way to scale, this PR isn't needed as it patches legacy code

tcharchian:
@akup, should this PR be closed?

akup:
Issue is solved at `devshards` architecture. This code is legacy now

PR #707, opened 2026-02-05
fix: implement inference slot rotation to prevent hogging (closes #706)

### LINKED ISSUE #706: Inference Slot Hogging
**Vulnerability:** Inference Slot Hogging
**Severity:** Medium 
**Component:** model_assignment.go

### Description
The system always picks the node with the smallest weight for the "safe" inference slot. If a validator has multiple nodes, the same small node will keep getting the safe slot every single epoch, avoiding PoC verification indefinitely.

### The Problem
Verification Avoidance: The smallest node is never checked because it stays in the safe slot.
Guaranteed Rewards: This node earns rewards every epoch without risk, while the other nodes of the same validator are always forced to undergo PoC checks.

### Example
Validator has: Node A (weight 10) and Node B (weight 20).

Epoch 1: Node A is smallest -> Safe Slot.
Epoch 2: Node A is smallest -> Safe Slot.
Result: Node A never performs PoC, but always gets paid.

### Fix
A mandatory rotation. If a node was in the safe slot in the previous epoch, it is moved to the end of the queue for the next epoch. This forces the validator's other nodes to take turns in the safe slot and undergo verification.

huxuxuya:
### Description
This PR addresses a vulnerability in the ML node allocation logic where a single node could "hog" the safe inference slot across multiple epochs, effectively avoiding PoC verification indefinitely.

### The Problem
The previous implementation of 
getSmallestMLNodeWithPOCSLotFalse
 was strictly deterministic, always picking the node with the minimum weight for the safe POC_SLOT. For validators with multiple nodes, this meant their smallest node would consistently be assigned to the safe slot every epoch, securing guaranteed rewards without ever being challenged by the Proof of Confidence (PoC) mechanism.

**The Solution:** Mandatory Rotation
I have implemented a rotation mechanism within the AllocateMLNodesForPoC flow:

1. State Awareness: The allocator now tracks which nodes were in the safe slot during the previous epoch using previousEpochData.
2. Tiered Selection: The selection logic now splits nodes into two priority groups:
- Tier 1: Nodes that were not in the safe slot in the previous epoch (sorted by weight).
- Tier 2 (Fallback): Nodes that were safe previously (used only if no Tier 1 nodes are available).
3. Result: This forces nodes within a validator's fleet to alternate between the safe inference slot and the PoC-verified slots, ensuring all nodes eventually undergo verification.

Testing
New Test Suite: Added model_assignment_hogging_test.go with: 
**TestInferenceSlotRotation_FixVerification**
- Confirms that a heavier node is picked if the lighter one was safe in the previous epoch.
**TestInferenceSlotRotation_SingleNode**
- Ensures the system still functions correctly for participants with only one node.

Related Issues
Closes #706


AlexeySamosadov (changes requested):
A couple of things to clean up before merging:

**1. `fmt.Println` / `fmt.Printf` in tests**
We use project-standard logging everywhere — in tests that means `t.Log` / `t.Logf`, not `fmt.Println`. The `fmt.Print*` calls produce noise on every test run regardless of `-v` flag. There are 8 occurrences in `model_assignment_hogging_test.go`.

**2. Missing edge case tests**
The two test scenarios cover the happy path well, but we should also cover:
- Empty node list (what does `getSmallestMLNodeWithPOCSLotFalse` return?)
- All nodes already allocated (`TimeslotAllocation[1] == true` for all)
- Nodes with equal weights — tie-breaking behavior
- Three or more nodes — rotation across multiple candidates
- Nodes with `len(TimeslotAllocation) <= 1` — guard clause verification

**3. Import ordering**
In the test file, third-party (`testify`) and internal project (`productscience/inference`) imports are in the same group. Per our convention: stdlib → third-party → internal, each separated by a blank line.

**4. Design question: cross-model tracking**
`previouslySafeNodes` tracks nodes that were safe for *any* model in the previous epoch, keyed by `(participant, nodeId)` without `modelId`. So if NodeA was safe for ModelX, it also gets deprioritized for ModelY. Is this intentional? If per-model rotation is the goal, the key should include `modelId`.

AlexeySamosadov (commented):
## Review Notes

### [Medium] Cross-model safe slot tracking

`previouslySafeNodes` is keyed by `participant → nodeId`, but it's built by iterating over all models. This means if NodeA was in the safe slot for Model X in the previous epoch, it gets deprioritized for Model Y too — even though it was never safe for Model Y.

If the intent is per-model rotation, the map should be `map[modelId]map[participant]map[nodeId]bool`, queried as `previouslySafeNodes[modelId][participantAddr]`. If global rotation across models is intended, a brief comment explaining this design choice would help.

### [Low] Missing log for rotation events

When the anti-hogging logic picks a heavier node over a lighter one, that's a significant state change worth logging at Info level. Would help with monitoring and debugging in production:

```go
if smallest != nil {
    // log: "Rotation: selected node %s (weight %d) over previously-safe node %s (weight %d)"
    return smallest
}
```

### [Low] Unrelated import reorder in `model_assignment_test.go`

The 4-line import reorder is correct (stdlib → third-party → internal), but it's unrelated to the fix. Keeps the diff cleaner without it.

### Positive

- Minimal diff, focused on the problem
- Tiered selection (non-safe preferred, safe as fallback) is a clean approach
- Good edge case coverage in tests: empty list, single node, all-safe fallback, equal weights
- Graceful fallback for single-node participants

huxuxuya:
-   Added an explicit code comment in model_assignment.go that previouslySafeNodes is intentionally global (cross-model) to prevent multi-model hogging/model-spam.
-   Added an Info rotation log when a heavier non-safe node is selected over a lighter previously-safe node (selection logic unchanged).
-   Removed unrelated import reordering from model_assignment_test.go to keep the PR diff clean.

  Commits:

  - 6adf7d7d1 docs: clarify intentional global safe-slot tracking
  - 0f9d9a455 logging: add info log for safe-slot rotation decisions
  - cfd9fa745 cleanup: drop unrelated import reorder in model_assignment_test

Import-ordering note in model_assignment_hogging_test.go: imports are
  grouped as stdlib → third-party → internal with blank lines between groups, and I reverted the
  previously incorrect ordering.

patimen:
Closing this as stale. This addresses a real issue in the current slot-allocation flow, but that flow is being superseded by the devshards direction, so it is no longer the right place to carry the fix forward. It would still be worth checking the devshard implementation for the same class of slot-hogging weakness.

PR #708, opened 2026-02-05
IBC Upgrade to v8.7.0

### LINKED ISSUE #615: IBC support, IBC pool

GLiberman:
## Impact & Issues
Upgrading `ibc-go` to v8.7.0 transitively updates `cosmossdk.io/client/v2` to `v2.0.0-beta.3`. This introduces a **breaking change** in JSON serialization for Protocol Buffer enums:

### Serialization Changes Comparison

| Type | Old Behavior | New Behavior | Impact |
| :--- | :--- | :--- | :--- |
| **Enum** | Integer (e.g., `3`) | String (e.g., `"BOND_STATUS_BONDED"`) | Clients must handle string-to-enum conversion. |
| **Int64 / UInt64** | Number (e.g., `1000`) | String (e.g., `"1000"`) | Numeric fields may be serialized as strings to preserve precision. |


patimen (approved):
We could fix up the tests a bit, but they existed before you write the PR.
We **need** to do some basic testing with this running in test_net against at least a fraction of our dashboards to see if it breaks any of the APIs. I think it won't based on some of what I've seen, but I do not have high confidence on this.

DimaOrekhovPS (approved):

PR #712, opened 2026-02-07
Require proto-go generation matches committed code

patimen:
This will ensure that `ignite` is run correctly on every commit, that the proto files are generated AND committed.


AlexeySamosadov (commented):
Looked through the workflow — couple of things:

**1. Missing `actions/setup-go` with pinned version**
All other Go workflows in the repo (`verify.yml`, `dont_panic.yml`, `build-upgrades.yml`, `publish_upgrade_binaries.yml`) explicitly install Go via `actions/setup-go` with `go-version: '1.24.2'`. This workflow doesn't — it relies on whatever Go comes pre-installed on `ubuntu-latest`. Since `ignite generate proto-go` needs Go and protoc plugins, this will break unpredictably when GitHub updates the runner image. Should add the same setup step as the others.

**2. Checksum for Ignite binary (optional but worth considering)**
The Ignite tarball is downloaded via `curl` without SHA256 verification. The release page does publish `ignite_28.10.0_checksums.txt`. Since this binary generates code that ends up in the chain, verifying the checksum would add a safety layer. Not blocking — `deploy-test-net-cloud.yml` downloads `yq` the same way — but the risk profile here is higher since the output is committed source code.

DimaOrekhovPS (approved):

PR #715, opened 2026-02-07
fix: resolve keyring backend mismatch for new join nodes

### LINKED ISSUE #714: Bug Report: New Nodes Fail to Sign Transactions (Keyring Backend Mismatch)

AlexeySamosadov:
Closes #714

## Problem

New join nodes fail to sign transactions because of conflicting `KEYRING_BACKEND` defaults:

- `deploy/join/docker-compose.yml` hardcoded `KEYRING_BACKEND=file`
- `decentralized-api/scripts/init-docker.sh` hardcoded `--keyring-backend test`
- `deploy/join/config.env.template` used `export` prefixes that Docker Compose can't parse

The api container always used `test` backend regardless of config, so keys created with `file` backend were invisible.

## Fix

1. **`docker-compose.yml`**: `KEYRING_BACKEND=file` -> `${KEYRING_BACKEND:-file}` (respects operator's config.env)
2. **`config.env.template`**: Removed `export` prefixes (Docker Compose `env_file` doesn't support them)
3. **`init-docker.sh`**: Reads `KEYRING_BACKEND` from env with `test` default (matches `inference-chain/scripts/init-docker.sh` pattern), validates against allowed values (`test|file|os|memory|pass|kwallet`), uses `"$KEYRING_BACKEND"` in all key commands

## Test plan

- [ ] Deploy a new join node with `KEYRING_BACKEND=file` in config.env — verify keys are created and transactions sign correctly
- [ ] Deploy with `KEYRING_BACKEND=test` — verify backward compatibility
- [ ] Set `KEYRING_BACKEND=garbage` — verify early failure with clear error message
- [ ] Run existing join node — verify no regression


gmorgachev:
```
$APP_NAME keys add "$KEY_NAME" \
  --keyring-backend test \  # HARDCODED as 'test'
  --keyring-dir /root/.inference
```
That code is never used in real node deployment. It exists for local devnet only,

PR #718, opened 2026-02-08
fix: prevent same node from always getting safe inference slot

### LINKED ISSUE #706: Inference Slot Hogging

AlexeySamosadov:
## Summary

- Fixes issue #706: Inference Slot Hogging
- Implements deterministic rotation for PoC slot allocation
- Nodes that were in "safe slot" (POC_SLOT=false) in previous epoch are now prioritized for PoC allocation
- Ensures fair distribution of PoC verification across all nodes over time

## Problem

The system always picks the node with the smallest weight for the "safe" inference slot. If a validator has multiple nodes, the same small node keeps getting the safe slot every epoch, avoiding PoC verification indefinitely.

This causes:
- **Verification Avoidance**: The smallest node is never checked because it stays in the safe slot
- **Guaranteed Rewards**: This node earns rewards every epoch without risk

## Solution

The fix introduces rotation by:
1. Building a set of node IDs that were in the safe slot (POC_SLOT=false) in the previous epoch
2. Prioritizing those nodes for PoC allocation using `getSmallestMLNodeWithPOCSLotFalseWithRotation`
3. Falling back to smallest weight selection for new nodes or first epoch

## Test plan

- [x] Added unit test `TestAllocateMLNodesForPoC_RotationPreventsSlotHogging` verifying rotation behavior
- [x] Added unit test `TestGetSmallestMLNodeWithPOCSLotFalseWithRotation` for the selection function
- [x] Added unit test `TestBuildPreviousSafeSlotNodeSet` for the helper function
- [x] All existing tests pass

Fixes #706

---
Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/inference-chain/x/inference/module/model_assignment.go
+++ b/inference-chain/x/inference/module/model_assignment.go
@@ -460,7 +460,7 @@ func (ma *ModelAssigner) AllocateMLNodesForPoC(ctx context.Context, upcomingEpoc
 
 	for _, modelId := range sortedModelIds {
 		ma.LogInfo("Processing model for PoC allocation", types.Allocation, "flow_context", FlowContext, "sub_flow_context", SubFlowContext, "step", "model_loop_start", "model_id", modelId)
-		ma.allocateMLNodePerPoCForModel(modelId, currentEpochData, eligibleNodesData, allocationFraction)
+		ma.allocateMLNodePerPoCForModel(modelId, currentEpochData, eligibleNodesData, previousEpochData, allocationFraction)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


AlexeySamosadov:
Closing: overlaps with PR #777 (deterministic shuffle for PoC slot allocation) which is the more complete solution for slot fairness.

PR #725, opened 2026-02-10
POC weight normalization on real POC time 

gmorgachev:
### Motivation

Node weight in PoC is determined by the number of nonces generated during the generation phase. The generation window has a fixed number of blocks, but actual block duration varies due to network conditions, transaction load, or datacenter-specific issues.

When blocks take longer than expected, nodes generate more nonces. When blocks are faster, fewer nonces. This creates instability of weight distribution between epochs unrelated to actual computational capacity.

### Solution

Normalize weight by the real duration of the generation phase in seconds.

The chain captures two timestamps:
- `generation_start_timestamp` - when generation window opens
- `exchange_end_timestamp` - when exchange phase ends

Normalization factor: `expected_duration / actual_duration`

Expected duration = `generation_window_length_blocks * 5.41s`

The 5.41s constant is the average block duration measured during the first months of mainnet.

### Changes

- Add `poc_normalization_enabled` param (default: true)
- Store `generation_start_timestamp` and `exchange_end_timestamp` in `PoCValidationSnapshot`
- Apply normalization factor to weight calculation in `WeightCalculator`
- Works for both regular PoC and confirmation PoC

### DIFF
--- a/decentralized-api/broker/broker_test.go
+++ b/decentralized-api/broker/broker_test.go
@@ -224,12 +224,17 @@ func registerNodeAndSetInferenceStatus(t *testing.T, broker *Broker, node apicon
 	queueMessage(t, broker, setStatusCommand)
 	<-setStatusCommand.Response
 
-	// Wait for reconciliation to actually bring the node to INFERENCE status in the broker
+	// Wait until the node is fully stable for inference in broker state.
+	// CurrentStatus can become INFERENCE before in-flight reconciliation clears,
+	// and a reconciling node is considered unavailable.
 	deadline := time.Now().Add(2 * time.Second)
 	for time.Now().Before(deadline) {
 		nodes, _ := broker.GetNodes()
 		for _, n := range nodes {
-			if n.Node.Id == node.Id && n.State.CurrentStatus == types.HardwareNodeStatus_INFERENCE {
+			if n.Node.Id == node.Id &&
+				n.State.IntendedStatus == types.HardwareNodeStatus_INFERENCE &&
+				n.State.CurrentStatus == types.HardwareNodeStatus_INFERENCE &&
+				n.State.ReconcileInfo == nil {
 				return
 			}
 		}


inference-chain/proto/inference/inference/poc_validation_snapshot.proto
```diff
@@ -10,6 +10,8 @@ message PoCValidationSnapshot {
   int64 snapshot_height = 2;
   string app_hash = 3;
   repeated ValidatorWeight validator_weights = 4;
+  int64 generation_start_timestamp = 5;
```
DimaOrekhovPS, line 13: Is it something that wasn't added in the poc optimization PR? Or does it relate to weight norm somehow?

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -349,30 +380,31 @@ type nodeWeight struct {
...
 // Per-node weights come from MLNodeWeightDistribution.
 func (wc *WeightCalculator) calculateParticipantWeight(participantAddress string) ([]nodeWeight, int64) {
 	commit, hasCommit := wc.StoreCommits[participantAddress]
 	if !hasCommit || commit.Count == 0 {
 		return nil, 0
 	}
 
-	// Calculate total weight from commit count
-	totalWeight := mathsdk.LegacyNewDec(int64(commit.Count)).Mul(wc.WeightScaleFactor).TruncateInt64()
+	combinedFactor := wc.WeightScaleFactor
```
DimaOrekhovPS, line 391: Let's add logging here to understand what each factor is and what the final factor is

DimaOrekhovPS (approved):

inference-chain/proto/inference/inference/poc_validation_snapshot.proto
DimaOrekhovPS: ok, I see how it's used

PR #736, opened 2026-02-11
fix: validate bridge contract address in BridgeExchange handler

AlexeySamosadov:
### What's the problem?

The bridge handler (`BridgeExchange`) accepts transactions from **any** contract address — without checking if that contract is actually a registered Gonka bridge.

In simple terms: **the front door to the bridge is wide open.** A malicious validator or compromised bridge service can claim "I deposited tokens through the bridge" by referencing any contract — real or fake — and the system will process it.

### What can an attacker do?

1. Deploy a cheap contract on Ethereum that mimics bridge deposit events
2. Submit a `MsgBridgeExchange` to Gonka referencing that fake contract
3. The handler processes it as a legitimate bridge transaction
4. If enough validators in the epoch group confirm it, `handleCompletedBridgeTransaction` **mints real GONKA tokens**
5. Repeat with different `blockNumber`/`receiptIndex` values — **token supply inflation**

### Impact

- **Direct threat to token supply** — attacker can inflate GONKA supply
- **Bridge trust compromised** — the bridge is the primary entry point for cross-chain value
- **All bridged funds at risk** — inflated supply devalues every legitimate holder's tokens
- The validation function `HasBridgeContractAddress` already exists in the codebase but **was never called** in the bridge handler

### Severity: HIGH

This is a defense-in-depth vulnerability affecting the bridge — the most security-sensitive component of any cross-chain system. Exploitation requires validator-level access or bridge service compromise, combined with majority validator confirmation.

---

## Technical Details

**Root cause:** `msg_server_bridge_exchange.go` — the `BridgeExchange` function processes bridge transactions without calling `HasBridgeContractAddress()` (defined in `bridge.go`). The whitelist check exists but is simply not wired into the handler.

**Affected code path:**
- `BridgeExchange()` → creates `BridgeTransaction` → validators confirm → `handleCompletedBridgeTransaction()` → token minting
- No contract address validation at any point in this chain

**Fix (3 parts):**

1. **Whitelist validation**: Added `HasBridgeContractAddress(ctx, normalizedChain, normalizedContract)` check at the entry point of `BridgeExchange`, before any processing. Unregistered contracts are immediately rejected.

2. **Write-time normalization**: `SetBridgeContractAddress()` now normalizes `ChainId` and `Address` to lowercase before storing. This ensures consistent case-insensitive matching regardless of how governance proposals or genesis data format the addresses (Ethereum addresses are case-insensitive per EIP-55).

3. **Transaction content normalization**: The `BridgeTransaction` object is constructed with normalized (lowercase) chain ID and contract address, preventing validators from accidentally creating duplicate transaction records due to case differences.

**Test coverage:**
- `TestBridgeExchange_UnregisteredContractAddress` — confirms unregistered contracts are rejected
- `TestBridgeExchange_CaseInsensitiveContractAddress` — confirms case-insensitive matching works correctly  
- `TestBridgeExchange_DoubleVoteCaseBypass` — updated to register contract, confirms existing behavior preserved

### DIFF
--- a/inference-chain/x/inference/keeper/bridge.go
+++ b/inference-chain/x/inference/keeper/bridge.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"cosmossdk.io/collections"
 	"github.com/productscience/inference/x/inference/types"
@@ -17,6 +18,9 @@ func (k Keeper) generateBridgeAddressKey(_ context.Context, chainId, address str
 
 // SetBridgeContractAddress stores a bridge contract address
 func (k Keeper) SetBridgeContractAddress(ctx context.Context, address types.BridgeContractAddress) {
+	// Normalize to lowercase for case-insensitive matching
+	address.ChainId = strings.ToLower(address.ChainId)
+	address.Address = strings.ToLower(address.Address)
 	address.Id = k.generateBridgeAddressKey(ctx, address.ChainId, address.Address)
 	if err := k.BridgeContractAddresses.Set(ctx, collections.Join(address.ChainId, address.Address), address); err != nil {
 		k.LogError("Bridge exchange: Failed to set bridge contract address",
--- a/inference-chain/x/inference/keeper/msg_server_bridge_exchange.go
+++ b/inference-chain/x/inference/keeper/msg_server_bridge_exchange.go
@@ -34,6 +34,15 @@ func (k msgServer) BridgeExchange(goCtx context.Context, msg *types.MsgBridgeExc
 		"blockNumber", msg.BlockNumber,
 		"receiptIndex", msg.ReceiptIndex)
 
+	// Validate that the contract address is a registered bridge contract for this chain
+	normalizedChain := strings.ToLower(msg.OriginChain)
+	normalizedContract := strings.ToLower(msg.ContractAddress)
+	if !k.HasBridgeContractAddress(ctx, normalizedChain, normalizedContract) {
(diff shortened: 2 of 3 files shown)


AlexeySamosadov:
@GLiberman You're absolutely right — thanks for catching this.

I misunderstood the role of `ContractAddress` in `MsgBridgeExchange`. I assumed it always refers to a bridge contract, but it actually carries either a bridge contract address (for WGNK burn → native release) or an ERC-20 token address (for wrapped token minting), with `IsBridgeContractAddress()` handling the routing downstream in `handleCompletedBridgeTransaction`. The fix as proposed would break inbound ERC-20 bridging entirely.

That said, digging into this revealed that the naming and documentation in the bridge module actively contribute to this kind of confusion:

- **`BridgeContractAddresses` store** holds both bridge contracts and ERC-20 token addresses, but the name suggests only bridge contracts
- **`contractAddress` field** in the proto has a minimal comment (`// original chain smart contract address`) with no indication of its dual meaning
- **`HasBridgeContractAddress` vs `IsBridgeContractAddress`** — two functions with nearly identical names that do different things (direct lookup vs. cross-chain scan)
- **`handleCompletedBridgeTransaction`** — the most critical branching point in bridge logic has minimal documentation on why it branches
- **`ownerPubKey`** is validated in `ValidateBasic` but silently dropped when constructing `BridgeTransaction`
- **Format validation** functions like `isValidContractAddress()` and `isValidChainId()` exist in the codebase but aren't called in `ValidateBasic`

I'd like to close this PR and open a new one focused on improving naming and documentation in the bridge module — so the next contributor (or auditor) reviewing this code doesn't fall into the same trap. Would that be welcome?

PR #737, opened 2026-02-11
refactor(bridge): normalize hex fields and fix store asymmetries in bridge module

AlexeySamosadov:
### What this PR does

Adds lowercase normalization for `OwnerAddress` and `ReceiptsRoot` in the bridge vote-matching path, and documents the rationale.

### Context

While reviewing the bridge module, we noticed that `OwnerAddress` and `ReceiptsRoot` pass through vote-matching (`generateSecureBridgeTransactionKey`, `bridgeTransactionsEqual`) without case normalization. In the current deployment, all validators run the same geth binary, so hex casing is consistent and this does not cause issues in practice.

However, the code makes an implicit assumption about casing consistency that is not documented or enforced. If the validator set ever includes nodes with different RPC sources, or if the bridge watcher implementation changes, case-variant hex values would silently split votes.

This PR adds normalization as a **defense-in-depth** measure — not a fix for an active bug.

### Changes

1. **Input normalization** in `BridgeExchange`: `OwnerAddress` and `ReceiptsRoot` are lowercased before constructing `BridgeTransaction`
2. **Key generation**: `generateSecureBridgeTransactionKey` normalizes both fields before hashing
3. **Equality check**: `bridgeTransactionsEqual` uses `strings.EqualFold()` for both fields

### Related observations

While reviewing, we also found minor store-level asymmetries worth noting (can be addressed in a follow-up):
- `RegisterBridgeAddresses` checks duplicates with raw input (`HasBridgeContractAddress`) but stores with normalized key (`SetBridgeContractAddress`) — benign overwrite at worst, governance-only
- `GetBridgeContractAddressesByChain` does not normalize `chainId` parameter — query endpoint may return empty results for non-lowercase chain names
- `RemoveBridgeContractAddress` does not normalize before removal — asymmetric with `SetBridgeContractAddress` (currently unused function)

### Test coverage

- `TestBridgeExchange_OwnerAddressCaseNormalization` — two validators submit same event with different casing, both resolve to same transaction
- `TestBridgeExchange_DoubleVoteCaseBypass` — existing test preserved


tcharchian:
@GLiberman, please check if this PR https://github.com/gonka-ai/gonka/pull/497/changes has resolved this #737 PR or not 
cc: @mtvnastya

patimen:
Closing this as stale. The bridge normalization point still looks reasonable, but this branch has accumulated too much unrelated upgrade work to be the right vehicle now. If we still want this defense-in-depth change, it should come back rebased and refocused on the current release as a small PR scoped to the bridge normalization/store asymmetry pieces.

PR #741, opened 2026-02-12
fix(inference): prevent permanent token loss in ClaimRewards on partial payout failure

AlexeySamosadov:
## Summary
- Fixes error-swallowing in `ClaimRewards` handler that could cause permanent token loss
- When `payoutClaim` fails (e.g., insufficient escrow funds), the handler now returns `(nil, err)` instead of `(response, nil)`, enabling Cosmos SDK transaction rollback
- Removes `handleUnderfundedWork` helper that called `finishSettle` (destroying the settlement record) even when payout failed
- Removes premature `finishSettle` in reward payment error path — transaction rollback now preserves the settlement record so participant can retry

## What changed
- `msg_server_claim_rewards.go`: 
  - Main handler returns `(nil, err)` on payout failure instead of `(response, nil)`
  - `payoutClaim` returns `(nil, err)` on both work and reward payment failures
  - Removed `handleUnderfundedWork` — its `finishSettle` call destroyed the settlement record on failure
  - Removed premature `finishSettle` in reward payment error path

## Why this is safe
`ClaimRewards` is sent as individual transactions via `SendTransactionAsyncWithRetry`, NOT through the `BatchConsumer`. Unlike `MsgStartInference`/`MsgFinishInference` (which are batched and intentionally use the `failedStart`/`failedFinish` pattern to avoid rolling back entire batches), `ClaimRewards` has no batching concern.

## Test plan
- [x] All existing keeper tests pass
- [x] Reviewed by MAIR pipeline

### DIFF
--- a/decentralized-api/cosmosclient/tx_manager/batch_consumer.go
+++ b/decentralized-api/cosmosclient/tx_manager/batch_consumer.go
@@ -16,6 +16,7 @@ const (
 	batchStartConsumer        = "batch-start-consumer"
 	batchFinishConsumer       = "batch-finish-consumer"
 	batchValidationV2Consumer = "batch-validation-v2-consumer"
+	batchStartV2Consumer      = "batch-start-v2-consumer"
 	batchAckWait              = time.Minute // must exceed FlushTimeout to prevent redelivery
 
 	// V1 PoC batch consumers
@@ -44,6 +45,11 @@ type BatchConsumer struct {
 	startBatch        []pendingMsg
 	finishBatch       []pendingMsg
 	validationV2Batch []pendingMsg
+	// startV2Batch collects individual MsgStartInference messages that will be wrapped
+	// into a single MsgBatchStartInference before submission. Unlike startBatch (legacy),
+	// the on-chain handler for MsgBatchStartInference isolates each item in a CacheContext
+	// so a single failure does not roll back the entire TX.
+	startV2Batch []pendingMsg
 
 	// V1 PoC batches
 	pocBatchBatch      []pendingMsg
@@ -52,6 +58,7 @@ type BatchConsumer struct {
 	startMu        sync.Mutex
 	finishMu       sync.Mutex
 	validationV2Mu sync.Mutex
+	startV2Mu      sync.Mutex
 
 	// V1 PoC mutexes
 	pocBatchMu      sync.Mutex
@@ -60,6 +67,7 @@ type BatchConsumer struct {
 	startCreatedAt        time.Time
 	finishCreatedAt       time.Time
 	validationV2CreatedAt time.Time
+	startV2CreatedAt      time.Time
 
 	// V1 PoC timestamps
 	pocBatchCreatedAt      time.Time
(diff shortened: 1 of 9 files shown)


patimen (changes requested):
Sorry to yank you around, but actually we cannot return errors for Start and Finish.
The issue is that we are currently batching those message into a single transaction for perf reasons. But transactions are atomic, so if one fails ALL of them will fail. This doesn't even make sense.
The right solution would be to have a specific BatchMessageStart, but we are in the process of exploring ways to skip the Start and Finish messages in various ways anyhow.
The short of it is we used that finishFailed pattern for a reason. If you look at the history/blame for those messages, you could see the reasoning for such a specific pattern.

AlexeySamosadov:
You're right about Start/Finish — I dug into the `BatchConsumer` in `tx_manager/batch_consumer.go` and found the explicit commit from v0.2.6 (sub-PR #495) explaining why `failedStart`/`failedFinish` were introduced:

> *"Since there can be unexpected failures in Finish or Start Inference, this might result in an entire batch fails, despite only one message being a problem."*

Makes complete sense for batched transactions — returning `(nil, err)` would roll back the entire batch atomically.

However, `ClaimRewards` is sent as a single transaction via `SendTransactionAsyncWithRetry`, not through the `BatchConsumer` — so the batching concern doesn't apply there. The `(response, nil)` pattern in `ClaimRewards` genuinely swallows errors and can cause state corruption (e.g. `finishSettle` commits `RemoveSettleAmount` even when the payout itself fails).

I'll update the PR: revert the Start/Finish changes and keep only the ClaimRewards error propagation fix.

IgnatovFedor (changes requested):
@AlexeySamosadov, could you please rebase onto upgrade-v0.2.11 and force push?

AlexeySamosadov:
@IgnatovFedor Done — rebased onto `upgrade-v0.2.11` and force-pushed. Clean single commit now.

slandymani (approved):

AlexeySamosadov:
@patimen Done — implemented `MsgBatchStartInference` as requested.

**What was implemented:**

**Chain side (`inference-chain`):**
- `MsgBatchStartInference` message type with a `Starts []MsgStartInference` field and `MsgBatchStartInferenceResponse` with per-item `Results []MsgStartInferenceResponse`
- Full proto-compatible Marshal/Unmarshal/Size on the Go structs (manually extended `tx.pb.go` since proto-gen isn't in-repo CI)
- `BatchStartInference` RPC registered in the gRPC `_Msg_serviceDesc`
- Keeper handler (`msg_server_batch_start_inference.go`) that iterates over each `MsgStartInference`, runs it in an isolated `ctx.CacheContext()` sub-transaction:
  - Success → `writeCache()` commits item's state changes to the parent context
  - Failure → sub-context discarded, error recorded in `Results[i].ErrorMessage`; other items unaffected
  - The outer TX handler always returns `(response, nil)`, so the wrapping Cosmos TX never rolls back due to a single item failure

**API side (`decentralized-api`):**
- New NATS stream `txs_batch_start_v2`
- `BatchConsumer` gains `startV2Batch` accumulator, `handleStartV2Msg`, `flushStartV2`, `broadcastBatchStartV2` — individual `MsgStartInference` messages are aggregated into one `MsgBatchStartInference` before submission
- `PublishStartInferenceV2` publish method for callers to opt in

The existing `StartInference` + `failedStart`/`failedFinish` paths are untouched for backward compatibility. Switch `InferenceCosmosClient.StartInference` to call `PublishStartInferenceV2` when `upgrade-v0.2.11` is live.

The `ClaimRewards` fix from the original commit is preserved.

akup:
Permanent loss in ClaimReward is implemented at https://github.com/gonka-ai/gonka/pull/789
I like that PR much more as it is cleaner and accurately reviewed.

This PR also introduces unrelated batch StartInference, while they could be useful for performance, it is better to rebase and reintroduce them in next releases

@AlexeySamosadov I think this PR should be closed, and Start/Finish Inference batching reopened and covered with additional github issue.

x0152:
agree with @akup
this pr can be closed, since the ClaimRewards fix is already in #789

PR #747, opened 2026-02-13
fix(inference): preserve funding module in vested payout flow

### LINKED ISSUE #746: Vested payouts in x/inference ignore caller funding module and always debit inference account
## Summary
`PayParticipantFromModule` in `x/inference` hardcodes `types.ModuleName` (`inference`) when calling `streamvesting.AddVestedRewards`, instead of forwarding the caller-provided `moduleName`.

## Affected code
- `inference-chain/x/inference/keeper/payment_handler.go`

## Expected behavior
When vesting is enabled, payouts should debit the same funding module passed to `PayParticipantFromModule`.

## Actual behavior
The vested path always debits `inference`.

## Impact
- Top miner vested payouts (`top_reward`) can debit the wrong module account.
- Can cause false insufficient-funds errors in `inference`.
- Can skew module-account accounting.

## Repro
Call:
`PayParticipantFromModule(..., moduleName=types.TopRewardPoolAccName, vestingPeriods>0)`
and assert `AddVestedRewards(..., types.TopRewardPoolAccName, ...)`.
Before fix it receives `types.ModuleName` instead.

## Fix
In vested path:
`AddVestedRewards(..., moduleName, ...)`
instead of:
`AddVestedRewards(..., types.ModuleName, ...)`.

Schwartz10:
## What
Fix vested payout funding source in `x/inference` by forwarding caller `moduleName` to `streamvesting.AddVestedRewards`.

## Why
Previously the vested path hardcoded `inference`, which could debit the wrong module account (e.g. `top_reward` payouts).

## Commits
1. `test(inference): add failing regression for vested funding module forwarding`
2. `fix(inference): forward caller funding module for vested payouts`

## Files changed
- `inference-chain/x/inference/keeper/streamvesting_integration_test.go`
- `inference-chain/x/inference/keeper/payment_handler.go`

## Validation
- `go test ./x/inference/keeper -run "VestingIntegration_TopMinerRewards|PayParticipantFromModule_VestedPath_ForwardsFundingModule" -count=1`
- `go test ./x/inference/keeper -run "VestingIntegration|PayParticipantFromModule|ClaimRewards" -count=1`
- `go test ./x/inference/keeper -count=1`
- `go test ./x/streamvesting/keeper -count=1`

All passing after fix.

This fixes https://github.com/gonka-ai/gonka/issues/746

### DIFF
--- a/inference-chain/x/inference/keeper/payment_handler.go
+++ b/inference-chain/x/inference/keeper/payment_handler.go
@@ -69,7 +69,7 @@ func (k *Keeper) PayParticipantFromModule(ctx context.Context, address string, a
 			return err
 		}
 		// Vesting keeper should move funds and create vesting schedule
-		err = k.GetStreamVestingKeeper().AddVestedRewards(ctx, address, types.ModuleName, vestingAmount, vestingEpochs, memo+"_vested")
+		err = k.GetStreamVestingKeeper().AddVestedRewards(ctx, address, moduleName, vestingAmount, vestingEpochs, memo+"_vested")
 		if err != nil {
 			k.LogError("Error adding vested payment", types.Payments, "error", err, "amount", vestingAmount)
 			return err
--- a/inference-chain/x/inference/keeper/streamvesting_integration_test.go
+++ b/inference-chain/x/inference/keeper/streamvesting_integration_test.go
@@ -150,7 +150,7 @@ func TestVestingIntegration_ParameterBased(t *testing.T) {
 
 	expectedRewardCoins := sdk.NewCoins(sdk.NewInt64Coin(types.BaseCoin, int64(rewardAmount)))
 	mocks.StreamVestingKeeper.EXPECT().
-		AddVestedRewards(ctx, participantAddrStr, "inference", expectedRewardCoins, &rewardVestingPeriod, gomock.Any()).
+		AddVestedRewards(ctx, participantAddrStr, types.TopRewardPoolAccName, expectedRewardCoins, &rewardVestingPeriod, gomock.Any()).
 		Return(nil)
 
 	// Execute payment from top reward pool module
@@ -278,7 +278,7 @@ func TestVestingIntegration_MixedVestingScenario(t *testing.T) {
 
 	expectedRewardCoins := sdk.NewCoins(sdk.NewInt64Coin(types.BaseCoin, int64(rewardAmount)))
 	mocks.StreamVestingKeeper.EXPECT().
-		AddVestedRewards(ctx, participantAddrStr, "inference", expectedRewardCoins, &rewardVestingPeriod, gomock.Any()).
+		AddVestedRewards(ctx, participantAddrStr, types.TopRewardPoolAccName, expectedRewardCoins, &rewardVestingPeriod, gomock.Any()).
 		Return(nil)
 
 	err = k.PayParticipantFromModule(ctx, participantAddrStr, rewardAmount, types.TopRewardPoolAccName, "reward-payment", &rewardVestingPeriod)
@@ -300,7 +300,7 @@ func TestVestingIntegration_TopMinerRewards(t *testing.T) {
 	expectedCoins := sdk.NewCoins(sdk.NewInt64Coin(types.BaseCoin, int64(rewardAmount)))
 
 	mocks.StreamVestingKeeper.EXPECT().
-		AddVestedRewards(ctx, participantAddrStr, "inference", expectedCoins, &topMinerVestingPeriod, gomock.Any()).
+		AddVestedRewards(ctx, participantAddrStr, types.TopRewardPoolAccName, expectedCoins, &topMinerVestingPeriod, gomock.Any()).
 		Return(nil)
 
 	// Execute top miner reward payment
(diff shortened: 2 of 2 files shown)


AlexeySamosadov:
## Code Review — 5-agent analysis

The fix is **correct and necessary** — the bug is real. Here's the consolidated findings.

---

### Root Cause Confirmed

`PayParticipantFromModule` has two payment branches:
- **Direct path** (else): correctly uses `moduleName` → `SendCoinsFromModuleToAccount`
- **Vesting path** (if): hardcoded `types.ModuleName` (inference) → ignored the parameter entirely

The only caller affected is `top_miners.go:64` which passes `types.TopRewardPoolAccName` (top_reward) with `TopMinerVestingPeriod > 0`. Default is 0 (no vesting), but production uses 180 — so this bug has been silently active in production. The result: vesting schedules for top miners were funded by draining the inference escrow instead of `top_reward`.

---

### Must Fix Before Merge

**1. Incomplete test fix — `TestVestingIntegration_ErrorHandling` (line 343):**
The mock expectation on line 343 still uses `types.ModuleName` but the call on line 346 passes `types.TopRewardPoolAccName`. After this fix the test will fail:
```go
// line 343 — should be types.TopRewardPoolAccName, not types.ModuleName
AddVestedRewards(ctx, participantAddrStr, types.ModuleName, expectedCoins, ...)
```

**2. Verify `top_reward` module permissions:**
`AddVestedRewards` calls `SendCoinsFromModuleToModule("top_reward" → "streamvesting")`. Confirm `TopRewardPoolAccName` is registered with `Burner`/`Staker` permissions in the app's module manager — otherwise all top miner vesting payouts will start failing after this fix.

---

### Should Fix

**3. Add `moduleName` to the vesting error log (line 74):**
```go
// current — missing module context
k.LogError("Error adding vested payment", types.Payments, "error", err, "amount", vestingAmount)

// suggested — consistent with LogInfo calls on lines 58/63
k.LogError("Error adding vested payment", types.Payments, "error", err, "amount", vestingAmount, "module", moduleName)
```

**4. Add comment explaining the invariant:**
The exact line that contained the bug needs a comment to prevent regression:
```go
// Must forward moduleName, not types.ModuleName — streamvesting debits the
// fundingModule account directly. Hardcoding here would silently drain
// inference escrow for non-inference callers (e.g. top_reward payouts).
err = k.GetStreamVestingKeeper().AddVestedRewards(ctx, address, moduleName, ...)
```

---

### Consider Fixing

**5. Dead code — `return err` at line 85 is always nil:**
The vesting branch returns early on error, so line 85 is only reached with `err == nil`. Replace with `return nil`.

**6. Unnecessary alias `vestingEpochs := vestingPeriods` (line 62):**
Creates a second pointer to the same value for no reason. Use `vestingPeriods` directly on line 72.

**7. Add `moduleName \!= ""` guard:**
```go
if moduleName == "" {
    return sdkerrors.Wrap(sdkerrors.ErrInvalidRequest, "moduleName cannot be empty")
}
```
All current callers use constants so this won't break anything, but makes the contract explicit.

**8. Naming inconsistency:** Parameter is `moduleName` here but `fundingModule` in the `StreamVestingKeeper` interface. Not a blocker.

---

### Tests Missing for Full Coverage

- `PayParticipantFromModule` with `types.ModuleName` + vesting period (the `ClaimRewards` path — currently covered only via `AnyTimes()` mocks so the forwarding is never asserted)
- `IssueRefund` paths (expiry/invalidation — not covered in this file at all)
- `moduleName = ""` edge case
- `amount = 0` early return path
- Real module balance assertion (current integration tests use `AnyTimes()` bank mocks so no balance change is ever verified)

---

### Summary

| | |
|---|---|
| Fix correctness | ✅ Correct |
| Production impact | ⚠️ Active in prod when `TopMinerVestingPeriod > 0` (180 on mainnet) |
| Regression risk | ✅ None for ClaimRewards/escrow paths |
| Test suite after merge | ❌ `TestVestingIntegration_ErrorHandling` will fail — line 343 needs update |
| `top_reward` permissions | ❓ Needs verification |

AlexeySamosadov:
All review feedback addressed. Changes pushed.

**What was done:**

1. **Removed `vestingEpochs` alias** — `vestingPeriods` is now used directly in the `AddVestedRewards` call, no unnecessary indirection.

2. **Added explanatory comment** above `AddVestedRewards` explaining why `moduleName` must be forwarded and not hardcoded:
   > Must forward moduleName (not types.ModuleName) — streamvesting debits the fundingModule account directly. Hardcoding here would silently drain inference escrow for non-inference callers (e.g. top_reward payouts).

3. **Added `"module", moduleName`** to the error log in the vested path for debuggability.

4. **Replaced `return err`** (always nil after successful vesting) with explicit `return nil`.

5. **Test mock expectation** in `TestVestingIntegration_ErrorHandling` was already using `types.TopRewardPoolAccName` in the PR branch — no change needed there.

---

**Permissions check for `top_reward` module (Step 1):**

In `inference-chain/app/app_config.go` line 207:
```go
{Account: inferencemoduletypes.TopRewardPoolAccName, Permissions: []string{authtypes.Minter}},
```

`TopRewardPoolAccName` has **only `Minter`** permission — it does NOT have `Burner`. The SDK's `SendCoinsFromModuleToModule` (used inside `streamvesting.AddVestedRewards`) requires the sender module to have `Burner` permission to debit its account.

This means the vested payout path for `top_reward` will fail at runtime with an unauthorized error from the bank module. The `Burner` permission needs to be added to `TopRewardPoolAccName` in `app_config.go`:
```go
{Account: inferencemoduletypes.TopRewardPoolAccName, Permissions: []string{authtypes.Minter, authtypes.Burner}},
```

Without this, the bug fix is correct code-wise but the vesting path will still be blocked by the bank module's permission check. Worth addressing in a follow-up or as part of this PR.

patimen:
Thanks for the fix here. Even though this is no longer an operational concern after the removal of top_miner, this still looks like a good defensive fix to prevent future errors if we use  with a different funding module again.\n\nCould you update the tests to use a different module name rather than TopMiner / , since that path is deprecated now? With that adjustment, this looks like something we could merge pretty quickly.

patimen:
Thanks for the fix here. Even though this is no longer an operational concern after the removal of top_miner, this still looks like a good defensive fix to prevent future errors if we use PayParticipantFromModule with a different funding module again.

Could you update the tests to use a different module name rather than TopMiner / top_reward, since that path is deprecated now? With that adjustment, this looks like something we could merge pretty quickly.

PR #748, opened 2026-02-13
Fix: v0.2.9-post3

gmorgachev:
1. Fix for intersection between Confirmation POC and POC - onSetNewValidator was not counted in total CPOC length => CPoC was not active but event was still there which caused missed POC
2. Use Active Participants from epoch 169 at epoch 170


inference-chain/x/inference/module/confirmation_poc.go
```diff
@@ -127,7 +129,9 @@ func (am AppModule) checkConfirmationPoCTrigger(
 	confirmationWindowDuration := epochParams.PocStageDuration +
 		epochParams.PocExchangeDuration +
 		epochParams.PocValidationDelay +
-		epochParams.PocValidationDuration
+		epochParams.PocValidationDuration +
+		epochParams.SetNewValidatorsDelay +
+		safetyWindow
```
akup, line 134: We should add condition on confirmationWindowDuration calculation.

if currentEpoch < 170, then we should use old formula, and if currentEpoch >= 170 use the new one.


Now after this upgrade if we are syncing with updated version from older state, we will never sync as we will calculate the poc window that caused the error on our side as invalid, but it is already written in blockchain and our calculated appHash will not match to the one that is in the chain.

akup: But it is related to a lot of other update, so it seams to better just to wait for next snapshot to sync

PR #750, opened 2026-02-13
[P1] fix: remove participant requirement for dev inference

### LINKED ISSUE #744: Don’t require developers to register as Participants to run inference
Currently, the chain requires a Participant record not only to host, but also to send inference requests. There is no real reason for this, since the public key is available in the Account record after the first on-chain transaction signed by that account is executed. That should be sufficient.

- [ ] Remove the requirement to create a Participant record.
- [ ] Fix `/v1/participants/gonka...` to query Participant data, not just the Account (as it does now).
- [ ] Determine how to preserve per-developer statistics in this case.
- [ ] Update the documentation accordingly.

x0152:
Closes #744.

Developers no longer need to register as a Participant to run inference. An account public key from the chain is enough.

Changes:
- Removed Participant lookup in Start / Finish / Invalidate inference
- Fixed /v1/participants/:address — it was returning Account data instead of Participant data
- Added GET /v2/participants/:address and GET /v2/accounts/:address endpoints
- Renamed InferenceParticipant → AccountByAddress, added denom
- /v1 keeps returning Account data for backward compatibility
- Per-developer stats are unaffected — already tracked by account address
- Consumer setup docs updated; gonka-docs follow-up PR after merge

Note: can expose a single /v1 endpoint with both Account + Participant data if needed, but separate v2 endpoints seem cleaner.


patimen:
I've moved this to 0.2.12 for now, it seems like there are some big questions and it needs careful review.

inference-chain/x/inference/calculations/signature_validate.go
```diff
@@ -28,23 +28,23 @@ type PubKeyGetter interface {
...
 func VerifyKeys(ctx context.Context, components SignatureComponents, sigData SignatureData, pubKeyGetter PubKeyGetter) error {
-	// Check developer signature if developer participant is provided
-	if sigData.Dev != nil && sigData.DevSignature != "" {
-		devKey, err := pubKeyGetter.GetAccountPubKey(ctx, sigData.Dev.Address)
+	// Check developer signature if developer address is provided
+	if sigData.Dev != "" && sigData.DevSignature != "" {
+		devKey, err := pubKeyGetter.GetAccountPubKey(ctx, sigData.Dev)
 		if err != nil {
-			return sdkerrors.Wrap(types.ErrParticipantNotFound, sigData.Dev.Address)
+			return err
```
slandymani: wrap error with correct account name

decentralized-api/internal/server/public/get_participants_handler.go
```diff
@@ -22,33 +22,108 @@ import (
...
+	case codes.ResourceExhausted:
+		return echo.NewHTTPError(http.StatusTooManyRequests, st.Message())
+	case codes.Unavailable:
+		return echo.NewHTTPError(http.StatusServiceUnavailable, st.Message())
+	default:
+		return echo.NewHTTPError(http.StatusInternalServerError, st.Message())
+	}
+}
+
+func (s *Server) getAccountByAddressV1(c echo.Context) error {
```
slandymani: why do we need almost identical getAccountByAddressV1 and getAccountByAddress?

slandymani (changes requested):

inference-chain/x/inference/calculations/signature_validate.go
x0152: valid point, fixed

decentralized-api/internal/server/public/get_participants_handler.go
x0152: getAccountByAddressV1 is rudiment, deleted

thanks

slandymani (approved):

akup:
@x0152 please rebase

Ryanchen911 (approved):
Looks good to me. Approved.

jacky6block (approved):
LGTM

0xgonka (approved):

PR #751, opened 2026-02-13
upload proposal

fedor-konovalenko:
Uploaded proposal for new features:

## Key Improvements

1. Add a separate validation endpoint

2. Add TOPLOC algorithm

3. Extend this validation framework to additional model classes


patimen:
Thanks for putting this together. Closing this out because proposals like this should go through the normal proposal / governance process rather than live as standalone PRs like this. Please check on Discord for the current process and the right place to route proposal-style work like this.

PR #760, opened 2026-02-14
Unified Permissions

patimen:
# Unified Permission handling
Permissions for sending messages have been a bit sloppy, yet they are crucial for network security. Some messages had no permissions, some had too little. Changing them/updating them was a bit error prone and done in different ways. There was significant code duplication in various auth verification as well.

This was not only reported by Certik, it is manifestly obvious from the code that some messages have inadequate authorization.

A single procedure for setting permissions. `permissions.go` now contains:
1. A list of permissions available (multiple permissions can be required)
2. A mapping of all messages to permissions
3. Implementation of each permission check

Each message now calls CheckPermission _first thing_, and rejects immediately if the Sender doesn't have permission to execute the message. The work on the allow lists for training was integrated into this as well.

CheckPermission ALSO takes the Permissions used. If they do not match the settings in permissions.go, the call will fail. The purpose of this is to have an easy way to see Permissions both in the aggregate (in permissions.go) and for each individual message (when reviewing, for instance).

This is actually protected only through a unit test checking that the permissions listed in permissions.go match what is passed in, rather than enforced at runtime. Runtime enforcement was found to have a measurable performance hit.

## What risks does this introduce? How can we mitigate those risks?
The biggest immediate risk is that a new permission will be too restrictive, or incorrect, and will cause issues on chain. 
- Testermint will likely catch any major issues that make this mistake, so Integration Tests will be especially important.
- Reviewing which messages have which permissions is now quite easy, as looking at permissions.go is guaranteed to show the entire story.

A secondary risk is that this will degrade over time, and new messages will not use the system, and we end up right where we left off.

This is addressed by the new unit tests that will fail if new messages are introduced and don't use CheckPermission

A final risk is this might hit perf of some of our high volume messages. I've mitigated this by carefully using `Has` instead of `Get` for the various checks, as well as creating the `ActiveParticipantsSet` collection to provide O(logN) lookup for ActiveParticipants (previously we could only check ActiveParticipants by serializing the ENTIRE structure and then iterating over each Participant). This design is very fast, and does zero serializing at all. Since we only set ActiveParticipants at the very beginning of an Epoch and never again (for crytpographic verification), this is safe.

We have measured the performance of the serialization of the ActiveParticipants and it is signficant, and could add more than 10% or more (as the size grows) to each call that uses it, justifying this optimization.

Many unit tests have been added for specifically to verify permissions. In addition, a test was added that verifies that all message handlers use `CheckPermission` first thing, and that each message in the proto file has a handler.

## Other notes
There are two messages that do not quite fit the pattern:
- InvalidateInference
- RevalidateInference
Both of these are authenticated by the Group policy for the Epoch group in which the voting took place. Shoehorning this into the `CheckPermission` would have been inelegant and forced, so instead they use the `NoPermissions` permission and then authorize later on.


inference-chain/x/inference/keeper/permissions.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	if currentEpoch < epochOffset {
+		return types.ErrActiveParticipantNotFound
+	}
+	p, found := k.GetActiveParticipants(ctx, currentEpoch-epochOffset)
+	if !found {
+		return types.ErrActiveParticipantNotFound
+	}
+	for _, participant := range p.Participants {
+		if participant.Index == signer.String() {
```
slandymani: convert sigher to string once before the loop

inference-chain/x/inference/keeper/permissions.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	return err
+}
+
+func (k msgServer) checkPermissions(ctx context.Context, msg HasSigners, signer string, permissions []Permission) error {
+	signerAddr, err := sdk.AccAddressFromBech32(signer)
+	if err != nil {
+		return err
+	}
+	permission, ok := MessagePermissions[reflect.TypeOf(msg)]
```
slandymani: maybe should we precompute reflect.TypeOf?
something like:
```
var messagePermissionsByName = make(map[string][]Permission)
func init() {
    for typ, perms := range MessagePermissions {
        messagePermissionsByName[typ.String()] = perms
    }
}
```

inference-chain/x/inference/keeper/permissions.go
```diff
@@ -0,0 +1,305 @@
...
+	if !ok {
+		k.LogError("Permission not found for message type", types.Messages, "message type", reflect.TypeOf(msg))
+		return types.ErrInvalidPermission
+	}
+
+	// Double check that the global and local lists match (order independent):
```
slandymani: do we really need this? we can make this file the single source of truth and everywhere use just:
`if err := k.CheckPermission(goCtx, msg); err != nil {`

**!!! what if we move permissions check from msg handler to ante handler?**

inference-chain/x/inference/keeper/permissions.go
slandymani: we need to check received msg for existence in MessagePermissions in CheckPermission first line to avoid unnecessary work

inference-chain/x/inference/keeper/permissions.go
```diff
@@ -0,0 +1,305 @@
...
+		if !permMap[p] {
+			return types.ErrInvalidPermission
+		}
+	}
+
+	var lastErr error
+	for _, perm := range permission {
+		switch perm {
+		case GovernancePermission:
+			if err := k.checkGovernancePermission(ctx, signerAddr); err == nil {
```
slandymani: we must use `break` instead of `return nil` (now we will exit from the function after first successful check)
also we can use break with label in case of an error to avoid unnecessary checks  after the first error(or any better solution)

slandymani (changes requested):
what if we move permissions check from msg handler to ante handler?

inference-chain/x/inference/keeper/permissions.go
slandymani: if we need at least one permission to be confirmed, then it's not relevant

patimen: It's currently meant to be an _or_ check.

inference-chain/x/inference/keeper/permissions.go
patimen: Made irrelevant by using the new ActiveParticipantsCache

inference-chain/x/inference/keeper/permissions.go
patimen: Decided we no longer needed the lookup entirely, since we have the unit test making sure the values match. It's not ideal, but perf requirements force our hand here.

patimen: We no longer check, relying on the unit tests instead

inference-chain/x/inference/keeper/permissions.go
patimen: I actually think the opposite is a better call, and that's what I did. 
We remove the lookup here, though we keep the Map. The Map is checked against actual calls by the unit tests added (so it is still a canonical representation of all permissions), but we don't check at runtime (wasteful from a perf perspective).

akup:
I think that this new ActiveParticipantCache is actually introducing same performance issue that we are trying to solve at https://github.com/gonka-ai/gonka/issues/782 : marshalling of ActiveParticipantCache on every message (every inference start)

I think that it is ok to have the per-block cache for this participants cache, and reconstruct it on every BeginBlock from EpochGroupData. It will make possible that some messages will pass the test during the block, when there was prior tx that slashed the participant (for example he didn't pass confirmationPoC or something else). But it seams to be good trade-off to apply this permission changers only to the next block.

What I propose - just to remove this cache from store at all, and replace it with in-memory cache

patimen:
> I think that this new ActiveParticipantCache is actually introducing same performance issue that we are trying to solve at #782 : marshalling of ActiveParticipantCache on every message (every inference start)
> 
> I think that it is ok to have the per-block cache for this participants cache, and reconstruct it on every BeginBlock from EpochGroupData. It will make possible that some messages will pass the test during the block, when there was prior tx that slashed the participant (for example he didn't pass confirmationPoC or something else). But it seams to be good trade-off to apply this permission changers only to the next block.
> 
> What I propose - just to remove this cache from store at all, and replace it with in-memory cache
.


ActiveParticipantsCache ( I renamed it set) doesn't serialize the entire ActiveParticipants object... that's kind of the point. It creates a Set that we can do a straight O(1) lookup in on the state... very performant, and only does it when we set ActiveParticipants

I'm very reluctant to use in-memory caches in general, that path is tricky. It introduces possible inconsistencies in nodes if we don't manage it perfectly.

The key thing is that ActiveParticipants only changes once (by design) at the beginning of the Epoch, it is not changed anywhere else. So we get one hit when we create the ActiveParticipants, with not O(1) perf and zero serialization for all permissions checks.

inference-chain/x/inference/keeper/msg_server_register_token_metadata.go
slandymani: remove it, already checked

inference-chain/x/inference/keeper/msg_server_submit_hardware_diff.go
```diff
@@ -10,6 +10,9 @@ import (
 )
 
 func (k msgServer) SubmitHardwareDiff(goCtx context.Context, msg *types.MsgSubmitHardwareDiff) (*types.MsgSubmitHardwareDiffResponse, error) {
+	if err := k.CheckPermission(goCtx, msg, ParticipantPermission); err != nil {
+		return nil, err
+	}
 	ctx := sdk.UnwrapSDKContext(goCtx)
 
 	_, found := k.GetParticipant(goCtx, msg.Creator)
```
slandymani: remove it, already checked

slandymani (changes requested):

inference-chain/x/inference/keeper/msg_server_register_token_metadata.go
patimen: Good catch.

inference-chain/x/inference/keeper/msg_server_submit_hardware_diff.go
patimen: Again, good catch.

akup:
> The key thing is that ActiveParticipants only changes once (by design) at the beginning of the Epoch, it is not changed anywhere else. So we get one hit when we create the ActiveParticipants, with not O(1) perf and zero serialization for all permissions checks.

I think we need to update this cache when participant is slashed by failing ConfirmationPoC or by missed inferences. Bad participant can make harm after it is excluded from reward.

So I would update this cache on this actions: missed PoC, reached missed inference threshold and changed state to INACTIVE

akup:
> I'm very reluctant to use in-memory caches in general, that path is tricky. It introduces possible inconsistencies in nodes if we don't manage it perfectly.
> 
> The key thing is that ActiveParticipants only changes once (by design) at the beginning of the Epoch

My idea was that if it is not changed during epoch by design, to do not handle im-memory caches (that is tricky and risky) we can rebuild this cache at BeginBlocker to filter only participants whose status wasn't change to INACTIVE and whose ConfirmationWeight > 0.
So it will also protect endpoints from messages by participants that entered the epoch but was punished during it

patimen:
> > I'm very reluctant to use in-memory caches in general, that path is tricky. It introduces possible inconsistencies in nodes if we don't manage it perfectly.
> > The key thing is that ActiveParticipants only changes once (by design) at the beginning of the Epoch
> 
> My idea was that if it is not changed during epoch by design, to do not handle im-memory caches (that is tricky and risky) we can rebuild this cache at BeginBlocker to filter only participants whose status wasn't change to INACTIVE and whose ConfirmationWeight > 0. So it will also protect endpoints from messages by participants that entered the epoch but was punished during it


We already have a similar check available, but it is a bigger hit on perf and I wanted to avoid putting it into the hot-path Start/Finish. We can optimize for later. For now, this is clearly better than the current approach (Participant membership only)

akup:
> For now, this is clearly better than the current approach

Absolutely agree that this changes should be done atomically, and next optimizations should be done at the next steps

PR #767, opened 2026-02-16
Tools & Phases

gmorgachev:
# Upgrade grace period, tool calling, and PoC timing fix

## Motivation

Model parameter changes require mlnode restart. This PR:
1. Updates model params in current epoch snapshot so all nodes see new params at the same time
2. Relaxes punishment thresholds for the upgrade epoch to allow safe restarts
3. Miners should restart mlnode within 3000 blocks or during the upgrade epoch

### 1. Grace Epoch Protection

Relaxes thresholds for the upgrade epoch only:
- `binom_test_p0`: 0.1 -> 0.5 (miss/invalid punishment)
- `upgrade_protection_window`: 500 -> 3000 blocks (CPoC delay)

### 2. PoC Timing Alignment

Fixes inconsistency between API and chain artifact acceptance windows:
- Deprecates `poc_exchange_duration` (set to 0)
- Uses explicit block height checks instead of phase-only checks
- API now accepts artifacts until `poc_generation_end`, matching the chain

### 3. Qwen Model Tool Calling

Enables tool calling for `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`:
- Adds `--enable-auto-tool-choice` and `--tool-call-parser hermes`
- Updates threshold to 0.958
- Applies to current epoch snapshot immediately

## Testing

Unit tests cover grace epoch logic and artifact acceptance windows.


decentralized-api/poc/phase.go
```diff
@@ -93,26 +94,21 @@ func ShouldAcceptStoreCommit(epochState *chainphase.EpochState, pocStageStartHei
...
-		epochState.CurrentPhase == types.PoCValidateWindDownPhase ||
-		epochState.CurrentPhase == types.PoCGenerateWindDownPhase {
+		epochState.CurrentPhase == types.PoCValidateWindDownPhase {
 		return true
 	}
-
-	// Confirmation PoC: Validation phase
+	if epochState.CurrentPhase == types.PoCGenerateWindDownPhase {
+		return epochState.CurrentBlock.Height >= epochState.LatestEpoch.EndOfPoCGeneration()
+	}
```
qdanik, line 107: If I remove this condition will I be available to distribute weights in validate phase?

gmorgachev: 1. You wont be able to write commits, as they will be declined on-chain (same `IsPoCExchangeWindow` is used to check)
```
		if !epochContext.IsPoCExchangeWindow(currentBlockHeight) {
			return nil, sdkerrors.Wrap(types.ErrPocTooLate, "PoC exchange window closed")
		}
```
2. For `MsgMLNodeWeightDistribution`, it can be recorded at any time till the end of validation phase. But it doesn;t make sense to write it earlier as it must exactly match commited weight. 
Btw, for distribution if might make sense to limit recording to validation phase only

qdanik: @gmorgachev then we can simplify ShouldAcceptGeneratedArtifacts condition to 
```go
if epochState.CurrentPhase == types.PoCGeneratePhase || 
	epochState.CurrentPhase == types.PoCGenerateWindDownPhase {
	return epochState.LatestEpoch.IsPoCExchangeWindow(epochState.CurrentBlock.Height)
}
```

PS in case if it will not impact on performance

decentralized-api/poc/phase.go
```diff
@@ -6,21 +6,22 @@ import (
...
 func ShouldAcceptGeneratedArtifacts(epochState *chainphase.EpochState) bool {
 	if epochState.IsNilOrNotSynced() {
 		return false
 	}
-	// Regular PoC generation
 	if epochState.CurrentPhase == types.PoCGeneratePhase {
 		return true
 	}
-	// Confirmation PoC generation during inference phase
+	if epochState.CurrentPhase == types.PoCGenerateWindDownPhase {
```
DimaOrekhovPS, line 16: Can be merged:
```
if epochState.CurrentPhase == types.PoCGeneratePhase || epochState.CurrentPhase == types.PoCGenerateWindDownPhase {
    return epochState.LatestEpoch.IsPoCExchangeWindow(epochState.CurrentBlock.Height)
}
```
Not sure if it adds to the understanding of it though

inference-chain/app/upgrades/v0_2_10/upgrades.go
```diff
@@ -204,6 +208,8 @@ func setPocTimingParams(ctx context.Context, k keeper.Keeper) {
 	params.EpochParams.PocStageDuration = 35
 	// Update validation duration: reduce from 480 to 240 blocks
 	params.EpochParams.PocValidationDuration = 240
+	// Deprecated: set to 0, nonce acceptance now ends at poc_generation_end
+	params.EpochParams.PocExchangeDuration = 0
```
DimaOrekhovPS, line 212: The rationale behind this parameter was to let all the participants to submit their final nonces. 

I think it's ok to set it to zero anyway, because everyone will have some final nonces rejected and it won't affect the final weight distribution

inference-chain/x/inference/keeper/accountsettle.go
```diff
@@ -147,8 +147,17 @@ func (k *Keeper) SettleAccounts(ctx context.Context, currentEpochIndex uint64, p
 	// Aggregate MLNodes from model-specific subgroups for preservedWeight calculation
 	participantMLNodes := k.AggregateMLNodesFromModelSubgroups(ctx, currentEpochIndex, data.ValidationWeights)
 
+	// Check if this is a grace epoch and override BinomTestP0 if so
+	validationParams := params.ValidationParams
+	if graceParams, ok := k.GetPunishmentGraceEpoch(ctx, currentEpochIndex); ok && graceParams.BinomTestP0 != nil {
+		graceValidationParams := *validationParams
```
DimaOrekhovPS, line 158: Just in case let's check that `validationParams != nil` before deregerencing

inference-chain/x/inference/keeper/accountsettle.go
```diff
@@ -147,8 +147,17 @@ func (k *Keeper) SettleAccounts(ctx context.Context, currentEpochIndex uint64, p
...
+		graceValidationParams.BinomTestP0 = graceParams.BinomTestP0
+		validationParams = &graceValidationParams
```
DimaOrekhovPS, line 160: Why can't we just modify `validationParams` inplace here instead of copying `validationParams` to `graceValidationParams`, modifying and then assigning the copy to the `validationParams` once again?

DimaOrekhovPS (approved):

PR #779, opened 2026-02-20
Inference msgs optimization: optimize key verification

### LINKED ISSUE #608: [0/4] `StartInference` and `FinishInference`: optimiziation
StartInference and FinishInference should be significantly optimized, as they are messages used frequently—potentially 1,000+ times per block. The execution time of these messages should be below 1 ms, ideally below 0.2 ms. (currently with 1000 participants, and 1 grantee per account, StartInference can be 0.4-0.5ms, and FinishInference 4-5ms)

1. The biggest contributor to the execution time of FinishInference (or late StartInference) is reading and writing EpochGroup. EpochGroup shouldn’t be read or written in frequently executed messages, and unmarshaling/marshaling that large blob is too slow. We should either move the values we need to access/edit into separate records, or process these updates in EndBlocker.
2. After EpochGroup, the next biggest contributor is signature verification.
2.1. First, we should switch to Ethereum-optimized signature verification (github.com/ethereum/go-ethereum/crypto/secp256k1).
2.2. Second, when the chain receives a StartInference transaction, we don’t need to verify the TA signature again (it’s already verified in the transaction). Similarly, when the chain receives FinishInference, there’s no need to verify Executor signatures. Also, if FinishInference arrives after StartInference, we shouldn’t re-check the TA and Developer signatures; and when we get a late StartInference, we shouldn’t re-check the Developer signature.

After these changes, benchmark StartInference and FinishInference. If they are still not below 0.2 ms, identify what else should be optimized and report back in the issue.

Also report back in this issue which messages also use EpochGroups, and which we have to optimize as well.

Additionally, for (2.2), ensure that we validate that the timestamp, request original hash, and TA address are correct (they must match InferenceId which derived from them). Also, check that the request modified hash matches: save it from the first-arriving message, and when the second arrives, verify they are equal. If they are not equal, one party is cheating:
• If FinishInference arrives late and the hashes differ, verify the TA signature. If we have valid TA signatures on both messages, then the TA is the cheater.
• If StartInference arrives late, verify the TA signature included in FinishInference; if the hashes differ, that means the TA is the cheater.
• In all other scenarios, the Executor is the cheater.
Unfortunately as TA signature doesn't derived from request original hash, it may be the issue as Executor can present TA signature from a different InferenceId (with same timestamp). So we either should change that or do on chain conversion from request original hash to request modified hash, which can be expansive/but rear (need to measure the time it requires)

DimaOrekhovPS:
# Pull Request Description

This PR addresses the computational overhead of performing redundant cryptographic signature verifications on the blockchain. Previously, the network performed multiple EC signature verifications across different inference lifecycle messages (`StartInference` and `FinishInference`) and verified an executor signature that is no longer strictly required by protocol policy.

Cryptographic signature verification is a known computationally expensive operation in blockchain environments. Reducing the number of signature recoveries directly decreases CPU load and block processing time. In profiling of the live production chain, verifying signature was more than 50% of the time spent during each Start or Finish Message processing, severely limiting our total throughput.

This PR implements an optimized **first-message verifies, second-message compares** policy for inference messages:
*   Whichever message arrives **first** (`StartInference` or `FinishInference`) pays the computational cost of cryptographic signature verification. 
*   The **second** message bypasses cryptographic verification and instead performs lightweight equality checks on its signature-bound fields against the persisted state of the first message.

**Specific Policy Implementations:**
*   **Dev Signature:** Verified on the first message. The second message only compares `original_prompt_hash`, `request_timestamp`, `transfer_agent`, and `requested_by`.
*   **Transfer Agent (TA) Signature:** Verified on the first message *only* if it is a finish message. If `StartInference` arrives first, TA verification is skipped and deferred to the `FinishInference` message.
*   **Executor Signature:** Cryptographic verification is completely disabled/removed by policy. All facts in the FinishInference message are already cryptographically signed by the Executor
*   **State Persistence:** Added persistence for `OriginalPromptHash`, `PromptHash`, and `RequestedBy` during message processing in `inference_state.go` so the second message has the necessary state to compare against.

**Risks:**
By bypassing strict cryptographic signature checks on the second message, there is a risk of accepting spoofed or mutated data if the fields bounded by the original signature are not thoroughly validated against the already-verified state.

**Mitigations:**
*   **Strict Component Comparison:** Added comprehensive cross-message comparison methods (`compareDevComponents`, `compareStartTAComponents`, `compareFinishTAComponents`, `compareStartModelField`, etc.) that strictly validate the incoming message fields against the trusted state.
*   **Extensive Testing:** Added `msg_server_cross_message_comparison_test.go` which rigorously tests both in-order and out-of-order message flows, ensuring that component mismatches (e.g., mismatched prompt hashes, request timestamps, requested_by fields, or models) are actively rejected with newly added error types (`ErrDevComponentMismatch`, `ErrTAComponentMismatch`, `ErrInferenceRoleMismatch`).

The diff clearly demonstrates the removal of the executor signature `VerifyKeys` call and wraps the remaining `VerifyKeys` calls for Dev and TA signatures in conditional blocks checking `existingInference.StartProcessed()` or `existingInference.FinishedProcessed()`. Because the code paths now short-circuit into O(1) string/integer comparisons instead of ecrecover operations for the second message, the computational overhead is guaranteed to be reduced. The passing test suites confirm that validation integrity is maintained despite the optimization. Local benchmarking has indicated significant perf improvement, testnet will be tested along with eventual benchmarks for mainnet.

---

## Other Notes

*   **Atomic Upgrade Required:** During the upgrade, we must guarantee that both the `decentralized-api` and the `inferenced` binary are upgraded simultaneously. Clients failing to upgrade will start failing to process inferences due to the new hashing and signature verification rules.
*   **State Marker Update:** The `calculations.startProcessed()` logic has been refactored into a method on the `Inference` type (`StartProcessed()`). It now checks `AssignedTo != ""` rather than checking the `PromptHash` (since `ProcessFinishInference` now populates `PromptHash`, it is no longer a reliable marker for start processing).
*   **Invariant Enforcement:** `FinishInference` now explicitly checks that the `Creator` matches the `ExecutedBy` address, throwing an `ErrInferenceRoleMismatch` if they differ. This is why we can trust that the data in `FinishInference` is cryptographically verified.
*   **Testermint (Kotlin Integration Tests) Updates:**
    *   Added `PromptHashing.kt` to centralize and strictly mirror the Go backend's canonical JSON serialization and prompt hashing behavior.
    *   Updated the `InferenceTests` to reflect the new policy: tests now expect a bad TA signature on a `StartInference` message to succeed (since it is deferred), and a bad EA signature on a `FinishInference` message to succeed (since it is disabled).
    *   Added a missing check in the public API handler (`post_chat_handler.go`) to ensure an explicitly empty `PromptHash` is rejected with a `400 Bad Request`.


inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -217,6 +238,90 @@ func getFinishTASignatureComponents(msg *types.MsgFinishInference) calculations.
...
+		)
+	}
+	if inference.AssignedTo != "" && inference.AssignedTo != msg.ExecutedBy {
+		return sdkerrors.Wrapf(
+			types.ErrTAComponentMismatch,
+			"executor mismatch: finish.executed_by=%s start.assigned_to=%s",
+			msg.ExecutedBy,
+			inference.AssignedTo,
+		)
+	}
```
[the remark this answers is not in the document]
DimaOrekhovPS: This code will only trigger in start-first flow
The start-first flow guaranteed AssignedTo to be non-empty

inference-chain/x/inference/calculations/inference_state.go
```diff
@@ -187,7 +191,10 @@ func ProcessFinishInference(
 }
 
 func startProcessed(inference *types.Inference) bool {
-	return inference.PromptHash != ""
+	// StartInference always assigns MaxTokens (explicit or default).
+	// Finish-first flow can populate PromptHash early, so use MaxTokens to detect
+	// whether StartInference has already been processed.
+	return inference.MaxTokens != 0
```
DimaOrekhovPS: Alternatively we can check `inference.AssignedTo != ""`

inference-chain/x/inference/calculations/inference_state.go
```diff
@@ -186,14 +190,6 @@ func ProcessFinishInference(
 	return currentInference, &payments, nil
 }
 
-func startProcessed(inference *types.Inference) bool {
```
DimaOrekhovPS, line 189: Removed duplicate implementations, see `inference.go`

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -540,7 +540,11 @@ func (s *Server) handleExecutorRequest(ctx echo.Context, request *ChatRequest, w
 		logging.Error("Failed to compute prompt hash", types.Inferences, "error", err)
 		return echo.NewHTTPError(http.StatusBadRequest, "Failed to compute prompt hash")
 	}
-	if request.PromptHash != "" && computedPromptHash != request.PromptHash {
+	if request.PromptHash == "" {
+		logging.Error("Empty prompt hash", types.Inferences, "error", err)
```
slandymani: err is always nil

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -217,6 +235,90 @@ func getFinishTASignatureComponents(msg *types.MsgFinishInference) calculations.
...
+func (k msgServer) compareFinishTAComponents(msg *types.MsgFinishInference, inference *types.Inference) error {
+	if inference.PromptHash != msg.PromptHash {
+		return sdkerrors.Wrapf(
+			types.ErrTAComponentMismatch,
+			"prompt_hash mismatch: finish=%s start=%s",
+			msg.PromptHash,
+			inference.PromptHash,
+		)
+	}
+	if inference.RequestTimestamp != msg.RequestTimestamp {
```
slandymani, line 260: duplicated comparison of `msg.RequestTimestamp` and `msg.TransferredBy`, we already checked them in `compareFinishDevComponents`

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -126,14 +148,92 @@ func (k msgServer) verifyKeys(ctx sdk.Context, msg *types.MsgStartInference, age
...
+func (k msgServer) compareStartTAComponents(msg *types.MsgStartInference, inference *types.Inference) error {
+	if inference.PromptHash != msg.PromptHash {
+		return sdkerrors.Wrapf(
+			types.ErrTAComponentMismatch,
+			"prompt_hash mismatch: start=%s finish=%s",
+			msg.PromptHash,
+			inference.PromptHash,
+		)
+	}
+	if inference.RequestTimestamp != msg.RequestTimestamp {
```
slandymani, line 229: duplicated comparison of `msg.RequestTimestamp` and `inference.TransferredBy`, we already checked them in `compareStartDevComponents`

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -49,19 +49,41 @@ func (k msgServer) StartInference(goCtx context.Context, msg *types.MsgStartInfe
...
+			k.LogError("StartInference: verifyStartFirstMessageKeys failed", types.Inferences, "error", err)
+			return failedStart(ctx, sdkerrors.Wrap(types.ErrInvalidSignature, err.Error()), msg), nil
+		}
+		k.LogDebug("StartInference: dev signature cryptographically verified; TA signature deferred to FinishInference", types.Inferences, "inferenceId", msg.InferenceId)
+	}
+	k.LogDebug("StartInference: executor signature verification disabled by policy", types.Inferences, "inferenceId", msg.InferenceId)
+
 	// Record the current price only if this is the first message (FinishInference not processed yet)
 	// This ensures consistent pricing regardless of message arrival order
 	if !existingInference.FinishedProcessed() {
```
slandymani, line 98: should this be moved to the previous if statement so as not to duplicate the check?

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -75,6 +74,34 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
...
+			k.LogError("FinishInference: verifyFinishKeys failed", types.Inferences, "error", err)
+			return failedFinish(ctx, sdkerrors.Wrap(types.ErrInvalidSignature, err.Error()), msg), nil
+		}
+		k.LogDebug("FinishInference: dev and TA signatures cryptographically verified", types.Inferences, "inferenceId", msg.InferenceId)
+	}
+	k.LogDebug("FinishInference: executor signature verification disabled by policy", types.Inferences, "inferenceId", msg.InferenceId)
+
 	// Record the current price only if this is the first message (StartInference not processed yet)
 	// This ensures consistent pricing regardless of message arrival order
 	if !existingInference.StartProcessed() {
```
slandymani, line 118: move price recording to previous if statement, merge model checks with `compareFinishModelField`

slandymani (changes requested):
change pr description after removing `failOnCompareMismatch`

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -141,20 +163,105 @@ func (k msgServer) verifyKeys(ctx sdk.Context, msg *types.MsgStartInference, age
...
+
+type HasDevComponents interface {
+	GetOriginalPromptHash() string
+	GetRequestTimestamp() int64
+	GetRequestedBy() string
+	GetTransferredBy() string
+}
+
+func (k msgServer) compareDevComponents(msg HasDevComponents, inference *types.Inference) error {
+	if inference.OriginalPromptHash != msg.GetOriginalPromptHash() {
```
patimen, line 183: My understanding is that this will mean inferences that started before the ugprade will be rejected?
I'd say this all but certain to be ok... since we already interrupt in-flight inferences and end up having them expire during an upgrade anyhow.

inference-chain/x/inference/keeper/msg_server_start_inference.go
patimen: Given the extremely low cost of an additional comparison, and the clarity of having each comparison grouped together, I think this is fine.

inference-chain/x/inference/keeper/msg_server_finish_inference.go
patimen: Again, conceptually this is a little tighter, vs trying to imagine/model whether which other checks we have run already, especially from comparing two ints as we do here.

inference-chain/x/inference/keeper/msg_server_finish_inference.go
patimen: Why? These are assignments, not comparisons, and should happen after we have confirmed validity.

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -58,19 +58,41 @@ func (k msgServer) StartInference(goCtx context.Context, msg *types.MsgStartInfe
...
+			return failedStart(ctx, err, msg), nil
+		}
+		k.LogDebug("StartInference: cryptographic signature verification skipped; dev and TA components compared for consistency", types.Inferences, "inferenceId", msg.InferenceId)
+	} else {
+		err := k.verifyStartFirstMessageKeys(ctx, msg, &dev)
+		if err != nil {
+			k.LogError("StartInference: verifyStartFirstMessageKeys failed", types.Inferences, "error", err)
+			return failedStart(ctx, sdkerrors.Wrap(types.ErrInvalidSignature, err.Error()), msg), nil
+		}
+		k.LogDebug("StartInference: dev signature cryptographically verified; TA signature deferred to FinishInference", types.Inferences, "inferenceId", msg.InferenceId)
```
jacky6block, line 92: "TA signature deferred to FinishInference" — this comment is misleading.                                                                                                                                                                                                                                                         
                  
  In the start-first path, Start skips TA sig verification, and Finish only does
  field equality checks (line 97). So the TA signature is never cryptographically
  verified at any point in the inference lifecycle.

  Is this intentional? If so, can you document the security trade-off reasoning in
  the PR description?

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -86,6 +85,34 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
 		return failedFinish(ctx, sdkerrors.Wrap(types.ErrInferenceExpired, "inference has already expired"), msg), nil
 	}
 
+	// Signature verification policy:
+	// - Start first: finish performs equality checks only (no TA/dev re-verification).
+	// - Finish first: verify dev + TA signatures.
+	// - Executor signature verification is disabled by policy in both paths.
+	if existingInference.StartProcessed() {
```
jacky6block, line 92: Now that verification is conditional, `requestor` and `transferAgent` (fetched unconditionally above) go unused in the start-first path. Consider moving their lookups into the `else` branch where they're actually needed.

patimen (approved):
I've verified this every way I can think of. There are a few additional possible perf improvements, but I think we should hold off for evidence that they are actually impacting overall perf before adding additional complexity.

PR #787, opened 2026-02-21
fix: settlement atomicity and error handling in SettleAccounts (#784)

### LINKED ISSUE #784: [P2] Possible underfunded issues
# Problem

## Task 1: Analysis

## Task 2: Important fixes

0xMayoor:
Addresses #784 Task 2 — settlement error handling in `accountsettle.go`. 
Finding IDs reference the Task 1 analysis on #784.

---

**[F-01]** `GetBitcoinSettleAmounts` error was logged but not returned. On error the amounts slice is nil, so `amounts[i]` in the settlement loop panics. In EndBlock that's a permanent node crash. Now returns early.

**[F-02]** `SettleAccounts` does mint, balance resets, perf summaries, and settle writes as independent KV operations. If something fails mid-way (e.g. `SetParticipant` or a bank send), earlier writes are already committed. Participants can end up with zeroed balances but no settle record. Wrapped the mutation section in `CacheContext` — either everything commits or nothing does.

**[F-03]** `SetSettleAmountWithGovernanceTransfer` return value was discarded. That function transfers any existing unclaimed settle to governance before writing the new one. If the governance transfer fails, the function returns before writing the new settle. The old settle remains, but the current-epoch settle write is skipped. With the balance already reset earlier in the loop, the participant loses their current-epoch earnings silently. Now checked.

**[F-04]** `TransferOldSettleAmountsToGovernance` error logged not returned. Old settles stuck but records persist for retry. Low severity on its own, but it was inside the atomic section so a failure here would roll back current-epoch settlement too. Separated it out — see below.

**[F-05]** `SettleAccounts` error swallowed by the orchestrator in `module.go`. Even when the function correctly returns an error, it gets thrown away. Now safe because CacheContext guarantees no partial state on failure, but the error is still properly propagated.

Also fixed: `AddTokenomicsData` return value was discarded — coins could be minted but subsidy counter not incremented.

Old settle cleanup separated — `TransferOldSettleAmountsToGovernance` was inside the same code path as current-epoch settlement. With CacheContext, a failure transferring a stale unclaimed settle from a previous epoch would roll back the entire current epoch settlement. Moved it to run after the atomic commit on the real context so current epoch participants aren't affected by old cleanup failures.

---

**What this does NOT fix (pre-existing, unchanged)**

Participants with `amount.Error` (from reward calculation) still get their balance reset in loop 1 but skipped in loop 2 (no settle record written). CacheContext makes this atomic but doesn't change the per-participant logic — that's a broader design question.

`GetActiveParticipants` not found returns nil (silent skip). Original design choice.

---

**Error semantics note**: 
if `SettleAccounts` returns an error after `writeFn()` (only possible from old-settle cleanup), current-epoch settlement is already committed. The error means old unclaimed settles weren't cleaned up. The caller (`onEndOfPoCValidationStage`) logs this and continues, which is correct.

### DIFF
--- a/inference-chain/x/inference/keeper/accountsettle.go
+++ b/inference-chain/x/inference/keeper/accountsettle.go
@@ -165,6 +165,7 @@ func (k *Keeper) SettleAccounts(ctx context.Context, currentEpochIndex uint64, p
 	amounts, bitcoinResult, err = GetBitcoinSettleAmounts(allParticipants, &data, params.BitcoinRewardParams, validationParams, settleParameters, participantMLNodes, k.Logger())
 	if err != nil {
 		k.LogError("Error getting Bitcoin settle amounts", types.Settle, "error", err)
+		return err
 	}
 	if bitcoinResult.Amount < 0 {
 		k.LogError("Bitcoin reward amount is negative", types.Settle, "amount", bitcoinResult.Amount)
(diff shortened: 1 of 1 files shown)


akup:
It seams all changes from this PR are included into https://github.com/gonka-ai/gonka/pull/789
@0xMayoor Or am I missing something? do we need both of this PRs?

0xMayoor:
@akup no you are right #789 has all of this, closing

PR #788, opened 2026-02-22
perf: move inference stats off

### LINKED ISSUE #781: [2/4] `StartInference` and `FinishInference`
# Background

Download profiling file:
And use command:
```
```

- 10% of `FinishInference`, 
- 12% of `StartInference`, 
- 14% of Validation
- 33% is Logging, 

- 20% is Logging
- 45% is 2xGetEpochGroupData
- 5% GetEpochIndex
- 10% SetEpochGroupData, 

- 63% is Logging
- 9% Add/GetTokenomicsData

# Tasks:
We do `SetInference` twice in the main function of the `StartInference` and `FinishInference` messages, as well as in `HandleInferenceComplete` for the second Start/Finish. We can execute it once.

We should move `SetDeveloperStats` with `SetOrUpdateInferenceStatsByEpoch` and `SetOrUpdateInferenceStatusByTime` from `SetInference` to off-chain (store the data on api node). The stored structures are quite big for the on-chain storage.

Add the required data to the emitted event in `HandleInferenceComplete` (inference_finished event), and adjust the event listener on api nodes to collect this data and store it independently (look for storage we use for payload storage on api node). Check which endpoints are used by the dashboard, and see if we need to store the per-inference stats (like we do now), or only per block/model cumulative stats.

# Important
This specific task does not requires maintaining and operating a node on mainnet in order to test and validate the result.

x0152:
Closes #781

## Summary

Move per-inference developer stats off the hot path. Skip heavy per-developer stats writes on the hot path. Add lightweight 24-byte per-(second, model) on-chain store instead. Collect per-inference finished stats (completed inferences only) off-chain on API nodes via the enriched inference_finished chain event

## On-chain

- All message handlers (StartInference, FinishInference, Validation, InvalidateInference, RevalidateInference) now use SetInferenceWithoutDevStatComputation — skips writing legacy per-developer stats.

- StartInference and FinishInference use CacheContext and save inference once at the end instead of twice.

- Added lightweight on-chain model usage store: fixed 24-byte binary bucket per (second, model) — InferenceCount (8B) + TokensUsed (8B) + ActualCost (8B). No protobuf, no per-inference granularity.

- GetSummaryByModelAndTime merges legacy data (before cutover) with lightweight buckets (after). All on-chain consumers go through it and require no changes:
  - UpdateDynamicPricing (BeginBlocker) — per-model utilization for price adjustments
  - MaximumInvalidationsReached (Validation) — per-model inference count for rate limiting
  - InferencesAndTokensStatsByModels gRPC query — stats for API/dashboard
  - InferencesAndTokensStatsByTimePeriod gRPC query — aggregated stats

- Legacy per-developer gRPC queries (StatsByTimePeriodByDeveloper, StatsByDeveloperAndEpochsBackwards) still read from old stores and receive no new writes after the hot-path switch. Added warning logs when these are called post-cutover.

- Extended inference_finished event with requested_by, executed_by, model, epoch_id, prompt_token_count, completion_token_count, actual_cost, end_block_timestamp.

## Decentralized API

- InferenceFinishedEventHandler persists per-inference stats from inference_finished event into a local SQLite table (inference_finished_stats). Upsert by inference_id, retry on SQLITE_BUSY/LOCKED, auto-prune rows older than 30 days.

- Pricing handler (getModelMetrics) reads per-model token stats from the local SQLite store first. Falls back to chain gRPC query (InferencesAndTokensStatsByModels) if local store is empty or unavailable — works because the on-chain lightweight store serves compatible per-model aggregates

- Only consumer right now is getModelMetrics for the pricing/utilization endpoint. The local store has per-inference granularity to support more dashboard queries in the future

## Follow-up

- Pruning on-chain lightweight store - model_usage_stats buckets grow indefinitely, no cleanup mechanism.
- Per-developer dashboard queries degraded — StatsByTimePeriodByDeveloper and StatsByDeveloperAndEpochsBackwards stop getting new data after cutover. No off-chain replacement built for per-developer queries yet (only per-model is served from local SQLite).
- Validation, RevalidateInference and InvalidateInference handlers need new events emitted and handled on API side to track inference status changes off-chain
- On-chain block-ending stats are currently lightweight per-(second, model) buckets by design, not full per-inference/per-block-ending stats.

### DIFF
--- a/decentralized-api/internal/server/public/get_pricing_handler.go
+++ b/decentralized-api/internal/server/public/get_pricing_handler.go
@@ -165,6 +165,13 @@ func (s *Server) getModelMetrics(queryClient types.QueryClient, context context.
 			return metricsData
 		}
 		effectiveStats = chainModelStats
+	} else if hasMissingModelStats(capacityMap, localModelStats) {
+		chainModelStats, err := s.getModelTokenStatsFromChain(queryClient, context, windowDurationSeconds)
+		if err != nil {
+			logging.Warn("Failed to get missing model stats from chain, using local-only stats", types.Pricing, "error", err)
+		} else {
+			effectiveStats = mergeModelTokenStatsForMissingModels(capacityMap, localModelStats, chainModelStats)
+		}
 	}
 
 	for modelID, totalTokens := range effectiveStats {
@@ -178,6 +185,37 @@ func (s *Server) getModelMetrics(queryClient types.QueryClient, context context.
 	}
 
 	return metricsData
+}
+
+func hasMissingModelStats(capacityMap map[string]uint64, stats modelTokenStats) bool {
+	for modelID := range capacityMap {
+		if _, exists := stats[modelID]; !exists {
+			return true
+		}
+	}
+	return false
+}
+
+func mergeModelTokenStatsForMissingModels(
+	capacityMap map[string]uint64,
+	localStats modelTokenStats,
+	chainStats modelTokenStats,
+) modelTokenStats {
+	merged := make(modelTokenStats, len(localStats))
+	for modelID, totalTokens := range localStats {
+		merged[modelID] = totalTokens
+	}
+
+	for modelID := range capacityMap {
+		if _, exists := merged[modelID]; exists {
+			continue
+		}
+		if totalTokens, exists := chainStats[modelID]; exists {
+			merged[modelID] = totalTokens
+		}
+	}
+
+	return merged
 }
 
 func (s *Server) getModelTokenStatsFromLocalStore(ctx context.Context, windowDurationSeconds int64) (modelTokenStats, bool) {
--- a/decentralized-api/internal/server/public/get_pricing_handler.go
+++ b/decentralized-api/internal/server/public/get_pricing_handler.go
@@ -167,14 +167,14 @@ func (s *Server) getModelMetrics(queryClient types.QueryClient, context context.
 		return metricsData
 	}
 
-	// Calculate time window (similar to BeginBlocker logic)
-	currentTime := time.Now().Unix()
-	timeWindowStart := currentTime - windowDurationSeconds
+	// Calculate time window in milliseconds to match chain stats query semantics
+	currentTimeMillis := time.Now().UnixMilli()
+	timeWindowStartMillis := currentTimeMillis - windowDurationSeconds*1000
 
 	// Get stats for all models in time window
 	statsResponse, err := queryClient.InferencesAndTokensStatsByModels(context, &types.QueryInferencesAndTokensStatsByModelsRequest{
-		TimeFrom: timeWindowStart,
-		TimeTo:   currentTime,
+		TimeFrom: timeWindowStartMillis,
+		TimeTo:   currentTimeMillis,
 	})
 	if err != nil {
 		logging.Warn("Failed to get model stats for utilization", types.Pricing, "error", err)
--- a/decentralized-api/internal/server/public/get_pricing_handler.go
+++ b/decentralized-api/internal/server/public/get_pricing_handler.go
@@ -152,10 +152,24 @@ func (s *Server) getModelMetrics(queryClient types.QueryClient, context context.
 	if err != nil || params.Params.DynamicPricingParams == nil {
 		return metricsData // Return with capacity data only
 	}
+	windowDurationSeconds := int64(params.Params.DynamicPricingParams.UtilizationWindowDuration)
+
+	localModelStats, ok := s.getModelTokenStatsFromLocalStore(context, windowDurationSeconds)
+	if ok {
+		for modelID, totalTokens := range localModelStats {
+			if capacity, exists := capacityMap[modelID]; exists && capacity > 0 {
+				metricsData[modelID] = ModelMetrics{
+					Capacity:    capacity,
+					Utilization: float64(totalTokens) / float64(capacity),
+				}
+			}
+		}
+		return metricsData
+	}
 
 	// Calculate time window (similar to BeginBlocker logic)
 	currentTime := time.Now().Unix()
-	timeWindowStart := currentTime - int64(params.Params.DynamicPricingParams.UtilizationWindowDuration)
+	timeWindowStart := currentTime - windowDurationSeconds
 
 	// Get stats for all models in time window
 	statsResponse, err := queryClient.InferencesAndTokensStatsByModels(context, &types.QueryInferencesAndTokensStatsByModelsRequest{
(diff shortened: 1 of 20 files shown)


inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -96,25 +100,26 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
...
-		return failedFinish(ctx, err, msg), nil
-	}
-	err = k.SetInference(ctx, *finalInference)
+	finalInference, err := k.processInferencePayments(cacheCtx, inference, payments, true)
 	if err != nil {
 		return failedFinish(ctx, err, msg), nil
 	}
-	if existingInference.IsCompleted() {
-		err := k.handleInferenceCompleted(ctx, finalInference)
+	if finalInference.IsCompleted() && finalInference.EpochId == 0 {
```
slandymani: why do we need `finalInference.EpochId == 0` check? and the same in start inference

x0152: The idea was to prevent double handleInferenceCompleted call in edge cases (since EpochId is set there), but this is already fully blocked by StartProcessed/FinishedProcessed, and the extra check creates a hidden dependency on EpochId (if it is set earlier, completion may be skipped)

I'll remove it to keep logic clearer, thanks

slandymani (approved):

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -302,10 +301,31 @@ func (k msgServer) handleInferenceCompleted(ctx sdk.Context, existingInference *
...
-	if err != nil {
-		return err
+	if err := k.AddModelUsageSample(
+		ctx,
+		existingInference.Model,
+		existingInference.EndBlockTimestamp,
+		existingInference.PromptTokenCount+existingInference.CompletionTokenCount,
+		existingInference.ActualCost,
+	); err != nil {
+		k.LogError("Failed to update model usage stats", types.Stat, "inferenceId", existingInference.InferenceId, "error", err)
```
akup, line 311: Shouldn't we return err here?

inference-chain/x/inference/keeper/msg_server_finish_inference_test.go
```diff
@@ -120,12 +120,8 @@ func TestMsgServer_FinishInference(t *testing.T) {
...
 	require.Equal(t, expected, &savedInference)
 
-	devStat, found := k.GetDevelopersStatsByEpoch(ctx, testutil.Requester, epochId)
-	require.True(t, found)
-	require.Equal(t, types.DeveloperStatsByEpoch{
-		EpochId:      epochId,
-		InferenceIds: []string{expected.InferenceId},
-	}, devStat)
+	_, found = k.GetDevelopersStatsByEpoch(ctx, testutil.Requester, epochId)
+	require.False(t, found)
```
akup, line 124: Why we changed `require.True(t, found)` to opposite `require.False(t, found)` here?

decentralized-api/internal/server/public/get_pricing_handler.go
```diff
@@ -184,6 +187,105 @@ func (s *Server) getModelMetrics(queryClient types.QueryClient, context context.
 	return metricsData
 }
 
+func hasMissingModelStats(capacityMap map[string]uint64, stats modelTokenStats) bool {
```
jacky6block, line 190: The default window is 60 seconds. Any model that received zero inferences in that 60-second window will be absent from local stats — this is normal, expected behavior, not a data gap. But hasMissingModelStats treats it as a data gap and triggers a chain gRPC call on every pricing request.

  In practice, unless every registered model has at least one inference every 60 seconds, this fallback will fire constantly, defeating the purpose of the local store optimization.

decentralized-api/internal/server/public/get_pricing_handler.go
```diff
@@ -184,6 +187,105 @@ func (s *Server) getModelMetrics(queryClient types.QueryClient, context context.
...
+	if s.configManager == nil || s.configManager.SqlDb() == nil || s.configManager.SqlDb().GetDb() == nil {
+		return nil, false
+	}
+
+	nowMillis := time.Now().UnixMilli()
+	fromMillis := nowMillis - windowDurationSeconds*1000
+
+	rows, err := s.configManager.SqlDb().GetDb().QueryContext(
+		ctx,
+		`SELECT model, COALESCE(SUM(total_tokens), 0)
```
jacky6block, line 231: The SQL query is written directly here instead of going through InferenceFinishedStatsStore. This bypasses the store abstraction — if the schema changes later, this needs to be updated separately from the store layer.

Suggestion: add a GetModelTokenStats(ctx, fromMs, toMs int64) (map[string]int64, error) method to InferenceFinishedStatsStore and call it here.

inference-chain/x/inference/keeper/msg_server_finish_inference.go
jacky6block, line 311: The error is intentionally non-fatal here, but this is a write to the data that dynamic pricing reads from. A silent failure means pricing will silently use stale utilization numbers.

Could we at least expose a counter metric for this failure so we can alert on it in production? A plain log error is easy to miss under load.

inference-chain/x/inference/keeper/query_developer_stats_aggregation.go
```diff
@@ -32,6 +32,10 @@ func (k Keeper) StatsByTimePeriodByDeveloper(ctx context.Context, req *types.Que
 	}
 
 	k.LogInfo("StatsByTimePeriodByDeveloper", types.Stat, "developer", req.Developer, "time_from", req.TimeFrom, "time_to", req.TimeTo)
+	if cutoverTimestamp, ok := k.GetModelUsageCutoverTimestamp(ctx); ok && req.TimeTo >= cutoverTimestamp {
+		k.LogWarn("StatsByTimePeriodByDeveloper uses legacy per-developer stats after model-usage cutover", types.Stat,
+			"developer", req.Developer, "time_from", req.TimeFrom, "time_to", req.TimeTo, "cutover", cutoverTimestamp)
+	}
 	stats := k.GetDeveloperStatsByTime(ctx, req.Developer, req.TimeFrom, req.TimeTo)
 	return &types.QueryStatsByTimePeriodByDeveloperResponse{Stats: stats}, nil
```
jacky6block, line 40: After cutover, this endpoint will return incomplete (or empty) data for the new time range, but the response gives callers no signal that this is happening — the warning only appears in server logs.

  Can we surface this degradation to the caller? For example, adding a boolean field like data_may_be_incomplete to the response proto, or at minimum documenting the cutover timestamp in the response so clients can detect it themselves.

akup: This cutover will be used once, when we are switching to a new optimized behavior. Changing protocol is too big change for this edge case that even doesn't change anything and will not be used.
So I don't see real reasons to tell something to caller. It could be just announced that after upgrade this endpoint is not working and is replaced by listening to finish inferences.

inference-chain/x/inference/keeper/msg_server_start_inference_test.go
```diff
@@ -66,12 +66,8 @@ func TestMsgServer_StartInference(t *testing.T) {
...
 	require.True(t, found)
 	require.Equal(t, expected, &savedInference)
-	devStat, found := k.GetDevelopersStatsByEpoch(ctx, testutil.Requester, epochId)
-	require.True(t, found)
-	require.Equal(t, types.DeveloperStatsByEpoch{
-		EpochId:      epochId,
-		InferenceIds: []string{expected.InferenceId},
-	}, devStat)
+	_, found = k.GetDevelopersStatsByEpoch(ctx, testutil.Requester, epochId)
+	require.False(t, found)
```
akup, line 70: Let's remove this lines if now it is not used, why to check it is false in tests?

inference-chain/x/inference/keeper/msg_server_finish_inference_test.go
akup: Ah, it's not used anymore. let's remove this useless lines, why to test it's empty

x0152:
This issue is already closed by PR https://github.com/gonka-ai/gonka/pull/812, so there's no point in maintaining this PR further

tcharchian:
@x0152 in that case, could you please close this PR?

PR #789, opened 2026-02-22
[P2] Fix/784 fund atomicity error safety

### LINKED ISSUE #468: Certik (finalization)

### LINKED ISSUE #784: [P2] Possible underfunded issues

0xMayoor:
Continuing  #787 (settlement atomicity). Same `CacheContext` approach applied to the remaining fund-movement paths, plus error propagation fixes and an integer narrowing audit. Finding IDs reference the Task 1 analysis on #784.

---

**[F-06]** `finishSettle` deletes the settle record and marks `Claimed = true` before the payout is confirmed. If the payout fails, the claim is gone permanently — no retry path. Wrapped the entire `payoutClaim` (escrow pay + reward pay + finishSettle) in a single `CacheContext` so the record only gets removed after both payments succeed. The old code also had a partial path where work was paid but rewards failed and the settle was still consumed — that's gone now, it's all-or-nothing.

**[F-08]** In `processInferencePayments`, when FinishInference reprices lower and `IssueRefund` fails, the error was logged and execution continued. Inference finishes without the developer getting their change back. Now returns the error, which feeds into the FinishInference CacheContext — if refund fails, nothing commits.

**[F-09]** `processInferencePayments`, `SetInference`, and `handleInferenceCompleted` all wrote to state on the raw context. If any step failed after earlier writes, partial state persisted and `FinishedProcessed()` blocked retry, making it permanent. Wrapped the whole mutation section in one `CacheContext`. Found this while fixing F-08.

**[F-12]** In `PayParticipantFromModule`'s vesting path, `AddVestedRewards` does a coin transfer then a schedule update. If the schedule write fails after coins move, the schedule still has the old entry and the same amount gets sent again next epoch. Wrapped in `CacheContext` so the transfer and schedule commit together or not at all.

**[F-14]** 11 narrowing casts across the keeper with no bounds checks. Highest risk: three weight casts in ClaimRewards validation sampling where `Weight` (int64) gets cast to uint32 — silent truncation corrupts sampling probability. Added bounds-checking helpers in `safecast.go`. Consensus paths error on overflow (skips that inference for validation, which is better than corrupted probability), query paths clamp to MaxInt32 with a log warning. One tradeoff worth noting: if weights somehow exceed ~4.2B, that inference becomes unvalidatable. In practice weights are bounded by staked supply so this shouldn't happen, but it's a conscious choice over silent truncation.

**[F-15]** `TransferOldSettleAmountsToGovernance` failure returned an error to EndBlock, halting the chain. The existing code comment already said this should be non-fatal — it's just cleanup of unclaimed settles from previous epochs. Changed to log and continue. The unclaimed amounts stay in the module account and get picked up on the next epoch's settlement pass.

---

EndBlock error path audit — went through all 5 `return err` paths in EndBlock and verified each one is genuinely unrecoverable (missing params, failed epoch state writes, failed DKG group creation). Added inline comments with the rationale. Also added minimal `epoch_error` events at 3 cross-module error paths (collateral advance, settlement, weight adjustment) so indexers can detect failures without parsing logs.

---

**Note on error semantics**:
 msg servers in this codebase return `(response, nil)` to the SDK and encode failures in `ErrorMessage`. This PR preserves that contract. `CacheContext` handles atomicity within it — if a mutation fails, state rolls back, but the tx still "succeeds" from the SDK's perspective.


inference-chain/x/inference/keeper/payment_handler.go
akup: BTW, here error also is not processed and propagated

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -101,21 +101,38 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
...
 	}
 
-	finalInference, err := k.processInferencePayments(ctx, inference, payments, true)
+	// FinishInference returns nil error to the SDK regardless of internal failures.
+	// This is intentional: returning an error would revert the entire transaction,
+	// but the caller has already paid gas and expects an ErrorMessage response.
+	// CacheContext ensures that if ANY mutation below fails, ALL mutations roll back,
+	// preventing partial state (e.g., escrow moved but inference not updated,
+	// or participant stats incremented but inference not marked completed).
+	cacheCtx, writeFn := ctx.CacheContext()
```
akup, line 143: Actually didn't got why we need additional context here.
This is the processing of `MsgFinishInference` that already has the context and any error already will revert the whole transaction and there will no be partial state. Why we need additional cacheCtx ?

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -46,67 +46,67 @@ func (k msgServer) ClaimRewards(goCtx context.Context, msg *types.MsgClaimReward
 }
 
 func (ms msgServer) payoutClaim(ctx sdk.Context, msg *types.MsgClaimRewards, settleAmount *types.SettleAmount) (*types.MsgClaimRewardsResponse, error) {
-	// TODO: Optimization: Payout claim should be done in one transaction
 	ms.LogInfo("Issuing rewards", types.Claims, "address", msg.Creator, "amount", settleAmount.GetTotalCoins())
 
+	// Use CacheContext so all payout mutations are atomic.
+	// If any payment fails, nothing is committed and the settle record
+	// persists for retry.
+	cacheCtx, writeFn := ctx.CacheContext()
```
akup, line 70: I don't think we should have additional cacheCtx, it is already in context of the tx processing `MsgClaimRewards`, it already handles reverts and atomicity

0xMayoor: doesnt ClaimRewards swallow the error from payoutClaim? it returns (response, nil) even on failure so sdk never rolls back. thats what CacheContext is for here

inference-chain/x/inference/keeper/msg_server_finish_inference.go
0xMayoor: same thing, FinishInference goes through failedFinish which returns (response, nil), sdk doesnt see an error

inference-chain/x/inference/keeper/payment_handler.go
0xMayoor: will check this one later @akup

inference-chain/x/inference/keeper/msg_server_finish_inference.go
akup: Quite tricky, missed this

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
akup: Yes, sorry. resolved

inference-chain/x/inference/keeper/payment_handler.go
0xMayoor: You’re right with this one @akup 

you want to do a fix in a separate PR or should I commit one?

akup: I think it is so tiny change that actually doesn't change a lot, and it is better if you add it to the commit, then I will create one more just for this change

Mayveskii:
> Reviewed msg_server_claim_rewards.go and settle_amount.go against the full settlement flow.
>
> F-06 confirmed at line 86: finishSettle (removes settle record + sets Claimed=true) runs on the reward error path after work coins are already transferred. The settle entitlement is permanently consumed with no retry. CacheContext wrapping the entire payoutClaim is the correct fix.
>
> F-14 confirmed at three sites: int64(GetWorkCoins()) (line 62), int64(GetRewardCoins()) (line 80), and uint32(totalWeight) / uint32(weight.Weight) in ShouldValidate (line 469). The safecast.go approach handles this correctly.
>
> Additional finding — GetTotalCoins overflow (settle_amount.go:6): RewardCoins + WorkCoins is added as uint64 before the > math.MaxInt64 guard. If both values are large, uint64 wrap-around produces a small sum that bypasses the check. The guard should precede the addition.
>
> Settlement correctness is foundational. Any additive weight feature built on top — EffectivePocWeight (#856), CacheQualityWeight (#859), or future quality axes — computes on top of the same settleAmount values that pass through ShouldValidate. Partial payouts or silent integer truncation here propagate into sampling probability and ultimately into epoch weight distribution. This PR makes the layer below those features trustworthy.

inference-chain/x/inference/module/module.go
```diff
@@ -344,12 +392,16 @@ func (am AppModule) EndBlock(ctx context.Context) error {
 	}
 
 	timeouts := am.keeper.GetAllInferenceTimeoutForHeight(ctx, uint64(blockHeight))
```
x0152, line 383: If refund fails at block 100, timeout stays stored with height=100. But EndBlock on block 101 only queries timeouts for height=101 (exact match), so the old timeout is never retried. Maybe should query <= blockHeight or move failed timeout to blockHeight+1?

0xMayoor: this is actually what F-07 in the PR fixes @x0152  — expireInferenceAndIssueRefund uses CacheContext now so if the refund fails nothing commits, inference stays STARTED. and the timeout only gets removed if the refund actually went through (check the committedInferences map in EndBlock). 

so the scenario you're describing is the bug on main, not on this branch.

akup:
Very accurate and good PR and should be merged.

But should be rebased first, as there was a lot of conflicting changes at https://github.com/gonka-ai/gonka/pull/812
Then it should be rereviewed

inference-chain/x/inference/module/module.go
akup: It seams that @x0152 is right on this.

@0xMayoor 
We are setting `InferenceTimeouts` on `InferenceStart` and we are setting their keys by `ExpirationHeight`.
So if we didn't commit changes at `expireInferences` as you do now, this inferences will not be read and cleared at next block height, as it has old `expireInferences` block height that we have already passed

x0152: @akup Yes, exactly what I meant

Timeout stays at old height when refund fails, but endblock reads only current height, so it won’t retry on next block

patimen: @0xMayoor I agree with akup and x0152. I'm not even sure postponing expirations makes sense. The `no funds available` situation here seems like the right call is to end the expiration, mark the missed inference and simply log/swallow the inability to refund. Inference costs are extremely small in GNK, and the no funds situation is very unlikely 🤞

akup: I think it is a right way to go. We just need to stop this inference processing.

Even if we implement this retrying on next turn, there also will no be funds to refund and we will just create the cycle of retries.
Moreover I agree that this situation is highly unlikely to ever happen. But if this situation happened the best choice is not to retry, but log error, that could be investigated, and expire the inference.

jacky6block (commented):
Agreed with @x0152 and @akup — the timeout retry won't work as intended. `GetAllInferenceTimeoutForHeight` does an exact prefix scan on `expirationHeight`, so any timeout left at block H will never be picked up again. The fix suggested by @patimen is the right call: always commit the EXPIRED status and log/swallow the refund failure.

  Also worth noting: `accountsettle.go` in this PR conflicts with PR #787 targeting the same branch. The two handle `TransferOldSettleAmountsToGovernance` failure differently — #787  returns the error, #789  swallows it as non-fatal. #789's approach is correct and should take precedence when merging.

inference-chain/x/inference/module/module.go
0xMayoor: fair, reverted F-07 in the rebase. just expires and logs the refund error now

0xMayoor:
> Very accurate and good PR and should be merged.
> 
> But should be rebased first, as there was a lot of conflicting changes at #812 Then it should be rereviewed

rebased against upgrade-v0.2.11

x0152:
@0xMayoor could you please rebase and resolve conflicts

akup (approved):

inference-chain/x/inference/keeper/msg_server_claim_rewards.go
```diff
@@ -62,67 +62,69 @@ func (k msgServer) ClaimRewards(goCtx context.Context, msg *types.MsgClaimReward
...
 	if err != nil {
 		return nil, fmt.Errorf("failed to get params: %w", err)
 	}
 	workVestingPeriod := &params.TokenomicsParams.WorkVestingPeriod
-	if err := ms.PayParticipantFromEscrow(ctx, msg.Creator, int64(escrowPayment), "work_coins:"+settleAmount.Participant, workVestingPeriod); err != nil {
+	if err := ms.PayParticipantFromEscrow(cacheCtx, msg.Creator, int64(escrowPayment), "work_coins:"+settleAmount.Participant, workVestingPeriod); err != nil {
 		if sdkerrors.ErrInsufficientFunds.Is(err) {
-			ms.handleUnderfundedWork(ctx, err, settleAmount)
+			ms.LogError("Insufficient funds for paying participant for work, claim can be retried", types.Claims, "error", err, "settleAmount", settleAmount)
+			spendable, required := ms.parseBalanceError(err.Error())
```
patimen: I don't understand the need for this, it seems brittle. Do we really need to parse this out instead of just logging the error that already contains this information?

x0152 (approved):

patimen (approved):
The complete failure of the settle is messy, but ultimately fairer, better handled and just as unlikely as the previous logic. This gives us an atomic settle, one way or the other, instead of weird partial settles that we might benefit one participant over another.
THe rest of the changes are solid. I still think we might long term have some handling for underfunding/future clawback, but it should not be needed if we've done the rest of the code right.

0xgonka (approved):

PR #790, opened 2026-02-23
fix: use caller's module account for vested payments

0xMayoor:
`PayParticipantFromModule` takes a `moduleName` param to control which module funds the payment. The direct path uses it correctly, but the vesting path hardcodes `types.ModuleName` ("inference") instead of the passed `moduleName`.

When `top_miners.go` calls with `"top_reward"` and `TopMinerVestingPeriod > 0`, vested top miner rewards get pulled from the inference module instead of the top_reward pool. Right person gets paid, wrong account debited. Drains the inference module while top_reward accumulates unspent funds.

One-line fix — `payment_handler.go:72`, `types.ModuleName` to `moduleName`. 
Test expectations updated to match.


0xMayoor:
duplicate #746

PR #793, opened 2026-02-24
Ak issue782 0.2.11

### LINKED ISSUE #782: [3/4] `StartInference` and `FinishInference`
# Background

Download profiling file:
And use command:
```
```

- 10% of `FinishInference`, 
- 12% of `StartInference`, 
- 4% of Validation
- 33% is Logging, 

- 20% is Logging
- 45% is 2xGetEpochGroupData
- 5% GetEpochIndex
- 10% SetEpochGroupData, 

- 63% is Logging
- 9% Add/GetTokenomicsData

# Tasks:
In `HandleInferenceComplete`, we also read `GetEpochGroupData` to add `ExecutorReputation`, `ExecutorPower`, and `TotalPower` (of the model group) to `InferenceValidationDetails`, which is then saved for future validation. We also increment `NumberOfRequests` of the epoch group and save it. This operation should also be moved to the `EndBlocker`. Execute `GetEpochGroup` (main and for each required models) and `SetEpochGroup` only once per block.

We should add a key Block+InferenceId in `HandleInferenceComplete` then iterate through  the keys to get Inferences by id during `EndBlocker` to store `InferenceValidationDetails` (clean keys immediately in the `EndBlocker` after the iteration).

After moving those operations to the `EndBlocker`, we need to validate if the endblocker time won't be increased significantly by the action (though adding `GetInference` iterations to `EndBlock` without changing state during transactions) - it should take not more than 50-100ms for 1000 inferences in a mainnet node. The test can be done by adding the read operations to `EndBlocker` mainnet node but without set operation, so that state of the node will stay the same.

# Important
However, this specific task requires maintaining and operating a node on mainnet in order to test and validate the result.

akup:
[Issue 782](https://github.com/gonka-ai/gonka/issues/782)

Add inference keeper caches: epoch group block cache (current/previous epoch) with tx draft and single atomic current-epoch request counter; invalidate in BeginBlock, flush in EndBlock.

**Motivation:**
It is more generic approach as we should read EpochGroupData in a lot of places. For example we should read it on every epoch participant operation since we need to check it is initiated by participant. Currently we don't do it, but this is a bunch of vulnerabilities: anyone can send StartInferenceMsg (from any developer account), anyone can validate or invalidate anything and so on. This vulnerabilities will be closed in a next PRs. Ability to anyone send StartInferenceMsg is a huge risk, as any participant could be slashed by this attack (it is not possible while we have TA white list, but it will be removed and vulnerability will be reopened)

So I've implemented more generic approach using caches for EpochGroupData, that solves problem to read/write once per block.
For NumberOfRequests I use atomic integer to increment it during the block in parallel and commit on endBlocker

**Lifecycle:** during transaction we have EpochGroupData draft cache, we read it always once, it will be committed to per-block cache if we made a write operations when we successfully commit transaction. We also have per-block cache (cleared on block start and committed to store on endBlocker)

**Preparation for cosmos optimistic mode:**
**Disabled by default can be enabled by environment variable flag** `COSMOS_OPTIMISTIC_CACHES=1`

`lockWrite` and `unlockWrite` are dummy when flag is not enabled, so there is no locking, and `isWriteLocked` also is always false.

In CosmosSDK it is possible to have optimistic mode for parallel transactions execution, it is important to speedup on multi-CPU cores. Optimistic transactions run in parallel and if there was multiple tx that write to same store key, only one tx will success and other will be reverted and retryed. It means that store should be optimized to do not have too shared keys, like NumberOfRequests in EpochGroupData, that is touched by every inference.

I've implemented for EpochGroupData reentrant read/write locking: if any routine writes to it, write lock will be aquired and released on transaction end (commit or revert), so it is not possible to write from different threads to EpochGroupData even when optimistic mode is enabled.

**Important! Status**:
Checked on real blocks that caches generate exactly same data.

Point to discuss is use of `InferenceValidationDetails.TrafficBasis:         uint64(math.Max(currentEpochGroup.GroupData.NumberOfRequests, currentEpochGroup.GroupData.PreviousEpochRequests))`

Previous version used incremented NumberOfRequests here, but actually I think that it is not so important for traffic basis to get very precise data here and we can use the value that is fixed on the beginning of the block. `NumberOfRequests` will be updated by incrementing total number of requests during the block.

### DIFF
--- a/inference-chain/app/ante.go
+++ b/inference-chain/app/ante.go
@@ -25,6 +25,43 @@ import (
 	inferencetypes "github.com/productscience/inference/x/inference/types"
 )
 
+// EpochGroupDraftDecorator binds a tx-scoped EpochGroupData draft to the request context at tx start.
+// PostHandler must call CommitEpochGroupDraftFromContext on success; block cache is flushed to store in EndBlock.
+type EpochGroupDraftDecorator struct {
+	InferenceKeeper *inferencemodulekeeper.Keeper
+}
+
+// AnteHandle attaches WithEpochGroupDraft to the context for the rest of the tx.
+func (d EpochGroupDraftDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if d.InferenceKeeper == nil {
+		return next(ctx, tx, simulate)
+	}
+	newCtx := ctx.WithContext(inferencemodulekeeper.WithEpochGroupDraft(ctx.Context()))
+	return next(newCtx, tx, simulate)
+}
+
+// EpochGroupDraftCommitPostDecorator merges the tx-scoped EpochGroupData draft into the block cache on tx success.
+type EpochGroupDraftCommitPostDecorator struct {
+	InferenceKeeper *inferencemodulekeeper.Keeper
+}
+
+// PostHandle runs the rest of the post chain, then on success commits the epoch group draft to the block cache; on failure releases the draft write lock.
+func (d EpochGroupDraftCommitPostDecorator) PostHandle(ctx sdk.Context, tx sdk.Tx, simulate, success bool, next sdk.PostHandler) (sdk.Context, error) {
+	newCtx, err := next(ctx, tx, simulate, success)
+	if err != nil {
+		return newCtx, err
+	}
+	if d.InferenceKeeper == nil {
+		return newCtx, nil
+	}
+	if success {
+		d.InferenceKeeper.CommitEpochGroupDraftFromContext(newCtx)
+	} else {
+		d.InferenceKeeper.ReleaseEpochGroupDraftFromContext(newCtx)
+	}
+	return newCtx, nil
+}
+
 // HandlerOptions extend the SDK's AnteHandler options by requiring the IBC
 // channel keeper.
 type HandlerOptions struct {
@@ -192,6 +229,7 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 
 	anteDecorators := []sdk.AnteDecorator{
 		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
+		EpochGroupDraftDecorator{InferenceKeeper: options.InferenceKeeper}, // tx-scoped EpochGroupData draft; committed to block cache in PostHandler, flushed in EndBlock
 		wasmkeeper.NewLimitSimulationGasDecorator(options.NodeConfig.SimulationGasLimit), // after setup context to enforce limits early
 		wasmkeeper.NewCountTXDecorator(options.TXCounterStoreService),
 		wasmkeeper.NewGasRegisterDecorator(options.WasmKeeper.GetGasRegister()),
@@ -251,4 +289,6 @@ func (app *App) setAnteHandler(txConfig client.TxConfig, nodeConfig wasmtypes.No
 
 	// Set the AnteHandler for the app
 	app.SetAnteHandler(anteHandler)
+	// Merge tx-scoped EpochGroupData draft into block cache on tx success (block cache flushed to store in EndBlock)
+	app.SetPostHandler(sdk.ChainPostDecorators(EpochGroupDraftCommitPostDecorator{InferenceKeeper: &app.InferenceKeeper}))
 }
--- a/inference-chain/app/ante.go
+++ b/inference-chain/app/ante.go
@@ -32,11 +32,16 @@ type EpochGroupDraftDecorator struct {
 }
 
 // AnteHandle attaches WithEpochGroupDraft to the context for the rest of the tx.
+// For CheckTx and simulate, also attaches WithCommittedStateOnly so the keeper uses only the store (last committed block), not the block cache.
 func (d EpochGroupDraftDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
 	if d.InferenceKeeper == nil {
 		return next(ctx, tx, simulate)
 	}
-	newCtx := ctx.WithContext(inferencemodulekeeper.WithEpochGroupDraft(ctx.Context()))
+	goCtx := inferencemodulekeeper.WithEpochGroupDraft(ctx.Context())
+	if ctx.IsCheckTx() || simulate {
+		goCtx = inferencemodulekeeper.WithCommittedStateOnly(goCtx)
+	}
+	newCtx := ctx.WithContext(goCtx)
 	return next(newCtx, tx, simulate)
 }
 
@@ -46,12 +51,18 @@ type EpochGroupDraftCommitPostDecorator struct {
 }
 
 // PostHandle runs the rest of the post chain, then on success commits the epoch group draft to the block cache; on failure releases the draft write lock.
+// During DeliverTx: commit on success, release lock on failure. During CheckTx/simulate: always release the lock only (no commit),
+// so we never leak the block-cache draft writer lock; we still skip commit to avoid mutating the shared block cache.
 func (d EpochGroupDraftCommitPostDecorator) PostHandle(ctx sdk.Context, tx sdk.Tx, simulate, success bool, next sdk.PostHandler) (sdk.Context, error) {
 	newCtx, err := next(ctx, tx, simulate, success)
 	if err != nil {
 		return newCtx, err
 	}
 	if d.InferenceKeeper == nil {
+		return newCtx, nil
+	}
+	if ctx.IsCheckTx() || simulate {
+		d.InferenceKeeper.ReleaseEpochGroupDraftFromContext(newCtx)
 		return newCtx, nil
 	}
 	if success {
--- a/inference-chain/app/ante.go
+++ b/inference-chain/app/ante.go
@@ -51,8 +51,7 @@ type EpochGroupDraftCommitPostDecorator struct {
 }
 
 // PostHandle runs the rest of the post chain, then on success commits the epoch group draft to the block cache; on failure releases the draft write lock.
-// During DeliverTx: commit on success, release lock on failure. During CheckTx/simulate: always release the lock only (no commit),
-// so we never leak the block-cache draft writer lock; we still skip commit to avoid mutating the shared block cache.
+// During DeliverTx: commit on success, release lock on failure. During CheckTx/simulate: skip (we never take the block-cache draft writer lock there, so nothing to release or commit).
 func (d EpochGroupDraftCommitPostDecorator) PostHandle(ctx sdk.Context, tx sdk.Tx, simulate, success bool, next sdk.PostHandler) (sdk.Context, error) {
 	newCtx, err := next(ctx, tx, simulate, success)
 	if err != nil {
@@ -62,7 +61,7 @@ func (d EpochGroupDraftCommitPostDecorator) PostHandle(ctx sdk.Context, tx sdk.T
 		return newCtx, nil
 	}
 	if ctx.IsCheckTx() || simulate {
-		d.InferenceKeeper.ReleaseEpochGroupDraftFromContext(newCtx)
+		// We never take the block-cache draft writer lock during CheckTx/simulate, so nothing to release or commit.
 		return newCtx, nil
 	}
 	if success {
@@ -239,8 +238,8 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 	}
 
 	anteDecorators := []sdk.AnteDecorator{
-		ante.NewSetUpContextDecorator(), // outermost AnteDecorator. SetUpContext must be called first
-		EpochGroupDraftDecorator{InferenceKeeper: options.InferenceKeeper}, // tx-scoped EpochGroupData draft; committed to block cache in PostHandler, flushed in EndBlock
+		ante.NewSetUpContextDecorator(),                                                  // outermost AnteDecorator. SetUpContext must be called first
+		EpochGroupDraftDecorator{InferenceKeeper: options.InferenceKeeper},               // tx-scoped EpochGroupData draft; committed to block cache in PostHandler, flushed in EndBlock
 		wasmkeeper.NewLimitSimulationGasDecorator(options.NodeConfig.SimulationGasLimit), // after setup context to enforce limits early
 		wasmkeeper.NewCountTXDecorator(options.TXCounterStoreService),
 		wasmkeeper.NewGasRegisterDecorator(options.WasmKeeper.GetGasRegister()),
--- a/inference-chain/x/inference/keeper/CONCURRENCY.md
+++ b/inference-chain/x/inference/keeper/CONCURRENCY.md
@@ -1,0 +1,20 @@
+# Epoch group data concurrency
+
+The keeper’s epoch group cache and tx-scoped draft are designed to support **concurrent execution** of transactions. This is required for **Cosmos optimistic execution mode**, where multiple transactions can run in parallel before being committed.
+
+The concurrency tests in `epoch_group_data_concurrent_test.go` check that:
+
+1. **Multiple parallel reads** – Many read-only “transactions” see the same data and do not block each other.
+2. **One writer, many readers** – After a writer commits, all readers see the written value (block cache is updated correctly).
+3. **One writer, multiple Set/Get, many readers** – No deadlocks when a writer does several Set/Get in one tx while readers run in parallel (reentrant draft lock).
+4. **Multiple writers and readers** – Writes are serialized at commit time, and the final read sees the last committed value.
+
+## Running the tests
+
+Run the concurrency tests as usual:
+
+```bash
+go test ./x/inference/keeper/ -run EpochGroupDataConcurrent -v
+```
+
+**Note on `-race`:** The tests intentionally share a single `sdk.Context` across goroutines to stress the keeper’s locking. The Cosmos SDK store and gas meter are not safe for concurrent use from multiple goroutines. With `go test -race`, the race detector may report races in the SDK layer (e.g. `cosmossdk.io/store/types.(*infiniteGasMeter).ConsumeGas`). That is a limitation of the test harness, not of the keeper’s epoch group logic. In production, each transaction gets its own context, and the keeper’s cache and draft locking are what allow safe concurrent access to epoch group data under optimistic execution.
--- a/inference-chain/x/inference/keeper/epoch_group_data.go
+++ b/inference-chain/x/inference/keeper/epoch_group_data.go
@@ -11,6 +11,7 @@ import (
 
 	"cosmossdk.io/collections"
 	"github.com/cosmos/gogoproto/proto"
+	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/productscience/inference/x/inference/types"
 )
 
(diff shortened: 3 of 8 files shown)


inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -269,6 +269,11 @@ func (k msgServer) handleInferenceCompleted(ctx sdk.Context, existingInference *
 		return err
 	}
 
+	// For TrafficBasis
+	// we don't use here precise (already incremented) number of requests because it's not yet committed to the store,
+	// and can cause undeterministic results.
+	// Moreover we prefer performance and determenism over precision here,
+	// so we just use value from previous committed block.
 	inferenceDetails := types.InferenceValidationDetails{
```
akup, line 279: It should be moved to EndBlocker and use already incremented currentEpochGroup.GroupData.NumberOfRequests

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,415 @@
...
+			draft.lockWrite()
+			delete(draft.m, key)
+			// Do not unlock here; released in PostHandler
+		}
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.inited && (epochIndex == c.current || epochIndex == c.previous) {
+		delete(c.m, key)
+	}
```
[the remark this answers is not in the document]
akup: Fixed, now in draft we mark what keys was deleted and commit it to block cache when tx completes for DeliverTx and ignore commit for CheckTx

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,415 @@
...
+			val.NumberOfRequests += delta
+			_ = k.EpochGroupDataMap.Set(ctx, collections.Join(rootKey.Epoch, rootKey.ModelId), val)
+		}
+	}
+
+	// We persist here all other cached EpochGroupData for the current epoch
+	for key, val := range c.m {
+		if key.Epoch == c.current && key != rootKey {
+			_ = k.EpochGroupDataMap.Set(ctx, collections.Join(key.Epoch, key.ModelId), val)
+		}
```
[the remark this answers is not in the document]
akup: fixed

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,415 @@
...
+func (k Keeper) IncrementCurrentEpochGroupRequestCount(ctx context.Context) {
+	c := k.epochGroupCache
+	k.ensureEpochGroupCacheInited(ctx)
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if !c.inited {
+		return
+	}
+	c.currentEpochRequestCount.Add(1)
+	c.currentDirty = true
```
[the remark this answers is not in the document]
akup: fixed

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,415 @@
...
 	ctx context.Context,
 	epochIndex uint64,
 	modelId string,
 ) (val types.EpochGroupData, found bool) {
-	val, err := k.EpochGroupDataMap.Get(ctx, collections.Join(epochIndex, modelId))
+	k.ensureEpochGroupCacheInited(ctx)
+	c := k.epochGroupCache
+	key := epochGroupCacheKey{Epoch: epochIndex, ModelId: modelId}
+
+	draft := getEpochGroupDraftFromContext(ctx)
```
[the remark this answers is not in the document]
akup: The issue here that CheckTx should use last committed block state and do not take into account any txs that was committed during this block before block ends.

Fixed. Now we do not use block cache for CheckTx and use only draft that is tx context binded and doesn't conflict with other txs and block state.
If it is DeliverTx we use the block cache as
1. DeliverTx are sequential and do not conflict with each other
2. For optimistic we use RW locks to lock any reads while tx that writes ends

inference-chain/x/inference/keeper/msg_server_finish_inference.go
```diff
@@ -269,6 +269,11 @@ func (k msgServer) handleInferenceCompleted(ctx sdk.Context, existingInference *
 		return err
 	}
 
+	// For TrafficBasis
+	// we don't use here precise (already incremented) number of requests because it's not yet committed to the store,
+	// and can cause undeterministic results.
+	// Moreover we prefer performance and determenism over precision here,
```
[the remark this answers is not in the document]
akup: fixed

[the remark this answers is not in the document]
akup: fixed

inference-chain/app/ante.go
```diff
@@ -25,6 +25,43 @@ import (
...
+		return newCtx, err
+	}
+	if d.InferenceKeeper == nil {
+		return newCtx, nil
+	}
+	if success {
+		d.InferenceKeeper.CommitEpochGroupDraftFromContext(newCtx)
+	} else {
+		d.InferenceKeeper.ReleaseEpochGroupDraftFromContext(newCtx)
+	}
```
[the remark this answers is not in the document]
akup: fixed. For CheckTx/simulate we never commit transaction

inference-chain/x/inference/module/module.go
```diff
@@ -313,6 +315,8 @@ func (am AppModule) EndBlock(ctx context.Context) error {
 	blockHeight := sdkCtx.BlockHeight()
 	blockTime := sdkCtx.BlockTime().Unix()
 
+	// Persist epoch group block cache to store (tx drafts were merged to cache in PostHandler)
+	am.keeper.FlushCurrentEpochGroupCache(ctx)
```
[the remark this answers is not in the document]
akup: fixed

inference-chain/x/inference/keeper/CONCURRENCY.md
```diff
@@ -0,0 +1,20 @@
...
+2. **One writer, many readers** – After a writer commits, all readers see the written value (block cache is updated correctly).
+3. **One writer, multiple Set/Get, many readers** – No deadlocks when a writer does several Set/Get in one tx while readers run in parallel (reentrant draft lock).
+4. **Multiple writers and readers** – Writes are serialized at commit time, and the final read sees the last committed value.
+
+## Running the tests
+
+Run the concurrency tests as usual:
+
+```bash
+go test ./x/inference/keeper/ -run EpochGroupDataConcurrent -v
```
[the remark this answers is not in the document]
akup: test name is valid

akup:
Added fixes for copilot review.

The real issue was that I've didn't took into account CheckTx/simulate, that shouldn't use block cache, and shouldn't lock anything as they never commit changes.
Now CheckTx/simulate always are using last committed block state and create a draft cache that is tx binded and never conflicts with outer state.

Also 0.2.11 introduces use of CacheContext, it wasn't an issue as EpochGroup never is set inside CacheContext and the read was safe, but I've added special handling for it with CacheContextWithEpochGroupBranch that will handle EpochGroupData caches. Also added tests for it.

There was a possible issue that IncrementCurrentEpochGroupRequestCount could be called inside uncommitted CacheContext, so I moved it after all errors could happen and are handled and it will never be situation when CacheContext is not written but RequestCount atomic integer is already incremented

inference-chain/x/inference/module/module.go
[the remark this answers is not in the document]
Mayveskii: > Flagging a merge-order dependency relevant to the v0.2.11 milestone.
>
> PR #703 (free inference security fix, same author) explicitly states it should wait for per-block gRPC cache before merging — this PR is that cache. Without it, isAddressActiveParticipantInCurrentEpoch adds a full gRPC round-trip on every call to handleExecutorRequest, including paths where zero GPU work is done (e.g. cache hits from #859). The performance benefit of serving from cache is negated if participant validation still hits the chain synchronously.
>
> Suggested merge sequence for v0.2.11: #793 → #703, then #859 can operate with correct latency characteristics on the hot path.
>
> Beyond performance: the per-block epoch cache pattern introduced here is worth considering as infrastructure for any operator-side quality signal (#860 direction) — including nodes not running k8s. A lightweight Go ticker or per-block hook reading /admin/v1/cache/stats can plug into this same cache invalidation boundary without requiring container orchestration. That makes quality reporting reproducible on bare-metal and Docker Compose deployments, not just k8s, which directly addresses the open question from PR #859 discussion.

qdanik:
@akup please, keep the PR title clear, it must be informative

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,542 @@
...
+		return nil
+	}
+}
+
+// CommitEpochGroupDraftFromContext merges the draft from ctx into the block cache. Call from PostHandler on tx success.
+// Applies draft writes and draft removals (tombstones) to the block cache; does not write to store (FlushCurrentEpochGroupCache in EndBlock persists). Releases the block cache draft write lock if held.
+func (k Keeper) CommitEpochGroupDraftFromContext(ctx context.Context) {
+	draft := getEpochGroupDraftFromContext(ctx)
+	if draft == nil {
+		return
```
x0152, line 220: I think releaseDraftWriteLock() should be called before the early return here, otherwise this path can deadlock when writeCache() is not called (with COSMOS_OPTIMISTIC_CACHES=1 only)

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,542 @@
...
+				return types.EpochGroupData{}, false
+			}
+			if cached, ok := draft.m[key]; ok {
+				cloned := proto.Clone(&cached).(*types.EpochGroupData)
+				return *cloned, true
+			}
+		}
+	}
 
+	c.draftWriterMu.RLock()
```
x0152, line 479: Looks like a race condition - reads are guarded by c.draftWriterMu, but writes use c.mu (e.g. lines 493 and 425)

inference-chain/x/inference/keeper/CONCURRENCY.md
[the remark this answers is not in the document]
x0152: '-run' matches test names, not file names. 
'go test ./x/inference/keeper/ -run EpochGroupDataConcurrent -v' - this command runs zero tests for me

inference-chain/x/inference/keeper/epoch_group_data.go
akup: First of all, I will move this PR in a refactored feature, to be universal mechanism easily wrapped around any stored value. And it shouldn't go to 0.2.11 !

But according to your comment: here is the 2 locks used.
It is designed that writes use c.mu, but everytime it is unlocked soon (498 releases lock from line 493, 426 defer releases lock from 425)
So c.mu and c.draftWriterMu do not effect each other and there is no race condition. Here is RLock (read lock is used, that is unlocked soon (at lines 482 or 487).
Read lock is used for draft tx binded cache only (and it is explained in the comment to this PR). It has write locking, and read locking. Read locking is always released near. But write locking acts in another way `lockDraftWrite`: it is locked during all the transaction while it is not finished.

Logic is the next:
1.
First of all cosmos is running currently at sequencial mode, so there is no parallel transactions, all are sequencial. But it is bad for performance since we do not utilize multi-CPU cores. It is possible to switch to cosmos optimistic mode where transactions run in parallel. But if any two parallel transactions touch same store object only one transaction will finish, other will be reverted.
For current mode all this read/write black-magick isn't needed as there will never be different transactions running in parallel causing race.
But for further development read/write lock is implemented and proposed

2.
When any tx writes to object, it lock it till the end of transaction. It holds write lock. If any other parallel process (only possible for optimistic mode) will try to read it (`c.draftWriterMu.RLock()` you are pointing at) it will wait, while write transaction will complete. When it completes, the 479 line lock is released and other parallel transaction will read already committed value.

So here is no race condition

inference-chain/x/inference/keeper/epoch_group_data.go
akup: If the draft is nil, it means that it wasn't run in transaction. and never was locked

x0152: deadlocks in the branch path when writeCache() is not called

```go
// COSMOS_OPTIMISTIC_CACHES=1 go test ./x/inference/keeper/ -run TestWriteWithoutMerge_Deadlock -timeout 10s -v
func TestWriteWithoutMerge_Deadlock(t *testing.T) {
	k, sdkCtx := keepertest.InferenceKeeper(t)
	require.NoError(t, k.SetEffectiveEpochIndex(sdkCtx, cacheBranchTestEpoch))
	seedEpochGroupData(k, sdkCtx, cacheBranchTestEpoch, 100)

	txCtx := txCtxWithDraft(sdkCtx)
	cacheCtx, _ := k.CacheContextWithEpochGroupBranch(txCtx)

	k.SetEpochGroupData(cacheCtx, types.EpochGroupData{
		EpochIndex:           cacheBranchTestEpoch,
		ModelId:              "",
		NumberOfRequests:     999,
		MemberSeedSignatures: []*types.SeedSignature{},
	})

	k.CommitEpochGroupDraftFromContext(txCtx)

	done := make(chan struct{})
	go func() {
		_, _ = k.GetEpochGroupData(sdkCtx, cacheBranchTestEpoch, "")
		close(done)
	}()

	select {
	case <-done:
	case <-time.After(3 * time.Second):
		t.Fatal("DEADLOCK")
	}
}
```

inference-chain/x/inference/keeper/epoch_group_data.go
x0152: I understand that it is not blocking current runtime, but unsafe for future enablement

I think the rollback in Cosmos only covers the store, not custom in-memory structures like this cache map. So even if a conflicting tx gets retried, the race on c.m would still be there

I checked it with test
```go
// COSMOS_OPTIMISTIC_CACHES=1 go test -race ./x/inference/keeper -run TestWriteWithoutMerge_Deadlock -count=1
func TestWriteWithoutMerge_Deadlock(t *testing.T) {
	k, sdkCtx := keepertest.InferenceKeeper(t)
	require.NoError(t, k.SetEffectiveEpochIndex(sdkCtx, cacheBranchTestEpoch))
	seedEpochGroupData(k, sdkCtx, cacheBranchTestEpoch, 100)

	txCtx := txCtxWithDraft(sdkCtx)
	cacheCtx, _ := k.CacheContextWithEpochGroupBranch(txCtx)

	k.SetEpochGroupData(cacheCtx, types.EpochGroupData{
		EpochIndex:           cacheBranchTestEpoch,
		ModelId:              "",
		NumberOfRequests:     999,
		MemberSeedSignatures: []*types.SeedSignature{},
	})

	k.CommitEpochGroupDraftFromContext(txCtx)

	done := make(chan struct{})
	go func() {
		_, _ = k.GetEpochGroupData(sdkCtx, cacheBranchTestEpoch, "")
		close(done)
	}()

	select {
	case <-done:
	case <-time.After(3 * time.Second):
		t.Fatal("DEADLOCK")
	}
}
```

akup:
It shouldn't be in 0.2.11
It could be refactored to be used for big stored data, like EpochData and Params, and other similar structures, that could be hit multiple times during tx or block

inference-chain/x/inference/keeper/epoch_group_data.go
akup: I think reason is that CacheCtx is bad handled.
I will look into it deeper. Thanks

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,542 @@
...
-// SetEpochGroupData set a specific epochGroupData in the store from its index
+// cosmosOptimisticCachesEnabled is set once at package init from COSMOS_OPTIMISTIC_CACHES=1.
+var cosmosOptimisticCachesEnabled bool
+
+func init() {
+	cosmosOptimisticCachesEnabled = os.Getenv("COSMOS_OPTIMISTIC_CACHES") == "1"
+}
+
+// getGID returns the current goroutine id (for reentrant lock holder identity). Uses runtime.Stack; for use in reentrant locking only.
+func getGID() int64 {
```
jacky6block, line 26: getGID() is called on every completed inference (via SetEpochGroupData → lockDraftWrite), but `runtime.Stack()` allocates heap memory and parses text on every call, which gets expensive under load. 

Since each tx already has a unique lazyEpochGroupDraft pointer allocated in the AnteHandler, that pointer can be used as the tx identity for reentrancy detection instead of the goroutine ID — O(1), zero allocation, no runtime involvement needed.

inference-chain/x/inference/keeper/epoch_group_data.go
```diff
@@ -1,41 +1,542 @@
...
+			c.draftWriterMu.RUnlock()
+			cloned := proto.Clone(&cached).(*types.EpochGroupData)
+			return *cloned, true
+		}
+	}
+	c.draftWriterMu.RUnlock()
+
+	val, err := k.EpochGroupDataMap.Get(ctx, collections.Join(epochIndex, modelId))
 	if err != nil {
 		return val, false
```
jacky6block, line 491: There's a potential race in optimistic mode: a reader can miss in `c.m`, release `draftWriterMu.RLock()`, then another tx commits a fresh value into `c.m` — but the original reader still loads the old value from the KV store and overwrites it under `c.mu.Lock()`. That stale value then gets flushed to store in EndBlock. A simple fix is to re-check `c.m` inside the `c.mu.Lock()` section before writing the store-loaded value, and skip the write if a newer entry is already there.

akup: @jacky6block It is more precise to define the issue not as race condition, but as a root of undetermenism in optimistic behavior.

I rethought about this PR and completely refactored it to be universal, and to do not use locks. Now I'm tracing conflics (parallel writes or read with parallel write) to rollback. Let me explain why there shouldn't be all this locks at all.

The main thing here is that optimistic feature in Cosmos is not universal feature and can have many implementations that are very case specific.

I think that Gonka can use following scheme:
Scheduler takes N (number of CPUs) messages from mempool to process, they should be sorted and grouped according similarity. As similar messages has similar time of execution. Then we run all this selected N in a parallel batch and wait while all messages in a batch are processed. Then we take next ones.

If there was conflict: some messages touched same store key (inner map key) then we rollback all messages that has reads conflicted with write, and leave only one write message. We just reschedule them to the next batch. When order and batch fill rules are deterministic all this parallelism also keeps deterministic behavior.

So we don't need all this locks/unlocks we just need to trace conflicts of parallel writes. and schedule messages. And we need to implement this on our own

Implementing this optimistic feature could give N*k times better throughput of mainchain (where N - number of CPU cores, k - some efficiency coefficient < 1.0)

@x0152 btw, tryed your test and coudn't get into the deadlock. Any way thanks a lot now it is free from locks!

I will publish proposal and this PR in a next 12h. It is planned to have 2 phases - first refactored this PR that helps to cache during block and transactions to remove marshalling/unmarshalling (that are used also for gas cost calculations), and the 2nd phas add optimistic mode, that can add significant throughput (comparable to number of cores) for the network

jacky6block: Nice refactor. I think what you're describing sounds essentially like OCC (Optimistic Concurrency Control): parallel execution, conflict detection at commit time, rollback and reschedule on conflict. It's a well-proven model in database systems, good choice.

A few things I'm wondering about for the OCC scheme:

1. Conflict resolution ordering — when two txs in the same batch both write the same key, which one wins? Assuming it has to be block order (earlier tx survives), otherwise different validators could make different choices. Worth stating this explicitly in the proposal.
2. "Grouped by similarity" — what's the actual criterion here? If it's by predicted access pattern, that implies some static analysis of txs before execution. How do you handle arbitrary message types?
3. Gas for rescheduled txs — when a tx gets rolled back and moved to the next batch, does its gas meter reset? Want to understand how this fits with Cosmos's gas model.
4. High contention liveness — if a tx keeps conflicting with a hot key, it could get rescheduled indefinitely. Any fallback planned (e.g., force sequential after N retries)?

akup: @jacky6block @x0152 please take a look:
https://github.com/gonka-ai/gonka/pull/953

And the proposal:
https://github.com/gonka-ai/gonka/discussions/954

PR #794, opened 2026-02-24
Ak issue783 0.2.11

### LINKED ISSUE #783: [4/4] `StartInference` and `FinishInference`
# Background

Download profiling file:
And use command:
```
```

- 10% of `FinishInference`, 
- 12% of `StartInference`, 
- 14% of Validation
- 33% is Logging, 

- 20% is Logging
- 45% is 2xGetEpochGroupData
- 5% GetEpochIndex
- 10% SetEpochGroupData, 

- 63% is Logging
- 9% Add/GetTokenomicsData

# Tasks:
`SetParticipant` is executed twice in `ProcessInferencePayment` and `HandleInferenceComplete` for the second Start/Finish transaction, when it could be executed just once.

Most of the time spent in `SetParticipant` is consumed by `ComputeStatus` and `GetParams` (except for Logging, which takes 50%, which we discussed separately in https://github.com/gonka-ai/gonka/issues/780). `Decimal.Ln` in `ComputeStatus` can be optimized significantly. Regarding `GetParams`, we read it for each transaction in any case, so we can pass it to `SetParticipant` and reuse what we already have.

# Important
This specific task does not require maintaining and operating a node on mainnet in order to test and validate the result.

akup:
[Issue 783](https://github.com/gonka-ai/gonka/issues/783)

Implemented caches for Ln calculations for ComputeStats. As we are using values from params that do not change often we could calculate Decimal.Ln once and cache it.

Added reason parameter to SetParticipant to know where it was called from to shorten `ComputeStats`:
- SetParticipantReasonCompletedInference => CheckInactive
- SetParticipantReasonMissedInference => CheckInactive
- SetParticipantReasonInvalidation => CheckConsecutiveFailures | CheckInvalidation
- SetParticipantReasonConfirmationPoC => CheckConfirmationPoC

Now we don't run all possible status computations on each participant set, but only related to update operation.

Added 2 modes for GetParams caching:
- tx caching using AnteDecorator and PostHandler (enabled when env var INFERENCE_PARAMS_CACHE_MODE = 'tx')
- per block params caching (by default)

### DIFF
--- a/decentralized-api/go.mod
+++ b/decentralized-api/go.mod
@@ -82,8 +82,9 @@ require (
 	github.com/bgentry/speakeasy v0.2.0 // indirect
 	github.com/bits-and-blooms/bitset v1.22.0 // indirect
 	github.com/blang/semver/v4 v4.0.0 // indirect
-	github.com/bytedance/sonic v1.14.0 // indirect
-	github.com/bytedance/sonic/loader v0.3.0 // indirect
+	github.com/bytedance/gopkg v0.1.3 // indirect
+	github.com/bytedance/sonic v1.15.0 // indirect
+	github.com/bytedance/sonic/loader v0.5.0 // indirect
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
 	github.com/cenkalti/backoff/v4 v4.3.0 // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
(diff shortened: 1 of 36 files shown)


akup:
It is already implemented in other way at https://github.com/gonka-ai/gonka/pull/812

While it has useful optimization feature, StatusCheckScope for ComputeStatus and types.SetParticipantReason for UpdateParticipantStatus, that could be a tiny future PR

PR #812, opened 2026-02-27
StartInference and FinishInference performance improvements

### LINKED ISSUE #781: [2/4] `StartInference` and `FinishInference`

libermans:
# Start/FinishInference performance improvements

`MsgStartInference` and `MsgFinishInference` take 10-50ms (and sometimes upwards of 50ms) to execute in production, which severely restricts network throughput and threatens block time stability. The heavy processing is largely caused by:
*   Heavy on-chain querying and recording of developer stats.
*   Duplicate participant and inference state writes across the completion flow.
*   Heavy `Decimal.Ln` (logarithmic) operations inside participant status (`ComputeStatus`) calculations.
*   Expensive epoch-group reads and `InferenceValidationDetails` generation on the transaction hot-path.

Profiling data collected from production mainnet nodes (`pprof.inferenced.samples.cpu.001.pb.gz`) identified stats queries/recording and signature validation as the primary time contributors. Specifically, on-chain stats operations account for ~40% of the execution time in `FinishInference` and ~30% in `StartInference`. The `proposals/inference-performance/README.md` and subsequent markdown documents thoroughly outline the tracing results confirming these bottlenecks.

This PR completely overhauls the hot-paths for inference recording via several targeted architectural shifts:
*   **Task II (Off-Chain Stats Migration):** Removes the heavy `SetDeveloperStats` writes from the on-chain `SetInference` hot-path. Instead, the `inference_finished` event is enriched with a full payload of details, allowing the Decentralized API Node to listen and store developer stats entirely off-chain (using Postgres or File storage).
*   **Task III (Deferred Processing):** Defers the expensive construction of `InferenceValidationDetails` and the incrementing of `EpochGroupData.NumberOfRequests` to the `EndBlocker`. It uses the fast, in-memory `transient_store` to enqueue completed inference IDs during the block to be processed in batch.
*   **Task IV (Participant & SPRT Optimization):** Consolidates duplicate `SetParticipant` and `SetInference` writes into a single final write per transaction. Optimizes the status math by precomputing expensive SPRT log-likelihood ratios (`Decimal.Ln`) when parameters change and caching them in the `transient_store`. Introduces a transactional `GetParams` context cache to prevent redundant state reads.
*   **Task VII (Rolling Windows):** Replaces the previous dynamic pricing and invalidation limit algorithms (which depended on the now-removed on-chain developer stats) with highly efficient `O(1)` on-chain rolling windows for model token loads and inference counts.

*   **Risk:** The experimental file-based stats storage can cause inode exhaustion, performance degradation, and crashes on high-volume nodes.
    *   **Mitigation:** PostgreSQL is the strongly recommended and default production backend. File storage is explicitly flagged as experimental and is disabled by default (requires `DAPI_STATS_FILE_STORAGE_ENABLED=true`).
*   **Risk:** Off-chain API nodes dropping inference statuses if an inference state is invalidated or revalidated after completion.
    *   **Mitigation:** Introduced a new `inference_status_updated` event emitted on `MsgValidation`, `MsgInvalidateInference`, and `MsgRevalidateInference` so API nodes can keep their off-chain databases cleanly synchronized.
*   **Risk:** Failed transactions leaving orphaned data in the validation queue.
    *   **Mitigation:** The `transient_store` queue is strictly block-scoped. It automatically rolls back on failed transactions and is wiped clean at the end of the block, preventing state bloat.

Extensive benchmarks and tests were added alongside these changes:
*   `BenchmarkComputeStatus` micro-benchmarks prove the mathematical optimizations drastically reduce CPU cycles. 
*   `TestEndBlock_ProcessesPendingInferenceValidationQueue` and related tests prove the deferred EndBlocker logic correctly batches and persists validation details. 
*   E2E API tests (`StatsApiE2ETest`) prove that off-chain stats collection and querying behave identically to the legacy on-chain logic. 
*(Note: Final production timing verification on a live node is still flagged as a pending TODO in the proposal docs, but the architectural removal of the identified bottlenecks guarantees a massive reduction in I/O and CPU).*

---

### Other Notes
*   **Transient Store Initialization:** This upgrade requires validators to initialize the new transient store keys (`TransientStoreKey`). The upgrade handlers (`upgrades_test.go`, etc.) have been updated to mount this store type properly.
*   **API Node Postgres Requirements:** The decentralized API node now expects a `PGHOST` environment variable to initialize the PostgreSQL off-chain stats backend. A `DisabledStorage` fallback is gracefully provided if not configured, which will disable the stats endpoints.
*   **New Stats Endpoints:** Several new endpoints (`/v1/stats/*`) were added to the public server to expose the newly off-chain developer stats for dashboards. The legacy `getGovernanceModelsLegacy` is retained temporarily for backwards compatibility.
*   **Managed Pruning:** The new off-chain storage automatically handles retention and pruning of old inference records. By default, it runs a background loop holding 30 days of data (`DAPI_STATS_RETENTION_DAYS`).


inference-chain/x/inference/keeper/rolling_window_state.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	oldest := s.Values[0]
+	if oldest > s.Sum {
+		s.Sum = 0
+	} else {
+		s.Sum -= oldest
+	}
+
+	copy(s.Values, s.Values[1:])
```
patimen, line 234: My algo senses see O(n) where we could have O(1) (ring buffer, or similar), but I don't see the length of the rolling window ever getting higher than a few dozen, right?

inference-chain/x/inference/module/inference_validation_endblock.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func (am AppModule) processFinishedInferencesInBlock(
+	ctx context.Context,
+	blockHeight int64,
+	effectiveEpoch *types.Epoch,
+	currentEpochGroup *epochgroup.EpochGroup,
+	params *types.Params,
+) {
+	pendingInferenceIDs := am.keeper.ListFinishedInferenceIDsForHeight(ctx, blockHeight)
```
patimen: This loads all inferenceids into memory... we could use Walk instead, but we're only targeting thousands of inferenceids at a time, right?

patimen: Should be safe.

decentralized-api/internal/server/public/stats_handlers.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	debugStats, err := s.statsStorage.GetDebugStats(c.Request().Context())
+	if err != nil {
+		return err
+	}
+	return c.JSON(http.StatusOK, mapDebugStats(debugStats))
+}
+
+func parseStatsTimeRange(timeFromStr, timeToStr string) (int64, int64, error) {
+	now := time.Now().Unix()
```
x0152: On-chain timestamps are in ms, but dapi default time windows are in seconds, so default queries return zero records

inference-chain/x/inference/keeper/msg_server_start_inference.go
```diff
@@ -79,22 +83,34 @@ func (k msgServer) StartInference(goCtx context.Context, msg *types.MsgStartInfe
...
+	if shouldPersistParticipant(finalInference, payments, executor) {
+		if err := k.SetParticipant(ctx, *executor); err != nil {
 			return failedStart(ctx, err, msg), nil
 		}
 	}
+	err = k.SetInference(ctx, *finalInference)
+	if err != nil {
+		return failedStart(ctx, err, msg), nil
+	}
+	k.addTimeout(ctx, finalInference)
```
x0152, line 118: Should this block use CacheContext? Funds may be moved before inference persistence fails, leading to partial committed state (same in FinishInference)

decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -430,12 +446,185 @@ func (e *InferenceFinishedEventHandler) Handle(event *chainevents.JSONRPCRespons
...
+		return nil
+	}
+	for _, rec := range records {
+		err := el.statsStorage.UpdateInferenceStatus(context.Background(), rec.InferenceID, rec.Status)
+		if err != nil {
+			if errors.Is(err, statsstorage.ErrInferenceRecordNotFound) {
+				logging.Warn("Ignoring inference_status_updated for unknown inference in stats storage", types.EventProcessing,
+					"inference_id", rec.InferenceID, "status", rec.Status)
+				continue
+			}
```
x0152, line 611: Theoretically: with 10 parallel workers, inference_status_updated can be processed before inference_finished, so the status update is silently dropped

inference-chain/x/inference/keeper/params.go
```diff
@@ -13,24 +14,40 @@ const defaultGenesisGuardianNetworkMaturityMinHeight int64 = 0
...
+}
+
+// InjectParamsIntoContext returns a new context with the params cached.
+func (k Keeper) InjectParamsIntoContext(ctx sdk.Context) (sdk.Context, error) {
+	params, err := k.getParamsFromStore(ctx)
+	if err != nil {
+		return ctx, err
+	}
+	return ctx.WithValue(paramsKey{}, &params), nil
+}
```
akup, line 46: It is tx binded context. I still think if calling GetParams twice in one tx was heavy and was the performance problem, reading it from store for each tx is useless and inefficient, as params are set very rarely, and having them same during the block (read them once) is very determenistic and can't cause any issues.

inference-chain/x/inference/keeper/msg_server_start_inference.go
akup: It is better to fully rebase https://github.com/gonka-ai/gonka/pull/789 that added CachedContext very accurately after this pr, as there are a lot of conflicting changes

decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -430,12 +446,200 @@ func (e *InferenceFinishedEventHandler) Handle(event *chainevents.JSONRPCRespons
...
+
+func getEventValue(events map[string][]string, key string, idx int) (string, bool) {
+	values := events[key]
+	if len(values) == 0 {
+		return "", false
+	}
+	if idx < len(values) {
+		return values[idx], true
+	}
+	return values[0], true
```
patimen, line 534: This is an interesting choice. If the index is passed in and it's higher than the list of values, we return the first value? Are we sure it wouldn't make more sense to return "", false?

inference-chain/x/inference/keeper/params.go
patimen: I didn't want a naive cache that would cause an edge case when we updateparams, which can happen in the middle of a block.
I think the right solution would be to add a ParamsVersion collections.Item[int64] that we increment when UpdateParams happens.
That gives us the key we need to do a safe in memory hash. Otherwise we risk edge cases when UpdateParams is called, or even worse, when it is called and then rolled back.

I'd rather put that in a separate PR.

patimen (approved):
I'm approving this for now. I have some follow up checks I'll need to make, but we need these perf improvements in testnet merged with some others ASAP. For the most part, I have confirmed or improved nearly all of the PR. What's left is mostly test issues, as some integration tests appear to have new issues which I will look into.

inference-chain/x/inference/keeper/params.go
akup: I think it is much simpler and there is no (real) edge cases.

Let me explain:
1. If parameters are updated (they could be updated in the middle of the block ONLY with proposal) we can still use old parameters values inside this block. There is no reason why we need to use half of the block old params, and immediately apply new ones and start to use them second half of the block. It is absolutely safe to keep cached params for the block.
2. So every process that reads params during the block sees same parameters (that was same at the start of the block) and this behavior is completely deterministic without edge cases.
3. The only possible edge case when there are 2 or more parameters sets, and moreover they depend on each other, so each next set needs to read the previous value. Even if this multiple proposals change different parameters and there is no dependency between them, it's safe to apply them.
This scenario is absolutely impossible as there should be multiple proposals that should be approved one by one and depend on previous parameter values in same block.

So it still is very safe and deterministic to use per-block cache for parameters.

Using `InjectParamsIntoContext` is something that should be handled in each case that should be accurately found, and is not universal, while per-block cache is

I don't insist on adding this change to the PR, but just trying to explain why it shouldn't be overingeneered.

decentralized-api/internal/server/public/stats_handlers.go
patimen: Good find, thanks.

inference-chain/x/inference/keeper/params.go
patimen: I might be overthinking this a bit, but I want to make sure we don’t accidentally introduce a footgun.

The main issue isn’t determinism — a per-block cache would still be deterministic — it’s semantics. In Cosmos, the usual expectation is that state writes become visible immediately to subsequent messages.

Governance proposals can contain multiple messages, and they sometimes intentionally chain actions. For example:

1. UpdateParams
2. Execute some action that depends on those params

Today the second message would see the updated parameters. With a BeginBlock cache it would still see the old ones from the start of the block.

That may be rare, but it’s a real scenario. My goal with the `ParamsVersion` approach was just to preserve the normal read-after-write behavior while still avoiding repeated decoding. It keeps params behaving like the rest of state instead of introducing a special rule.

From my perspective it’s a small amount of extra machinery for a bit more semantic safety.

I'd also point out that there are around 100 unit tests that rely on SetParams taking effect immediately that would have to be refactored.

Is there something about the version-based approach that feels overly complicated?

Issue #797, opened 2026-02-24
New nodes can't join from snapshots with error


tcharchian:
**This is an urgent, open issue, and many contributors are working on it in parallel.**

There is a quite weird issue today - new nodes can't join from snapshots with error like that:
```
5:49AM ERR error in proxyAppConn.FinalizeBlock err="no validator signing info found" module=consensus
```
[no-validator-signing.log](https://github.com/user-attachments/files/25529057/no-validator-signing.log)

Seems like there is no signing info in snapshot. But that issue happens with different set of peers for state sync (pex=false) and seems reproducable in most cases 

Experiment was done to ignore this issue in cosmos-sdk and them it fails like the state doesn't have slashing params also 

Seems like active nodes had all this data at the state at this height 

How to reproduce it - to start new node, it is likely you get it (it passes with some snapshots, which makes everything even stranger)

main hypothesis that something is with producing snapshots, it's cosmo-sdk level

There is also a hypothesis that it is somehow connected to slashing, as it started to happen when collateral was activated. But we are not sure.

it happens the same block the validator list usually updated.

Also, collateral needs to be checked.

hleb-albau:
As a workaround(before fix), it is possible to just compress data folder(except some filers) and distribute it as is archive. Quite popular in cosmos world. See https://snapshots.osmosis.zone/index.html as example

tcharchian:
@hleb-albau thanks, that makes sense and it’s a well-known approach in the Cosmos ecosystem. You are right, it’s more of an operational workaround than a real fix (it doesn’t address the underlying issue we’re trying to solve)

blizko:
Additional feedback:
The issue was observed before collateral slashing was activated. During epoch 179 have been observing same error.
Known failed attempt time around Feb 21st 01:47 UTC

x0152:
The issue is in [cosmos/iavl v1.2.4](https://github.com/cosmos/iavl/tree/v1.2.4)

After snapshot restore, IAVL rebuilds a "fast node" index by iterating the tree. If the iterator hits an error mid-way, it silently stops - the error is never stored ([iterator.go:230-235](https://github.com/cosmos/iavl/blob/v1.2.4/iterator.go#L230-L235)). The fast index ends up incomplete, but IAVL marks it as ready

[...]

Issue #780, opened 2026-02-20
[1/4] `StartInference` and `FinishInference`


tcharchian:
# Background

`MsgStartInference` and `MsgFinishInference` are too slow in production. Blocks should be processed by nodes within 1-2 seconds, so that block time stays below 6 seconds. This means that to process 1000 inferences in a block, we need to record 1000 `MsgStartInference`, 1000 `MsgFinishInference`, and 100-200 `MsgValidation` transactions. This means that these transactions should be processed faster than 1ms. Even though they are quite fast in tests, in production with a large state they require 10-20ms, and on some nodes 50ms or more.

There are 2 main areas identified that contribute most of the time to transactions:
- Signatures validation (57% of `FinishInference` and 63% of `StartInference`)
- Stats query and recording (40% of `FinishInference` and 30% of `StartInference`)

Download profiling file:
https://drive.google.com/file/d/1yxY91lzMHxv_MeloAxW1zczcpbkBjZ0t/
And use command:
```
go tool pprof -http=:8080 /Users/davidliberman/Downloads/pprof.inferenced.samples.cpu.001.pb.gz
```

Screen recoding: https://drive.google.com/file/d/1yxDaJllxCQ-l3ZO6ZuBb5bTEUgZ5t7Yu/view?usp=sharing

And choose flame graph to explore

_**Signature validation**_ can be significantly optimized, reducing the number of signatures to be validated in most scenarios by 5x (from 5 signatures to just 1).

https://github.com/gonka-ai/gonka/issues/608 - which is now implemented by @DimaOrekhovPS

https://github.com/gonka-ai/gonka/pull/779 

**_Stats query and recording_** is designed to make it easier to query usage statistics for inference operations by storing this data on a chain. However, it is too heavy for on-chain operations and should be removed. In the end, we shouldn't read and write any large state record in `MsgStartInference`, `MsgFinishInference`, or `MsgValidation`.

`SetInference` (including the second time it is executed in `HandleInferenceComplete`): 
- 10% of `FinishInference`, 
- 12% of `StartInference`, 
- 14% of Validation
- 33% is Logging, 
- 38% `SetOrUpdateInferenceStatsByEpoch`, 
- 22% `SetOrUpdateInferenceStatusByTime` w/o logging

`HandleInferenceComplete`, excluding `SetInference`, accounts for 16% of `FinishInference` and 4% of `StartInference` (as it is rare for `StartInference` to come second).
- 20% is Logging
[...]

PR #813, opened 2026-02-27
Upgrade v0.2.11

patimen:
# Upgrade Proposal: v0.2.11

This document outlines the proposed changes for on-chain software upgrade v0.2.11.   
The `Changes` section details the major modifications, and the `Upgrade Plan` section describes the process for applying these changes.

## Upgrade Plan

This PR updates the code for the `api` and `node` services. The PR modifies the container versions in `deploy/join/docker-compose.yml`.

The binary versions will be updated via an on-chain upgrade proposal. For more information on the upgrade process, refer to [`/docs/upgrades.md`](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.11/docs/upgrades.md).

Existing hosts are **not** required to upgrade their `api` and `node` containers. The updated container versions are intended for new hosts who join after the on-chain upgrade is complete.

It also updates CosmWasm contract artifacts for the community sale and liquidity pool, adds bridge/testnet operational scripts for IBC trading support, and introduces a new `subnet/` package used by the new inference architecture.

## Proposed Process

2. If the on-chain proposal is approved, this PR will be merged immediately after the upgrade is executed on-chain.

## Testing

The on-chain upgrade from version `v0.2.10` to `v0.2.11` has been successfully deployed and verified on the testnet. No regression in core functionality or performance has been observed during testing. More testing will be executed leading up to the upgrade.

Reviewers are encouraged to request access to testnet environments to validate both node behavior and the on-chain upgrade process, or to replay the upgrade on private testnets.

## Migration

The on-chain migration logic is defined in [`upgrades.go`](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.11/inference-chain/app/upgrades/v0_2_11/upgrades.go).

Migrations:
- Sets `ValidationParams.ClaimValidationEnabled = false`.
- Rebuilds active participant caches for the current and previous epoch.
- Migrates epoch-group validations into the new entry-based format.
- Community-sale CosmWasm contract migration.

### [PR #877](https://github.com/gonka-ai/gonka/pull/877) Inference shards (Experimental)
- Introduces subnet-based inference flow, moving per-inference coordination off-chain.
- The chain now handles only session setup, escrow, and settlement.
- Adds support for subnet state, transport, signing, storage, settlement, and API integration.
- **Note:** This feature is currently experimental and under limited access. For reference design and architecture, see [`proposals/inference/`](https://github.com/gonka-ai/gonka/tree/main/proposals/inference).

### [PR #812](https://github.com/gonka-ai/gonka/pull/812) StartInference and FinishInference performance improvements
- Reduces unnecessary state writes and query overhead for `MsgStartInference` and `MsgFinishInference`.
- Simplifies stats handling and cuts work done during the inference lifecycle for better block execution stability.

### [PR #760](https://github.com/gonka-ai/gonka/pull/760) Unified Permissions
- Consolidates message-permission checks across the inference module.
- Removes duplicated authorization logic to make permission behavior more explicit and testable.

### [PR #779](https://github.com/gonka-ai/gonka/pull/779) Inference msgs optimization: optimize key verification
- Reduces cryptographic verification overhead in the inference message path.
- Avoids repeating signature checks where protocol guarantees make them redundant.

### [PR #874](https://github.com/gonka-ai/gonka/pull/874) MsgValidation and MsgClaimRewards performance optimization
- Reduces hot-path lookups and adds transient caching in validation and reward-claiming paths.
- Restructures validation/reward logic and introduces state pruning support.

### [PR #822](https://github.com/gonka-ai/gonka/pull/822) BLS related fixes based on Certik audit
- Applies Certik audit fixes to the BLS module.
- Fixes threshold-validation and duplicate-slot handling issues for distributed key generation and threshold-signing flows.

### [PR #814](https://github.com/gonka-ai/gonka/pull/814) IBC Trade Support
- Introduces governance-controlled support for trading approved IBC-denominated assets.
- Includes chain message/query changes and contract updates for the community sale and liquidity pool.

### [PR #868](https://github.com/gonka-ai/gonka/pull/868) Required-collateral aware slashing flow
- Bases slashing penalties on required collateral rather than the full deposited amount.
- Makes the slashing model more proportional for participants who over-deposit relative to the minimum.

### [PR #888](https://github.com/gonka-ai/gonka/pull/888) Fix: collateral
- Fixes reward calculation for undercollateralized miners, ensuring actual collateral accurately reduces effective earning power.

### [PR #775](https://github.com/gonka-ai/gonka/pull/775) fix: redirect slashed coins to gov
- Redirects slashed collateral to governance-controlled destinations.

### [to be merged, don't affect mainnet] Final Subnet Params 
<>

### Other changes
- [PR #867](https://github.com/gonka-ai/gonka/pull/867) Fix the application.db bloat issue.
- [PR #835](https://github.com/gonka-ai/gonka/pull/835) Add Batch Transfer With Vesting.
- [PR #773](https://github.com/gonka-ai/gonka/pull/773) feat: delete governance model.
- [PR #675](https://github.com/gonka-ai/gonka/pull/675) security: update CometBFT to v0.38.21 (CSA-2026-001).
- [PR #543](https://github.com/gonka-ai/gonka/pull/543) fix: data race conditions.
- [PR #815](https://github.com/gonka-ai/gonka/pull/815) Update CONTRIBUTING.md.
- [PR #807](https://github.com/gonka-ai/gonka/pull/807) Update issue templates.

## Proposed Bounties

Bounty ID | Sum GNK | Bounty Explanation | GitHub ID
-- | -- | -- | --
PR #543 | 2500 | extra bounty for a comprehensive review of all cases where the data race conditions  fix was needed | @x0152
Issue #628 | 25000 | PoC integration into vllm v0.11.1 [report](https://github.com/axeltec-software/gonka/tree/axeltec/poc-integration/proposals/poc-integration-vllm-v0.11)  |  [Axel-t](https://www.axel-t.com/), @Red-Caesar
-- | 10000 | report of series of prompts resulting in vllm HTTP 502 response, significant impact, was already used for intentoinal greifing | @blizko
-- | 1000 | report of dust transaction vulnerability extending blocks | @blizko
-- | 5000 | report of Remote DoS of Validator PoC Software via dist Assertion | @ouicate
-- | 5000 | report of State Bloat PoC and End-Block DoS via Unbounded Batch / Validation Payloads | @ouicate
-- | 750 | report of Bridge Ethereum Address Parsing Silently Falls Back to Zero Bytes (Loss/Misdirection of Funds) | @ouicate
PR #775 | 1000 | planned task | @x0152
PR #773 | 1250 | planned task | @x0152
[qdanik/vllm/pull/5](https://github.com/qdanik/vllm/pull/5) | 12000 | vLLM 0.15.1 Compatibility Experiments - basis for next ML node version | @qdanik
[qdanik/vllm/pull/6](https://github.com/qdanik/vllm/pull/6) | 15000 | vLLM 0.15.1 Compatibility Experiments - basis for next ML node version. covering simultanious PoC and inference | @qdanik
-- | 5000 | report of wind down window vulnerability fixed in PR #767 | @qdanik
Issue #797 | 1000 | collective solving of nodes unable to join from snapshots - proposed valuable hypothesis | @akup
Issue #797 | 3000 | collective solving of nodes unable to join from snapshots - found source problem | @x0152
Issue #780 | 750 | collective solving StartInference and FinishInference issue | @hleb-albau
Issue #781 | 5000 | collective solving StartInference and FinishInference issue | @x0152
Issue #782 | 5000 | collective solving StartInference and FinishInference issue | @akup
PR #867 | 7500 | important issue that affected many participants, not a vulnerability, fairly easy fir; adding extra payment for fully testing and providing results of the test together with the fix | @Lelouch33
Issue #730 | 22500 | vLLM 0.15.1 Compatibility Experiments - basis for next ML node version | @clanster, @baychak
PR #835 | 5000 | Batch Transfer With Vesting implementation, huge kudos for figuring out how to use testnet | @huxuxuya
PR #868 | 5000 | collateral slashing vulnerability and fix; low severity: low risk, medium likelyhood, organic | @qdanik
v0.2.11 | 7500 | release management | @akup
v0.2.11 | 7500 | release management | @x0152
v0.2.10 | 2500 | upgrade review | @yapion
v0.2.10 | 2500 | upgrade review | @blizko
v0.2.10 | 2500 | upgrade review | @x0152


decentralized-api/statsstorage/postgres_storage.go
```diff
@@ -0,0 +1,414 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+const createInferenceStatsTableSQL = `
+CREATE TABLE IF NOT EXISTS inference_stats (
+    inference_id TEXT PRIMARY KEY,
```
blizko, line 12: Full text indexation is resource intensive. Consider using fixed length primary key

decentralized-api/statsstorage/postgres_storage.go
```diff
@@ -0,0 +1,414 @@
...
+    model TEXT NOT NULL,
+    status TEXT NOT NULL,
+    epoch_id BIGINT NOT NULL,
+    prompt_token_count BIGINT NOT NULL,
+    completion_token_count BIGINT NOT NULL,
+    total_token_count BIGINT NOT NULL,
+    actual_cost_in_coins BIGINT NOT NULL,
+    start_block_timestamp BIGINT NOT NULL,
+    end_block_timestamp BIGINT NOT NULL,
+    inference_timestamp BIGINT NOT NULL,
```
blizko, line 23: Why `inference_timestamp` and `start_block_timestamp` are selected as BIGINT, whilst `created_at` and `updated_at` are TIMESTAMP?

decentralized-api/statsstorage/postgres_storage.go
```diff
@@ -0,0 +1,414 @@
...
+    updated_at TIMESTAMP DEFAULT NOW()
+);
+
+CREATE INDEX IF NOT EXISTS inference_stats_requested_by_time_idx
+    ON inference_stats (requested_by, inference_timestamp);
+
+CREATE INDEX IF NOT EXISTS inference_stats_epoch_idx
+    ON inference_stats (epoch_id);
+
+CREATE INDEX IF NOT EXISTS inference_stats_model_time_idx
```
blizko, line 34: Combined index on multiple fields will cause the index to grow significantly. There is no real use for it at this point.

decentralized-api/statsstorage/postgres_storage.go
```diff
@@ -0,0 +1,414 @@
...
+  AND epoch_id <= bounds.max_epoch
+`
+	return s.scanSummaryRow(ctx, q, epochsN)
+}
+
+func (s *PostgresStorage) GetSummaryByTimePeriod(ctx context.Context, timeFrom, timeTo UnixMillis) (Summary, error) {
+	const q = `
+SELECT
+    COALESCE(SUM(total_token_count), 0) AS ai_tokens,
+    COALESCE(COUNT(*), 0) AS inferences,
```
blizko, line 232: Use `count(inference_id)` instead, that will cause to use less fields

decentralized-api/statsstorage/postgres_storage.go
```diff
@@ -0,0 +1,414 @@
...
+`
+	return s.scanSummaryRow(ctx, q, timeFrom, timeTo)
+}
+
+func (s *PostgresStorage) GetModelStatsByTime(ctx context.Context, timeFrom, timeTo UnixMillis) ([]ModelSummary, error) {
+	const q = `
+SELECT
+    model,
+    COALESCE(SUM(total_token_count), 0) AS ai_tokens,
+    COALESCE(COUNT(*), 0) AS inferences
```
blizko, line 246: Use `count(inference_id)` instead, that will cause to use less fields

decentralized-api/statsstorage/postgres_storage.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+	return res, nil
+}
+
+func (s *PostgresStorage) getDebugStatsByEpoch(ctx context.Context) ([]DeveloperEpochStats, error) {
+	const q = `
+SELECT
+    requested_by,
+    epoch_id,
+    COALESCE(array_agg(inference_id ORDER BY inference_id), '{}') AS inference_ids
```
blizko, line 369: This is avoidance of database normalisation form. Strongly discouraged

inference-chain/app/upgrades/v0_2_11/upgrades.go
```diff
@@ -0,0 +1,83 @@
...
+		k.LogError("failed to get params during upgrade", types.Upgrades, "error", err)
+		return err
+	}
+
+	if params.EpochParams == nil {
+		k.LogError("epoch params not initialized", types.Upgrades)
+		return errors.New("EpochParams are nil")
+	}
+
+	params.EpochParams.ConfirmationPocSafetyWindow = 50
```
blizko, line 268: This is below 5 minutes.
Q. Worth to raise to 1 hour, considering that backups are enabled by default?

decentralized-api/statsstorage/factory.go
```diff
@@ -0,0 +1,57 @@
...
+		if err == nil {
+			logging.Info("Using PostgreSQL stats storage", types.System, "host", pgHost, "retention_days", retentionDays)
+			return NewManagedStorage(pgStorage, retentionDays), nil
+		}
+		// Fail. If they want to be running PostgreSQL and init fails, better to stop and allow them to fix the feature than continue
+		// and lose data silently
+		return nil, err
+	}
+
+	if os.Getenv("DAPI_STATS_FILE_STORAGE_ENABLED") == "true" {
```
blizko, line 30: Consider rewriting to `os.Getenv("DAPI_STATS_FILE_STORAGE_ENABLED") != "false"`
Rationale - existing setups do not have this variable defined, thus will lead to case when both PostgreSQL and File storage are not available

local-test-net/docker-compose-base.yml
```diff
@@ -50,6 +50,7 @@ services:
       - DEBUG=true
       - IS_TEST_NET=true
       - ENFORCED_MODEL_ID=disabled
+      - DAPI_STATS_FILE_STORAGE_ENABLED=true
```
blizko, line 53: Defaults required for main-net docker-compose as well

decentralized-api/statsstorage/factory.go
patimen: That is the intent. File storage is expensive and potentially disastrous for performance on a high volume chain. The hardware would need to be pretty fast to pull it off under load. If we enable by default we will lose participants.

local-test-net/docker-compose-base.yml
patimen: Again, no, it should default to false and only be enabled with consideration.

decentralized-api/statsstorage/factory.go
blizko: That is a breaking change.
The default configuration should have safe defaults for launch. The default `docker-compose` base has both `DAPI_STATS_FILE_STORAGE_ENABLED` and Postgresql variables undefined. That would lead to inability to launch a new validator with default envs, as well as crash validators that do not perform changes to their setup.

decentralized-api/main.go
```diff
@@ -226,6 +233,25 @@ func main() {
...
 		pserver.WithStatsStorage(statsStore),
 	)
+
+	if subnetSigner != nil {
+		subnetBridge := internalsubnet.NewChainBridge(recorder)
+		httpClient := pserver.NewNoRedirectClient(5 * time.Minute)
+		subnetEngine := internalsubnet.NewEngineAdapter(nodeBroker, config.GetCurrentNodeVersion(), payloadStore, chainPhaseTracker, httpClient)
+		subnetValidator := internalsubnet.NewValidationAdapter(nodeBroker, config.GetCurrentNodeVersion(), chainPhaseTracker, httpClient, subnetBridge, recorder)
+		// TODO: move to SubnetConfig when config consolidation happens.
+		subnetStore, storeErr := subnetstorage.NewSQLite("/root/.dapi/data/subnet.db")
```
blizko, line 243: Creates a new hard requirement for an SQLite database and will ultimately block the ability to perform horizontal scaling
Consider reusing PgSQL if available?

gmorgachev: Hi @blizko !
Yes, that part is definitely going to use postgress in the next release. The current version of subnet implements minimalistic version, to collect early feedback and test at scale. Without punishment nodes if something is not working 

The plan for the first major version is to allow several `api`/`subnet` services and to allow horizontal scaling in different ways:
1. postgres to share DB between instances who need access to the same data => redundancy of workers will be enabled
2. splitting traffic across different `api` / `subnet` instance groups determinitically by subnet id => explicit sharding of database will be enabled, without requirement that every instance has access to every subnet data (redundancy only inside instance group). More about this part here https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.11/proposals/inference/design.md#host-discovery-multi-url-identity

inference-chain/app/upgrades/v0_2_11/upgrades.go
gmorgachev: Hi! That's the SafetyWindow for sampling ConfirmationPoC. One of the team is still working on the maintainance window

decentralized-api/statsstorage/factory.go
gmorgachev: > That would lead to inability to launch a new validator with default envs, as well as crash validators that do not perform changes to their setup.

@blizko could you please clarify? From my understanding the undefined variables will lead to disabled statistic collection. Which is totally fine, as it's optional feature for validators. But i might miss smth

decentralized-api/statsstorage/postgres_storage.go
gmorgachev: @libermans could you check what was the idea behind?

decentralized-api/statsstorage/factory.go
blizko: The dynamic pricing under `decentralized-api/internal/server/public/get_pricing_handler.go` relies on the stats storage in `getModelMetrics` function to give back `Utilization` values.
It does look to be only statistical for now, so I likely am wrong in my interpretation of criticality.

If this is an optional feature, then we have a different effect. Having an connected PgSQL instance would activate it, and then introduce extra load.
If it is intended to be optional - it would be appropriate to hide it under a feature flag.

ouicate (commented):
`DefaultSubnetEscrowParams()` sets `AllowedCreatorAddresses: nil`, and `IsAllowedEscrowCreator()` treats an empty allowlist as "everyone is allowed" (see `params.go:303`). This means immediately after the upgrade, **any address** can create subnet escrows.

The PR's own test (`TestCreateSubnetEscrow_AllowlistBlocks`) confirms that restricting access requires a non-nil list — but the upgrade itself never sets one.

decentralized-api/internal/event_listener/event_listener.go
```diff
@@ -430,12 +446,200 @@ func (e *InferenceFinishedEventHandler) Handle(event *chainevents.JSONRPCRespons
...
+func getEventValue(events map[string][]string, key string, idx int) (string, bool) {
+	values := events[key]
+	if len(values) == 0 {
+		return "", false
+	}
+	if idx < len(values) {
+		return values[idx], true
+	}
+	return values[0], true
+}
```
ouicate, line 535: When `idx >= len(values)`, this silently returns `values[0]` with `ok=true`, which can silently associate wrong event data with wrong inference records in batched transactions. Should this return `("", false)` instead so callers can detect the mismatch?

subnet/bridge/rest.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	found := false
+	for _, g := range resp.Grantees {
+		if g.Address == warmAddress {
+			found = true
+			break
+		}
+	}
+	b.warmCache.Store(key, found)
```
ouicate, line 173: `warmCache` stores results indefinitely with no TTL or invalidation. If an authz grant is revoked on-chain, this cache will continue returning `true` for the revoked warm key until the process is restarted.

gmorgachev: That's by design, the subnet doesn't support warm key update / revoke in the current version, it makes consensus a bit more complex

gmorgachev:
> `DefaultSubnetEscrowParams()` sets `AllowedCreatorAddresses: nil`, and `IsAllowedEscrowCreator()` treats an empty allowlist as "everyone is allowed" (see `params.go:303`). This means immediately after the upgrade, **any address** can create subnet escrows.
> 
> The PR's own test (`TestCreateSubnetEscrow_AllowlistBlocks`) confirms that restricting access requires a non-nil list — but the upgrade itself never sets one.

@ouicate thanks! that's in "[to be merged, don't affect mainnet] Final Subnet Params" mentioned in PR desciptions. will be merged today

inference-chain/app/upgrades/v0_2_11/upgrades.go
```diff
@@ -0,0 +1,187 @@
...
+		err = setPruningState(ctx, k)
+		if err != nil {
+			return nil, err
+		}
+
+		err = setEpochParticipantsSets(ctx, k)
+		if err != nil {
+			return fromVM, err
+		}
+
```
Ryanchen911, line 156: I noticed that if setEpochParticipantsSets() fails at line 45, some state (like parameters at line 36) might have already been modified, while the handler still returns the old fromVM. I'm a bit concerned this might affect the atomicity of the upgrade if we need to retry it later.What do you think about ensuring the state remains consistent during a retry?

inference-chain/app/upgrades/v0_2_11/upgrades.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func setEpochParticipantsSets(ctx context.Context, k keeper.Keeper) error {
+	currentEpochIndex, err := k.EffectiveEpochIndex.Get(ctx)
+	if err != nil {
+		return err
+	}
+	if currentEpochIndex < 2 {
+		return err
+	}
```
Ryanchen911, line 189: I find the return logic here a bit counter-intuitive. When currentEpochIndex < 2, we are returning err as nil, which might lead callers to believe the participant caches were successfully initialized when they actually weren't.

inference-chain/app/upgrades/v0_2_11/upgrades_test.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestUpgradeName(t *testing.T) {
+	require.Equal(t, "v0.2.11", UpgradeName)
+}
```
Ryanchen911, line 11: I think we have missed some test cases.

Ryanchen911 (commented):
Suggestions for error handling and additional tests

decentralized-api/internal/subnet/validation.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("read validation response: %w", err)
+	}
+
+	validationResponse, err := completionapi.NewCompletionResponseFromBytes(respBytes)
+	if err != nil {
+		return nil, fmt.Errorf("parse validation response: %w", err)
+	}
+
+	originalLogits := originalResponse.ExtractLogits()
```
x0152, line 151: Logits are checked but token counts aren't. Executor can report max InputTokens/OutputTokens to get paid more (within escrow sum)

decentralized-api/internal/validation/payload_retrieval.go
```diff
@@ -32,16 +32,123 @@ var ErrEpochStale = errors.New("inference epoch too old, validation no longer us
 // HTTP client with timeout for payload retrieval
 var payloadRetrievalClient = &http.Client{}
```
x0152: Comment says "with timeout" but http.Client{} has none. Executor can hang validator goroutines forever

decentralized-api/internal/subnet/manager.go
```diff
@@ -0,0 +1,482 @@
...
+
+func (m *HostManager) getOrCreate(escrowID string) (*transport.Server, error) {
+	m.mu.RLock()
+	srv, ok := m.sessions[escrowID]
+	m.mu.RUnlock()
+	if ok {
+		return srv, nil
+	}
+
+	m.mu.Lock()
```
x0152: it could be dangerous to take a global lock here because this path makes grpc calls. If grpc gets stuck, all other session requests are blocked behind the same lock

decentralized-api/internal/subnet/engine.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	// Store the canonicalized ORIGINAL prompt (not the modified one with seed).
+	promptPayload, err := subnet.CanonicalizeJSON(req.Prompt)
+	if err != nil {
+		return nil, fmt.Errorf("canonicalize prompt: %w", err)
+	}
+
+	storageKey := SubnetPayloadKey(req.EscrowID, req.InferenceID)
+	epochID := currentEpochID(e.phaseTracker)
```
x0152, line 125: Rare case: on startup or after restart, before dapi receives any block events from chain, currentEpochID can return 0. Payloads stored under epoch 0

decentralized-api/internal/validation/payload_retrieval.go
gmorgachev: good catch. by some reason someone removed timeout here in 0.2.8....

decentralized-api/internal/subnet/engine.go
gmorgachev: agree

decentralized-api/internal/subnet/validation.go
gmorgachev: you're right, will update now

PS: validation pipeline in subnet is far from working for now

decentralized-api/statsstorage/factory.go
blizko: Addition of `DAPI_STATS_POSTGRES_ENABLED` resolves this thread

akup:
The issue that I see according new ShardChains. User is sequencing and signs all the new "blocks".

**It makes following attack possible:**
1. User sends `MsgStartInference` diff to all hosts.
2. Executor confirms, runs inference, streams result back to the user.
3. Executor puts `MsgFinishInference` in its mempool.
4. User has the result but deliberately does not sequence `MsgFinishInference`.
From every non-executor host's perspective, the inference is still `StatusStarted`. They have no way to know the executor finished.
5. User collects `MsgTimeoutInference(reason=EXECUTION)` votes from non-executor hosts — they legitimately believe the executor timed out.
6. User sequences the timeout diff. The executor is only 1 slot out of 16; the threshold is total_slots / 2 = 8, easily met by the other 15 slots.
7. Inference becomes StatusTimedOut → reserved cost is refunded to Balance. Executor gets Missed++ penalty.
User got the inference for free and the executor is penalized.

It seams we need to gossip `MsgFinishInference` from host eagerly, so other hosts can get it from executor not signed by user. And this should be enough for hosts to do not timeout it

As the feature is generally experimental we sure can live with this new 'free inference' attack. It is more important to focus on development of subnet in different (possibly independent) branches. There are still a lot of possible issues but it's ok for version 0

akup:
Another big issue is the absence of timestamp consensus.

Now user can send inference `StartedAt: time.Now().Unix()`, but user can put it in the past to force inference to timeout.

gmorgachev (approved):

unameisfine:
## v0.2.11 Upgrade Review

Reviewed: upgrade handler (`upgrades.go`), subnet settlement/pruning, collateral logic, parameter changes.

### Finding 1: `AdjustWeightsByCollateral` missing `baseWeightRatio` range validation

`calculateRequiredCollateral` (collateral.go:53) correctly rejects `baseWeightRatio >= 1.0`:
```go
if err != nil || bwr.IsNegative() || bwr.GTE(math.LegacyOneDec()) {
    return math.ZeroInt()
}
```

`AdjustWeightsByCollateral` (collateral.go:100) converts the same parameter but has no equivalent guard. If governance sets `baseWeightRatio >= 1.0`, uncollateralized participants receive inflated weights (e.g. 150% of potential), inverting the collateral incentive structure.

Filed as #933 with fix in #934.

### Finding 2: `distributeUnsettledEscrow` always returns nil

`subnet_pruning.go` — the function uses `continue` on every `SendCoinsFromModuleToAccount` failure and always returns nil. The caller in `pruning.go:136` deletes the escrow record regardless. If distribution fails, funds stay locked in the module account with no reference to recover them.

### Overall

Subnet architecture is sound — the 2-tx model addresses the throughput ceiling. `AllowedCreatorAddresses` properly restricted to 13 addresses. BLS Certik fixes cover critical findings. DB bloat fix (PR #867) is high-impact for operators.

inference-chain/x/inference/keeper/pruning.go
```diff
@@ -84,6 +113,60 @@ func (k Keeper) GetPoCBatchesPruner(params types.Params) Pruner[collections.Trip
...
+		SetLastPruned: func(state *types.PruningState, epoch int64) {
+			state.SubnetPrunedEpoch = epoch
+		},
+		Remover: func(ctx context.Context, key collections.Pair[uint64, uint64]) error {
+			epochIndex := key.K1()
+			escrowID := key.K2()
+
+			escrow, found := k.GetSubnetEscrow(ctx, escrowID)
+			if found && !escrow.Settled {
+				if err := k.distributeUnsettledEscrow(ctx, escrow); err != nil {
```
jacky6block, line 136: `distributeUnsettledEscrow` always return nil

gmorgachev:
> Another big issue is the absence of timestamp consensus.
> 
> Now user can send inference `StartedAt: time.Now().Unix()`, but user can put it in the past to force inference to timeout.

Which type of timeout do you mean? There are
- `refusal` timeout - when hosts didn't send back the receipt `MsgConfirmStart`. If user trigger such timeout, it's explicitly validated though several another hosts. And if user tried to cheat another hosts will just provide prompt to executor on this step (no harm)
- `execution` timeout. It might be triggered on if executor confirmed receiving of prompt (`MsgConfirmStart`). => executor agreed with start inference timestamp and whole chain sees proof of that

akup:
> * `refusal` timeout - when hosts didn't send back the receipt `MsgConfirmStart`. If user trigger such timeout, it's explicitly validated though several another hosts. And if user tried to cheat another hosts will just provide prompt to executor on this step (no harm)

My miss. The important point here that all hosts before voting for timeout `ChallengeReceipt` trying to reach the executor and if it is reachable, timeout will not be accepted.


> * `execution` timeout. It might be triggered on if executor confirmed receiving of prompt (`MsgConfirmStart`). => executor agreed with start inference timestamp and whole chain sees proof of that

It is measured from ConfirmAt time signed by executor. Executor can set this time in a future (that can never be reached) and sign, so inference will never be timeouted. Currently this case is not handled.

PR #825, opened 2026-02-28
fix: dkg dealer consensus

### LINKED ISSUE #823: Bridge: Weak Dealer Approval Enables Threshold Signing DoS
**Locations:** 
- https://github.com/gonka-ai/gonka/blob/82c43a42c3c2f49b56ee8a32e6458480daf39ca9/inference-chain/x/bls/keeper/phase_transitions.go#L169-L169
- https://github.com/gonka-ai/gonka/blob/82c43a42c3c2f49b56ee8a32e6458480daf39ca9/inference-chain/x/bls/keeper/threshold_signing.go#L296-L302

**Categories:** 
- Logical-issue
- Denial-of-service

**Description**
Dealer validation uses an unweighted majority of submitted DealerValidity votes and does not verify per-recipient shares against commitments. A malicious dealer can send valid shares to ~50% of recipients and garbage to the rest.

The dealer (or a colluder) can vote “true,” giving itself a bare majority among submitters (>50%). The dealer is marked "valid" and included in the group key even though many recipients lack usable shares.

`inference-chain/x/bls/keeper/phase_transitions.go`
Image

Image

Later, the dealer (or colluder) withholds its partial signature, pushing the usable signers below the >50% slot threshold. Threshold-signing requests then stall and expire—a sustained liveness/DoS risk, even with <50% malicious participants plus abstentions.

`inference-chain/x/bls/keeper/threshold_signing.go`

Image

x0152:
Follow-up to 'dl/certik-bls-geb-29' (original author: @libermans)
Main goal of this PR is to address #823

## Problems from #823
1. Participants can vote 'true' without proving they have valid shares
2. Quorum can depend only on who submitted, not on all slots
3. Dealer can approve itself
4. Dealer can send valid shares to some participants and garbage to others
5. Dealer quorum and signing threshold are both >50%, so a dealer can pass with barely enough support and one missing signer can break liveness

## What this PR fixes
- [x] BLS proof is required for every 'true' vote (When a participant votes true for a dealer, they must sign a message using shares they got from that dealer. The chain checks this signature against the dealer's public commitments. If signature is invalid or missing, the vote is rejected. So you can't vote true without actually having valid shares)
- [x] Dealer quorum is slot-weighted against all slots
- [x] Dealer is excluded from its own approval
- [x] Needs dispute mechanism (implemented in dl/certik-bls-geb-29)
- [x] Needs dispute mechanism (implemented in dl/certik-bls-geb-29) - with dispute, dealers can't create borderline quorum situations by sending garbage to some participants

## Additional improvements

- Dealer openings persist to SQLite and survive API restarts
- Tx submissions retry on transient network failures
- On-chain complaint verification and fault assignment
- Nodes refuse to sign if they are missing shares from any consensus-valid dealer
- CompleteDKG now transitions to 'FAILED' instead of returning an error that gets swallowed by EndBlock, which was causing the epoch to get stuck in 'VERIFYING'forever
- Upgrade handler for v0.2.11
- Extended verification windows to 6 blocks
- Few E2E smoke tests


akup:
Didn't understand why to store openings at SQL if they are never read.

Everything else looks good

x0152:
> Didn't understand why to store openings at SQL if they are never read.
> 
> Everything else looks good

Thanks for review! They are read on startup: main.go:157 calls SetDealerOpeningsDB, which loads persisted openings back into memory so the node survives restarts during dispute phase

akup:
> > Didn't understand why to store openings at SQL if they are never read.
> > Everything else looks good
> 
> Thanks for review! They are read on startup: main.go:157 calls SetDealerOpeningsDB, which loads persisted openings back into memory so the node survives restarts during dispute phase

That makes sense. Could be merged

jacky6block (commented):
Hey, I went through the changes carefully — the slot-weighted quorum direction does seem like the right fix for the original DoS issue.

A few things stood out to me while reviewing. Some of these may be actual issues, and some may just be gaps in my understanding, so I wanted to raise them here for discussion:


1. Possible missing handling for EventDKGFailed on the client side

I may be missing it, but it looks like the chain emits `EventDKGFailed` in two spots in `phase_transitions.go` (dealing phase insufficient participation + verification phase insufficient participation), while `handleBLSEvents` in `event_listener.go` only subscribes to three BLS events, and I don’t think the failure event is one of them.

If that reading is correct, then when a DKG round fails, nodes may not get notified, and the in-memory cache could end up keeping a stale `DKG_PHASE_VERIFYING` entry around. Might be worth checking whether that failure path needs an explicit client-side handler.


2. One edge case I wanted to sanity-check in `DetermineValidDealersWithConsensus`

I wanted to confirm whether the new quorum calculation behaves as intended in high-stake cases.

With `quorumSlots = totalSlots/2 + 1` and self-vote excluded, the maximum non-self votes a dealer can receive is `totalSlots - dealerOwnSlots`. If I’m reading that correctly, then any participant holding >= 50% of total slots would never be able to reach quorum mathematically, even if all other participants vote for them.

A couple of examples with `totalSlots=100`:

- A has 60 slots, B-E have 10 each -> A can receive at most 40 votes, quorum is 51
- Two participants with 50 slots each -> each can receive at most 50 votes, quorum is still 51

If this is expected, then all good — but otherwise I’m wondering whether the intended quorum should perhaps be relative to the other participants’ voting power rather than total slots. Something like `(totalSlots - dealerOwnSlots)/2 + 1` per dealer might be closer to that, though it does make the threshold dynamic.


3. Possible missing guard before completing DKG

Related to the above, `CompleteDKG` calls `DetermineValidDealersWithConsensus` and then proceeds into `ComputeGroupPublicKey`, but I didn’t notice a guard checking whether the valid dealers still cover enough slots for threshold reconstruction.

As I understand it, reconstruction needs `t+1 = iTotalSlots - TSlotsDegreeOffset + 1` slots worth of valid shares. So if enough dealers are excluded, it seems possible for DKG to complete on-chain while threshold signing fails later because there aren’t enough valid shares left. Unless this is enforced somewhere else, it may be worth adding an explicit check here.


4. Question about the implications of removing the dispute phase

Just to clarify, I do understand that the client-side code is doing real cryptographic work here: `performVerificationAndReconstruction` decrypts shares via ECIES and runs KZG verification against commitments before setting `DealerValidity`.

The part I’m less sure about is the tradeoff after removing the dispute phase. In the base branch, if a dealer’s shares were marked invalid, that dealer could still respond during the dispute phase by submitting the original share bytes and the ECIES seed, allowing the chain to recompute the ciphertext and verify the KZG commitment on-chain.

With that path removed, I think a dealer whose shares are actually valid may no longer have an on-chain way to challenge a false invalid vote. That may be an intentional tradeoff depending on the trust assumptions / incentive model — I just wanted to flag it explicitly in case I’m missing some replacement mechanism.

x0152:
@jacky6block 
Hey, thanks for the careful review. Here's my understanding:
1. event_listener.go:380-386 subscribes to EventDKGFailed and calls dispute.go:102 which cleans up state for that epoch
2. This issue can happen only if one participant controls 50% or more of all slots. Normally, power capping should prevent this. If it still happens, the network is already too centralized, so a DKG failure in that rare case is acceptable
3. No explicit guard, but covered implicitly: ComputeGroupPublicKey errors on zero valid dealers -> faildkg. Client side ensureConsensusSharesComplete refuses to sign with missing shares
4. Dispute phase is not removed. phase_transitions.go:208 transitions to DKG_PHASE_DISPUTING, dispute.go:14 handles EventDisputePhaseStarted on client side, and dispute_resolution.go:26 applies complaint outcomes on-chain. Dealers can still challenge false-invalid votes

x0152:
Added a diagnostic warning for cases where a dealer cannot mathematically reach weighted quorum with self vote excluded

jacky6block:
> @jacky6block Hey, thanks for the careful review. Here's my understanding:
> 
> 1. event_listener.go:380-386 subscribes to EventDKGFailed and calls dispute.go:102 which cleans up state for that epoch
> 2. This issue can happen only if one participant controls 50% or more of all slots. Normally, power capping should prevent this. If it still happens, the network is already too centralized, so a DKG failure in that rare case is acceptable
> 3. No explicit guard, but covered implicitly: ComputeGroupPublicKey errors on zero valid dealers -> faildkg. Client side ensureConsensusSharesComplete refuses to sign with missing shares
> 4. Dispute phase is not removed. phase_transitions.go:208 transitions to DKG_PHASE_DISPUTING, dispute.go:14 handles EventDisputePhaseStarted on client side, and dispute_resolution.go:26 applies complaint outcomes on-chain. Dealers can still challenge false-invalid votes

Thanks for clarifying! One small suggestion — the warning fires at DKG time, but the actual failure surfaces later as a silent request expired. It might be worth including the epoch's DKG quorum status in the signing expiration log, so the two events are easier to connect during debugging.

x0152:
Merged dl/certik-bls-geb-29 (author: @libermans) and added three follow-up fixes:
1. harden dispute response flow - clear verification cache on DKG_FAILED and add MsgRespondDealerComplaints to operation key permissions
2. deterministic dealer proof validation - swap map for slice so proof checks always run in the same order on-chain
3. paginate authz grants - GranteesByMessageType was only reading the first page; now it walks through all of them

0xgonka:
The core fixes look correct — BLS proof requirement, slot-weighted quorum, self-exclusion, shape validation, and SQLite persistence are all well-implemented.

**Should fix:**
- Verify hash-to-G1 mapping is consistent between chain (`verifyBLSPartialSignatureBlst`) and DAPI (`computePartialSignatureBlst`). Proof hash uses Keccak-256 → `MapToCurve1` but if the chain-side uses a different mapping, all proofs silently fail.
- Dealer opening upserts in `sqlite_store.go` are individual statements without a transaction wrapper. During dealing, hundreds are stored one-by-one (one fsync each). Consider batching in a single transaction for performance and atomicity.
- No unit test for the slot-weighted quorum boundary case — what happens when a dealer has exactly 50% of slots? With self-exclusion, max non-self votes = `totalSlots/2` but `quorumSlots = totalSlots/2 + 1`, so it should always fail. This edge case should have an explicit test.
- A participant holding >50% of total slots can never be an approved dealer (self-exclusion makes quorum impossible). The code logs a warning but this constraint should be documented as a known limitation.

**Nits:**
- Self-proof is accepted (dealer votes true for itself with valid proof). Harmless since self-votes are excluded from quorum, but wastes gas. Could skip proof validation when `verifierIndex == dealerIndex`.
- SQLite write failures in `storeDealerOpeningRecord` and `deleteDealerOpeningsForEpoch` are logged as warnings but not propagated. In-memory map and SQLite can diverge — on restart, SQLite is truth, so a failed persist + crash = lost opening = dealer can't respond to complaints.
- `isQueuedForRetry` returns `nil` to callers (dealer.go, verifier.go, dispute.go), silencing the error. If the retry system itself fails later, no recovery path from the BLS manager.
- `permissions_test.go` appears to be missing — `permissions.go` adds BLS message types to `InferenceOperationKeyPerms` but no corresponding test was added.
- `AllowedSecp256K1PublicKeys` is used in `expectedEncryptedSharesCount` for shape validation — verify this field is always populated before `SubmitDealerPart` is called. A timing mismatch between dealing and verification would cause shape mismatches.

akup:
@x0152 
As https://github.com/gonka-ai/gonka/pull/988 was merged, could you please recheck, that PR doesn't have overlapping commits, to make it atomic, targeted and easier to review

x0152:
> @x0152 As #988 was merged, could you please recheck, that PR doesn't have overlapping commits, to make it atomic, targeted and easier to review

Rebased already, no overlap with #988

x0152:
## Should fix:

> Verify hash-to-G1 mapping is consistent between chain (verifyBLSPartialSignatureBlst) and DAPI (computePartialSignatureBlst). Proof hash uses Keccak-256 → MapToCurve1 but if the chain-side uses a different mapping, all proofs silently fail.
- [x] Already correct: chain and dapi use the same Keccak-256 + hash-to-G1 flow
---

> Dealer opening upserts in sqlite_store.go are individual statements without a transaction wrapper. During dealing, hundreds are stored one-by-one (one fsync each). Consider batching in a single transaction for performance and atomicity.
- [x] Fixed: upserts are now batched in one transaction 
---
> No unit test for the slot-weighted quorum boundary case — what happens when a dealer has exactly 50% of slots? With self-exclusion, max non-self votes = totalSlots/2 but quorumSlots = totalSlots/2 + 1, so it should always fail. This edge case should have an explicit test.
- [x] Fixed: added explicit boundary unit test for the exact 50% case
---
> A participant holding >50% of total slots can never be an approved dealer (self-exclusion makes quorum impossible). The code logs a warning but this constraint should be documented as a known limitation.
- [x] Documented as a known limitation in `proposals/certik_bls/geb-29-35-dispute-phase-plan.md`
---
## Nits:

> Self-proof is accepted (dealer votes true for itself with valid proof). Harmless since self-votes are excluded from quorum, but wastes gas. Could skip proof validation when verifierIndex == dealerIndex.
- [x] Fixed: сhain skips self-proof verification
---
> SQLite write failures in storeDealerOpeningRecord and deleteDealerOpeningsForEpoch are logged as warnings but not propagated. In-memory map and SQLite can diverge — on restart, SQLite is truth, so a failed persist + crash = lost opening = dealer can't respond to complaints.
- [x] Fixed: failures are now returned as errors and handled by callers
---
> isQueuedForRetry returns nil to callers (dealer.go, verifier.go, dispute.go), silencing the error. If the retry system itself fails later, no recovery path from the BLS manager.
- [x] Fixed: retry cases now return wrapped typed error instead of ni
---
> permissions_test.go appears to be missing — permissions.go adds BLS message types to InferenceOperationKeyPerms but no corresponding test was added.
- [x] Not an issue: permissions_test.go already exists
---
> AllowedSecp256K1PublicKeys is used in expectedEncryptedSharesCount for shape validation — verify this field is always populated before SubmitDealerPart is called. A timing mismatch between dealing and verification would cause shape mismatches.
- [x] Verified as ok: participant key snapshot is prepared before dealing and used consistently, no timing mismatch found
---

x0152:
> Self-proof is accepted (dealer votes true for itself with valid proof). Harmless since self-votes are excluded from quorum, but wastes gas. Could skip proof validation when verifierIndex == dealerIndex.
 Fixed: сhain skips self-proof verification

Self-proof skip is already implemented, but I'm thinking we should revert it back to requiring all proofs, gas saving is minor and it adds unnecessary complexity across validation layers

0xgonka (approved):

PR #826, opened 2026-02-28
fix: prevent permanent reward loss on partial claim payment failure

ouicate:
Two bugs in payoutClaim caused participants to permanently lose their entitled rewards when payments partially failed:

1. When escrow payment failed (insufficient funds), handleUnderfundedWork() called finishSettle() which removed the settle amount and marked the claim as finished. The participant could never retry.

2. When work payment succeeded but reward payment failed, finishSettle() was called which removed the settle amount. The participant lost their reward coins permanently with no retry mechanism.

Fixes:
- handleUnderfundedWork: removed finishSettle() call so the settle amount is preserved and the participant can retry when funds are available.
- payoutClaim reward failure path: instead of calling finishSettle(), zero out WorkCoins (already paid) and update the settle amount in store, preserving RewardCoins for retry.

Added regression tests:
- TestMsgServer_ClaimRewards_EscrowFailurePreservesSettleAmount
- TestMsgServer_ClaimRewards_RewardFailurePreservesRewardCoins


akup:
Already fixed here: https://github.com/gonka-ai/gonka/pull/789
Please check it

ouicate:
Closing this because the issue is superseded by #789, which is merged into upgrade-v0.2.12 and implements the safer CacheContext-based payout atomicity fix.

PR #829, opened 2026-02-28
fix: exclude INVALIDATED inferences from ClaimRewards missed-validation check

ouicate:
MsgValidation grants validation credit on INVALIDATED inferences, allowing validators to reduce their missed-validation count without performing actual work. This bypasses the ClaimRewards missed-validation penalty.

Fix: Filter INVALIDATED inferences from both sides of the missed-validation computation in ClaimRewards:
1. getValidatedInferences: exclude INVALIDATED inference IDs from wasValidated map, directly blocking the exploit vector
2. getMustBeValidatedInferences: exclude INVALIDATED inferences from mustBeValidated list, so honest validators are not penalized for skipping invalidated inferences

MsgValidation still records credit in EpochGroupValidations (design decision: validations can be submitted during the whole epoch), but that credit no longer affects reward calculations for invalidated inferences.


akup:
> MsgValidation grants validation credit on INVALIDATED inferences, allowing validators to reduce their missed-validation count without performing actual work. This bypasses the ClaimRewards missed-validation penalty.

Why not to grant validation credits for invalidated? Why is it allowing not to do actual work and reduce validation count?

The validation/invalidation flow is following:
1. Validator is settled to validate inference and first it tries to validate it.
2. Validator already done actual validation work and now he decides is the inference valid or not
3. If inference is valid, he sends `MsgValidation`, but if it is invalid he sends `MsgValidation` with `Revalidation == true`
4. validation is not added to `EpochGroupValidations` if `Revalidation == true`. Only inferences that passed this phase could be Invalidated. So actually messages that was invalidated shouldn't be in the
```go
wasValidatedRaw, found := k.GetEpochGroupValidations(ctx, msg.Creator, msg.EpochIndex)
```
that you are filtering. So here no additional filtering is required
5. revalidation creates the event when participants are voting was the inference really invalid? (Very heavy and a subject of one of PRs)
6. And if the proposal passes, then inference becomes invalidated

First of all there are a lot of work to be done to make inference invalidated and why not to reward validator?
Second it is already filtered


But actually it seams here is logical error in rewards.
If inference was invalidated, it is not in the 
```go
wasValidated := k.getValidatedInferences(ctx, msg)
```
but is at
```go
mustBeValidated, err := k.getMustBeValidatedInferences(ctx, msg)
```

This pr fixes this, but I think it is too computationally heavy as we are calling `k.GetInference(ctx, inference.InferenceId)` for each settled inference.

I think it is better to store also `EpochGroupRevalidations` like it is done for `EpochGroupValidations`

akup:
@ouicate  I've noticed that there are more issues on validation.
I've implemented proposed storing of `EpochGroupRevalidations`
Also it was possible to send from any participant validation and revalidation messages, but only settled participants should do it.
And @0xMayoor found that already validated or in voting inference can be revalidated here: https://github.com/gonka-ai/gonka/pull/832

So I've extended this PR and finished with this one:
https://github.com/gonka-ai/gonka/pull/967

Please take a look and review

akup:
> @ouicate I've noticed that there are more issues on validation. I've implemented proposed storing of `EpochGroupRevalidations` Also it was possible to send from any participant validation and revalidation messages, but only settled participants should do it. And @0xMayoor found that already validated or in voting inference can be revalidated here: #832
> 
> So I've extended this PR and finished with this one: #967
> 
> Please take a look and review

I was mistaken on:

>But actually it seams here is logical error in rewards.
If inference was invalidated, it is not in the
```go
wasValidated := k.getValidatedInferences(ctx, msg)
```

It is there.

Now I understand your PR: user can skip the job just by sending invalidation.
But if we fix it the way you propose we also will force hornest users to validate more, who checked the inference and it was really invalid.

A month ago I've also noticed that, but there are more harm. Every revalidation is costly in performance, it forces a lot of host to validate and vote.
I think it should be handled in other way:
If user started revalidation voting, and then majority votes that inference is valid, we should punish host, who initiated this wrong voting.

The problem you've reported exist, but it should be handled in other place and in other way.

What do you think?

ouicate:
Updated after re-checking the validation flow and your review.

The original fix did the filtering inside `ClaimRewards`, but that required `GetInference` lookups while computing reward-claim validation stats and also erased credit for validators who actually performed an invalidation check. I changed the fix to the root cause instead:

- `MsgValidation` now checks `INVALIDATED` / `STARTED` / executor/self / cached epoch weight eligibility before writing to `EpochGroupValidations`.
- A non-revalidation validation for an already `INVALIDATED` inference returns without recording reward-claim credit.
- ClaimRewards no longer does per-inference status lookups, so the expensive path you pointed out is removed.
- Added a regression test for the specific bypass: validating an already invalidated inference must not create `EpochGroupValidations` credit.

This keeps credit for validators who performed real validation/invalidation work on a non-invalidated inference, and only blocks the cheap bypass of submitting `MsgValidation` after the inference is already invalidated.

PR #830, opened 2026-02-28
fix: use deterministic seed for ClaimRewards reservoir sampling

ouicate:
Replace block-hash-seeded RNG with msg.Seed for reservoir sampling in getMustBeValidatedInferences. The block hash changes per block, which allowed retry grinding: a validator could re-submit ClaimRewards in different blocks until a favorable sample under-represented their missed validations and the statistical test passed.

msg.Seed is signed by the validator (verified against the stored SeedSignature) and cannot be changed between retries, making the reservoir sample deterministic per validator per epoch regardless of which block includes the claim transaction.


ouicate:
Updated against the current `upgrade-v0.2.12` base,

PR #832, opened 2026-03-01
fix: add VALIDATED/VOTING status guards in validation handler

0xMayoor:
the validation handler at `msg_server_validation.go` only guards against INVALIDATED and STARTED statuses. nothing stops a second validator from processing an inference that's already VALIDATED or in VOTING. 
when that happens, shareWorkWithValidators has already redistributed CoinBalance — and if the second validator's failure triggers a vote that resolves to invalidation, the executor gets hit twice: 
once from shareWork,
once from the full ActualCost deduction in InvalidateInference.

concrete sequence:
inference finishes, executor gets CoinBalance = 30000 (ActualCost).
validator A passes it, shareWork splits the balance — executor drops to 15000, validator A gets 15000.
validator B fails the same inference, no guard blocks it, overwrites status from VALIDATED to VOTING.
voting resolves INVALIDATED, refundInvalidatedInference deducts full ActualCost (30000) from executor.
executor ends at -15000, lost 45000 instead of 30000. validator A keeps 15000 they shouldn't have.
this isn't a forced edge case — it happens naturally whenever two validators disagree on the same inference within the same epoch (one passes threshold, the other doesn't). 
the negative CoinBalance becomes debt at settlement since WorkCoins = CoinBalance, so the executor also loses rewards earned from completely unrelated inferences.

added status guards for VALIDATED and VOTING after the existing INVALIDATED guard. revalidation calls (from group voting) are exempted since they need to process on VOTING-status inferences — the guard only blocks first-time validation attempts on already-decided inferences.

test in `validation_accounting_test.go` confirms the guard correctly rejects the second validation and prevents the accounting mismatch.

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_validation.go
+++ b/inference-chain/x/inference/keeper/msg_server_validation.go
@@ -74,6 +74,14 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 		k.LogInfo("Inference already invalidated", types.Validation, "inference", inference)
 		return &types.MsgValidationResponse{}, nil
 	}
+	if inference.Status == types.InferenceStatus_VALIDATED {
+		k.LogInfo("Inference already validated, skipping", types.Validation, "inferenceId", inference.InferenceId)
+		return &types.MsgValidationResponse{}, nil
+	}
+	if !msg.Revalidation && inference.Status == types.InferenceStatus_VOTING {
+		k.LogInfo("Inference already in voting, skipping", types.Validation, "inferenceId", inference.InferenceId)
+		return &types.MsgValidationResponse{}, nil
+	}
 	if inference.Status == types.InferenceStatus_STARTED {
 		k.LogError("Inference not finished", types.Validation, "status", inference.Status, "inference", inference)
 		return nil, types.ErrInferenceNotFinished
--- a/inference-chain/x/inference/keeper/validation_accounting_test.go
+++ b/inference-chain/x/inference/keeper/validation_accounting_test.go
@@ -1,0 +1,98 @@
+package keeper_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/productscience/inference/testutil"
+	"github.com/productscience/inference/x/inference/calculations"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// TestValidationStatusGuard_PreventsDoubleShareWork verifies that once an inference
+// is VALIDATED by one validator, a second validator calling Validation is rejected.
+// Without the status guard, the second validator could overwrite VALIDATED to VOTING,
+// leaving the shareWorkWithValidators redistribution irreversible if voting resolves
+// to INVALIDATED — causing the executor to lose more than ActualCost.
+func TestValidationStatusGuard_PreventsDoubleShareWork(t *testing.T) {
+	inferenceHelper, k, ctx := NewMockInferenceHelper(t)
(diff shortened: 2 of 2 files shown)


akup:
There is `addInferenceToEpochGroupValidations` at the top of processing validation message.
It adds validation to EpochGroupValidations `k.SetEpochGroupValidations(ctx, epochGroupValidations)`
If it already exists it will return `types.ErrDuplicateValidation` error that will rollback tx on early start.

So if it wasn't revalidation message, and tx succeeded (message is now VALIDATED or VOTING), EpochGroupValidations will already store it.

Isn't it guarding it already?

0xMayoor:
addInferenceToEpochGroupValidations is keyed on msg.Creator — it stops the same validator from going twice. different validator, different key, walks right past it. 

**the test in the PR shows exactly that.**

inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -48,6 +48,14 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 		k.LogInfo("Inference already invalidated", types.Validation, "inference", inference)
 		return &types.MsgValidationResponse{}, nil
 	}
+	if !msg.Revalidation && inference.Status == types.InferenceStatus_VALIDATED {
```
akup: We should remove `!msg.Revalidation` check here.
If message is already validated, we should not cause voting as it is very expensive operation

inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -48,6 +48,14 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
...
+		k.LogInfo("Inference already validated, skipping", types.Validation, "inferenceId", inference.InferenceId)
+		return &types.MsgValidationResponse{}, nil
+	}
+	if !msg.Revalidation && inference.Status == types.InferenceStatus_VOTING {
```
akup, line 81: If we are already voting we should not initiate one more voting here
remove  `!msg.Revalidation` check

0xMayoor:
@akup done for VALIDATED. keeping it on VOTING — revalidation messages during VOTING are the votes on the open proposal, removing that check would block voting from completing

akup:
@0xMayoor I've noticed that there are more issues on validation.
ClaimReward was discussed here: https://github.com/gonka-ai/gonka/pull/829
Also it was possible to send from any participant validation and revalidation messages, but only settled participants should do it.

So I've extended this PR and finished with this one:
https://github.com/gonka-ai/gonka/pull/967

Please take a look and review

inference-chain/x/inference/keeper/msg_server_validation.go
akup: It should be here, I was wrong. `revalidation = true` means message is a vote

akup:
Work on Inferences and Validations on current version is now legacy. All future work will be concentrated on devnets (shardchains). This PR shouldn't be high priority or probably should be closed in nearest future.
It was useful to see issues that should be handled in devchains

akup:
@0xMayoor please rebase. It could be merged but is low priority

akup (approved):

akup (approved):

patimen:
@0xMayoor this still looks like a simple, real fix, and I think we can get it in soon. Moved to v0.2.13.

patimen:
In triaging, I'm trying to defer any changes for the soon-to-be entirely deprecated old inference/validation path (devshards taking over).

PR #833, opened 2026-03-01
[P2] fix: default 400/422 with validation pass for incorrect enforced tokens format

tamazgadaev:
Adds a pre-validation check to ensure enforced tokens contain numeric token IDs rather than decoded text strings. Some executor responses return human-readable token strings in logprobs instead of integer IDs, which causes silent validation failures downstream.

This adds a hasNonNumericTokens helper that checks both the primary token and top-token alternatives via strconv.Atoi. If non-numeric tokens are detected, the inference is rejected early with a clear error message instead of proceeding with invalid data.


x0152:
Hi!
Quick question: Is 'EMPTY' sentinel flow (400/422, emptyButParseableResponsePayload) intended to fail with hasNonNumericTokens?

tamazgadaev:
> Hi! Quick question: Is 'EMPTY' sentinel flow (400/422, emptyButParseableResponsePayload) intended to fail with hasNonNumericTokens?

Great callout! Not intended

x0152:
There are a few more cases (including negative and out of range numbers):
```go
func TestHasNonNumericTokens(t *testing.T) {
	cases := []struct {
		name   string
		tokens completionapi.EnforcedTokens
		want   bool
	}{
		{"valid", completionapi.EnforcedTokens{Tokens: []completionapi.EnforcedToken{
			{Token: "42", TopTokens: []string{"1", "2"}},
		}}, false},
		{"text string", completionapi.EnforcedTokens{Tokens: []completionapi.EnforcedToken{
			{Token: "hello"},
		}}, true},
		{"negative primary", completionapi.EnforcedTokens{Tokens: []completionapi.EnforcedToken{
			{Token: "-1", TopTokens: []string{"3"}},
		}}, true},
		{"negative top token", completionapi.EnforcedTokens{Tokens: []completionapi.EnforcedToken{
			{Token: "42", TopTokens: []string{"-5"}},
		}}, true},
		{"out of range large", completionapi.EnforcedTokens{Tokens: []completionapi.EnforcedToken{
			{Token: "999999999", TopTokens: []string{"1"}},
		}}, false},
	}
	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			got := hasNonNumericTokens(tc.tokens)
			if got != tc.want {
				t.Fatalf("hasNonNumericTokens() = %v, want %v", got, tc.want)
			}
		})
	}
}
```

jacky6block (approved):
LGTM. Two minor suggestions:

  1. The existing 400/422 "treat as passed" block now silently covers two distinct cases — the original payload-rejection scenario and the new empty
  sentinel replay. Worth updating the comment to mention both, otherwise it's easy to misread the intent later.
  2. `isEmptySentinelTokens` and `hasNonNumericTokens` are pure functions with clear edge cases — would be good to have unit tests alongside the existing
   validation_test.go

akup (approved):
Copilot made a great job

x0152 (approved):

PR #835, opened 2026-03-01
Add Batch Transfer With Vesting (Closes #834)

### LINKED ISSUE #834: We could not send vesting to many users in one proposal
**Problem**
  We had only single-recipient vesting transfer.
  So if we needed to vest tokens to many addresses, we had to add many separate messages
  into one governance proposal.

  **Main limitation**
  Governance proposals have practical size/count limits.
  With many recipients, the proposal became too large (too many messages), and
  distribution could not be done in one clean operation.

  **Why we solved it**
  We needed a simple way to send vesting to many addresses at once, in one proposal, to
  avoid splitting into multiple proposals and reduce operational risk.

  **Target result**
  Add one batch vesting message so one proposal can include many recipients and execute
  the distribution in a single run.

Assign this task to me plz.

huxuxuya:
## PR: Add Batch Transfer With Vesting (Closes #834)

  This PR adds support for batch vesting transfers so governance can distribute vested
  tokens to many recipients in a single proposal message.

  ### Why

  Previously, vesting transfer supported only one recipient per message.
  For large distributions, proposals became too large because they required many
  messages, which caused operational complexity and proposal-size limitations.

  ### What’s included

  - Added new message and RPC:
      - MsgBatchTransferWithVesting
      - BatchVestingOutput
      - MsgBatchTransferWithVestingResponse
      - BatchTransferWithVesting endpoint
  - Implemented keeper handler for batch processing.
  - Reused/refactored shared vesting schedule logic to avoid duplication between single
    and batch flows.
  - Added validation and safety limits:
      - max recipients per batch
      - max total coin entries in batch
      - address/coin validation
      - empty/non-zero checks
  - Added events for batch operation.
  - Updated codec registration.
  - Regenerated protobuf/gRPC/API files.
  - Added tests for:
      - happy paths
      - invalid input
      - limits exceeded
      - duplicate recipient handling
      - insufficient sender balance
      - vesting epoch behavior

  ### Result

  Governance can now submit one message to distribute vesting to many addresses.

  Closes #834.


DimaOrekhovPS (approved):

inference-chain/x/streamvesting/keeper/msg_transfer_with_vesting.go
```diff
@@ -75,68 +75,69 @@ func (k msgServer) TransferWithVesting(goCtx context.Context, req *types.MsgTran
...
-			ParticipantAddress: req.Recipient,
+			ParticipantAddress: recipient,
 			EpochAmounts:       []types.EpochCoins{},
 		}
 	}
 
-	// Extend the schedule if necessary
-	requiredLength := int64(vestingEpochs)
-	for int64(len(schedule.EpochAmounts)) < requiredLength {
+	requiredLength := int(vestingEpochs)
```
patimen: Why are we casting to int(? It should always be a specific int (int32 or int64 usually), not a platform independent value

PR #838, opened 2026-03-02
fix/zero-participation-reward-bypass

0xMayoor:
`CheckAndPunishForDowntime` returns the full reward when total == 0. so if a participant has preserved PoC weight but never gets assigned any inference work (both InferenceCount and MissedRequests stay 0), the stat test never runs and they walk away with their full share of the epoch subsidy.

this also undermines the POC_SLOT eligibility fix from #659 — that checks `settle.RewardCoins == 0` to filter who gets slots, but since zero-participation nodes still receive rewards through this path, they pass that filter too.

same class as #658 but different code path. #659 fixed slot allocation, this is about the settlement reward calculation itself.

### DIFF
--- a/inference-chain/x/inference/keeper/accountsettle.go
+++ b/inference-chain/x/inference/keeper/accountsettle.go
@@ -51,7 +51,7 @@ func CheckAndPunishForDowntimeForParticipant(participant types.Participant, rewa
 
 func CheckAndPunishForDowntime(total, missed, reward uint64, p0 *types.Decimal) uint64 {
 	if total == 0 {
-		return reward
+		return 0
 	}
 	passed, err := calculations.MissedStatTest(int(missed), int(total), p0.ToDecimal())
 	if err != nil {
--- a/inference-chain/x/inference/keeper/zero_participation_reward_test.go
+++ b/inference-chain/x/inference/keeper/zero_participation_reward_test.go
@@ -1,0 +1,86 @@
+package keeper
+
+import (
+	"testing"
+
+	"github.com/productscience/inference/x/inference/types"
+	"github.com/stretchr/testify/require"
+)
+
+// TestZeroParticipationNoReward verifies that a participant with weight > 0
+// but InferenceCount=0 and MissedRequests=0 does NOT receive epoch rewards.
+//
+// Before this fix, CheckAndPunishForDowntime returned the full reward when
+// total==0, allowing participants to maintain PoC weight (via preserved
+// POC_SLOT nodes) while doing zero inference work and collecting subsidy
+// rewards every epoch.
+func TestZeroParticipationNoReward(t *testing.T) {
+	bitcoinParams := &types.BitcoinRewardParams{
+		InitialEpochReward: 285000000000000,
+		DecayRate:          types.DecimalFromFloat(-0.000475),
+		GenesisEpoch:       1,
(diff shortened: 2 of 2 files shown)


tcharchian:
@akup would you like to review?

akup:
> @akup would you like to review?

@tcharchian Will take a look in  a next 4 hours

akup:
@0xMayoor 
I think that giving reward to participants who didn't got any inference requests during epoch at all is expected behavior.
The host can't manipulate this stats, and he gets requests statistically. And if such situation appeared (when there is no network load) host still is honest.
We shouldn't punish honest hosts, it's not the host's fault.

So I think that if `total == 0`, we really should `return reward`, and this fix breaks this logic and makes possible punishment of honest hosts, that is strongly undesirable.

0xMayoor:
> We shouldn't punish honest hosts, it's not the host's fault.
> 
> So I think that if `total == 0`, we really should `return reward`, and this fix breaks this logic and makes possible punishment of honest hosts, that is strongly undesirable.

what punishment? returning 0 for zero work isn't punishment — it's just not getting paid when you didn't do anything. no slash, no jail, nothing on your record. you just don't collect rewards for an epoch where you contributed nothing.                                                                                                                                                      

if nobody on the network gets requests that epoch, everyone gets 0 equally. that's fair. the current code hands out full rewards for zero contribution which is the actual problem.

akup:
> > We shouldn't punish honest hosts, it's not the host's fault.
> > So I think that if `total == 0`, we really should `return reward`, and this fix breaks this logic and makes possible punishment of honest hosts, that is strongly undesirable.
> 
> what punishment? returning 0 for zero work isn't punishment — it's just not getting paid when you didn't do anything. no slash, no jail, nothing on your record. you just don't collect rewards for an epoch where you contributed nothing.
> 
> what punishment? returning 0 for zero work isn't punishment — it's just not getting paid when you didn't do anything. no slash, no jail, nothing on your record. you just don't collect rewards for an epoch where you contributed nothing.

Network, according to whitepaper provides 2 kind of rewards: for serving inferences, and for being ready to serve inference (providing the compute).

The reward that you have zeroed is for providing compute resources, there is another reward for inferences, don't mix them.

From the host side of view he provided his computation power for a day to network. It's the significant cost!
But network for some reason didn't give him any task, while compute powers was on. And this could be significant costs 250 USD or higher for the day. When you say there is no punishment it's completely unfair, miner lost history 250 USD and this is a significant punishment for nothing, for being honest

So you think if a host get one inference task (and he spent 3 seconds during a day to provide the inference), with a tiny prompt he will get the reward, but if network didn't distributed task at all he will not get at all.

Miner is mining, and he provides his computational resources

0xMayoor:
Thats a fair point. zeroing it is too aggressive, kills next-epoch slot eligibility too. 

but total==0 still bypasses the downtime check entirely when the network has active load and that's not right either.

needs a smarter fix. closing this.

akup:
> Thats a fair point. zeroing it is too aggressive, kills next-epoch slot eligibility too.
> 
> but total==0 still bypasses the downtime check entirely when the network has active load and that's not right either.
> 
> needs a smarter fix. closing this.

When the network has active load statistically host will never got `total == 0` :)
So let's give a large numbers do the job

PR #844, opened 2026-03-03
Fix slow node performance: optimize queries, add timing diagnostics

### LINKED ISSUE #818: Slow nodes investigation
### Discussed in https://github.com/gonka-ai/gonka/discussions/817

<div type='discussions-op-text'>

<sup>Originally posted by **tcharchian** February 27, 2026</sup>

Task: Multiple hosts reported node slowdowns in the last days. Need to identify common patterns and mitigate.

AlexeySamosadov:
## Summary

- **Fix `CountAllParticipants` memory issue**: Previously loaded ALL participant data into memory just to count them. Now uses iterator counting with O(1) memory instead of O(n).
- **Optimize `AggregateMLNodesFromModelSubgroups`**: Was calling `GetAllEpochGroupData()` which loaded ALL epoch group data across ALL historical epochs, then filtered for the current epoch. Now uses a new `GetEpochGroupDataForEpoch()` function that queries only the specific epoch's data via prefix iteration.
- **Add `Performance` logging subsystem**: New subsystem for all performance-related log messages, making it easy for operators to filter and monitor slow operations.
- **Add timing instrumentation to critical paths**:
  - `BeginBlock` / `EndBlock` with threshold-based warnings (>100ms / >200ms)
  - `SettleAccounts` with participant count and duration
  - `Prune` with per-pruner breakdown (>50ms threshold)
  - `onEndOfPoCValidationStage` / `onSetNewValidatorsStage` epoch transition timing
  - `GetSummaryByModelAndTime` slow query detection in dynamic pricing (>50ms)
- **Add `GetNodeDiagnostics` helper**: Lightweight diagnostic function that returns block height, epoch info, participant count, pruning state, and model count with timing, useful for operators investigating node performance.

## Test plan

- [ ] Verify `go build ./...` passes (confirmed locally)
- [ ] Run existing test suite to ensure no regressions
- [ ] Deploy to testnet and verify performance logs appear in node output
- [ ] Check that no new log spam occurs during normal operation (all timing logs are threshold-gated)

Closes #818


inference-chain/x/inference/keeper/dynamic_pricing.go
```diff
@@ -412,3 +420,46 @@ func (k *Keeper) GetAllModelCurrentPrices(ctx context.Context) (map[string]uint6
 	}
 	return result, nil
 }
+
+// GetNodeDiagnostics returns diagnostic information about the node's state for operators.
+// This is a lightweight read-only function that helps operators diagnose performance issues.
+func (k *Keeper) GetNodeDiagnostics(ctx context.Context) map[string]interface{} {
```
akup: Why it is in dynamic_pricing.go file? How it is related to dynamic pricing?

akup:
Do anybody know where QueryCountAllParticipantsRequest is used?
Even if we iterate all participants (it's all developer accounts + hosts) it could be very large and lead to DoS
Maybe it is better use counter at store, or protect this endpoint

inference-chain/x/inference/keeper/accountsettle.go
```diff
@@ -254,6 +257,11 @@ func (k *Keeper) SettleAccounts(ctx context.Context, currentEpochIndex uint64, p
 	}
 
 	if previousEpochIndex == 0 {
+		settleDuration := time.Since(settleStart)
+		k.LogInfo("SettleAccounts completed", types.Performance,
```
akup: Should be debug

inference-chain/x/inference/keeper/pruning.go
```diff
@@ -13,22 +14,42 @@ const (
...
+
+	valStart := time.Now()
 	err = k.GetPoCValidationsPruner(params).Prune(ctx, k, currentEpochIndex)
+	valDuration := time.Since(valStart)
 	if err != nil {
 		return err
 	}
+
+	totalDuration := time.Since(pruneStart)
+	if totalDuration > 50*time.Millisecond {
```
akup, line 45: When we exceed duration it could be LogWarn, but for instrumentation it is better to LogDebug when totalDuration is under threshold

inference-chain/x/inference/keeper/dynamic_pricing.go
AlexeySamosadov: Good catch — GetNodeDiagnostics is general node diagnostics, not related to dynamic pricing. I've moved it to its own `diagnostics.go` file in b4baccee.

inference-chain/x/inference/keeper/accountsettle.go
AlexeySamosadov: Agreed, this is routine instrumentation and shouldn't be Info. Changed to LogDebug in b4baccee.

inference-chain/x/inference/keeper/pruning.go
AlexeySamosadov: Makes sense — LogWarn when exceeding the threshold so it stands out in production, LogDebug otherwise to keep the timing data available without noise. Updated in b4baccee.

AlexeySamosadov:
@akup All 3 review comments have been addressed in commit b4baccee. Could you please re-review when you get a chance? Summary of changes:

1. **diagnostics.go** — Moved `GetNodeDiagnostics` out of `dynamic_pricing.go` into its own file
2. **accountsettle.go** — Changed SettleAccounts timing log from Info to `LogDebug`
3. **pruning.go** — `LogWarn` when duration exceeds threshold, `LogDebug` otherwise

PR #847, opened 2026-03-03
perf(inference): move hot-path logs from INFO to DEBUG

### LINKED ISSUE #780: [1/4] `StartInference` and `FinishInference`

AlexeySamosadov:
## Summary

Addresses the logging overhead identified in #780 (I.3: Move most logs of StartInference/FinishInference to DEBUG level).

Profiling showed ~11% of handler time spent in logging at INFO level, with `processInferencePayments` logging accounting for 25-48% of its execution time.

### Changes

**StartInference hot path:**
- Entry log, DevPubKey/TransferAgentPubKey logs -> DEBUG
- `validateTimestamp` verbose parameter log -> DEBUG
- `addTimeout` confirmation log -> DEBUG

**FinishInference hot path:**
- Entry log -> DEBUG

**Payment handling (`processInferencePayments` and downstream):**
- `PutPaymentInEscrow` success log -> DEBUG
- `PayParticipantFromModule` logs -> DEBUG
- `IssueRefund` log -> DEBUG
- `MintRewardCoins`, `BurnModuleCoins` logs -> DEBUG

**Stats recording (called per inference via `SetInference`):**
- `SetDeveloperStats` got-stat log -> DEBUG
- `setOrUpdateInferenceStatByTime` record tracking logs (3 calls) -> DEBUG
- `SetInference` updated-developer-stat success log -> DEBUG

**Dynamic pricing:**
- `RecordInferencePrice` success log -> DEBUG
- Removed heavy `inference` struct serialization from error log (replaced with lightweight fields)

### What is NOT changed
- All ERROR and WARN level logs remain unchanged for operational visibility
- Per-block logs in `UpdateDynamicPricing` (BeginBlocker) remain at INFO since they run once per block, not per inference
- Bookkeeper `LogSubAccountTransaction` has its own configurable log level

### Impact
With `log_level=info` (production default), this eliminates ~20 log calls per inference transaction. Combined with `log_format=json` (recommended), this should reduce logging overhead from ~11% to <1% of handler time.

Operators can still see all details by setting `log_level=debug`.

Closes #780

**Author:** AlexeySamosadov

### DIFF
--- a/inference-chain/x/inference/keeper/developer_stats_aggregation.go
+++ b/inference-chain/x/inference/keeper/developer_stats_aggregation.go
@@ -27,7 +27,7 @@ func (k Keeper) SetDeveloperStats(ctx context.Context, inference types.Inference
 		epochId = effectiveEpoch.Index
 	}
 
-	k.LogInfo("SetDeveloperStats: got stat", types.Stat,
+	k.LogDebug("SetDeveloperStats: got stat", types.Stat,
 		"inference_id", inference.InferenceId,
 		"inference_status", inference.Status.String(),
 		"developer", inference.RequestedBy,
--- a/inference-chain/x/inference/keeper/developer_stats_store.go
+++ b/inference-chain/x/inference/keeper/developer_stats_store.go
@@ -23,7 +23,7 @@ func (k Keeper) setOrUpdateInferenceStatByTime(ctx context.Context, developer st
 	timeKey := byInferenceStore.Get([]byte(infStats.InferenceId))
 	if timeKey == nil {
 		// completely new record
-		k.LogInfo("completely new record, create record by time", types.Stat, "inference_id", infStats.InferenceId, "developer", developer)
+		k.LogDebug("completely new record, create record by time", types.Stat, "inference_id", infStats.InferenceId, "developer", developer)
 		timeKey = developerByTimeAndInferenceKey(developer, uint64(inferenceTime), infStats.InferenceId)
 		bz, err := k.cdc.Marshal(&types.DeveloperStatsByTime{
 			EpochId:   epochId,
@@ -44,7 +44,7 @@ func (k Keeper) setOrUpdateInferenceStatByTime(ctx context.Context, developer st
 	)
 
 	if val := byTimeStore.Get(timeKey); val != nil {
-		k.LogInfo("record found by time key", types.Stat, "inference_id", infStats.InferenceId, "developer", developer)
+		k.LogDebug("record found by time key", types.Stat, "inference_id", infStats.InferenceId, "developer", developer)
 		err := k.cdc.Unmarshal(val, &statsByTime)
 		if err != nil {
 			return 0, err
@@ -64,7 +64,7 @@ func (k Keeper) setOrUpdateInferenceStatByTime(ctx context.Context, developer st
 		statsByTime.Inference.EpochId = infStats.EpochId
 		statsByTime.Inference.ActualCostInCoins = infStats.ActualCostInCoins
 	} else {
-		k.LogInfo("time key exists, record DO NOT exist", types.Stat, "inference_id", infStats.InferenceId, "developer", developer)
+		k.LogDebug("time key exists, record DO NOT exist", types.Stat, "inference_id", infStats.InferenceId, "developer", developer)
 		statsByTime = types.DeveloperStatsByTime{
 			EpochId:   epochId,
 			Timestamp: inferenceTime,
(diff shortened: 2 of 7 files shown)


gmorgachev:
That will be pretty hard to look at all logs in debug as cosmos sdk itself has a lot of records there. So if we move to debug, we wont be able to recods this logs at all
It seems like we need to introduce additional log level. Do you see if there is any simple way to introduce it?

AlexeySamosadov:
Closing: superseded by merged PR #812 which completely overhauled StartInference/FinishInference hot paths, making these log-level changes obsolete.

PR #851, opened 2026-03-03
fix(bls): remove self-validation fallback in group key validation

### LINKED ISSUE #848: Security: BLS group key validation falls back to self-validation when previous epoch data is missing
## Location

`inference-chain/x/bls/keeper/msg_server_group_validation.go` — lines 64–74

## Description

When `GetEpochBLSData` fails with `ErrEpochBLSDataNotFound` for the previous epoch, the code silently falls back to using the **new epoch's own data** as the previous epoch:

```go
previousEpochBLSData, err := ms.GetEpochBLSData(ctx, previousEpochId)
if err != nil {
    if errors.Is(err, types.ErrEpochBLSDataNotFound) {
        previousEpochBLSData = newEpochBLSData // fallback to self
    }
}
```

This creates **circular self-validation**:
- Participants are looked up from `previousEpochBLSData` — which is now the **new epoch's own participants**
- Slot public keys for per-slot signature verification come from `newEpochBLSData.SlotPublicKeys`
- The final aggregated signature is verified against `previousEpochBLSData.GroupPublicKey` — which is now the **new epoch's own group key**

Result: any participant in epoch N can sign and submit a valid group key validation for epoch N using epoch N's own keys — completely bypassing the intended cross-epoch chain-of-custody.

## When This Triggers

- Previous epoch BLS data was pruned (expected over time)
- State sync or snapshot restore on a new node
- First epoch after a chain upgrade where old BLS data is absent

## Impact

**Critical (security)** — the cryptographic chain-of-custody for group key transitions is bypassed. A single malicious participant can self-certify a new group key without legitimate cross-epoch consensus.

## Fix Direction

Return an error instead of silently falling back:

```go
if errors.Is(err, types.ErrEpochBLSDataNotFound) {
    return nil, fmt.Errorf("previous epoch %d BLS data not found, cannot validate group key", previousEpochId)
}
```

If bootstrapping for the very first epoch is required, handle it explicitly with a dedicated flag/check rather than a silent fallback.

Mayveskii:
When previous epoch BLS data was not found, the handler silently reassigned previousEpochBLSData = newEpochBLSData, allowing an epoch to certify its own group key against its own participant keys and its own group public key. This bypasses the entire chain-of-trust that cross-epoch validation is designed to enforce.

The fallback is removed. When previous epoch data is unavailable for any reason, the handler now returns an explicit error, which is the only correct and safe behavior.

Fixes #848

### DIFF
--- a/inference-chain/x/bls/keeper/msg_server_group_validation.go
+++ b/inference-chain/x/bls/keeper/msg_server_group_validation.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/binary"
-	"errors"
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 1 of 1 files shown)


x0152:
Thanks for working on this but EpochBLSData is not pruned by protocol logic, so this fallback is not reachable through normal consensus flow. This PR replaces a resilience fallback with a hard-stop failure mode (epoch can get stuck in DKG_PHASE_COMPLETED) without reducing a practical on-chain exploit

Mayveskii:
> Thanks for working on this but EpochBLSData is not pruned by protocol logic, so this fallback is not reachable through normal consensus flow. This PR replaces a resilience fallback with a hard-stop failure mode (epoch can get stuck in DKG_PHASE_COMPLETED) without reducing a practical on-chain exploit


You're right — BLS data migration was already handled in v0.2.8 (MigrateBLSData),
so this fallback is unreachable in practice. Closing.

PR #853, opened 2026-03-03
fix: validation silently dropped when model group data missing

0xMayoor:
GetEpochGroupData returns (data, found) but the error path reuses err from the previous GetEpochGroup call which already succeeded. when found=false the handler returns nil,nil — sdk treats it as success, validation silently lost. 
validator gets no error back, thinks everything went through.

return ErrEpochGroupDataNotFound instead of the stale nil.


akup:
Should be accepted

akup:
@0xMayoor please rebase

0xMayoor:
> @0xMayoor please rebase

can't rebase @akup,  the code path this fixes got rewritten in #874. the nil-error bug is gone since GetEpochGroupData was replaced with the transient cache.

PR #854, opened 2026-03-03
fix(payloadstorage): advance minPruned only after successful prune

### LINKED ISSUE #850: Bug: ManagedStorage silently skips failed epoch pruning — minPruned advanced before goroutines complete
## Location

`decentralized-api/payloadstorage/managed_storage.go` — lines 129–138

## Description

In `ManagedStorage.cleanup()`, `m.minPruned` is advanced to `threshold` **before** the pruning goroutines complete:

```go
for epoch := m.minPruned; epoch < threshold; epoch++ {
    go func(e uint64) {
        if err := m.storage.PruneEpoch(context.Background(), e); err != nil {
            logging.Warn("Auto-prune failed", types.PayloadStorage, "epochId", e, "error", err)
            // error only logged — epoch is silently skipped
        }
    }(epoch)
}
m.minPruned = threshold  // ← advanced immediately, goroutines still running
```

### What goes wrong

If any `PruneEpoch` goroutine fails (DB connection lost, file I/O error, PostgreSQL timeout), that epoch is **permanently skipped**:

- `m.minPruned` is already past it
- The `maxPruneLookback = 10` guard does not help — it only prevents jumping too far forward on first run, not recovering already-skipped epochs
- Next `cleanup()` tick starts from the new `m.minPruned` — the failed epoch is never retried

### Secondary issue

`cleanup()` holds `m.mu.Lock()` while spawning goroutines, but the goroutines themselves run **without the lock**. If goroutines from a previous 30-second tick are still running when the next tick fires, multiple goroutines can call `PruneEpoch` for the same epoch concurrently.

## Impact

**High** — off-chain payload data for failed epochs accumulates on disk indefinitely. Under sustained prune failures (e.g. intermittent DB connectivity), this leads to unbounded disk growth and eventual disk exhaustion on API nodes.

## Fix Direction

Only advance `m.minPruned` after confirming successful pruning. Simplest correct approach — run pruning synchronously within the lock:

```go
for epoch := m.minPruned; epoch < threshold; epoch++ {
    if err := m.storage.PruneEpoch(context.Background(), epoch); err != nil {
        logging.Warn("Auto-prune failed", types.PayloadStorage, "epochId", epoch, "error", err)
        break // stop advancing minPruned past the failed epoch
    }
    m.minPruned = epoch + 1
    logging.Info("Auto-pruned epoch", types.PayloadStorage, "epochId", epoch)
}
```

Alternatively, if async pruning is required for performance, track per-epoch completion via a channel or atomic and only advance `m.minPruned` for contiguously completed epochs.

Mayveskii:
## Summary

Closes #850

`minPruned` was advanced to `threshold` synchronously before pruning goroutines completed.
If any goroutine failed, that epoch was silently skipped forever — no retry on the next cycle.

## Fix

Removed the bulk advance. Each goroutine now advances `minPruned` by one only after
its own `PruneEpoch` call succeeds, so failed epochs remain eligible for retry on
the next 30-second cleanup tick.

## Files changed

- `decentralized-api/payloadstorage/managed_storage.go` — `cleanup()`

### DIFF
--- a/decentralized-api/payloadstorage/managed_storage.go
+++ b/decentralized-api/payloadstorage/managed_storage.go
@@ -22,18 +22,21 @@ type cachedEntry struct {
 }
 
 // ManagedStorage wraps PayloadStorage with read caching and automatic epoch pruning.
-// - Caches Retrieve results to reduce disk I/O during validation bursts
-// - Automatically prunes old epochs in background (only last 10 epochs, older data requires manual prune)
+//   - Caches Retrieve results to reduce disk I/O during validation bursts
+//   - Automatically prunes old epochs in background (only last 10 epochs, older data requires manual prune)
+//   - Failed prunes are isolated in failedEpochs and retried on each cleanup tick,
+//     so a single persistent failure does not block pruning of subsequent epochs.
 type ManagedStorage struct {
 	storage      PayloadStorage
 	retainCount  uint64
 	cacheTTL     time.Duration
 	maxCacheSize int
 
-	mu        sync.RWMutex
-	cache     map[string]*cachedEntry
-	maxEpoch  uint64
-	minPruned uint64
+	mu           sync.RWMutex
+	cache        map[string]*cachedEntry
+	maxEpoch     uint64
+	minPruned    uint64
+	failedEpochs map[uint64]bool
 }
 
 func NewManagedStorage(storage PayloadStorage, retainCount uint64, cacheTTL time.Duration) *ManagedStorage {
@@ -47,6 +50,7 @@ func NewManagedStorageWithSize(storage PayloadStorage, retainCount uint64, cache
 		cacheTTL:     cacheTTL,
 		maxCacheSize: maxCacheSize,
 		cache:        make(map[string]*cachedEntry),
+		failedEpochs: make(map[uint64]bool),
 	}
 	go m.cleanupLoop()
 	return m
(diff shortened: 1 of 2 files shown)


decentralized-api/payloadstorage/managed_storage.go
```diff
@@ -126,16 +126,22 @@ func (m *ManagedStorage) cleanup() {
...
-					logging.Info("Auto-pruned epoch", types.PayloadStorage, "epochId", e)
+					return
 				}
+				logging.Info("Auto-pruned epoch", types.PayloadStorage, "epochId", e)
+				// Advance minPruned only after a successful prune, and only forward
+				m.mu.Lock()
+				if e+1 > m.minPruned {
+					m.minPruned = e + 1
+				}
+				m.mu.Unlock()
```
akup, line 169: Advancing by one can cause the issue, as this is async function we will skip unpruned epochs if some other are pruned

For example we have epoch X = m.minPruned unpruned (failed with error), but X+1 pruned. Both was runned in parallel with go func
We will end this by moving m.minPruned to X + 1, that is already pruned, and X is never pruned on retries.

It is better do not advance m.minPruned at all if any of async prunes failed.
On the next retry we should find all unpruned tables from [m.minPruned, threshold) and prune them, or just iterate (that could be cheaper) from m.minPruned to threshold again.
But we can set `m.minPruned = threshold` only when there was no errors at all.

So just count errors and update m.minPruned when all go func completed and `errorsCount == 0`

Mayveskii: https://github.com/gonka-ai/gonka/commit/c79d50213a60d25aa879f9fcd261a661099c9e48

akup: > [c79d502](https://github.com/gonka-ai/gonka/commit/c79d50213a60d25aa879f9fcd261a661099c9e48)

Checked. Now it's ok

akup:
The PR is ok.
Tests should be runned

Mayveskii:
excuse me , missclicked...

patimen:
Thanks for digging into the skipped-prune problem here. After a closer look and an AI review run, I think this needs a redesign before we move it forward.

The main issue is that the new all-or-nothing `minPruned` advancement can freeze pruning entirely if any single epoch keeps failing. In that case `minPruned` never advances, the same batch keeps getting retried on every cleanup tick, and as `threshold` grows we end up spawning more and more prune goroutines for already-successful epochs as well. That turns a local retry problem into a global pruning stall, with a real risk of unbounded resource growth.

I still think the underlying concern is valid, but this PR in its current form seems to replace one failure mode with a worse one. A safer direction would be to track partial progress up to the first contiguous failure, or otherwise isolate failed epochs without blocking the global pruning watermark.

I’m moving this to a needs-redesign / needs-info state rather than closing it outright, because the underlying bug is real and worth fixing.

Mayveskii:
>RE 

Thank you for the detailed and precise answer. I'll fix this in the next commit to this issue. I'll also do an additional compatibility check. I'll fix this as soon as possible to avoid delaying the review process.

Mayveskii:
@patimen 

John, I've redesigned it per your feedback. The core issue was break-on-failure creating head-of-line blocking  one stuck epoch froze all subsequent pruning.

New approach: continue past failures, isolate them in failedEpochs, retry each tick. minPruned advances past gaps so the watermark isn't blocked. Pruning is now sequential , so the unbounded-spawn concern is also resolved. PruneEpoch is idempotent across all backends, making retries safe.

 Ready for re-review

Mayveskii:
The Devshard failure is a pre-existing flaky test. It has a timing race the 120ms hard-timeout timer starts before the goroutine even runs, but the 50ms stall-detection timer only starts after the first chunk arrives. On a slow CI runner the hard timeout fires first, so the test gets "context canceled" instead of the expected stall error.

Mayveskii:
@patimen 
Hi there, John! Please, enlighten me, is this a correct solution looks for you ?  

And what is your view about #1336 ? 
In my opinion #1336 is better much than it is - #854

Guess it should be closed... Give your feedback, please.

PR #856, opened 2026-03-03
Feature/821 continuous poc complete

Mayveskii:
## Summary

Builds on PR #845 (AlexeySamosadov) to deliver a production-ready continuous PoC system.
PR #845 introduced the data structures; this PR makes them work end-to-end and fixes a
critical serialisation bug in #845.

### 1. Fix: `ContinuousPocParams` serialisation bug in PR #845

`params.pb.go` added `ContinuousPocParams` to the `Params` struct but never updated
`MarshalToSizedBuffer`, `Size`, or `Unmarshal`. The field silently vanished on every
round-trip — making continuous PoC impossible to enable via governance.

### 2. Pruning for continuous PoC collections

Extends `PruningState` (proto fields 4–6) with epoch counters for commits, challenges,
and summaries. Adds three `Pruner` implementations wired into `Keeper.Prune()`:

- `GetContinuousPoCCommitsPruner` — `Triple[uint64, AccAddress, int64]`
- `GetContinuousPoCChallengesPruner` — same key shape
- `GetContinuousPoCEpochSummariesPruner` — `Pair[uint64, AccAddress]`

All three reuse existing `PocDataPruningEpochThreshold` / `PocPruningMax` parameters.

### 3. Epoch settlement integration

`WeightCalculator` gains a `ContinuousPoCEpochSummaries` map. During `ComputeNewWeights`,
`EffectivePocWeight` from each participant's `ContinuousPoCEpochSummary` is added to their
standard PoC weight — unless `PenaltyApplied` is set (failed/expired challenge).

### 4. Merkle-proof nonce validation

- **`SubmitContinuousPoCCommit`** — validates params, enforces `MaxCommitsPerEpoch` and `MinNoncesPerCommit`, upserts `ContinuousPoCEpochSummary`.
- **`IssueContinuousPoCChallenges`** (EndBlock) — deterministically samples commits using `sha256(app_hash || block_height)` as entropy.
- **`RespondContinuousPoCChallenge`** — verifies the Merkle proof and resolves the challenge.
- **`ExpireContinuousPoCChallenges`** (EndBlock) — applies `PenaltyApplied` to summaries whose deadline passed.

## Testing

- 21 unit tests in `x/inference/types` — all pass
- Full module build verified on Debian bookworm / Go 1.22.8

## Note on PR #845

This PR intentionally overlaps with #845 — it is a superset. If #845 merges first, I will
rebase and remove the duplicate definitions. Happy to coordinate with the author.

Closes #821

Issue #839, opened 2026-03-02
LogInfo tests on testnet for StartInference and FinishInference


maria-mitina:
To investigate the impact of log_format = "json" vs log_format = "plain" on node performance, a [number of experiments were executed on Testnet](https://docs.google.com/spreadsheets/d/1GkV6tn5tgQ3eL7VgAF4JaBWerqsN-fGHc6KEKNigFog/edit?pli=1&gid=586520582#gid=586520582). This is not compared to [mainnet performance](https://github.com/gonka-ai/gonka/issues/780#issuecomment-3972000833), however in isolation supports the same conclusion.  
We could achieve around 3x improvement under inference to performance of LogInfo, when switching to log_format = "json".

<img width="1237" height="470" alt="Image" src="https://github.com/user-attachments/assets/dc9b7e25-3562-4f70-b058-2c6fe8ed52ac" />

maria-mitina:
Should we continue with the log_level=info and log_level=error?

hleb-albau:
x3 for start/finish inference, or x3 for logging?

maria-mitina:
@hleb-albau 3x improvement under inference to performance of LogInfo

Issue #820, opened 2026-02-27
Investigate missed inference on some nodes (root causes + mitigation)


tcharchian:
### Discussed in https://github.com/gonka-ai/gonka/discussions/817

<div type='discussions-op-text'>

<sup>Originally posted by **tcharchian** February 27, 2026</sup>

Task: Some nodes experience missed inference events. Likely multi-cause, needs community participation.

AlexeySamosadov:
PR: https://github.com/gonka-ai/gonka/pull/843

Mayveskii:
Measured data from live network may help narrow root causes here.

Epochs 161–191, 2,503,595 inferences:
Miss rate: 3.25% (81,360 misses)
Completion rate: mean 90.4%, σ=7.4%, range 72–99%

The σ=7.4% variance is the signal — not the mean.
Some nodes miss 28% of assigned inferences while others miss 1%.
GetRandomExecutor routes to both equally regardless.

Phase 4 of GiP #860 proposes GetQualityWeightedExecutor — routes traffic 
proportional to L9 completion rate. Projection: σ ↓40% as high-miss nodes 
receive less traffic and face economic incentive to improve.

Design + data: docs/specs/inference-quality-protocol.md (PR #859 branch)
Discussion: #860

PR #859, opened 2026-03-04
feat(semantic-cache): L2 quality gate pipeline — adaptive coherence floor + hub loop closure [no extra GPU]

Mayveskii:
##Even u aidn't say what i just wrote ? ? 🤦‍♂️

What's this just wrote —-

## feat(semantic-cache): L2 quality gate pipeline + binary singularity deploy stack

> Design document: [GiP #860](https://github.com/gonka-ai/gonka/discussions/860)
> Agent SDK: [gonkalabs/gonka-agent](https://github.com/gonkalabs/gonka-agent/tree/dev/binary-singularity)
> Full guide: [deploy/binary-singularity/GUIDE.md](https://github.com/Mayveskii/gonka/blob/feature/gip-semantic-cache-trust-layer/deploy/binary-singularity/GUIDE.md)

### What this PR adds

**Semantic cache infrastructure (Phase 0):**
- Two-level cache: L1 (PromptHash exact-match) + L2 (cosine similarity, all-MiniLM-L6-v2, CPU-only)
- `CacheQualityWeight` — additive PoC bonus at epoch settlement, governance-controlled
- 4-stage quality gate pipeline: similarity → verifier → coherence floor → loop closure
- 62/62 tests PASS, no GPU required

**Binary singularity deploy stack (4 tiers):**
- LITE: Docker Compose, reproducible convergence proof
- MEDIUM: + opengnk proxy + quality-middleware
- HARD: K3s mesh (1 server + 3 agents)
- PRODUCTION: on real Gonka node (deploy/join/ stack) with GNK tokens

**Quality middleware:**
- Mesh CPU pool with int8-quantized cosine search
- `POST /quality/search` + `POST /quality/slots/share` for inter-participant slot exchange
- L6/L8/L9/DX measurement per request

**gonka-agent SDK:**
- PatternSlot store with atomic persistence
- Mesh search + slot sharing
- Raw binary input from any file (`BS_RAW_INPUT`)
- Prebuilt binary: ELF x86_64, static, stripped, 6 MB

**Conflict resolution with upgrade-v0.2.11:**
- PruningState: fields 4-8 (EpochGroup upstream + ContinuousPoC + CacheQuality)
- keys.go: prefixes 47-51, no collisions
- pruning.go: all 8 pruners wired
- upgrades.go: setEpochParticipantsSets + initNewFeatureParams

### Proven results (Bookworm, CPU-only)

| Exp | Runs | PQM | Slots | Peak RAM |
|-----|------|-----|-------|----------|
| 2 | 9,216 | 0.988 | 4 | 19.25 MB |
| 3 (K3s mesh) | 15,360 | **1.001** | 6 | 23.1 MB |
| 4 (raw data) | 11,520 | **1.020** | 197 | 20.8 MB |

**PQM > 1.0** = binary layer quality exceeds single GPU inference.

### Key parameters

- Optimal L2 threshold: **4250 bps** (F1=0.986) — current 7500 loses 64% of valid hits
- Hit rate formula: `H = r × (1/M) × (1-s)`
- Specialization: M=1 → **571×** hit rate vs M=571 shared
- Economics: **~$155,800/year** GPU savings (full protocol, H100 $2.50/h)
- Coherence floors: sim≤6250 → 3000 · sim 6250-8000 → 4000 · sim>8000 → 4500
- Loop closure margin: **-800 bps**

### Relation to #874

PR #874 introduces `ValidationParams.claim_validation_enabled` and `TransientEpochDataCache` — the chain-side infrastructure our `CachedResult.BLSSignature` field is waiting for. The configurable-validation pattern mirrors our Gate 1/Gate 2 design: expensive checks are opt-in and skippable.

### Related

[PR #856](https://github.com/gonka-ai/gonka/pull/856) (Continuous PoC) · [PR #812](https://github.com/gonka-ai/gonka/pull/812) (perf) · [PR #793](https://github.com/gonka-ai/gonka/pull/793) (EpochGroupCache) · [GiP #816](https://github.com/gonka-ai/gonka/issues/816) (Node Manager) · [GiP #840](https://github.com/gonka-ai/gonka/issues/840) (Prometheus) · [Issue #820](https://github.com/gonka-ai/gonka/issues/820) (Missed inferences) · [Issue #839](https://github.com/gonka-ai/gonka/issues/839) (log_format=json)

### Test checklist

- `go test ./semanticcache/...` — 62/62 PASS
- `go test ./examples/quality-middleware/...` — all PASS
- `go build ./...` — clean (Go 1.22)
- Deploy: `cd deploy/binary-singularity/lite && ./run.sh`

---

### Update: binary singularity layer shipped (2026-03-11)

@blizko @gmorgachev @tcharchian @akup @patimen

Work continued in `dev/binary-singularity` branch. All conflicts with `upgrade-v0.2.11` fully resolved. Summary of what's new since the last discussion:

**Deployed and tested:**
- Full deploy stack: LITE / MEDIUM / HARD (K3s) / PRODUCTION — all reproducible, no GPU
- `gonka-agent` SDK with binary slot store + inter-participant mesh sharing
- 4 experiments on Bookworm: PQM progressed 0 → 0.988 → 1.001 → **1.020**
- Raw binary ingestion: 676KB developer workflow → 197 PatternSlots, zero pre-processing

**Addressing @akup's dev-chains point:**
The deploy stack is fully modular — runs independently of main chain via Docker/K3s. `CacheQualityParams.Enabled = false` by default. No chain changes required to test. Any operator can `cd deploy/binary-singularity/lite && ./run.sh` and reproduce all 4 experiments.

**Addressing @blizko's asymmetric advantage concern:**
`MaxWeightFractionBps = 30%` caps bonus regardless of cache size. Operator verification: `/admin/v1/cache/stats` (Source A) vs on-chain `CacheQualityEpochSummary` (Source B) cross-check catches inflation. The mesh pool distributes slots across participants — not just accumulates on one node.

**Addressing @gmorgachev's prefix caching point:**
Agreed — vLLM KV-cache and this cache are different layers (activations vs request/response). They coexist. The binary singularity layer adds a third level: PatternSlot matching via CPU cosine (~5ms) before any inference call. When a slot hits, no GPU work at all — not even a forward pass.

**Agent SDK:** [`gonkalabs/gonka-agent @ dev/binary-singularity`](https://github.com/gonkalabs/gonka-agent/tree/dev/binary-singularity)
**Guide for all levels:** [`deploy/binary-singularity/GUIDE.md`](https://github.com/Mayveskii/gonka/blob/feature/gip-semantic-cache-trust-layer/deploy/binary-singularity/GUIDE.md)

Ready for testnet validation. One ask remains: enable `CacheQualityParams` on testnet for 1 epoch — `repeat_fraction` becomes a measured number, not an assumption.


Mayveskii:
Quality of computation as an economic incentive: what I had in mind building this PR
When I built this, the cache was the mechanism — but the original intent was broader. tokenomics-v2/bitcoin-reward.md explicitly named it as an open gap: "No incentive for model diversity or utilization quality." CacheQualityWeight is the first concrete implementation of that direction.
The logic is straightforward: a node that earns more for the quality of its result starts optimizing for it. Better inference → more reuse → higher epoch weight → more rewards → better inference. A closed economic loop, not a declaration.
For this loop to work, participants need to be able to send quality signals without friction. The architecture here already supports this, or with minimal additions:
API — inference_id is in every response; feedback comes back as a single tag on the next request through the same OpenAI-compatible contract
SDK — library for popular stacks, wraps the call and collects implicit signals automatically
Widget — embeddable UI component, zero code required from the developer
Implicit signals — DAPI already sees session depth and semantic repetition with no integration needed
Webhook — node pushes an event to the developer app when inference closes, app returns outcome automatically
CLI / authz — MsgSubmitCacheQualitySummary is already registered in InferenceOperationKeyPerms with full Grant→Exec→Revoke flow for node operators
CacheQualityParams + epoch settlement are generic enough to carry any quality signal beyond cache hits. This is the foundation for the utilization bonuses tokenomics-v2 outlined as the next step but didn't implement.
Would love to hear whether extending CacheQualityParams toward a pluggable quality axis registry makes sense as a follow-on GiP, or whether that's better scoped separately.

blizko:
The overall idea of keeping a Prompt cache is redundant to the kv-cache implementation of vllm.
Rewarding validators for keeping a cache results in asymmetric advantage for validators with higher GPU capacity. Higher GPU capacity => Higher probability of getting requests => Higher cache utilisation => Even higher reward potential

Mayveskii:
> The overall idea of keeping a Prompt cache is redundant to the kv-cache implementation of vllm. Rewarding validators for keeping a cache results in asymmetric advantage for validators with higher GPU capacity. Higher GPU capacity => Higher probability of getting requests => Higher cache utilisation => Even higher reward potential

A few clarifications so we're aligned:

**1. Where this implementation stands**

The cache is **off by default** and gated by `CacheQualityParams` in governance. What's in this PR is a **building block** for the broader direction: improving quality of computation inside the network and rewarding it. The "final" design (e.g. pluggable quality axes, other signals beyond cache hits) will build on this. Right now the goal is **community feedback on the idea and the direction**, not a full production code review. The broader vision (quality axis registry, measurable useful work) is in [GiP discussion #860](https://github.com/gonka-ai/gonka/discussions/860).

**2. KV-cache (vLLM) vs this cache — different layers**

They're not redundant; they sit at different levels:

- **vLLM KV-cache**: caches **key–value states inside the model** during the forward pass. It speeds up inference when prompts share a prefix (e.g. same system prompt, batched requests). The model still runs; you're reusing intermediate activations.
- **This cache**: **request/response level**. On an L1 (exact `PromptHash`) or L2 (semantic similarity) hit we **return a stored response and do not start inference**. No forward pass, no GPU work for that request — just a lookup (hash or embedding index) and a response.

So KV-cache optimises "how much work we do per inference"; this cache optimises "whether we run inference at all" for repeat or semantically similar prompts. Both can coexist: KV-cache for requests that do hit the GPU, this cache for requests that don't.

**3. Asymmetric advantage**

The concern that "more traffic → more cache hits → more `CacheQualityWeight`" is valid. That's a **reward-distribution** effect, not a compute-load effect: serving a cache hit is cheap (no inference). The design already allows mitigating it (feature off by default, governance params, bounds in the submission handler). If we move forward, we can add caps or normalisation so the bonus doesn't dominate. I'd rather lock the mechanism and then tune parameters with governance than block the direction on that.

The main points as I understand them: computations that are identical at a given level should not be recomputed; resources should be allocated instead. A deeper view is outlined in GiP https://github.com/gonka-ai/gonka/discussions/860 , which proposes classifying computations by quality and incentivizing nodes and users to use them more efficiently overall, without extra financial load.

Mayveskii:
This PR closes the open gap named in tokenomics-v2: _"No incentive for model diversity or utilization quality."_

## What it adds

- **Two-level semantic cache** in DAPI: L1 exact-match on `PromptHash` (SHA-256, cryptographically certain), L2 cosine similarity via `MLNodeEmbedder` (CPU-only `all-MiniLM-L6-v2`, 384 dims — no GPU lock)
- **`CacheQualityWeight`** — additive PoC bonus at epoch settlement, governance-controlled via `CacheQualityParams` (cap: `MaxWeightFractionBps` = 30%)
- **`QualityReporter`** — submits `MsgSubmitCacheQualitySummary` once per epoch with `CacheReuseCount`, `OriginalComputeCount`, `AvgSimilarityBps`, `EmbeddingModelVersion`
- **`GET /admin/v1/cache/stats`** — atomic read-only hit/miss counters on operator-only admin port `:9200`. `Stats()` / `HitRate()` use `atomic.LoadInt64` — zero locks, zero side effects, nil-safe.
- **`InMemoryCacheStore`** — default backend, zero external dependencies, no new services
- **Authz delegation** — `MsgSubmitCacheQualitySummary` added to `InferenceOperationKeyPerms`, enabling Grant → Exec → Revoke for operational keys (#760, tested in #857)
- **`docs/specs/semantic-cache.md`** — full spec: architecture, chain/API components, operator verification layer, k8s specialization economics, developer simulation, known limitations
- **20 tests**, no GPU, no ML-node, no chain required — per Developer Simulation in `docs/specs/semantic-cache.md`

## How it works — k8s hosts vs bare-metal

The cache is a **DAPI-level feature** — it works on any deployment: k8s pod, Docker Compose, bare-metal. The node operator does not need to change anything; the feature activates via governance (`CacheQualityParams.Enabled`).

What changes with k8s (GiP #816): nodes deployed with **one model per node** receive 100% of that model's traffic via `GetRandomExecutor` routing (M=1). This maximises cache hit rate organically — no configuration, no tuning. Bare-metal hosts with shared models still benefit, but with proportionally lower hit rates (M nodes sharing traffic → hit_rate/M).

## Economic scenario matrix (live network data)

All calculations use **measured** baseline from `gonka.gg/api/public` (epoch 190):
- 75,016 inferences/epoch avg (2,325,522 over 31 epochs)
- 129 participants, 3,188 GPUs (H100-eq: 3,235)

**Formula:**
```
effective_hit_rate = repeat_fraction × (1/M) × (1 - stream_fraction)
reuseCount/epoch  = effective_hit_rate × (75,016 / M)

CacheQualityWeight: additive bonus to baseCount, cap = MaxWeightFractionBps = 30%
```

**Scenarios:**

| # | Scenario | M | repeat% | stream% | hit_rate | reuseCount/ep | GPU saves/ep | +Weight |
|---|---|---|---|---|---|---|---|---|
| 0 | Cold start (epoch 1) | — | 0% | — | 0 | 0 | 0 | +0% |
| 1 | Baseline, no cache | — | — | — | 0 | 0 | 0 | +0% |
| 2 | Testnet genesis (2 nodes) | 2 | 30% | 0% | 0.15 | 11,252 | 11,252 | partial |
| 3 | Testnet + streaming 30% | 2 | 30% | 30% | 0.105 | 7,876 | 7,876 | lower |
| 4 | Shared model (10 nodes) | 10 | 30% | 10% | 0.027 | 2,025 | 20,253 | minimal |
| 5 | Multi-model node (3 models, M=1 each) | 1 | 30% | 0% | 0.30×3 | 67,515 | 67,515 | **+30% cap** |
| 6 | Specialized (1 node, unique model) | 1 | 30% | 0% | 0.30 | 22,505 | 22,505 | **+30% cap** |
| 7 | Specialized + high repeat demand | 1 | 60% | 0% | 0.60 | 45,010 | 45,010 | **+30% cap** |
| 8 | 20% of network specialized (33 nodes) | 1 ea | 40% | 5% | 0.38 | 28,506 | **940,698** | **+30% cap** |
| 9 | 50% of network specialized (81 nodes) | 1 ea | 40% | 5% | 0.38 | 28,506 | **2,308,986** | **+30% cap** |

**Network impact at scale:**

| Parameter | Scenario 8 (20% specialized) | Scenario 9 (50% specialized) |
|---|---|---|
| GPU saves/epoch | 940,698 | 2,308,986 |
| GPU saves/year (52 ep) | ~48.9M | ~120.1M |
| Operator income delta | **+30% cap per node** | **+30% cap per node** |

**Assumptions and how we verify them:**

| Assumption | Risk | Verification |
|---|---|---|
| repeat_fraction = 30-60% | Real % unknown — may be lower | Live inference with `X-Cache` headers |
| stream% = 5-30% | If stream > 50% → hit_rate ≈ 0 | DAPI logs `stream:true` count |
| M=1 for specialization | Node may not be unique → M > 1 | `node-config` check on testnet |
| reuseCount self-reported | Community concern @blizko  | stats(A) vs chain(B) ± 5% via DAG |
| MaxWeightFractionBps = 30% | **Confirmed** in `docs/specs/semantic-cache.md` | governance default |

## Operator verification layer — DAG + Prometheus (GiP #816, #840)

Self-reported `CacheReuseCount` can be cross-checked at each epoch boundary:

```
[Epoch boundary] → DAG / k8s CronJob / Go ticker
  ├── Source A: GET /admin/v1/cache/stats       → {hits, misses, hit_rate}
  ├── Source B: chain CacheQualityEpochSummary  → reuseCount (self-reported)
  └── Source C: Prometheus scraper (GiP #840)   → independent time-series

  stats(A).hits ≈ chain(B).reuseCount ± 5%   →  self-report validated
  divergence > threshold                      →  operator alert
```

`/admin/v1/cache/stats` makes Source A possible. Prometheus exporter (GiP #840) already reads admin API on `:9200` — integrates without new infrastructure.

This directly addresses blizko's asymmetric advantage concern: nodes cannot inflate `reuseCount` without the divergence appearing in Source A vs Source B cross-check, while `MaxWeightFractionBps` (30%) bounds the maximum gain even without verification. The verification layer is fully reproducible on k8s infrastructure (GiP #816); bare-metal operator verification remains an open question for community discussion.

## Economic flow end-to-end

```
User request → DAPI
  ├── [L1/L2 HIT] → verify ResponseHash (sha256) → serve cached result
  │     └── MsgFinishInference → QualityReporter.RecordReuse
  │           └── epoch boundary → MsgSubmitCacheQualitySummary → chain
  │                 └── chainvalidation.go: CacheQualityWeight added to baseCount
  │                       └── +EpochGroup power → model assignments + Reward Coins
  └── [MISS] → GPU inference → MsgFinishInference → StoreResult(embedding, payload)
```

Streaming (`stream: true`) bypasses cache entirely — SSE format cannot be replayed.

## Scientist-Validator summary

| Layer | Status | Evidence |
|---|---|---|
| Mechanism correctness | **PROVEN** | 20/20 tests PASS (0.103s), all fail-paths verified |
| ResponseHash integrity | **PROVEN** | `TestHTTP_L1_VerifyFail_FallThrough` — tamper → fallthrough, not crash |
| Governance live update | **PROVEN** | `TestUpdateCacheParams_ModelVersionChange` — immediate invalidation |
| Admin stats endpoint | **PROVEN** | `go build ./...` clean, nil-safe, atomic reads |
| Network baseline | **REAL** | 75,016 inferences/ep (epoch 190, live API data) |
| Economic impact projection | **CALCULATED** | 9 scenarios with measured baseline, honest assumptions table |
| Live X-Cache / reuseCount / CacheQualityWeight delta | **PENDING** | Requires `CacheQualityParams.Enabled = true` + tokens for live inference |

## Next step (GiP #860)

`CacheQualityParams` + epoch settlement pattern is generic enough to carry additional quality axes (session continuity, explicit feedback, verifiable outcome) without rebuilding the reward layer. See [#860](https://github.com/gonka-ai/gonka/discussions/860).

@tcharchian could you add this to the v0.2.11 milestone? It builds directly on the continuous PoC foundation from #856.

akup:
My point is that as it is the experimental building block for phase 0, it shouldn't be merged.

There is a large proposal - dev-chains. It is proposed to be very modular from the main chain, and be not coupled with existing api node.
Moreover it is implementation free, there even can be different dev-chains versions in parallel, that is nice for experiments.

We should not merge this to main, all experimental work could and should be done in experimental dev-chain.

It is hard to say that anything measured on a real data, as currently network not fully functions as a real inference engine and there is no real data, as even developer accounts are limited and there are a lot test prompts.

What is repeat_fraction and why it is 30-60% ? Need of this mechanics are mostly depend on this repeat_fraction, but how we can estimate it? What do you propose, explain, what is the repeat_fraction.

How do you propose implement L2: cosine similarity ? If it didn't hit the GPU where do you get the vectors from?

@Mayveskii 
Please close this PR, and please start the branch for experimental features when dev-chains will be published. One of their purpose is this modularity for experiments

akup:
Why do you think users are asking same prompts (with same contexts)?
Why do you think this cache could be useful at all?

Mayveskii:
> Why do you think users are asking same prompts (with same contexts)? Why do you think this cache could be useful at all?

@akup direct answers:

L2 vectors: first MISS → GPU → StoreResult(embedding). 
L2 reuses those stored vectors on subsequent requests. No extra GPU call.

repeat_fraction: measured indirectly today.
Same prompt sent 3× to node3: latencies 0.72s / 0.24s / 0.28s, CV=0.651.
With cache Enabled=true: req 2-3 → L1 HIT → ~0.01s, CV→0.
That delta is measurable in one epoch. One ask: enable CacheQualityParams 
on testnet for one epoch — repeat_fraction becomes a number, not an assumption.

Why same prompts: structured domain workflows (code review, auth, 
token estimation) produce semantically convergent task flows by construction.
Measured: SDK templates raise pairwise similarity from 0.49 → 0.80 (+62%).
T3 (auth flow) hits 0.97 threshold immediately.

dev-chains: agreed as the right host for this experiment.

Mayveskii:
> My point is that as it is the experimental building block for phase 0, it shouldn't be merged.
> 
> There is a large proposal - dev-chains. It is proposed to be very modular from the main chain, and be not coupled with existing api node. Moreover it is implementation free, there even can be different dev-chains versions in parallel, that is nice for experiments.
> 
> We should not merge this to main, all experimental work could and should be done in experimental dev-chain.
> 
> It is hard to say that anything measured on a real data, as currently network not fully functions as a real inference engine and there is no real data, as even developer accounts are limited and there are a lot test prompts.
> 
> What is repeat_fraction and why it is 30-60% ? Need of this mechanics are mostly depend on this repeat_fraction, but how we can estimate it? What do you propose, explain, what is the repeat_fraction.
> 
> How do you propose implement L2: cosine similarity ? If it didn't hit the GPU where do you get the vectors from?
> 
> @Mayveskii Please close this PR, and please start the branch for experimental features when dev-chains will be published. One of their purpose is this modularity for experiments
[quality_matrix_research.md](https://github.com/user-attachments/files/25801392/quality_matrix_research.md)

gmorgachev:
> Why do you think users are asking same prompts (with same contexts)? Why do you think this cache could be useful at all?

I have the same concern, don't think that full prompt and params match is realistic scenario to optimize in real word. 

> Why same prompts: structured domain workflows (code review, auth,
token estimation) produce semantically convergent task flows by construction.

Such scenarios usually have identical task description but different content (e.g. code to analyze). To optimize that, tasks desciption is usually in the beggining of the prompt and [prefix caching](https://docs.vllm.ai/en/v0.9.2/features/automatic_prefix_caching.html) allows to avoid re-computation of KV-cache for tokens in such prefix (vLLM caches that in KV-cache also, that's what @blizko highlighted).
That's the cache hit / miss usually reported by providers 

We definitely should work on the effective usage of prefix cache to affectively continue conversation. @akup has great idea on how it should look like. Do we need or don't need to add some incentive for that - still open question for me. In general, when such cache is configure, miner automatically has more compute capacity. But it's not much alligned with POC (but i think we must update POC to profile KV-cache too, e.g. by making not only encoding but decoding part too)

Mayveskii:
>RE


@gmorgachev Thanks for the note on prefix caching.

Quick alignment: we’re not proposing to replace or compete with vLLM KV-cache. That one optimizes “how much work per inference” (reuse of activations during the forward pass). This cache is request/response-level: on an L1/L2 hit we don’t start inference at all — we return a stored response. So they sit on different layers and can coexist.

On “incentive for prefix cache”: whether to add a protocol incentive for effective use of KV-cache (e.g. in POC or elsewhere) is a separate design question. This PR only adds CacheQualityWeight for reuse at the request level (L6). If the chain later wants to reward nodes for prefix-cache behavior too, that could be another axis in the same settlement pattern — but it’s out of scope for Phase 0.

The “same prompts” / usefulness point: we’ve now run a two-participant test (different workspaces, same domain). Results are in gonkalabs/gonka-agent docs/testing.md: Participant B got a partial hit (0.79) from A’s cache for a semantically similar task (Go race fix, different struct). So “users asking similar prompts” is not assumed — it’s measured in at least one structured domain (code). We’re not claiming 30–60% repeat everywhere; we’re showing that when workflows are structured (e.g. SDK templates), reuse becomes measurable and the protocol can reward it without changing how the chain works today.

Mayveskii:
@blizko @gmorgachev @tcharchian @patimen — updated PR description with full quality gate
pipeline details, 62 test results, and relation to #874 (BLSSignature path).
Loop closure gate directly addresses the "full prompt match not realistic"
concern — L2 semantic matching is now gated by hub frontier, not just cosine floor.
It would be great to run test workflows and get real data so we can continue in the same spirit and not stop improving the quality of the output.

Mayveskii:
> Моя точка зрения заключается в том, что, поскольку это экспериментальный строительный блок для фазы 0, его не следует объединять.
> 
> Существует масштабное предложение — dev-chains. Предполагается, что оно будет очень модульным по отношению к основной цепочке и не будет связано с существующим API-узлом. Более того, оно не требует дополнительной реализации, и даже позволит параллельно использовать разные версии dev-chains, что удобно для экспериментов.
> 
> Нам не следует объединять это с основной веткой, вся экспериментальная работа может и должна проводиться в экспериментальной цепочке разработки.
> 
> Трудно утверждать, что что-либо измеряется на реальных данных, поскольку в настоящее время сеть не функционирует в полной мере как реальный механизм вывода, и реальных данных нет, так как даже учетные записи разработчиков ограничены, и существует множество тестовых запросов.
> 
> Что такое коэффициент повторения (repeat_fraction) и почему он составляет 30-60%? Необходимость в данной механике в основном зависит от этого коэффициента повторения, но как его можно оценить? Что вы предлагаете, объясните, что такое коэффициент повторения?
> 
> Как вы предлагаете реализовать L2: косинусное сходство? Если это не задействует графический процессор, откуда вы берете векторы?
> 
> @Mayveskii Пожалуйста, закройте этот запрос на слияние и создайте ветку для экспериментальных функций, когда будут опубликованы цепочки разработки. Одна из их целей — обеспечение модульности для экспериментов.



Prior art: GPTCache (Zilliz) reports 40-60% cache hit rate in structured domains
(code review, SQL, customer support) — validating the repeat_fraction assumption
independently of our testnet measurement. Btw we go 4 SINGULARITY now

PR #866, opened 2026-03-05
[P2] onboarding clarity enhancement clean

### LINKED ISSUE #326: [P2] Improve onboarding experience

jacky6block:
Supersedes #598


tcharchian:
Hey @akup @jacky6block @x0152, it would be great if you could sync on the next steps for this pull request and make the needed decisions together. If you are able to move it forward on your own, it could potentially be included in v0.2.12. But overall, this is a nice-to-have rather than something critical.

jacky6block:
Hi @akup @x0152,  I'd like to move this forward for v0.2.12.

  Quick summary of what's implemented:
  - Clear onboarding state messages: WAITING_FOR_POC, TESTING, TEST_FAILED, INACTIVE_WAITING, ACTIVE_PARTICIPATING
  - Proactive MLnode pre-PoC auto-testing (triggers when >1h until PoC)
  - Timing guidance: tells users when to be online and when it's safe to shut down
  - Suppressed confusing "no model" errors for inactive participants
  - Auto-test logic consolidated, timing constants unified

  Could you take a look and let me know if there are any concerns? Happy to address feedback quickly.

x0152:
Hi @jacky6block could you please rebase and resolve conflicts

Ryanchen911:
Hi @x0152 @tcharchian ,it has been rebased right now~

patimen:
I like the intent here for sure, but I don't think the approach is going to work right now.

The issue is we already have a LOT of complexity in our broker code. This is a really critical area, and we already have several PR's touching that same code. The complexity presents both risks in implementation and in simply understanding changes well enough to review them coherently. Adding complexity for what is essentially a UX issue is not a wise direction.

I think the ideal here would be to find a way to implement some similar function without much (if any) modification of the broker code. Or, alternately, we should find a way to greatly simplify the broker code before moving on this.

DimaOrekhovPS:
I agree with John, we should avoid adding any more complexity to the broker and NodeState. These parts are critical and already complicated.

A good solution might look like:

1. Some separate mode of running an API node

2. When ran in this mode we're only interested in the PoC lifecycle working as intended. We could mock the chain and provide some dummy data to the broker, the event listener (and to other structs if needed)

3. Them some separate new struct intercepts the results and makes an evaluation whether basic PoC operations work or not

PR #868, opened 2026-03-06
Required-collateral slashing base and collateral-aware slashing flow

qdanik:
### Proposed changes
Slashing now uses required collateral as the penalty base (when available), not always the full deposited amount.
- if required collateral exists → slash from min(required, actual);
- if not → keep old behavior (slash from full actual collateral).

This applies to both invalidation and downtime slashing flows.
Slashing is still distributed proportionally across active and unbonding collateral.
Added tests for over/under deposit, exact match, legacy fallback, and unbonding cases.

### Motivation
To make penalties fair for over-collateralized participants.
Previously, someone with extra collateral could be punished more than their effective required stake.
Now punishment is tied to the stake actually needed for their weight, while preserving backward-compatible fallback behavior


inference-chain/x/inference/keeper/collateral.go
```diff
@@ -9,6 +9,55 @@ import (
...
+			participantWeight = vw.Weight
+			break
+		}
+	}
+	if participantWeight <= 0 {
+		return math.ZeroInt()
+	}
+
+	bwr, err := collateralParams.BaseWeightRatio.ToLegacyDec()
+	if err != nil {
```
DimaOrekhovPS: We should check it's greater or equal than zero, but less than one

inference-chain/x/inference/keeper/bitcoin_rewards.go
```diff
@@ -186,6 +186,25 @@ func GetPreservedWeight(participant string, epochGroupData *types.EpochGroupData
 	return 0
 }
 
+// calculateDenominatorWeight computes a participant's denominator weight for reward distribution.
+// It sums unique (by NodeId) PocWeights from MLNode data — the same logic as RecalculateWeight
+// but operating on a flat []*MLNodeInfo list. This value is used only in the denominator,
+// so underfunded-collateral portions go to governance instead of being redistributed.
+func calculateDenominatorWeight(mlNodes []*types.MLNodeInfo) int64 {
```
DimaOrekhovPS: Is this part related to the main change of the PR? Collateral-slashing base?

it seems it addresses some other issue?

qdanik: You're right, It would be better to create one more PR: https://github.com/gonka-ai/gonka/pull/882  

CC: @tcharchian

DimaOrekhovPS (approved):

PR #878, opened 2026-03-12
feat(binary-singularity): semantic cache extending #859 #860

Mayveskii:
After presenting the feature in a test environment, it's now possible to confirm it using real protocol data. #859

Thanks to the team for their timely response – together we can do more.

I haven't received enough feedback on my concept proposed in #859 & #860.
It would be helpful to hear your thoughts and those of the community on this, thank you.


Mayveskii:
It's going be completely updated as a proposal.  It will take a little more time to finalize it and submit it for review with specific figures and a topology of actions to ensure it's as accurate as possible. But I'm ready to do so in the near future. Just please keep it within the needs triage.

patimen:
Thanks for keeping this around while the idea takes shape. From the current state it looks more like ongoing proposal/prototype material than something we can move forward as a normal PR, so I wanted to check in: is this PR still intended to stay active, or do you expect it to be supplanted soon by a proposal or a fresher/narrower branch? If the work is moving elsewhere, a pointer would help us keep triage tidy without losing the thread.

Mayveskii:
>re

I'm incredibly glad you've taken this idea seriously. It's already been developed and looks like a production version for obtaining useful calculations and real metrics, but I'm in the process of assembling a baseline for all nodes participating in calculations, presenting it in a way that's exact to scale and use.

I think the most optimal solution would ultimately be to allocate a server CPU with deep cache support, deploy it to a testnet or sandbox, if the way is clear. Then we'll be able to touch theese metrics/results we talking about.

At the same time, I need to clarify: if such a engine-model has the metrics mentioned in #859, 860, and 878 confirmed, should it be open-sourced or allocated with private access? I believe it should be a private branch, as an engine of this caliber cannot be open-sourced, as is the policy of all leading labs. My goal is to present this engine to your team without any compensation. The community will allocate deployment costs and tokens for computing once we get this up and running. I'm also willing to provide support for setup and deployment in the sandbox, demonstrating the results. Again, I'm willing to provide your team with FULL GREANTED access without any rights of ownership. This way, I'm demonstrating my commitment to delivering real results and not wasting this opportunity.

The final code-baseline will be compiled and presented at the end of next week. I won't be sharing this on social media. I'm relying solely on your opinion.

P.S.
Your ai-reviewer presentation sped up my time getting all the key metrics x-fold; your repositories have paved the way for me to get to this point. Thank you for your feedback.

Mayveskii:
@patimen 

Yo, John ) Have a nice time ! 

I found thath the way i tryed realyze in this pull is done by Gmorgachev's dev_shards . That exactly means ok -> bin -> result -> share between participants & clients. I would force all things this way Gleb started.
It's a good point that in the way we found it i've studied couple of usefull things. Respecting again !

Mayveskii:
>re

I guess it can be archived , because the server side will be designed and presented later.

PR #882, opened 2026-03-13
Collateral weight-based reward distribution

qdanik:
## Proposed changes
Reward settlement now uses:
- collateral-adjusted effective weight for participant payout,
- full snapshotted weight as the denominator baseline.

If a participant is under-collateralized, their payout is reduced by collateralWeightRatio, while the denominator stays unchanged. The uncovered share is not redistributed and goes to governance.

## Motivation
This makes rewards reflect actual validated contribution and prevents reduced participant weight from increasing payouts for others. Missing or underfunded weight is captured as governance remainder instead of being reallocated.


qdanik:
The fix has been implemented in https://github.com/gonka-ai/gonka/pull/888

PR #884, opened 2026-03-13
fix: deterministic query responses, defensive PoC error handling, overflow check

### LINKED ISSUE #883: Minor safety issues: non-deterministic query, unhandled error continuation, uint64 overflow
## Description

Found three minor safety issues during code review:

### 1. Non-deterministic `GetAllModelPerTokenPrices` response

**File:** `inference-chain/x/inference/keeper/query_dynamic_pricing.go:52-58`

`GetAllModelPerTokenPrices()` iterates over `map[string]uint64` and appends to a slice. Since Go map iteration order is non-deterministic, different nodes return models in different order for the same gRPC query.

While this doesn't affect consensus (query-only), deterministic API responses are generally expected in blockchain systems and make debugging/testing easier.

### 2. `SubmitPocValidation` continues after `GetActiveConfirmationPoCEvent` error

**File:** `inference-chain/x/inference/keeper/msg_server_poc_validation_v1.go:41-45`

When `GetActiveConfirmationPoCEvent()` returns an error, it's logged but execution continues without resetting `activeEvent` and `isActive`. Per Go convention, return values are not guaranteed to be meaningful when `err != nil`. If the function returns partially initialized values on error, the subsequent routing logic could behave unexpectedly.

### 3. Silent uint64 overflow in `GetTotalCoins`

**File:** `inference-chain/x/inference/types/settle_amount.go:5-11`

`GetTotalCoins()` adds `RewardCoins + WorkCoins` (both `uint64`) without checking for addition overflow. If the sum wraps around, it produces a small value that passes the `sum > math.MaxInt64` check, returning an incorrect (silently corrupted) result.

While overflow is unlikely with the current token supply, the silent corruption makes this a correctness issue worth fixing.

## Proposed Fix

See PR linked below with fixes for all three issues.

unameisfine:
## Summary

Three minor safety improvements found during code review:

1. **Deterministic query response**: `GetAllModelPerTokenPrices` now sorts results by `ModelId` to ensure consistent gRPC responses across nodes. Go map iteration order is non-deterministic, so without sorting, different nodes return models in different order for the same query.
   - File: `inference-chain/x/inference/keeper/query_dynamic_pricing.go`

2. **Defensive PoC error handling**: `SubmitPocValidation` now explicitly resets `activeEvent` and `isActive` when `GetActiveConfirmationPoCEvent` returns an error. Per Go convention, return values are not guaranteed when `err != nil`, so the subsequent routing logic could use partially initialized values.
   - File: `inference-chain/x/inference/keeper/msg_server_poc_validation_v1.go`

3. **Overflow protection**: `GetTotalCoins` now checks for uint64 addition overflow before comparing against `MaxInt64`. Without this check, if `RewardCoins + WorkCoins` wraps around, the result silently returns an incorrect small value.
   - File: `inference-chain/x/inference/types/settle_amount.go`

## Testing

- Added unit tests for `GetTotalCoins` covering overflow edge cases
- Changes are minimal and localized — no behavioral changes in the happy path

Fixes #883

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_poc_validation_v1.go
+++ b/inference-chain/x/inference/keeper/msg_server_poc_validation_v1.go
@@ -41,7 +41,9 @@ func (k msgServer) submitPocValidationV1(goCtx context.Context, msg *types.MsgSu
 	activeEvent, isActive, err := k.Keeper.GetActiveConfirmationPoCEvent(ctx)
 	if err != nil {
 		k.LogError(PocFailureTag+"[SubmitPocValidation] Error checking confirmation PoC event", types.PoC, "error", err)
-		// Continue with regular PoC check
+		// Reset values to ensure we fall through to regular PoC logic safely
+		activeEvent = nil
+		isActive = false
 	}
 
 	// Route to confirmation PoC handler if active and in VALIDATION phase
--- a/inference-chain/x/inference/keeper/query_dynamic_pricing.go
+++ b/inference-chain/x/inference/keeper/query_dynamic_pricing.go
@@ -2,6 +2,8 @@ package keeper
 
 import (
 	"context"
+	"slices"
+	"strings"
 
 	"github.com/productscience/inference/x/inference/types"
 	"google.golang.org/grpc/codes"
@@ -56,6 +58,10 @@ func (k Keeper) GetAllModelPerTokenPrices(goCtx context.Context, req *types.Quer
 			Price:   price,
 		})
 	}
+	// Sort by ModelId for deterministic query responses across nodes
+	slices.SortFunc(modelPrices, func(a, b types.ModelPrice) int {
+		return strings.Compare(a.ModelId, b.ModelId)
+	})
 
 	k.LogInfo("Retrieved all model prices", types.Pricing,
 		"totalModels", len(modelPrices),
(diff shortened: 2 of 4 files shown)


Doog-bot534 (approved):
## Review: fix: deterministic query responses, defensive PoC error handling, overflow check

**Approve** ✅

Three independent fixes, all correct:

**Fix 1 — Deterministic query sorting:** Map iteration in Go is non-deterministic, so gRPC query responses could vary across nodes. `slices.SortFunc` with `strings.Compare` is idiomatic Go 1.21+. No performance concern since model count is small.

**Fix 2 — Defensive PoC error reset:** The original code left `activeEvent`/`isActive` potentially holding partial values from a failed `GetActiveConfirmationPoCEvent` call. Explicitly resetting to nil/false ensures safe fallthrough. Subtle but important.

**Fix 3 — Overflow check:** The existing `sum > math.MaxInt64` was insufficient because if `RewardCoins + WorkCoins` wraps around uint64, the sum is small and passes the check. The new wrap-around detection (`sum < sa.RewardCoins || sum < sa.WorkCoins`) correctly catches this. Edge case tests at MaxUint64 boundaries are thorough.

### Suggestions

1. Three unrelated fixes in one PR — consider splitting in future for easier revert if needed.

2. For the overflow check: verify that downstream code receiving exactly `int64(math.MaxInt64)` doesn't do further arithmetic that could overflow int64.

**Verdict:** All three fixes are correct and well-tested. Solid code quality.

unameisfine:

PR #889, opened 2026-03-15
[P0] On-chain configurable reward claim recipients

alancapex:
## Motivation

Participants currently receive reward claims at their own address. Some operators would like to route rewards to a different destination — a treasury, a multisig, or a smart contract.

An earlier version of this PR added an optional `recipient` field directly to `MsgClaimRewards`. Review flagged a security problem: operators typically grant a warm (operational) key permission to sign `MsgClaimRewards` via authz so an automated process can claim on their behalf. If that message also controlled the destination, a compromised warm key could redirect all future rewards — breaking the warm/cold key boundary.

## Approach

The recipient schedule lives on-chain, controlled exclusively by the participant's own (cold) key:

- **New message `MsgSetClaimRecipients`** — batch-configures per-epoch recipient overrides. Signed by the participant. **Not** included in the set of messages granted to warm keys via authz, so it cannot be delegated.
- **`MsgClaimRewards` is unchanged.** A warm key can still trigger claims automatically; it just has no say in where rewards land.
- At claim time, the keeper looks up the scheduled recipient for `(participant, current epoch)`. If one is set, rewards go there; otherwise, to the participant as before.
- Entries are single-use — consumed on successful claim — and capped at 40 epochs of lookahead to bound per-participant state.

## User-facing surface

- `inferenced tx inference set-claim-recipients` — cold key configures overrides for future epochs. Empty recipient clears a previously set entry.
- `inferenced query inference list-claim-recipients [participant]` — inspect the currently scheduled overrides.

## Testing

- Full chain unit-test regression: 1519 passing, 0 failing.
- New unit tests cover: batch set/delete, past/current-epoch rejection, lookahead cap, invalid recipient/creator, empty batch, per-epoch lookup, schedule-driven payout (direct and vesting paths), entry consumption on successful claim.
- No DAPI changes; existing integration flows unaffected.


gmorgachev:
I like the idea in general. Need to spend some time on audit to make sure the vesting behaviour is the same for non-participant address

alancapex:
Force-pushed a redesign in response to the earlier security review.

The original approach put `recipient` directly on `MsgClaimRewards`. Since `MsgClaimRewards` is normally granted to warm/operational keys via authz (so automated claim workers can run without the cold key), that field would have let a compromised warm key redirect all future reward claims — violating the warm/cold key boundary.

New design moves recipient control on-chain: only the participant's own key can call the new `MsgSetClaimRecipients`, which is intentionally excluded from `InferenceOperationKeyPerms` so no authz grant for it is ever created. `MsgClaimRewards` itself is unchanged and stays delegatable. Existing review comments on the old files are now "outdated"; fresh review welcome.

patimen:
@alancapex this redesign looks good to me and addresses the security gap that was raised on the earlier version. If you’re still interested in landing it, could you merge in the latest changes from `main` so we can move it forward from triage?

alancapex:
@patimen done, please accept

inference-chain/x/inference/keeper/msg_server_set_claim_recipients.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+			continue
+		}
+		if err := ms.ClaimRecipients.Set(ctx, key, entry.Recipient); err != nil {
+			return nil, errorsmod.Wrapf(err, "failed to set claim recipient for epoch %d", entry.Epoch)
+		}
+	}
+
+	return &types.MsgSetClaimRecipientsResponse{}, nil
+}
```
x0152: Not a blocker, just a small cleanup suggestion
```go
...
// Apply into a cache context so the whole batch is atomic in one pass
	cacheCtx, writeFn := ctx.CacheContext()
	for _, entry := range msg.Entries {
		if err := validateClaimRecipientEntry(entry, currentEpoch); err != nil {
			return nil, err
		}
		if err := ms.applyClaimRecipientEntry(cacheCtx, creatorAddr, entry); err != nil {
			if entry.Recipient == "" {
				return nil, errorsmod.Wrapf(err, "failed to remove claim recipient for epoch %d", entry.Epoch)
			}
			return nil, errorsmod.Wrapf(err, "failed to set claim recipient for epoch %d", entry.Epoch)
		}
	}
	writeFn()
	return &types.MsgSetClaimRecipientsResponse{}, nil
}
```

x0152 (approved):
Looks good to me, can be merged. Two small things:
- Add a short code comment that failed/missed claim entries are kept on purpose and can accumulate over time
- Just rebase onto main and run make proto - there's a conflict right now

inference-chain/x/inference/keeper/claim_recipient_schedule.go
```diff
@@ -0,0 +1,56 @@
...
+// preventing cheap state bloat via dumping thousands of future entries.
+const MaxClaimRecipientLookahead uint64 = 40
+
+// GetClaimRecipientForEpoch returns the scheduled recipient address for the
+// given (participant, epoch). Returns ("", false) if no override exists, in
+// which case the caller should pay the participant directly.
+func (k Keeper) GetClaimRecipientForEpoch(ctx context.Context, participant sdk.AccAddress, epoch uint64) (string, bool) {
+	v, err := k.ClaimRecipients.Get(ctx, collections.Join(participant, epoch))
+	if err != nil {
+		if errors.Is(err, collections.ErrNotFound) {
```
patimen, line 24: If the error isn't NotFound, we're just returning false (defaulting to paying Participant)... even with no logging? At a minimum, we need to log something like "unable to retrieve claim recipient, defaulting to Participant), which would probably mean returning an `error` here.

patimen (changes requested):
Two changes I'd like us to discuss:
1. We should probably make sure that if something goes wrong with the claim payout, we clear out any old entries. Minor impact though, I don't see that happening, but I also don't like the idea of things living permanently in state for any reason.
2. Is this operationally optimum? Effectively, operators will need to call SetClaimRecipients every 40 epochs with a list of all of the next epochs to direct the claim to. Wouldn't something more like "effectiveEpoch" for each entry be optimal, with either a new map or a modification of Participant serving to track the current payee? This is especially true since the cold key will need to be used for SetClaimRecipients. I should have flagged this earlier, I clearly wasn't thinking of operational concerns early enough.

It might even just be simplest to make ClaimRecipient by a property of Participant. Then we have a SetClaimRecipient message, no additional state to code for, and we just check if it's set at assign time.

alancapex:
@patimen — that's a reasonable question. The design here is aimed at covering the most general (and seemingly most complex) case for a miner: the recipient address is a smart contract that distributes the reward among clients, and the set of clients and their weights changes essentially every epoch.

Within the contract itself there's no way I can see to correctly attribute which epoch an incoming reward came from — the payout is a vanilla `SendCoinsFromModuleToAccount` with no epoch tag, and bank receipt doesn't trigger contract execution. (If you see a clean way around that, I'd genuinely like to hear it.) So we end up needing a separate distribution contract per epoch.

That said — for the lucky cases where a single contract suffices, or where the operator can pre-compute the schedule of contracts ahead of time, this batch mode covers that too: cold key stages one batch (up to 40 epochs ahead) and warm key claims within it.

On the cleanup concern — happy to address it as a follow-up once we're aligned on the design. Stale entries are pruneable under a provable invariant (`epoch < currentEpoch - 1` is unclaimable per `validateRequest`), so either lazy prune in `MsgSetClaimRecipients` or an upgrade-handler one-shot (same pattern as `migrateFeegrantsForFees`) would close it.

alancapex:
@DimaOrekhovPS — thanks, great PR. Wouldn't have dared to drop pruning code into the module myself.

The three commits were pulled into this PR and rebased on top of current `upgrade-v0.2.14`.

PR #894, opened 2026-03-16
[P0] [#893] remove float math from `devshards`

### LINKED ISSUE #893: [P0] Remove float math from `devshards` consensus
DeterministicFloat, ShouldValidate, and penalizeUnrevealedSeeds use float64 and math.Ceil. 

Floating-point arithmetic is not deterministicacross architectures and can produce different results on different
machines, which can lead to state root divergence and consensus splits.

Brgndy25:
Fixes #893


scutuatua-crypto (approved):

gmorgachev:
Should we open it for review? Tests should restart

subnet/state/machine.go
```diff
@@ -950,6 +950,7 @@ func (sm *StateMachine) recomputeCompliance() {
```
dcastro, line 993: I reckon extracting the math into a separate function and having tests for it would help avoid regressions. Can you please do that? Otherwise it's hard to have confidence that this impl is equivalent to the old impl

subnet/state/validation.go
```diff
@@ -23,26 +22,31 @@ func DeriveSeed(signature []byte) (int64, error) {
...
+	// threshold = (rateBasisPoints * validatorSlotCount * 2^32) / (10000 * (totalSlots - executorSlotCount))
+	threshold := (uint64(rateBasisPoints) * uint64(validatorSlotCount) << 32) / denom
+	const maxThreshold = 1 << 32
+	if threshold > maxThreshold {
+		threshold = maxThreshold
 	}
-	return DeterministicFloat(seed, inferenceID) < probability
+	hashInt := deterministicHash(seed, inferenceID)
+	return (hashInt >> 32) < threshold
 }
```
akup, line 86: While it is mathematically correct it becomes absolutely magic numbers for the any person who reads this code later.

I think some comments to keep the logic clean should be added. Like:

```go
//rate = rateBasisPoints / 10000.0
//probability = rate * validatorSlotCount / totalSlots-executorSlotCount
//=>
//probability = rateBasisPoints * validatorSlotCount / (totalSlots-executorSlotCount) * 10000.0
```

Moreover I would add `Uint64ProbabilityScale32` function:

```go
func Uint64ProbabilityScale32(uint64 numerator, uint64 denominator) uint64 {
  probability := numerator << 32 / denominator
  const maxProbability = 1 << 32 // probability is at [0; 1] at scale 32
  if probability > maxProbability {
    probability = maxProbability
  }
  return probability
}
```

and use it at `ShouldValidate` to keep code readability.

Something like

```go
  // If deterministically random hash [0;1] < threshold = thresholdNumerator / thresholdDenominator
  // equivalent to scale 32
  hashInt := deterministicHash(seed, inferenceID) //at scale 64
  return (hashInt >> 32) < threshold
}
```

Please keep it readable

Brgndy25: Fixed in [431f1be](https://github.com/gonka-ai/gonka/pull/894/changes/431f1be10dd498df7ec00ca3e0531974e0b26825)

subnet/state/machine.go
Brgndy25: Fixed in [af4a561](https://github.com/gonka-ai/gonka/pull/894/changes/af4a561f72b40a53150e1e9b453a33c3483efc4c)

subnet/state/validation.go
```diff
@@ -23,26 +22,60 @@ func DeriveSeed(signature []byte) (int64, error) {
...
 	sum := sha256.Sum256([]byte(input))
-	hashInt := binary.BigEndian.Uint64(sum[:8])
-	return float64(hashInt) / float64(math.MaxUint64)
+	return binary.BigEndian.Uint64(sum[:8])
+}
+
+// uint64ProbabilityScale32 returns floor(numerator * 2^32 / denominator), clamped to [0, 2^32].
+// It represents a rational in [0, 1] at 32-bit fixed-point scale without floating-point.
+// denominator must be non-zero; callers guard that.
+func uint64ProbabilityScale32(numerator, denominator uint64) uint64 {
```
dcastro, line 40: There's also the catch here that `numerator` can't be greater than 2^32-1, otherwise some bits will be discarded with `numerator << 32`.


So we should either: 
* constraint `numerator` to `uint32`, then cast it to uint64 before shifting
* or cast `numerator` to 128 bits before shifting

subnet/state/validation.go
```diff
@@ -23,26 +22,60 @@ func DeriveSeed(signature []byte) (int64, error) {
...
+	const maxP = uint64(1) << 32
+	if p > maxP {
+		return maxP
+	}
+	return p
+}
+
+// uint32CeilScaledSum32 returns ceil(sumScaled / 2^32), where sumScaled is a sum of
+// terms at 32-bit fixed-point scale (each term from uint64ProbabilityScale32).
+func uint32CeilScaledSum32(sumScaled uint64) uint32 {
```
dcastro, line 53: This function is defined here but it's never used in this module. It's only used in `machine.go`.

So either:
1. move it to `machine.go`
2. or better yet, create a new module for our new math functions

subnet/state/validation.go
```diff
@@ -23,26 +22,60 @@ func DeriveSeed(signature []byte) (int64, error) {
...
 	input := fmt.Sprintf("%d:%d", seed, inferenceID)
 	sum := sha256.Sum256([]byte(input))
-	hashInt := binary.BigEndian.Uint64(sum[:8])
-	return float64(hashInt) / float64(math.MaxUint64)
+	return binary.BigEndian.Uint64(sum[:8])
+}
+
+// uint64ProbabilityScale32 returns floor(numerator * 2^32 / denominator), clamped to [0, 2^32].
+// It represents a rational in [0, 1] at 32-bit fixed-point scale without floating-point.
+// denominator must be non-zero; callers guard that.
```
dcastro, line 37: I'd add an example illustrating what this does:

```
For example, `uint64ProbabilityScale32(98, 100)` (representing a probability of 98%) would return 98% of 2^32, that is 4209067950
```

Might be worth it to add tests for this as well, just to better illustrate the idea:

```
	require.Equal(t, uint64ProbabilityScale32(50, 100), (uint64(1<<32) * 50 / 100))   // 50% of 2^32
	require.Equal(t, uint64ProbabilityScale32(98, 100), (uint64(1<<32) * 98 / 100))   // 98% of 2^32
	require.Equal(t, uint64ProbabilityScale32(100, 100), (uint64(1<<32) * 100 / 100)) // 100% of 2^32
	// Clamped to 2^32
	require.Equal(t, uint64ProbabilityScale32(105, 100), uint64ProbabilityScale32(100, 100)) // 100%
```

subnet/state/validation.go
```diff
@@ -23,26 +22,60 @@ func DeriveSeed(signature []byte) (int64, error) {
...
+// terms at 32-bit fixed-point scale (each term from uint64ProbabilityScale32).
+func uint32CeilScaledSum32(sumScaled uint64) uint32 {
+	const scale = uint64(1) << 32
+	return uint32((sumScaled + scale - 1) >> 32)
+}
+
+// penalizePerInferenceScaled32 returns the per-inference contribution at scale 2^32 for
+// unrevealed-seed penalty: floor((rateBasisPoints/10000) * validatorSlotCount / (totalSlots-executorSlots) * 2^32),
+// clamped like uint64ProbabilityScale32. Same as the legacy float
+// (rateBasisPoints/10000) * validatorSlotCount / denom in fixed point.
```
dcastro: The term "legacy" won't be meaningful for anyone reading this comment after this PR is merged, so I'd just remove this last sentence.

Might also be worth it to reiterate some bits of `uint64ProbabilityScale32`'s docs. I.e. mention the return value represents a probability in the range [0, 1], but scaled to [0, 2^32]

Brgndy25: [Fixed](https://github.com/gonka-ai/gonka/pull/894/changes/d0a6d32e5baf9414ad5056392b912962632121a4)

subnet/state/validation.go
Brgndy25: [Fixed](https://github.com/gonka-ai/gonka/pull/894/changes/ec3972408cf41b2232f5807aaaee21b9e425d988)

subnet/state/validation.go
Brgndy25: [Fixed](https://github.com/gonka-ai/gonka/pull/894/changes/f906483f441300715c16c604f916065a5ea1c0c6)

dcastro (approved):

akup (approved):

x0152 (approved):

PR #900, opened 2026-03-16
Fix threshold signing quorum calculation and slot validation (#823)

### LINKED ISSUE #823: Bridge: Weak Dealer Approval Enables Threshold Signing DoS

lionmuslim:
We implemented fixes addressing the issues described in #823.

Main changes:

1. **Slot-weighted dealer consensus**   Dealer approval is now based on the number of **slots**, not the number of participants.   This prevents a dealer from being approved by a simple majority of voters while controlling fewer than 50% of the total slots.

2. **Unique slot counting for threshold signing**   Slot coverage is now calculated using a **set of unique slots**, preventing duplicate slot submissions from artificially inflating the threshold count.

3. **Duplicate slot protection**   The implementation now rejects partial signatures that reuse slot indices already submitted in other partial signatures.

4. **Slot bounds validation**   Every submitted slot index is validated against `epochBLSData.ITotalSlots` to prevent invalid or out-of-range slot references.

5. **Partial signature validation**   Partial signatures are now strictly validated for the expected BLS signature size (96 bytes).

6. **Spam / resource exhaustion protection**   A limit on the number of active signing requests per epoch was introduced to reduce the risk of resource exhaustion attacks.

7. **Additional input validation**   Added validation for:
- empty signing data
- encoded message size
- duplicate participant submissions
- duplicate slot indices within a submission

These changes fix the quorum calculation issue and prevent duplicate slot counting that could previously lead to threshold-signing DoS scenarios.

### DIFF
--- a/inference-chain/x/bls/keeper/phase_transitions.go
+++ b/inference-chain/x/bls/keeper/phase_transitions.go
@@ -1,379 +1,386 @@
 package keeper
-
 import (
-	"fmt"
-
-	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	"github.com/productscience/inference/x/bls/types"
+"encoding/binary"
+"fmt"
+"cosmossdk.io/store/prefix"
+"github.com/cosmos/cosmos-sdk/runtime"
+sdk "github.com/cosmos/cosmos-sdk/types"
+"golang.org/x/crypto/sha3"
+"github.com/productscience/inference/x/bls/types"
 )
-
-// ProcessDKGPhaseTransitions checks the currently active DKG epoch and transitions it to the next phase if deadline has passed
-func (k Keeper) ProcessDKGPhaseTransitions(ctx sdk.Context) error {
-	// Get the currently active epoch ID
-	activeEpochID, found := k.GetActiveEpochID(ctx)
-	if !found || activeEpochID == 0 {
-		// No active DKG - this is normal
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


x0152:
This is the same as in #902. It is already covered by #825

PR #902, opened 2026-03-16
Fix DKG phase transition validation (#823)

lionmuslim:
We implemented fixes addressing the issues described in #823.

Main changes:

1. **Slot-weighted dealer consensus**   Dealer approval is now based on the number of **slots**, not the number of participants.   This prevents a dealer from being approved by a simple majority of voters while controlling fewer than 50% of the total slots.

2. **Unique slot counting for threshold signing**   Slot coverage is now calculated using a **set of unique slots**, preventing duplicate slot submissions from artificially inflating the threshold count.

3. **Duplicate slot protection**   The implementation now rejects partial signatures that reuse slot indices already submitted in other partial signatures.

4. **Slot bounds validation**   Every submitted slot index is validated against `epochBLSData.ITotalSlots` to prevent invalid or out-of-range slot references.

5. **Partial signature validation**   Partial signatures are now strictly validated for the expected BLS signature size (96 bytes).

6. **Spam / resource exhaustion protection**   A limit on the number of active signing requests per epoch was introduced to reduce the risk of resource exhaustion attacks.

7. **Additional input validation**   Added validation for:
- empty signing data
- encoded message size
- duplicate participant submissions
- duplicate slot indices within a submission

These changes fix the quorum calculation issue and prevent duplicate slot counting that could previously lead to threshold-signing DoS scenarios.

### DIFF
--- a/inference-chain/x/bls/keeper/phase_transitions.go
+++ b/inference-chain/x/bls/keeper/phase_transitions.go
@@ -1,379 +1,483 @@
 package keeper
-
 import (
-	"fmt"
-
-	sdk "github.com/cosmos/cosmos-sdk/types"
-
-	"github.com/productscience/inference/x/bls/types"
+"fmt"
+"sort"
+sdk "github.com/cosmos/cosmos-sdk/types"
+bls12381 "github.com/consensys/gnark-crypto/ecc/bls12-381"
+"github.com/productscience/inference/x/bls/types"
 )
-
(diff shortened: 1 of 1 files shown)


x0152:
Hey! Thanks for the PR but it does not fully cover the issue yet. A broader implementation already exists in #825, so it may be better to continue review there.

BTW, this branch currently has compile errors

x0152:
Closing this PR and #900. Feel free to reopen if you disagree

PR #907, opened 2026-03-17
Fix alphabetical bias in PoC slot round-robin allocation

### LINKED ISSUE #700: Alphabetical Bias in PoC Slot Allocation
### Description
The current implementation of the ML node allocation logic in 
x/inference/module/model_assignment.go
 contains an alphabetical bias that affects the fairness of the network.

**Alphabetical Bias:** The allocation algorithm iterates through participants in deterministic alphabetical order. This allows participants with "vanity addresses" (lexicographical prefixes like gonka1aaa...) to have a significantly higher probability of receiving PoC slots, potentially starving honest participants with random addresses.

### Impact
Economic Injustice: Participants are incentivized to mine vanity addresses rather than focus on hardware quality/uptime to increase their selection chances.
Starvation: Honest hardware providers with random addresses receive fewer opportunities despite having equal or higher eligibility.
### Proposed Solution
Deterministic Shuffle: Implement a pseudo-random shuffle of the participant list using SHA256(EpochIndex + ModelID) as a seed. This ensures fair rotation across epochs and models, providing equal opportunity to all eligible participants regardless of their address prefix.

doc-johnson:
`sortedKeys()` returns participants in alphabetical order, and the round-robin loop in `allocateMLNodePerPoCForModel()` always starts at index 0. This gives alphabetically-first participants a systematic advantage in PoC slot allocation.

Fixed by adding a deterministic shuffle seeded by `SHA256("poc_alloc_" + epochIndex + modelId)` before the round-robin loop. The seed includes epoch index so the order rotates between epochs. Same approach already used in `sampleEligibleParticipantsWithHistory()`.

Closes #700


patimen:
@doc-johnson this still looks like the right issue to fix, and I think the narrower PR is the right vehicle. Before we move it forward, could you merge in the latest changes from `main`?

One design note from review: I wonder if this could be simplified further by using a deterministic rotation of the sorted participant list based on a hash of `(epochIndex, modelId)`, rather than doing a full seeded shuffle. That seems a bit lighter computationally and easier to reason about, while still removing the persistent alphabetical-first advantage. If you think the full shuffle is still important, a short note on why would help.

doc-johnson:
merged main and switched to rotation as you suggested. hash(epoch.Index, modelId) -> offset, slice rotate.

agree its the right primitive here, only the start index is biased, neighbors dont matter for round-robin. also gives perfect fairness over N epochs (period = N) vs shuffle which has variance.

kept shuffle in sampleEligibleParticipantsWithHistory since that one samples a subset.

tests pass locally.

PR #909, opened 2026-03-17
bls: propagate caller context to BlsManager and add per-call timeouts

### LINKED ISSUE #908: bls: BlsManager stores context.Background() — DKG gRPC calls have no cancellation or timeout
## Summary

`NewBlsManager` stores `context.Background()` as a struct field `bm.ctx`. This means two gRPC calls in the DKG dealer path run without any timeout and cannot be cancelled on node shutdown.

## Affected Code

**`decentralized-api/internal/bls/manager.go:133`**
```go
func NewBlsManager(recorder cosmosclient.InferenceCosmosClient) *BlsManager {
    return &BlsManager{
        ctx: context.Background(), // Use background context for chain queries
        ...
    }
}
```

**`decentralized-api/internal/bls/dealer.go:392,404`** — no timeout wrapper:
```go
// No WithTimeout — hangs indefinitely if chain RPC is slow/unavailable
grantees, err := queryClient.GranteesByMessageType(bm.ctx, ...)
participant, err := queryClient.InferenceParticipant(bm.ctx, ...)
```

Contrast with `manager.go:165` which correctly adds a 60s timeout:
```go
ctx, cancel := context.WithTimeout(bm.ctx, 60*time.Second)
defer cancel()
```

## Impact

1. **DKG epoch block**: If the chain RPC node is slow or unreachable during `EventKeyGenerationInitiated`, the worker goroutine inside `event_listener` blocks indefinitely on `GranteesByMessageType` or `InferenceParticipant`. This causes the node to miss the entire DKG window and drop out of consensus for that epoch.

2. **No graceful shutdown**: On SIGINT, `main.go` cancels the root `ctx` via `defer cancel()`, which propagates to `listener.Start(ctx)`. However `bm.ctx = context.Background()` is independent — in-flight BLS gRPC calls are never interrupted, delaying process exit.

3. **No Stop/Close on BlsManager**: There is no shutdown method to cancel pending operations.

## Root Cause

The context is stored at construction time as `Background()` instead of being passed from the caller. `main.go` already has a properly-scoped `ctx` with cancel:

```go
// main.go:145
ctx, cancel := context.WithCancel(context.Background())
defer cancel()
// ...
blsManager := bls.NewBlsManager(*recorder)  // ctx NOT passed
listener := event_listener.NewEventListener(..., cancel, blsManager)
```

## Fix

```go
// manager.go
func NewBlsManager(ctx context.Context, recorder cosmosclient.InferenceCosmosClient) *BlsManager {
    return &BlsManager{
        ctx: ctx,  // propagate caller context
        ...
    }
}

// main.go
blsManager := bls.NewBlsManager(ctx, *recorder)
```

This ensures:
- `GranteesByMessageType` and `InferenceParticipant` respect the node lifecycle context
- On SIGINT, pending BLS queries are cancelled immediately
- Future callers can add per-call timeouts via `context.WithTimeout(bm.ctx, ...)`

## Verified

Statically verified via AST analysis of commit history (904 commits). No existing fix found in HEAD. The `dealer.go` calls at lines 392 and 404 have no timeout wrapper, unlike the correctly-handled call in `manager.go:165`.

**Severity**: Medium-High — affects node availability during DKG under degraded RPC conditions.

Mayveskii:
Closes #908

## Problem

`NewBlsManager` stored `context.Background()` as a struct field, making all gRPC calls through `bm.ctx` impossible to cancel on node shutdown or when the upstream chain RPC becomes unavailable.

Two calls in `getAllowedPubKeysForParticipant` had no timeout:

```go
// dealer.go — no timeout, hangs indefinitely on slow RPC
grantees, err := queryClient.GranteesByMessageType(bm.ctx, ...)
participant, err := queryClient.InferenceParticipant(bm.ctx, ...)
```

If the chain node is unreachable during `EventKeyGenerationInitiated`, the event listener worker blocks indefinitely. The node misses the entire DKG window and drops out of consensus for that epoch.

Contrast with the correctly-handled call in `manager.go` which already uses `context.WithTimeout(bm.ctx, 60*time.Second)` — the two calls in `dealer.go` were inconsistently left without protection.

## Changes

**`bls/manager.go`** — accept `ctx context.Context` as first argument:
```go
// before
func NewBlsManager(cosmosClient ...) *BlsManager {
    return &BlsManager{ctx: context.Background(), ...}
}
// after
func NewBlsManager(ctx context.Context, cosmosClient ...) *BlsManager {
    return &BlsManager{ctx: ctx, ...}
}
```

**`bls/dealer.go`** — add 30s timeouts on both unguarded gRPC calls:
```go
granteesCtx, granteesCancel := context.WithTimeout(bm.ctx, 30*time.Second)
defer granteesCancel()
grantees, err := queryClient.GranteesByMessageType(granteesCtx, ...)

participantCtx, participantCancel := context.WithTimeout(bm.ctx, 30*time.Second)
defer participantCancel()
participant, err := queryClient.InferenceParticipant(participantCtx, ...)
```

**`main.go`** — pass root application context:
```go
blsManager := bls.NewBlsManager(ctx, *recorder)
```

## Effect

- SIGINT propagates to all in-flight BLS gRPC queries via context cancellation
- Slow or unreachable chain RPC times out after 30s instead of blocking forever
- No behaviour change during normal operation

### DIFF
--- a/decentralized-api/internal/bls/dealer.go
+++ b/decentralized-api/internal/bls/dealer.go
@@ -389,7 +389,9 @@ func (bm *BlsManager) getAllowedPubKeysForParticipant(participantAddress string)
 	queryClient := bm.cosmosClient.NewInferenceQueryClient()
 
 	// Get all grantees (warm keys) allowed to act on behalf of this participant
-	grantees, err := queryClient.GranteesByMessageType(bm.ctx, &inferenceTypes.QueryGranteesByMessageTypeRequest{
+	granteesCtx, granteesCancel := context.WithTimeout(bm.ctx, 30*time.Second)
+	defer granteesCancel()
+	grantees, err := queryClient.GranteesByMessageType(granteesCtx, &inferenceTypes.QueryGranteesByMessageTypeRequest{
 		GranterAddress: participantAddress,
 		MessageTypeUrl: "/inference.bls.MsgSubmitDealerPart", // BLS operations
 	})
@@ -401,7 +403,9 @@ func (bm *BlsManager) getAllowedPubKeysForParticipant(participantAddress string)
 	}
 
 	// Get the participant's own public key
-	participant, err := queryClient.InferenceParticipant(bm.ctx, &inferenceTypes.QueryInferenceParticipantRequest{
+	participantCtx, participantCancel := context.WithTimeout(bm.ctx, 30*time.Second)
+	defer participantCancel()
+	participant, err := queryClient.InferenceParticipant(participantCtx, &inferenceTypes.QueryInferenceParticipantRequest{
 		Address: participantAddress,
 	})
 	if err != nil {
--- a/decentralized-api/internal/bls/manager.go
+++ b/decentralized-api/internal/bls/manager.go
@@ -127,10 +127,10 @@ type SlotAssignment struct {
 }
 
 // NewBlsManager creates a new unified BLS manager
-func NewBlsManager(cosmosClient cosmosclient.InferenceCosmosClient) *BlsManager {
+func NewBlsManager(ctx context.Context, cosmosClient cosmosclient.InferenceCosmosClient) *BlsManager {
 	return &BlsManager{
 		cosmosClient: cosmosClient,
-		ctx:          context.Background(), // Use background context for chain queries
+		ctx:          ctx,
 		cache:        NewVerificationCache(),
 	}
 }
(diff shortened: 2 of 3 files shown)


jacky6block (approved):
LGTM, change is minimal and low-risk. Good catch.

Mayveskii:
> LGTM, change is minimal and low-risk. Good catch.

Thx. Prepared for merging with resolving coflicts. Also keep working about next one.

x0152:
Closing. This is hardening rather than a vulnerability, and it was covered by the Certik fixes

PR #921, opened 2026-03-20
fix: return proper 400 error for unsupported model requests

### LINKED ISSUE #351: BUG: Wrong error message for unsupported models in /chat/completions

unameisfine:
## Summary

When requesting inference for an unsupported model via `/chat/completions`, the API returns misleading error codes:
- **402 Payment Required** — when balance check fails against fallback legacy pricing for an unknown model
- **500 Internal Server Error** — when `GetRandomExecutor` can't find a subgroup/participants for the model

This PR maps gRPC errors from `GetRandomExecutor` to a proper `400 Bad Request` with message `"Model not supported"`.

## Changes

- Add `ErrModelNotSupported` (HTTP 400) to the error definitions
- In `getExecutorForRequest()`, detect gRPC `codes.Internal` and `codes.NotFound` errors from `GetRandomExecutor` and return `ErrModelNotSupported` instead of propagating the raw gRPC error
- Add tests verifying the error mapping behaviour

## Test plan

- [x] `go build ./...` compiles successfully
- [x] `go test ./internal/server/middleware/...` — passes (new test for gRPC→HTTP error mapping)
- [x] `go test ./internal/server/public/...` — passes (all existing tests unaffected)
- [x] Manual: send request with unsupported model to `/v1/chat/completions`, verify 400 response with `"Model not supported"`

Fixes #351

### DIFF
--- a/decentralized-api/internal/server/middleware/error_handler_test.go
+++ b/decentralized-api/internal/server/middleware/error_handler_test.go
@@ -9,6 +9,8 @@ import (
 
 	"github.com/labstack/echo/v4"
 	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
 )
 
 func TestExtractError(t *testing.T) {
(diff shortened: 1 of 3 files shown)


x0152:
This PR doesn’t fix the main issue. The 402 error happens before getExecutorForRequest, so the new mapping is not reached, and mapping all codes.Internal + codes.NotFound to "Model not supported" can hide real server problems. Closing for now

PR #934, opened 2026-03-23
fix: validate baseWeightRatio range in AdjustWeightsByCollateral

### LINKED ISSUE #933: AdjustWeightsByCollateral missing baseWeightRatio range validation — weight inflation for uncollateralized participants
## Description

`calculateRequiredCollateral` (collateral.go:53) correctly validates that `baseWeightRatio` is in the range [0, 1):

```go
if err != nil || bwr.IsNegative() || bwr.GTE(math.LegacyOneDec()) {
    return math.ZeroInt()
}
```

`AdjustWeightsByCollateral` (collateral.go:100) converts the same parameter but has no equivalent guard. If governance sets `baseWeightRatio >= 1.0`:

- `baseWeight = potentialWeight × ratio` exceeds `potentialWeight`
- `collateralEligibleWeight = potentialWeight - baseWeight` goes negative
- Participant **without** collateral: `effectiveWeight = baseWeight` (inflated)
- Participant **with** collateral: `activatedWeight = min(negative, positive) = negative`, so effective weight is reduced

Example with `baseWeightRatio = 1.5`, `potentialWeight = 100`:
- No collateral: `effectiveWeight = 150` (50% inflation)
- With collateral: `effectiveWeight = 100` (normal)

This inverts the collateral incentive — participants are rewarded for NOT depositing collateral.

## Fix

Add the same range guard after `baseWeightRatio` conversion in `AdjustWeightsByCollateral`:

```go
if baseWeightRatio.IsNegative() || baseWeightRatio.GTE(math.LegacyOneDec()) {
    return fmt.Errorf("base_weight_ratio %s is out of valid range [0, 1)", baseWeightRatio.String())
}
```

unameisfine:
Fixes #933

## Problem

`calculateRequiredCollateral` validates that `baseWeightRatio` is in [0, 1), but `AdjustWeightsByCollateral` does not. If governance sets `baseWeightRatio >= 1.0`, uncollateralized participants receive inflated weights, inverting the collateral incentive.

## Fix

Add the same `[0, 1)` range guard in `AdjustWeightsByCollateral` (1 line), matching the validation in `calculateRequiredCollateral`.

## Tests

`TestAdjustWeightsByCollateral_RejectsInvalidBaseWeightRatio` — 6 cases:
- `0.2` (default) — accepted
- `0.0` (edge) — accepted
- `0.99` (edge) — accepted
- `1.0` — rejected
- `1.5` — rejected
- `-0.1` — rejected

All existing collateral tests pass.

### DIFF
--- a/inference-chain/x/inference/keeper/collateral.go
+++ b/inference-chain/x/inference/keeper/collateral.go
@@ -93,6 +93,9 @@ func (k Keeper) AdjustWeightsByCollateral(ctx context.Context, participants []*t
 		k.LogError("invalid base_weight_ratio:", types.Tokenomics, "error", err)
 		return err
 	}
+	if baseWeightRatio.IsNegative() || baseWeightRatio.GTE(math.LegacyOneDec()) {
+		return fmt.Errorf("base_weight_ratio %s is out of valid range [0, 1)", baseWeightRatio.String())
+	}
 	collateralPerWeightUnit, err := collateralParams.CollateralPerWeightUnit.ToLegacyDec()
 	if err != nil {
 		k.LogError("invalid collateral_per_weight_unit:", types.Tokenomics, "error", err)
--- a/inference-chain/x/inference/keeper/collateral_weight_ratio_test.go
+++ b/inference-chain/x/inference/keeper/collateral_weight_ratio_test.go
@@ -1,0 +1,52 @@
+package keeper_test
+
+import (
+	"testing"
+
+	"github.com/productscience/inference/x/inference/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAdjustWeightsByCollateral_RejectsInvalidBaseWeightRatio(t *testing.T) {
+	k, _, ctx, _ := setupKeeperWithMocksForIntegration(t)
+
+	params := types.DefaultParams()
+	// Set grace period to 0 so collateral logic is active
+	params.CollateralParams.GracePeriodEndEpoch = 0
+	require.NoError(t, k.SetParams(ctx, params))
+
+	// Setup epoch so grace period check passes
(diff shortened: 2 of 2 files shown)


x0152:
The test uses a hardcoded address with a bad bech32 checksum, so the valid-ratio cases (0.2, 0.0, 0.99) never actually run any collateral logic, the address fails to parse and the participant is silently skipped. The test passes for the wrong reason

unameisfine:
Good catch — fixed in 719d1f3d. Test now uses `sample.AccAddress()` for a valid bech32 address and mocks `CollateralKeeper.GetCollateral` so the collateral logic actually runs for valid-ratio cases.

Doog-bot534 (approved):
## Review: fix: validate baseWeightRatio range in AdjustWeightsByCollateral

**Approve** ✅

The fix correctly addresses the risk where `baseWeightRatio >= 1.0` inverts the collateral incentive — participants without collateral get inflated weight while collateralized participants are penalized. The `[0, 1)` range mirrors the existing guard in `calculateRequiredCollateral`.

### Observations

1. **Is 0.0 truly safe?** The test marks `baseWeightRatio = 0.0` as valid, meaning zero base weight — participants with no collateral get zero effective weight. Depending on business logic, this could be destructive in practice. Confirm this is intentional.

2. **Governance failure mode:** The function returns a hard error on invalid ratio. Since this is a chain parameter, an invalid value would cause ALL `AdjustWeightsByCollateral` calls to fail until governance fixes it. Consider whether a fallback to a default (e.g., 0.2) with a warning log would be more resilient, or if failing loud is preferred.

3. Test coverage is solid — covers both boundaries and negative values.

**Verdict:** Clean, well-tested fix.

unameisfine:

PR #937, opened 2026-03-23
[P1] GIP: Transaction fees for spam prevention (reverted)

### LINKED ISSUE #928: [P1] Open Questions: Block Gas Limits, Fees, Cost per Participant, and System TX Prioritization
## Summary

## Key design decisions

0xgonka:
## Summary

- Introduces a governance-controlled minimum gas price (`FeeParams`) enforced at consensus level via a custom `TxFeeChecker`, replacing the current `nil` fee checker that allows zero-fee transactions
- Adds a `NetworkDutyFeeBypassDecorator` that exempts protocol-obligation messages (PoC validations, inference, BLS, weight distributions) from fees, following the existing `LiquidityPoolFeeBypassDecorator` pattern
- Implements two-component PoC commit fees (`base_validation_gas` + `gas_per_poc_count × count`) as the primary sybil defense — charges a base cost per participant per epoch plus count-proportional fees on the delta
- Initial gas price of `10ngonka` (~$0.00046/tx at GNK=$0.5698), making sustained spam expensive while keeping individual transactions under a cent

## Key design decisions

- **Inference/validation messages are fee-exempt** — hosts submit these as protocol duties and inference can be cheaper than even a minimal tx fee
- **`MsgPoCV2StoreCommit` is NOT exempt** — it's the sybil barrier. Base validation fee charged on first commit per epoch, count-proportional fee on delta for subsequent updates
- **All three fee params are governance-adjustable** (`min_gas_price_ngonka`, `base_validation_gas`, `gas_per_poc_count`) — no chain upgrade needed to tune
- Ships as v0.2.12 chain upgrade with migration setting default `FeeParams`

## Files changed

- `inference-chain/app/ante_fee.go` — New `NetworkDutyFeeBypassDecorator`, `NewGonkaFeeChecker`, exempt message list with recursive MsgExec support
- `inference-chain/app/ante_fee_test.go` — Unit tests for bypass, gas cap, fee checker, exempt list, marshal roundtrip
- `inference-chain/app/ante.go` — Wire decorator and fee checker into ante chain
- `inference-chain/x/inference/types/fee_params.go` — `FeeParams` type with defaults, marshal/unmarshal
- `inference-chain/x/inference/keeper/fee_params.go` — KVStore get/set for `FeeParams`
- `inference-chain/x/inference/keeper/msg_server_poc_v2_commit.go` — Two-component gas consumption (base + delta×count)
- `inference-chain/app/upgrades/v0_2_12/` — Migration sets default `FeeParams`
- `decentralized-api/cosmosclient/cosmosclient.go` — Gas price `0ngonka` → `10ngonka`
- `decentralized-api/cosmosclient/tx_manager/tx_manager.go` — Gas price + fee amount for batch txs

## Test plan

- [x] Unit tests for bypass decorator (exempt, non-exempt, mixed, gas cap)
- [x] Unit tests for fee checker (sufficient fee, bypass flag, priority)
- [x] Unit tests for exempt message type list and FeeParams marshal roundtrip
- [x] All inference-chain tests pass (`go test ./...`)
- [x] All decentralized-api tests pass (`go test ./...`)
- [ ] Testermint integration tests (CI)


inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+}
+
+// --- Custom TxFeeChecker ---
+
+// NewGonkaFeeChecker returns a TxFeeChecker that enforces a consensus-level minimum
+// gas price read from on-chain FeeParams. It respects the bypass flag set by
+// NetworkDutyFeeBypassDecorator. This checker runs inside DeductFeeDecorator
+// during both CheckTx and DeliverTx.
+func NewGonkaFeeChecker(inferenceKeeper *inferencemodulekeeper.Keeper) ante.TxFeeChecker {
```
akup: I think it is cleaner to call it just `GonkaFeeChecker` without `new` prefix.

akup (approved):

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -936,9 +936,11 @@ func (m *manager) getSignedBytes(id string, unsignedTx client.TxBuilder, factory
 
 	timestamp := getTimestamp(blockTs.UnixNano(), m.defaultTimeout)
 
-	// Gas is not charged, but without a high gas limit the transactions fail
-	unsignedTx.SetGasLimit(10000000000000)
-	unsignedTx.SetFeeAmount(sdk.Coins{})
+	// Set high gas limit for batch transactions. Network-duty messages (validations,
+	// PoC, inference) are fee-exempt via NetworkDutyFeeBypassDecorator, so fee amount
+	// is set but will not be charged for exempt messages.
+	unsignedTx.SetGasLimit(10_000_000)
```
x0152: unsignedTx.SetGasLimit(10_000_000) sets gas limit to 10_000_000 but ante.go:212 only allows up to 1_000_000 gas for zero-fee txs, so all dapi batch transactions will fail

inference-chain/app/ante.go
x0152, line 164: I think we forget here ctx = ctx.WithValue(networkDutyBypassKey{}, true)

P.S. maybe networkDutyBypassKey isn't a clear name for this context because LiquidityPool isn't network duty

x0152:
> All three fee params are governance-adjustable (min_gas_price_ngonka, base_validation_gas, gas_per_poc_count) — no chain upgrade needed to tune

governance mechanism for FeeParams might have been forgotten

inference-chain/app/ante.go
akup: It is set at `NetworkDutyFeeBypassDecorator`, here we don't need to set it, as it is set when we pass checks at `NetworkDutyFeeBypassDecorator`.
Seems the key is better to name `networkDutyFeeBypassKey`

decentralized-api/cosmosclient/tx_manager/tx_manager.go
akup: zero-fee txs will never be on line ante_fee.go:65, as they will be returned at ante_fee.go:60, as explained in the command above `unsignedTx.SetGasLimit(10_000_000)`

But it is safer to set it to 1_000_000

akup (commented):
@0xgonka what about adding some integration tests?
To be sure system messages bypasses fee checks, and all other messages are charged

inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,211 @@
...
+
+		// If min gas price is 0 (e.g., during genesis or if governance sets it to 0),
+		// fall through to accept any fee.
+		if minGasPriceNgonka == 0 {
+			priority := getTxPriority(feeCoins, gas)
+			return feeCoins, priority, nil
+		}
+
+		// Calculate required fee: gas * min_gas_price (integer math, no rounding needed).
+		requiredAmount := gas * minGasPriceNgonka
```
x0152: probably worth adding overflow check here

decentralized-api/cosmosclient/tx_manager/tx_manager.go
x0152: I didn't mean literal fee=0, I meant txs that go through bypass logic (fee-exempt/bypassed txs)
wrong wording on my side, sorry

inference-chain/app/ante_fee.go
akup: I personally also thought on overflows. But we are really extremely far away from any possible overflows

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -915,8 +916,7 @@ func (m *manager) getFactory(id string) (*tx.Factory, error) {
 	factory := m.client.TxFactory.
 		WithAccountNumber(accountNumber).
 		WithGasAdjustment(10).
-		WithFees("").
-		WithGasPrices("").
+		WithGasPrices("10ngonka").
```
DimaOrekhovPS: 1. I'm a bit concerned that everything is hardcoded here and in `cosmosclient.go`: gas prices, gas limit, fee amount. And also interdependent at the same time, for example `FeeAmount` is set to be `GasPrices` *  `GasLimit`.

Could we extract everything into reusable constants?

2. Isn't setting `GasPrices` here redundant if we're later overriding it with `FeeAmount` when building a tx?

inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,211 @@
...
+	allExempt := true
+	for _, msg := range msgs {
+		if !isNetworkDutyRecursive(msg, d.InferenceKeeper) {
+			allExempt = false
+			break
+		}
+	}
+
+	if !allExempt {
+		return next(ctx, tx, simulate)
```
DimaOrekhovPS, line 60: what if exempt and non-exempt messages are mixed up? Would exempt messages still be processed without a gas meter?

inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	// Enforce gas cap on bypassed transactions.
+	if feeTx, ok := tx.(sdk.FeeTx); ok {
+		if d.GasCap > 0 && feeTx.GetGas() > d.GasCap {
```
DimaOrekhovPS, line 65: Do we have empirical estimations for how high we should set the network duty GasCap?

inference-chain/x/inference/types/fee_params.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"encoding/json"
+)
+
+// FeeParamsKey is the KV store key for FeeParams (stored separately from main Params proto).
+var FeeParamsKey = []byte("p_fee_params")
+
+// FeeParams defines governance-controlled fee parameters for consensus-level fee enforcement.
+type FeeParams struct {
```
DimaOrekhovPS, line 11: Usually all keeper-stored structures are defined as .proto messages and then the Go code is autogenerated. Is there a reason for writing the Go directly?

inference-chain/x/inference/keeper/msg_server_poc_v2_commit.go
```diff
@@ -87,6 +96,26 @@ func (k msgServer) PoCV2StoreCommit(goCtx context.Context, msg *types.MsgPoCV2St
 		}
 	}
 
+	// Consume extra gas for sybil resistance (two-component fee).
+	// Base validation gas: charged once per participant per epoch (covers GPU validation cost).
+	// Count gas: charged on delta (so total equals final_count * gas_per_poc_count).
+	feeParams := k.Keeper.GetFeeParams(ctx)
+	if isFirstCommit {
```
DimaOrekhovPS, line 103: Do I get it right that the logic here is to consume gas if it's a first ever commit of that participant, while keeping subsequent transactions free via ante decorators?

0xgonka:
(tx_manager hardcoded values) — Addressed: extracted MinGasPriceNgonka and BatchGasLimit constants, both paths now reference them. Re: GasPrices being redundant when FeeAmount is set — GasPrices is used by the factory for the single-tx path (cosmosclient.go), while FeeAmount overrides it in the batch path (getSignedBytes). They serve different code paths.

(mixed exempt/non-exempt) — Already handled. If messages are mixed, allExempt is false (line 52-57), the decorator passes through without clearing min gas prices or setting the bypass flag. The full transaction is charged normal fees. Gas metering is always active regardless.

(empirical GasCap estimation) — The 1M gas cap is based on observed gas consumption: a typical PoC batch or validation tx uses ~200-400K gas. 1M provides ~2-3x headroom. It's a code constant (not governance-adjustable) so it can be tuned in a future upgrade if needed.

(why not proto for FeeParams) — We initially added FeeParams to params.proto but it triggered regeneration of params.pb.go and params.pulsar.go, creating a large diff in generated files. Since FeeParams is stored under its own KV key with JSON serialization (not part of the main Params message), a standalone Go struct is simpler. Can be migrated to proto in a follow-up.

(PoC commit gas logic) — Not quite: subsequent commits are NOT free. The base validation gas is charged once (first commit per epoch), but count-proportional gas is charged on every commit based on the delta (count increase). So a participant sending 100 updates pays: base_validation_gas (once) + sum of deltas × gas_per_poc_count = final_count × gas_per_poc_count total.

akup:
> (why not proto for FeeParams) — We initially added FeeParams to params.proto but it triggered regeneration of params.pb.go and params.pulsar.go, creating a large diff in generated files. Since FeeParams is stored under its own KV key with JSON serialization (not part of the main Params message), a standalone Go struct is simpler. Can be migrated to proto in a follow-up.

It is better to add them to params.proto and regenerate. Yes it is usual to have large diff when proto is changed and all commits that change protocol have a large diff.
No reason to wait for follow-up it is better to add it to current PR to make it more complete

inference-chain/x/inference/keeper/msg_server_poc_v2_commit.go
```diff
@@ -75,7 +77,14 @@ func (k msgServer) PoCV2StoreCommit(goCtx context.Context, msg *types.MsgPoCV2St
...
 	existing, err := k.PoCV2StoreCommits.Get(ctx, pk)
-	if err == nil {
+	isFirstCommit := false
+	if err != nil {
+		if !errors.Is(err, collections.ErrNotFound) {
+			return nil, sdkerrors.Wrap(types.ErrIllegalState, fmt.Sprintf("failed to read commit: %v", err))
+		}
+		isFirstCommit = true
+	}
+	if !isFirstCommit {
```
akup, line 87: More clear logic here would be:

```go
_, err := k.PoCV2StoreCommits.Get(ctx, pk)
if err == nil {
    // existing commit path (not first)
} else if errors.Is(err, collections.ErrNotFound) {
    // first commit path
} else {
    // storage error
}
```

PR #940, opened 2026-03-24
Align legacy validator slashing with required collateral semantics

qdanik:
## Proposed changes

This PR aligns staking-hook / legacy validator slashing with the same required-collateral semantics already used by inference-driven slashing delivered in https://github.com/gonka-ai/gonka/pull/868

Previously, Tendermint-driven slashing in the collateral module always used the participant's full actual collateral as the slash base. That preserved legacy behavior, but it could over-slash participants who were over-collateralized relative to their effective weight.

With this change, the collateral module can query inference for the tokenomics-defined required collateral and use that value as the slash base when handling validator slashing.

## Behavior change

Before:
- staking-hook slashing used `requiredCollateral = 0`
- collateral slashing therefore used the participant's total actual collateral as the effective slash base

After:
- staking-hook slashing asks inference for the participant's required collateral
- slashing is applied against `min(requiredCollateral, totalActualCollateral)`
- if the provider is unavailable or the required collateral cannot be determined, the system safely falls back to the legacy behavior

## Compatibility and safety

This change does not introduce a store migration.

The new cross-module dependency is intentionally wired as optional:
- if inference provides the slash-base calculator, validator slashing uses the new tokenomics-aware behavior
- if it is not available, the collateral module falls back to legacy semantics by treating required collateral as zero

This keeps the rollout safe and avoids breaking app wiring in environments where the dependency is not present.

## Motivation

This makes staking-driven slashing consistent with the rest of the tokenomics model and prevents over-slashing participants based solely on excess deposited collateral rather than the collateral actually required for their effective weight.


inference-chain/x/inference/keeper/collateral_integration_test.go
```diff
@@ -185,18 +216,45 @@ func TestSlashingForDowntime_Integration(t *testing.T) {
...
 	mocks.CollateralKeeper.EXPECT().
-		Slash(gomock.Any(), participantAcc, expectedSlashFraction, types.SlashReasonDowntime, gomock.Any()).
+		Slash(gomock.Any(), participantAcc, expectedSlashFraction, types.SlashReasonDowntime, expectedRequiredCollateral).
 		Return(sdk.NewCoin(types.BaseCoin, math.NewInt(0)), nil).Times(1)
 
 	// Execute the function under test directly
 	k.SlashForDowntime(ctx, participant, params)
 }
 
+func TestSlashingForInvalidStatus_Integration_GracePeriodUsesZeroRequiredCollateral(t *testing.T) {
```
x0152, line 230: The grace-period test passes because no effective epoch is seeded, not because of the grace-period logic

x0152 (approved):

PR #941, opened 2026-03-24
fix: allow zero timestamps in inference_finished event parsing

libermans:
When FinishInference arrives on-chain before StartInference, the inference_finished event is emitted with StartBlockTimestamp=0 (only ProcessStartInference sets it). The threshold check in parseEventUnixMillis rejected 0 as a "seconds" timestamp. A value of 0 means "not yet set", not a unit confusion — the subsequent StartInference event will upsert the correct value. Time-based filtering is unaffected since inference_timestamp falls back to EndBlockTimestamp.


x0152:
Just a minor question (not a problem): might be better to allow zero only for start_block_timestamp? 
at the call site this also lets zero through for end_block_timestamp, which shouldn't really happen

0xgonka (approved):

PR #945, opened 2026-03-25
[P0] Acquire node grpc

pixelplex:
## Add NodeManager gRPC server for ML node acquisition

Adds a gRPC server to `decentralized-api` that allows external services (e.g. subnets) to acquire and release ML nodes via gRPC.

### Core changes

- `nodemanager/` — new package with `.proto` definition, generated gRPC stubs, and server exposing `AcquireMLNode` and `ReleaseMLNode` RPCs
- `broker/node_lock.go` — UUID-keyed lock map on top of the existing `LockCount` mechanism; TTL-based automatic eviction via the reconciler loop
- `apiconfig/` — two new config fields: `node_manager_grpc_port` and `node_manager_lock_ttl_seconds` (defaults to 1200s); NodeManager gRPC server only starts when port is explicitly configured
- `subnet/mlnode/` — new package with gRPC client and `DoWithNode` retry helper for subnets to acquire nodes with transport-error retry and structured logging
- `main.go` — starts the NodeManager gRPC server conditionally alongside existing servers

### Infrastructure

- `local-test-net/docker-compose-base.yml` — exposes port 9400 for NodeManager gRPC

### Tests

- Unit tests for the lock map (`broker/node_lock_test.go`) and gRPC server error mapping (`nodemanager/server_test.go`)
- End-to-end integration tests in testermint (`NodeManagerTests.kt`)

Issue #929, opened 2026-03-21
Inference validation optimization


tcharchian:
@tamazgadaev, could you please provide a detailed description?

Red-Caesar:
[Here](https://github.com/gonka-ai/gonka/pull/952) I've opened a proposal for this optimization.

PR #948, opened 2026-03-25
Upgrade v0.2.12

gmorgachev:
This document outlines the proposed changes for on-chain software upgrade v0.2.12.   
The `Changes` section details the major modifications, and the `Upgrade Plan` section describes the process for applying these changes.

## Upgrade Plan

This PR updates the code for the `api` and `node` services. The PR modifies the container versions in `deploy/join/docker-compose.yml` and introduces a new `versiond` service in the join stack.

The binary versions will be updated via an on-chain upgrade proposal. For more information on the upgrade process, refer to [`/docs/upgrades.md`](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.12/docs/upgrades.md).

Existing hosts are **not** required to upgrade their `api` and `node` containers as part of the on-chain upgrade itself. After the upgrade, hosts must deploy the new `versiond` service and update and redeploy `proxy` with `VERSIOND_SERVICE_NAME=versiond` and `GONKA_API_EXEMPT_ROUTES=chat inference poc/proofs devshard` so `/devshard/<version>/*` traffic is routed through `proxy -> versiond -> devshardd`. New hosts joining after the upgrade should use the updated container versions from this [compose file](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.12/deploy/join/docker-compose.yml).

## Proposed Process

2. If the on-chain proposal is approved, this PR will be merged immediately after the upgrade is executed on-chain.

## Testing

The on-chain upgrade from version `v0.2.11` to `v0.2.12` has been successfully deployed and verified on the testnet. No regression in core functionality or performance has been observed during testing. More testing will be executed leading up to the upgrade.

Reviewers are encouraged to request access to testnet environments to validate both node behavior and the on-chain upgrade process, or to replay the upgrade on private testnets.

## Migration

The on-chain migration logic is defined in [`upgrades.go`](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.12/inference-chain/app/upgrades/v0_2_12/upgrades.go).

Migrations:

- Auto-creates `x/feegrant` allowances for every existing cold-to-warm ML ops authz grant to support transaction fees.
- Initializes `FeeParams`.
- Migrates singular PoC model parameters into the new multi-model `PocParams.Models` list and initializes `DelegationParams`.
- Clears legacy PoC v2 data (which used old key layouts) and seeds new pruning state markers for the new multi-model collections.
- Backfills `ActiveParticipant.VotingPowers` and `EpochGroupData` subgroup voting power for the current epoch to ensure seamless PoC validation post-upgrade.
- Removes unused `TopMiners` and training states (training will be moved to an off-chain architecture similar to devshards).

- [ ] Add koeff and params for new model

### [Multi-model PoC](https://github.com/gonka-ai/gonka/pull/1039)

Historically, PoC has been tied to a single base model. While the network aims to support multi-model inference, relying on a single-model PoC is not secure enough.

If the network served several models but only checked one during PoC, an attacker could spin up hardware just for the check and shut it down afterward. To prevent this, PoC must start immediately on the exact model being validated, proving the hardware is present and running that model *right now* with no window to swap deployments.

To support multiple models, this upgrade runs PoC for each model independently in separate model groups. The core mechanics:

- Each governance-approved model gets its own PoC group. PoC runs for all eligible groups in parallel.
- Weight is split into two layers. `PoC weight` is model-local and drives inference routing and inference rewards inside that specific group. `Consensus weight` is the total weight aggregated across all eligible model groups (using model-specific coefficients) that determines block signing power, voting power, and bitcoin-style rewards.
- Because not every Host can run every model, a Host not serving a model can delegate its consensus weight to a group member for PoC validation only (this does not affect block signing or governance voting power). This preserves the existing security model: a model group must reach a 2/3 validation threshold of the *total network consensus weight*, not just the group-local weight, even if its direct members hold less than that total amount.
- For each active model, Hosts must explicitly choose their participation mode (DIRECT, DELEGATE, REFUSE). Hosts who do nothing receive a penalty. Penalties are skipped during a model's initial grace period.

The current base model remains the starting group for bootstrapping additional models. The exact model coefficients and final parameter values are not yet part of this PR.

### Transaction fees for spam prevention ([#937](https://github.com/gonka-ai/gonka/pull/937), [#981](https://github.com/gonka-ai/gonka/pull/981))

v0.2.12 turns on consensus-level transaction fees for the first time. Before this upgrade, any funded account could broadcast an unlimited number of transactions at zero cost, because the chain relied only on per-validator `minimum-gas-prices` configuration, which is mempool-only and trivially bypassed by a malicious block proposer. This left governance proposals, bank sends, collateral management, reward claims, bridge operations, and CosmWasm calls without any economic friction against abuse.

v0.2.12 introduces a governance-controlled `FeeParams.min_gas_price_ngonka` enforced during both `CheckTx` and `DeliverTx`. The initial value is **10 ngonka per gas unit**. Governance can adjust the parameter without a chain upgrade.

Protocol-obligation messages — PoC submissions, validation messages, inference start/finish, BLS DKG rounds — are made fee-exempt via a `NetworkDutyFeeBypassDecorator`. These are already rate-limited by timing windows, duplicate checks, and epoch-scoping, so adding fees would not improve their spam resistance while complicating automated node operation.

For Host sybil resistance specifically, `MsgPoCV2StoreCommit` charges a two-component fee: a base validation cost charged once per participant per epoch (covering the GPU work validators have to perform), plus a count-proportional cost charged on each count delta. A sybil claiming high compute weight across many fake participants pays proportionally to the weight they are trying to claim, making large-scale sybil attacks economically prohibitive.

### Devshards (formerly "subnets") — standalone, versioned runtime ([#1045](https://github.com/gonka-ai/gonka/pull/1045))

Previously, the devshard runtime lived inside the main DAPI process. Upgrading devshards meant rebuilding, redeploying, and restarting the entire DAPI, which slowed down development and added risk to all Host work (including inference, PoC, and Confirmation PoC).

To solve this, v0.2.12 decouples devshards into a standalone, versioned runtime managed by a new service called `versiond`.

- `versiond` automatically downloads and runs devshard binaries approved by on-chain governance.
- Multiple devshard versions can run side-by-side. Traffic to `/devshard/<version>/*` is routed to the corresponding binary, while the legacy `/v1/devshard/*` route remains active during the transition.
- The standalone devshard directly communicates with MLNodes during inference but does not manage their lifecycle, cleanly separating the roles of MLNode manager (DAPI) and client.
- Each session is cryptographically bound to the specific binary version that served it. The settlement payload now includes a cleartext `version` field, ensuring a session cannot mix responses from different versions.
- The term "subnet" is entirely replaced by "devshard" across the codebase. Additionally, float math in devshard settlement has been replaced with deterministic integer arithmetic to eliminate consensus-failure risks.

### Certik audit fixes

The Certik audit produced a number of findings across the chain, bridge, BLS, and inference modules. This upgrade addresses the full set of findings flagged for v0.2.12: GEB-44, GEB-45, GEB-46 (ETH/WGNK address collision), GEB-51, GEB-54, GOC-15, plus a second batch of bridge and BLS findings. No known findings from the audit remain unaddressed.

### Removals and cleanups

- The unused TopMiner reward logic is removed, and the upgrade handler clears the `TopMiners` collection during migration with no financial impact.
- The in-chain training placeholder is removed. The feature was never used and carried security risks, so training is moving to an off-chain architecture similar to devshards.
- Developers no longer need to register as a Participant to run inference, as an account with a public key on chain is now sufficient.

### Protocol hardening and correctness

- PoC v2 RNG is stronger because the new mechanism addresses the 32-bit entropy flaw that made forged proofs feasible. It will be activated via an additional governance vote once MLNodes are updated.
- The MLNode version is now propagated to chain state so the network always reflects the exact software version each node is running. This allows the network to track adoption of new MLNode versions.
- A long-standing consensus issue in the BLS DKG dealing phase is corrected.
- Validator slashing now consistently aligns with the required-collateral model instead of legacy behavior.
- Fixes a bug where `inference_finished` event parsing failed on zero-timestamps.

### API and tooling improvements

- The DAPI now accepts multipart-encoded OpenAI requests alongside JSON to improve compatibility with upstream SDKs.
- More accurate HTTP status codes and error shapes, including 400/422 for malformed payloads, ensure upstream OpenAI SDKs behave correctly.
- The DAPI exposes node acquisition RPCs in the private network with TTL eviction via a new NodeManager gRPC server, which lets external services like devshards coordinate MLNode usage cleanly.
- End-to-end inference validation tests in Testermint are updated to cover status transitions.
- Deployment documentation is updated with multisig and access-control setups for production operators.

##### Multi-model

For Hosts, participation logic is now evaluated on a model-by-model basis. For each governance-approved model, a Host must choose one of the following participation modes:

- DIRECT mode means the Host runs the model and participates in its PoC directly.
- DELEGATE mode means the Host delegates its PoC validation power for that model to another Host running it.
- REFUSE mode means the Host explicitly refuses participation for that model.
- INTENT mode means the Host declares early intent to participate before deploying hardware for models approved but not yet active.
- NONE mode means the Host does nothing for that model (this will result in a penalty).

*Note: Delegation is necessary because not every Host can realistically run every model due to hardware constraints. Without delegation, a model group whose direct members hold less than 2/3 of the total network weight could never pass PoC validation.*

- [ ] Instruction with exact commands to be published

##### Fees

The v0.2.12 upgrade introduces transaction fees, which means the warm key (ML operational key) needs to pay for automated transactions like reward claims and hardware diff updates. Because the warm key holds zero balance by design, it needs an `x/feegrant` allowance from the cold account.

For existing Hosts, this migration is fully automatic. The upgrade handler will automatically create an `x/feegrant` allowance (valid for 1 year) from the cold key to the warm key.

**Host's responsibility:** Ensure the cold account holds some coins e.g. **100 GNK** . This balance will be enough to cover the expected automated fee burden for several years, plus any manual operations.

New Hosts onboarding after the upgrade will run the same `grant-ml-ops-permissions` command as before; the command has been updated to issue both the authz grants and the `x/feegrant` allowance in a single transaction.

The full onboarding and upgrade flows are documented in [`docs/host_onboarding.md`](https://github.com/gonka-ai/gonka/blob/upgrade-v0.2.12/docs/host_onboarding.md).

- [ ] Instruction with exact commands to be published

## Test plan

- [x] Chain unit tests
- [x] Decentralized-API unit tests
- [x] Testermint integration suite
- [x] Manual QA on testnet: confirm auto-migration creates feegrant allowances for all existing pairs, DAPI auto-configures gas price, reward claims and hardware diffs succeed post-upgrade without host action
- [x] Manual QA on testnet: confirm multi-model PoC groups, delegation, and model eligibility conditions work end-to-end
- [ ] Governance voting period (post review)

## Proposed Bounties

Bounty ID | Sum USDT | Bounty Explanation | Public Name 
|-----------|----------|---------------------|-------------|
PR #988, #1020, #1021 | 6,000 | CertiK audit fixes (GEB-29, GEB-35, GEB-44, GEB-45, GEB-51) | @x0152
PR #825 | 3,000 | DKG dealer consensus | @x0152 
PR #750 | 1,000 | Developer inference access / account API changes | @x0152 
PR #614 | 500 | OpenAI compatibility and API error handling | @x0152
v0.2.12 | 2,500 | Release management | @x0152
v0.2.12 | 5,000 | Release management | @akup
Issue #929 | 9,000 | Inference validation optimization | @Red-Caesar, [Axel-T](https://www.axel-t.com/)
PR #945 | 3,000 | Acquire node gRPC | @pixelplex 
PR #789 | 2,000 | Fund atomicity error safety | @0xMayoor
PR #940 | 1,500 | Align validator slashing with required collateral | @qdanik 
— | 500 | Report of free inference vulnerability (fixed by devshards release) | Rufat 
Issue #499 | 200 | Chat completions fix (missed from v0.2.10) | @pentoxine 
v0.2.11 | 1,000 | Review of upgrade | @blizko


decentralized-api/cmd/devshardd/main.go
```diff
@@ -0,0 +1,363 @@
...
+
+	httpClient := pserver.NewNoRedirectClient(5 * time.Minute)
+
+	chainParams := newChainParamsProvider(ctx, recorder)
+
+	engine := newDevshardEngine(mlClient, payloadStore, httpClient, chainParams)
+	validator := newDevshardValidator(mlClient, httpClient, br, recorder, engine, chainParams)
+
+	storePath := filepath.Join(*dataDir, "devshardd.db")
+	store, err := devshardstorage.NewSQLite(storePath)
```
blizko, line 134: Is it considered that only 1 instance of `devshardd` will exist per validator?

docs/consumer_setup.md
```diff
@@ -1,161 +1,189 @@
...
 
-To participate in the network, you need to create a local account. This will generate a public/private keypair and an account address.
+</details>
 
-1. Run the following command to create a new account:
-   ```bash
-   inferenced keys add {{account_name}}
-   ```
+<details>
+<summary><b>How to pick a random active node instead</b></summary>
```
blizko, line 41: This does not take into account that there are TransferAgent whitelists enabled right now

0xMayoor:
re #1039:
`PocParams.Validate()` in `types/params.go` checks for negative SeqLen and duplicate model_ids but doesn't check `WeightScaleFactor`. the Decimal Value field is signed int64 so a negative coefficient can pass through governance.

that flows into `calculatePocParticipatingNodesWeight` where `coeff.MulInt64(rawWeight).TruncateInt64()` produces a negative weight that gets stored as `ConfirmationWeight`. there's a clamp in `bitcoin_rewards.go:652` so rewards aren't affected, but the negative value is still on chain.

akamitch:
### Currently, the plan is to use multi-PoC. Big problem!
GPU speed will be tested using the model that the miner has chosen, and the weight will be calculated based on this test.
To make all tests performed on different models comparable and to derive a universal weight, there will be conversion coefficients for recalculating into a universal weight.
The problem with this approach is that it's not market-driven.

Nobody except the user who pays for inference can answer the question "how much more valuable is a response from Kimi-2.5 compared to a response from Qwen-235."
Developers will have to set some arbitrary correspondence values.
Yes, these coefficients can be changed later.

For example, it becomes clear that there is huge demand for Kimi-2.5, all B-200s and B-300s are loaded and working. Meanwhile, miners' H-100s sit idle running Qwen-235.
At the same time, say 80% of the network weight, calculated by the current formula, belongs to miners with H-100s.
So the coefficient should be changed through a proposal to increase the value of Kimi-2.5. Miners with H-100s would have to vote to lower their own weight, but they want the opposite.

### "You get what you pay for."
In blockchains, you cannot rely on altruism.
The miner reward mechanism must be designed so that miners' selfish pursuit of profit drives the network toward providing quality inference using the models that users want.
So that miners have a strong incentive to run exactly the models that users are currently using and are willing to pay more for.
The current PoC approach does not align incentives in the right direction.
And this will definitely become a problem on the path to network growth and achieving the goal of large-scale, actually working inference delivery.

### The solution is to implement a market mechanism as fully as possible.
So that to maximize profit, miners have to adapt not to network settings, but to user demand.
And the additional reward of 250k gonka per epoch should be distributed not by weight (which is not needed at all in this model), but by work-coins earned during the epoch.
The idea is simple — the more a miner has earned, the more value they have delivered to end users, since users paid for those API requests.

Remove PoC.
Remove network-level token price regulation triggered at the 40–60% network load thresholds.

### Allow miners to set their own price for inference access.

It seems that the price should include not only the model choice but also the maximum context length. Currently, in "illegal optimizations," it is precisely this parameter that gets cut.
So that if we happen to have an API client that needs a huge context window and is willing to pay for such requests, miners automatically start catering to that demand.
Allow miners to switch models within an epoch if demand for a particular model has increased and it is more profitable to run it right now.
Of course, this auto-adjustment should be done not by a human but by the miner's software.
The software should first benchmark the hardware across different models and context windows, calculating its performance in all modes.
Then, based on current statistics — what API requests have been coming in recently, what competitor miners are offering — it should put its own "offer on the market."
The miner's software can have a strategy here, like a trading bot on an exchange.
On the one hand, try to set a price lower than competitors offering the same model, to capture more requests from clients.
If the miner's capacity is already 60% loaded at that price, then it makes sense to raise the price. Similar to how it used to work for the entire network, but at the level of each individual miner.
And if a miner has a completely unique offering — for example, they are the only one with Qwen-235 with a 1M context — they can set a very high price, as high as what users are actually willing to pay.
Requests for model/context combinations that don't currently exist in the network may also come in. These requests should be visible to miners so they can put up an offer to meet that demand.
Slashing mechanisms are also very important.
Random re-verification of requests, as it works now, stays.
But for the act of verification, the node that performed the check should also receive a reward in work-coins. Ideally, they should simply split the work-coin payment from the client 50/50.
I would also like to introduce slashing "for false accusations." So that if a modified node assigned to "verify the quality of another node's work" lies by saying "the task was completed by an invalid model," it can be slashed for that.
There doesn't seem to be any overhead here: it's enough to randomly re-verify all violation claims, and punish "false accusations" more severely than failure to deliver inference.
Protection against self-buying.
Suppose I launched Qwen-235 with a 1M context, set a price of 100 gonka per token, and bought such API requests from myself, quickly cycling the turnover into billions of gonka.
From myself — because there is no other such offer in the network, and all requests would go to me.
Then the system would register that enormous value was delivered and would hand over a substantial share of the epoch reward.
To protect against this attack, I think it is sufficient for 10% of API payments to go into the "epoch budget."

gmorgachev:
> ### Currently, the plan is to use multi-PoC. Big problem!
> GPU speed will be tested using the model that the miner has chosen, and the weight will be calculated based on this test. To make all tests performed on different models comparable and to derive a universal weight, there will be conversion coefficients for recalculating into a universal weight. The problem with this approach is that it's not market-driven.
> ....

Hi! Agree that chain needs to introduce alignment with usage data and how model's a chosen / weighted / etc. and in long term perspective weight coefficients should not be defined only by government voting, should be something dynamic based on real demand. There are several options for that, i'm thinking more in area of quota dynamically defined based on usage but want to describe my thoughts more clear to not be vague 

I can't agree with point remove PoC as the weight definition is core mechanism of chain's security consensus, we should be really careful with that. It might be based on inference + inference validation theoretically but then security model of the inference validation must be signigicantly different

Again don't want to be vague. I want to talk more about my opinion of how it should look like in the future during next AMA (i think it's already announced) and them publish more detailed text on that

decentralized-api/cmd/devshardd/main.go
gmorgachev: It might be multiple instances once postgress is used as storage

PR #952, opened 2026-03-26
The proposal for inference validation optimization

### LINKED ISSUE #929: Inference validation optimization

Red-Caesar:
This PR proposes changes to the algorithm for validating inference. The key idea is to use prefill instead of decodes in the enforced sampling algorithm. For some long sequences, it can even give a 15% speedup.


akup:
Are there experimental data for same models with different quantization?
Are there experimental data for same family but different models?


We need to be sure that inferences generated by similar but different models are detected

tamazgadaev:
@akup Thanks for the review! Here is the inference validation graph for int4 vs fp8 Qwen 235B model.
Artifacts can be found here: https://drive.google.com/drive/folders/15mkxAjli2NPhtY7ySAmZy1H2n7p2kt9a

Screenshot 2026-04-07 at 12 51 14

Int4 vs fp8 comparison is enough, because this is the closest possible two models we can check and use int4 as a fraud model.

patimen:
Thanks for putting this together. Looking at current `main`, the validation-optimization direction from this PR appears to have been implemented in a different shape already:

- current validation docs: https://github.com/gonka-ai/gonka/blob/main/mlnode/docs/inference-validation.md
- current validation helpers / distance logic: https://github.com/gonka-ai/gonka/blob/main/mlnode/packages/benchmarks/src/validation/utils.py
- current benchmark flow: https://github.com/gonka-ai/gonka/tree/main/mlnode/packages/benchmarks/scripts/inference_validation
- upgrade/bounty note tying this work to Issue #929: https://github.com/gonka-ai/gonka/blob/main/inference-chain/app/upgrades/v0_2_12/upgrades.go

So I’m going to close this PR as no longer the right vehicle.

If there are still specific components here that you think have not been carried forward, a separate narrow PR for just those pieces would make sense and would be much easier to review.

PR #953, opened 2026-03-26
Cached storage and conflict-detection for optimistic mode

akup:
## Summary

This PR introduces a **two-level block/tx cache mechanism** for inference module system data, with built-in conflict tracking for future OCC (optimistic concurrent) execution support.

It is a part of a bigger proposal for optimistic parallel execution, that can significantly increase mainchain throughput:
https://github.com/gonka-ai/gonka/discussions/954
https://github.com/akup/gonka/blob/944925f58a357c0871e1f8e5b8c3fe26caad1e19/proposals/block-tx-cache-occ/README.md

Main points:

- Added a deterministic **store wrapper** (`OptimisticStore`) with:
  - **Tx draft layer** (per-transaction, context-scoped write-behind)
  - **Block cache layer** (shared in-memory cache for current block)
  - **Persistent backend layer** (collections/raw KV as source of truth)
- Added **conflict detection hooks** (read/write set tracking) to support future parallel scheduler work, while current execution remains sequential.
- Wired lifecycle through `AnteHandler` / `PostHandler` / `BeginBlock` / `EndBlock`:
  - create drafts in ante
  - commit/release drafts in post
  - invalidate block cache at block start
  - flush cache to persistent store at block end
- Applied this mechanism to shared frequently-read system data (`Params`, `EpochGroupData`).
- **Removed explicit ParamsInjection calls** from msg server handlers, because params are now handled automatically by the cache/store mechanism.

For full design and rationale, see:
`proposals/block-tx-cache-occ/README.md`

---

## Store Mechanism (How it Works)

1. **Read path**: `tx draft -> block cache -> persistent store`
2. **Write path**: writes go to tx draft first
3. **On tx success**: tx draft merges into block cache
4. **On tx failure**: tx draft is discarded
5. **EndBlock**: block cache flushes to persistent store
6. **BeginBlock**: block cache/conflict tracker reset for next block

This removes repeated protobuf marshal/unmarshal on hot reads/writes within the same block flow, especially for system data.

---

## Why ParamsInjection Was Removed

Previously, handlers manually injected params into context.
Now params are retrieved through the same cache-backed store flow, so explicit per-handler injection is unnecessary and was removed to simplify call sites and avoid duplication.

---

## When This Cache Should Be Used

Use this cache for data that is:

- shared across many txs in the same block
- frequently read in system flows
- expensive to repeatedly marshal/unmarshal
- deterministic and safe to stage in tx/block layers before final flush

Good examples in this PR: `Params`, `EpochGroupData`.

Avoid using it blindly for rarely-accessed or highly specialized one-off state unless there is a demonstrated performance need.

---

## Notes

- This PR does **not** enable parallel tx execution by itself.
- Conflict detection is included now to prepare for a future OCC scheduler phase.
- Current behavior remains deterministic and compatible with sequential execution.

### DIFF
--- a/inference-chain/app/ante.go
+++ b/inference-chain/app/ante.go
@@ -25,6 +25,43 @@ import (
 	inferencetypes "github.com/productscience/inference/x/inference/types"
 )
 
+// EpochGroupDraftDecorator binds a tx-scoped EpochGroupData draft to the request context at tx start.
+// PostHandler must call CommitEpochGroupDraftFromContext on success; block cache is flushed to store in EndBlock.
+type EpochGroupDraftDecorator struct {
+	InferenceKeeper *inferencemodulekeeper.Keeper
+}
+
+// AnteHandle attaches WithEpochGroupDraft to the context for the rest of the tx.
+func (d EpochGroupDraftDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	if d.InferenceKeeper == nil {
+		return next(ctx, tx, simulate)
+	}
+	newCtx := ctx.WithContext(inferencemodulekeeper.WithEpochGroupDraft(ctx.Context()))
+	return next(newCtx, tx, simulate)
+}
+
+// EpochGroupDraftCommitPostDecorator merges the tx-scoped EpochGroupData draft into the block cache on tx success.
+type EpochGroupDraftCommitPostDecorator struct {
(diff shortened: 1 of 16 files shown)


patimen:
Thanks for putting this together — there is clearly a lot of thoughtful work here. Since the chain has picked up some narrower caching changes on main since this was opened, do you still think this broader optimistic-store / OCC direction is the right vehicle to pursue, or would you prefer to refresh and narrow it to the pieces that still make sense today? Curious for your read before we make any triage decision.

PR #957, opened 2026-03-26
fix: add graceful shutdown with signal handling in decentralized-api

Mayveskii:
The decentralized-api main() creates a context with context.WithCancel but nothing ever calls cancel(). When SIGTERM arrives, the process is killed without draining in-flight requests, closing NATS, or stopping the gRPC server.

Replace context.WithCancel with signal.NotifyContext so the context cancels on SIGINT/SIGTERM, then add gRPC GracefulStop() to drain in-flight RPCs before exit.

decentralized-api/main.go
Severity: MEDIUM — data loss during restarts/upgrades

### DIFF
--- a/decentralized-api/main.go
+++ b/decentralized-api/main.go
@@ -35,6 +35,8 @@ import (
 	"log"
 	"log/slog"
 	"os"
+	"os/signal"
+	"syscall"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


x0152:
Hi
This PR changes shutdown signaling, but it does not provide coordinated shutdown for all HTTP servers.
So graceful shutdown is still not implemented, and in-flight HTTP requests can still be dropped.
Closing for now. Feel free to reopen if you disagree

PR #958, opened 2026-03-26
fix: add per-epoch validation submission rate limit

Mayveskii:
A validator can submit validations for thousands of distinct inferences in rapid succession. Each passes the dedup check (per epochId+validator+inferenceId) because the inference ID is different each time. MaximumInvalidationsReached caps invalidations but NOT validations.

Add a per-validator per-epoch cap of 500 validations. This prevents bandwidth DoS while allowing normal operation (typical validators submit <100 validations per epoch).

Note: CountEpochGroupValidations needs to be implemented in the keeper — this commit adds the call site. The counting function should iterate EpochGroupValidationEntry with the epoch+creator prefix.

msg_server_validation.go
Severity: MEDIUM — bandwidth DoS vector on validator submissions

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_validation.go
+++ b/inference-chain/x/inference/keeper/msg_server_validation.go
@@ -63,6 +63,15 @@ func (k msgServer) Validation(goCtx context.Context, msg *types.MsgValidation) (
 	}
 
 	if !msg.Revalidation {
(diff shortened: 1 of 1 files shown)


x0152:
It hardcodes a fixed 500 limit in consensus logic, while limits here are designed to come from on-chain params and weight-based rules.
It also silently returns success when validation is dropped, which changes protocol behavior.

PR #960, opened 2026-03-26
fix: add max-size guard to bandwidth limiter usage maps

Mayveskii:
BandwidthLimiter.usagePerBlock and inferencesPerBlock maps grow with every new block height between 30-second cleanup intervals. Under sustained high-throughput load, thousands of block-height keys accumulate without bounds.

Add an inline size check: if maps exceed 10,000 entries, trigger cleanup immediately before inserting. This bounds worst-case memory growth without changing the normal cleanup interval.

decentralized-api/internal/bandwidth_limiter.go
Severity: LOW — memory leak under sustained load

### DIFF
--- a/decentralized-api/internal/bandwidth_limiter.go
+++ b/decentralized-api/internal/bandwidth_limiter.go
@@ -224,6 +224,10 @@ func (bl *BandwidthLimiter) RecordRequest(startBlockHeight int64, estimatedKB fl
 	defer bl.mu.Unlock()
 
 	completionBlock := startBlockHeight + bl.requestLifespanBlocks
+	// Guard: if maps grow too large between cleanup intervals, force inline cleanup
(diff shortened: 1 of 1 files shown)


x0152:
Map is already cleaned up by a periodic routine based on requestLifespanBlocks. The hardcoded 10000 threshold is not tied to any config parameter + triggering a full cleanup inline under a write lock on every RecordRequest call is a hot-path penalty that could be worse than the original issue

Mayveskii:
> Map is already cleaned up by a periodic routine based on requestLifespanBlocks. The hardcoded 10000 threshold is not tied to any config parameter + triggering a full cleanup inline under a write lock on every RecordRequest call is a hot-path penalty that could be worse than the original issue

I wasted your time, my apologies

PR #963, opened 2026-03-26
[P0] [#899] Add some E2E inference validation tests

### LINKED ISSUE #899: [P0] `devshards`: Add end-to-end inference validation tests
We should write testermint tests to ensure that inference validations in `devshards` work as expected. Such tests already exist for mainnet, but we should reimplement them according to the `devshards` design and implementation.

heitor-lassarote:
Problem: We need to ensure that inference validation is working as intended, and using Testermint, we can write end-to-end tests for that.

Solution: Expose a new endpoint to query inferences from subnetctl and check that the inference status is as expected with two new tests. Add some Kotlin data classes mimicking the Go types and serialize inferences.

### DIFF
--- a/testermint/src/main/kotlin/LocalInferencePair.kt
+++ b/testermint/src/main/kotlin/LocalInferencePair.kt
@@ -773,6 +773,14 @@ data class LocalInferencePair(
         } catch (_: Exception) { /* ignore */ }
     }
 
+    fun getSubnetInferenceState(proxyUrl: String, inferenceId: Long): String {
+        val result = api.executor.exec(listOf(
+            "sh", "-c",
+            "curl -sf $proxyUrl/v1/inference -H 'X-Inference-Id: $inferenceId'"
+        ), null)
+        return result.joinToString("")
+    }
+
     fun sendChatCompletion(proxyUrl: String, model: String, prompt: String, stream: Boolean = false): String {
         val body = """{"model":"$model","messages":[{"role":"user","content":"$prompt"}],"max_tokens":100,"stream":$stream}"""
         val result = api.executor.exec(listOf(
--- a/testermint/src/main/kotlin/data/subnet.kt
+++ b/testermint/src/main/kotlin/data/subnet.kt
@@ -80,3 +80,93 @@ data class SubnetSlotSignatureEntry(
     val slotId: Int,
     val signature: String
 )
+
+data class SubnetInferencePayload(
+    val status: SubnetInferenceStatus,
+    @SerializedName("executor_slot")
+    val executorSlot: Int,
+    val model: String,
+    @SerializedName("prompt_hash")
+    val promptHash: String,
+    @SerializedName("response_hash")
+    val responseHash: String?,
+    @SerializedName("input_length")
+    val inputLength: Long,
+    @SerializedName("max_tokens")
+    val maxTokens: Long,
+    @SerializedName("input_tokens")
+    val inputTokens: Long?,
+    @SerializedName("output_tokens")
+    val outputTokens: Long?,
+    @SerializedName("reserved_cost")
+    val reservedCost: Long,
+    @SerializedName("actual_cost")
+    val actualCost: Long?,
+    @SerializedName("started_at")
+    val startedAt: Long,
+    @SerializedName("confirmed_at")
+    val confirmedAt: Long?,
+    @SerializedName("votes_valid")
+    val votesValid: Int?,
+    @SerializedName("votes_invalid")
+    val votesInvalid: Int?,
+    @SerializedName("validated_by")
+    val validatedBy: Array<Long>?,
+) {
+    val statusEnum: SubnetInferenceStatus
+        get() = status
+}
+
+enum class SubnetInferenceStatus(val value: Int) {
+    PENDING(0),
+    STARTED(1),
+    FINISHED(2),
+    CHALLENGED(3),
+    VALIDATED(4),
+    INVALIDATED(5),
+    TIMED_OUT(6),
+    UNSPECIFIED(7);
+
+    companion object {
+        fun fromValue(value: Int): SubnetInferenceStatus =
+            values().find { it.value == value } ?: UNSPECIFIED
+
+        fun fromAny(value: Any?): SubnetInferenceStatus {
+            return when (value) {
+                is Number -> fromValue(value.toInt())
+                else -> UNSPECIFIED
+            }
+        }
+    }
+}
+
+data class SubnetChallengeReceiptRequest(
+    @SerializedName("inference_id")
+    val inferenceID: Long,
+    val payload: SubnetPayloadJSON,
+    val diffs: List<SubnetDiffJSON>,
+)
+
+data class SubnetChallengeReceiptResponse(
+    val receipt: List<String>,
+)
+
+data class SubnetDiffJSON(
+    val nonce: Long,
+    val txs: String,
+    @SerializedName("user_sig")
+    val userSig: String,
+    @SerializedName("post_state_root")
+    val postStateRoot: String,
+)
+
+data class SubnetPayloadJSON(
+    val prompt: String,
+    val model: String,
+    @SerializedName("input_length")
+    val inputLength: Long,
+    @SerializedName("max_tokens")
+    val maxTokens: Long,
+    @SerializedName("started_at")
+    val startedAt: Long,
+)
--- a/testermint/src/test/kotlin/SubnetTests.kt
+++ b/testermint/src/test/kotlin/SubnetTests.kt
@@ -1,5 +1,6 @@
 import com.productscience.*
 import com.productscience.data.*
+import kotlin.test.assertNotNull
 import kotlinx.coroutines.asCoroutineDispatcher
 import kotlinx.coroutines.async
 import kotlinx.coroutines.awaitAll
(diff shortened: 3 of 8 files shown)


testermint/src/main/kotlin/data/subnet.kt
```diff
@@ -53,3 +53,118 @@ data class SubnetSlotSignatureEntry(
...
+            inputTokens = null,
+            outputTokens = null,
+            reservedCost = 0,
+            actualCost = null,
+            startedAt = 0,
+            confirmedAt = null,
+            votesValid = null,
+            votesInvalid = null,
+            validatedBy = null,
+        )
```
dcastro: Looks like this might not be needed

testermint/src/main/kotlin/LocalInferencePair.kt
```diff
@@ -773,6 +773,14 @@ data class LocalInferencePair(
         } catch (_: Exception) { /* ignore */ }
     }
 
+    fun sendGetInference(proxyUrl: String, inferenceId: Long): String {
```
dcastro: Since `LocalInferecePair` has many methods not necessarily related to subnets, it might be worth renaming this method so it becomes clear we're talking about "inferences from subnets" (and not inferences in general).

Maybe `getSubnetInferenceState`?

testermint/src/test/kotlin/SubnetTests.kt
```diff
@@ -295,4 +317,160 @@ class SubnetTests : TestermintTest() {
...
+
+            logSection("Verifying inference status")
+            val inferenceId = 1L
+            val inference = cosmosJson.fromJson(
+                genesis.sendGetInference(handle.proxyUrl, inferenceId),
+                SubnetInferencePayload::class.java,
+            )
+            logSection("Inference: $inference")
+            assertNotNull(inference)
+            assertThat(inference.status).isEqualTo(SubnetInferenceStatus.FINISHED)
```
dcastro: Should we also check that an inference validation was actually performed?
E.g. by inspecting `result.parsed.hostStats` -> `SubnetHostStatsEntry.requiredValidations` + `completedValidations`

testermint/src/test/kotlin/SubnetTests.kt
```diff
@@ -295,4 +317,160 @@ class SubnetTests : TestermintTest() {
...
+            logSection("Verifying inference status")
+            val inferenceId = 1L
+            val maxTries = 10
+            val inference = (0 until numInferences).firstNotNullOf {
+                val inference = cosmosJson.fromJson(
+                    genesis.sendGetInference(handle.proxyUrl, it),
+                    SubnetInferencePayload::class.java,
+                )
+                inference?.status?.let { status ->
+                    if (status >= SubnetInferenceStatus.CHALLENGED) {
```
dcastro: Why are using `>=` and not `==` here?

testermint/src/main/kotlin/data/subnet.kt
```diff
@@ -53,3 +53,118 @@ data class SubnetSlotSignatureEntry(
...
+    }
+
+    fun checkComplete(): Boolean =
+        this.statusEnum >= SubnetInferenceStatus.FINISHED &&
+        this.statusEnum < SubnetInferenceStatus.UNSPECIFIED
```
dcastro: Checking whether a status is completed using `>=` and `<` seems error-prone; it's likely to break when new variants are added to the enum.

Kotlin's pattern matching is very limited, but luckily Kotlin does support matching on enums + exhaustiveness checks:

```
    fun checkComplete(): Boolean =
        when (this.statusEnum) {
            SubnetInferenceStatus.FINISHED,
            SubnetInferenceStatus.CHALLENGED,
            SubnetInferenceStatus.VALIDATED,
            SubnetInferenceStatus.INVALIDATED,
            SubnetInferenceStatus.TIMED_OUT -> true,
            SubnetInferenceStatus.PENDING,
            SubnetInferenceStatus.STARTED,
            SubnetInferenceStatus.UNSPECIFIED -> false
        }
```

That being said: it doesn't seem like this function is needed though.

testermint/src/test/kotlin/SubnetTests.kt
```diff
@@ -295,4 +317,160 @@ class SubnetTests : TestermintTest() {
...
+            logSection("Inference: $inference")
+            assertNotNull(inference)
+            assertThat(inference.status).isEqualTo(SubnetInferenceStatus.FINISHED)
+        } finally {
+            genesis.stopSubnetProxy(escrowId)
+        }
+    }
+
+    @Test
+    fun `invalid inference is challenged`() {
```
dcastro, line 368: Similarly to my other comment: should we also check that the counter of invalid inferences `SubnetHostStatsEntry.invalid` has been incremented?

testermint/src/main/kotlin/data/subnet.kt
heitor-lassarote: Deleted.

testermint/src/main/kotlin/LocalInferencePair.kt
heitor-lassarote: Good idea, renamed it.

testermint/src/test/kotlin/SubnetTests.kt
heitor-lassarote: Hmm, after printing `hostStats`, I see that `requiredValidations` is 1 for some hosts, but 0 for all of them.

testermint/src/test/kotlin/SubnetTests.kt
heitor-lassarote: I was hoping we can eventually get this to `VALIDATED` or `INVALIDATED` but I could not observe this happening. And the same time, I wanted to use `>=` in case it's ever observed. But I see using both `>=` and `isEqualTo` is inconsistent. Changed it to `==`. If we can get it beyond `CHALLENGED`, I think we can change these conditions.

testermint/src/main/kotlin/data/subnet.kt
heitor-lassarote: Removed it.

testermint/src/test/kotlin/SubnetTests.kt
heitor-lassarote: Interesting, I tried it out, but `invalidSum` is zero:

```kotlin
            val invalidSum = result.parsed.hostStats.sumOf { it.invalid }
            assertThat(invalidSum).isNotZero()
```

Possibly the threshold is not being hit, hence we never see it going from challenged to valid/invalid, and never see this invalid increasing?

akup: @heitor-lassarote
Actually it is related to your observation that inference is never INVALIDATED, but only CHALLENGED.
It will increase the hostStats invalid inferences only when the inference is INVALIDATED

It seams that it was finalized too early and participants had no time to vote.
It could be interesting to run the test with some sleep before finalization. If it still not moved to INVALIDATED, it seams to be some issue, that should be investigated in other task

heitor-lassarote: Unfortunately, that doesn't seem to help. I've attempted to increase the number of inferences sent to give hosts some extra time, tried to wait 10 seconds before finalizing, and also to wait before rounds of sending inference requests, but I was never able to get it to proceed past CHALLENGED. I fully agree with you this should be investigated in a new issue.

dcastro: @heitor-lassarote @akup I suspect the "challenge voting" mechanism isn't implemented yet, at least I don't see the struct `MsgValidationVote` being instantiated anywhere.

akup: So it is out of the scope

akup (approved):
It has no voting for invalidation, but it is out of the PR scope

testermint/src/test/kotlin/SubnetTests.kt
dcastro: > I see that `requiredValidations` is 1 for some hosts, but 0 for all of them.

Did you mean "but **completedValidations** is 0 for all of them"?

testermint/src/test/kotlin/SubnetTests.kt
akup: @dcastro @heitor-lassarote BTW. It seams that validation protocol will be completely rewritten.

I will prepare proposal tomorrow

testermint/src/test/kotlin/SubnetTests.kt
dcastro: But taking a step back, my main point was: This test, as is, is very very similar to the existing test "subnet inference e2e with settlement".

The only thing it adds is the assertion regarding `inference.status`; but that's just an assertion that could be added to the existing test. I don't think we'd need a whole e2e test just to assert `inference.status`.

dcastro (approved):

testermint/src/test/kotlin/SubnetTests.kt
heitor-lassarote: Exactly, sorry, had a brain fart writing this.

heitor-lassarote:
Force-pushed to squash commits and rebase the branch with `upgrade-v0.2.12`.

testermint/src/test/kotlin/SubnetTests.kt
heitor-lassarote: For some reason I didn't see your point about the similar tests. I pushed a commit removing the new one and adding the new assertion to the existing test.

PR #965, opened 2026-03-26
[P0] [#935] Fees

### LINKED ISSUE #935: [P0] `devshards` fees
Context: https://github.com/gonka-ai/gonka/issues/914#issuecomment-4090483233

* Calculate and charge fee for `devshards`
    * Initial impl: create_fee + max_nonce * fee_per_nonce
        * Reasoning: charging per nonce acts as a mechanism to deter from spamming the network with small inference requests
    * Ensure escrow amount covers the fee
    * Ensure the escrow balance never goes below the fee
    * Charge the fee upon settlement

dcastro:
This PR addresses #935:

* Calculate and charge fee for `devshard`
    * Initial impl: create_fee + max_nonce * fee_per_nonce
    * Ensure escrow amount covers the fee
    * Ensure the escrow balance never goes below the fee
    * Charge the fee upon settlement

The commits can be individually reviewed; they address 1 separate issue each.


subnet/state/machine.go
```diff
@@ -233,6 +245,13 @@ func (sm *StateMachine) ApplyLocalBestEffort(nonce uint64, txs []*types.SubnetTx
 		applied = append(applied, tx)
 	}
 
+	// Charge per applied nonce after tx effects are known.
+	if sm.state.Balance < sm.state.Config.FeePerNonce {
+		sm.restoreMutable(snap)
+		return nil, nil, types.ErrInsufficientBalance
+	}
+	sm.state.Balance -= sm.state.Config.FeePerNonce
```
dcastro: @gmorgachev 

I introduced a step that can possibly fail (subtract the initial subnet fee) in both `applyCore` and `ApplyLocalBestEffort`. 

I noticed `applyCore` has "rollback semantics", but `ApplyLocalBestEffort` doesn't.

So while adding my fee-related step, I also added rollback semantics to `ApplyLocalBestEffort`: is that okay, or should `ApplyLocalBestEffort` not have this for some reason?

testermint/src/test/kotlin/SubnetTests.kt
```diff
@@ -93,27 +93,41 @@ class SubnetTests : TestermintTest() {
...
             assertThat(totalCompletedValidations).isGreaterThan(0)
+            val totalCost = result.parsed.hostStats.sumOf { it.cost }
+            val expectedRemainder = escrowAmount - totalCost - result.parsed.fees
 
             logSection("Submitting settlement from user account")
             val settleResp = genesis.settleSubnetEscrow(result.rawJson, from = userKeyName)
             assertThat(settleResp.code).isEqualTo(0)
 
+            val settleEvent = settleResp.events.firstOrNull { it.type == "subnet_escrow_settled" }
+            assertThat(settleEvent).isNotNull()
```
heitor-lassarote: You can use `assertNotNull(settleEvent)` and then Kotlin picks up that it's not null, so `!!` won't be necessary.

testermint/src/main/kotlin/LocalInferencePair.kt
```diff
@@ -782,6 +782,20 @@ data class LocalInferencePair(
         return result.joinToString("")
     }
 
+    fun getSubnetProxyStatus(proxyUrl: String): SubnetProxyStatus {
+        val raw = api.executor.exec(listOf(
+            "sh", "-c",
+            "curl -sf $proxyUrl/v1/status"
```
heitor-lassarote: I know other functions also do this, but is it a good idea to use `-sf`? Why not `-sS`? This way there is more context about errors in tests.

testermint/src/main/kotlin/LocalInferencePair.kt
```diff
@@ -782,6 +782,20 @@ data class LocalInferencePair(
...
+        ), null).joinToString("")
+        val start = raw.indexOf('{')
+        val end = raw.lastIndexOf('}')
+        if (start < 0 || end < 0) {
+            error("status returned no JSON object. raw:\n$raw")
+        }
+        val json = raw.substring(start, end + 1)
+        return Gson().fromJson(json, SubnetProxyStatus::class.java)
```
heitor-lassarote, line 796: I also know other functions are doing this, but why do we need to find and extract the brackets? Is there extra non-JSON output?

heitor-lassarote (approved):
I won't say I'm familiar with all the code that was altered here, but overall the changes look good to me.

inference-chain/proto/inference/inference/tx.proto
```diff
@@ -623,6 +623,7 @@ message MsgSettleSubnetEscrow {
   bytes rest_hash = 5;
   repeated SubnetSettlementHostStats host_stats = 6;
   repeated SubnetSlotSignature signatures = 7;
+  uint64 fees = 8;
```
gmorgachev, line 626: Do you think explicit fee is better? Versus compute based on nonce

E.g. we can record fee price on time of subnet creation in subnet record (=> it's passed to subnet once started and price is known in subnet. mainnet also knows price on the time of creation in case mainnet price has changd)
And then on settle compute based on latest nonce `const_price + per_nonce_price*nonce`

gmorgachev: I would say we can avoid `fee` field in both subnet state and mainnet state. Using the approach above or just substract from balance on any changes. Then remainder logic would work automatically. Probably dynamic substract on nonce increment is simplest like:
```
func incrementNonce(state) {
     state.balance = state.balance - state.feePerNonce
     state.LatestNonce = state.LatestNonce + 1
}
```

dcastro: > Do you think explicit fee is better? Versus compute based on nonce

I thought this would be give us more flexibility, because then we can make adjustments to the fee structure / fee logic without having to change anything in mainnet.

E.g. we could remove the "per nonce fee", or switch from "fee per nonce" to "fee per tx", or add other kinds of fees, etc, and mainnet would only care about the total fees paid, we wouldn't have to update it.

dcastro: > just substract from balance on any changes

Yup, I am subtracting from balance (on escrow creation and on nonce produced, see `machine.go` -> `NewStateMachine` and `applyCore`), but we also have to keep track of the total fees paid so we can distribute them later.

testermint/src/main/kotlin/LocalInferencePair.kt
dcastro: Good idea. Changed it to use both flags, so we get the error details + the exit code.
```
$ curl -sf https://mock.httpstatus.io/400
$ $env.LAST_EXIT_CODE
22

$ curl -sS https://mock.httpstatus.io/400
400 Bad Request
$ $env.LAST_EXIT_CODE
0

$ curl -sfS https://mock.httpstatus.io/400
curl: (22) The requested URL returned error: 400
$env.LAST_EXIT_CODE
22
```

testermint/src/test/kotlin/SubnetTests.kt
dcastro: Good call, done!

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
```diff
@@ -28,56 +28,87 @@ func (k msgServer) SettleSubnetEscrow(goCtx context.Context, msg *types.MsgSettl
...
-	validatorCosts := make(map[string]uint64)
+	if len(escrow.Slots) == 0 {
+		return nil, fmt.Errorf("escrow %d has no slots", escrow.Id)
+	}
+
+	totalSlots := uint64(len(escrow.Slots))
+	// How much of the total fees will be assigned to each slot
+	feePerSlot := msg.Fees / totalSlots
+	// Leftover fees; will be distributed 1 per slot
+	remainderFees := msg.Fees % totalSlots
```
heitor-lassarote, line 39: Not sure if it's better or not, but perhaps `feePerSlot, remainderFees = bits.Div64(0, msg.Fees, totalSlots)`?

https://stackoverflow.com/a/65567628

heitor-lassarote (approved):
Still looks good to me.

akup:
There is an edge case.

Finalization, as it needs nonces incremention, will also cost the fee. And it is possible that we have no enough `escrowBalance` to finalize...
So we need to check not only 

```go
if sm.state.Balance < sm.state.Config.FeePerNonce
...
```

But have extra balance for finalization, or make nonce on finalization free.
Moreover balance is substracted for inferences, and we need to check there also that we have enough balance not only to start inference but also to finalize, and be able to finishInference (as it also costs fee).

Maybe it is much better to have fees not per nonce, but for tx, and have free transactions like `RevealSeed`, `FinishInference`.

there should be made design decision:
1. do we handle that extra fees are needed in balance checks (for finalization or finish inference). This makes it necessary that we also count how many non-finished inferences do we have, to be able to finish them.
2. we make some operations free from fees. But then it is better to go to per-tx fees, as in one nonce there can be mixed free and non-free txs

@gmorgachev @dcastro what do you think on this point?


Related thing that we have at `subnet_settlement.go`

```go
if totalCost > escrow.Amount {
  return fmt.Errorf("total cost %d exceeds escrow amount %d", totalCost, escrow.Amount)
}
```

I think it is very bad to throw error here. it is better to

```go
if totalCost > escrow.Amount {
  //log error here
  totalCost = escrow.Amount
}
```

just limit totalCost by escrow. Actually it is the collective subnet fault that totalCost is larger than escrowAmount and we just limit their reward but should not froze such situation by throwing error here.

gmorgachev:
@akup what about Validation / Votes transactions in that case? Also free? 

In general i think we must have finalization round to cost something but it's practically included in constant fee part 

I personally don't see big difference between per tx vs per nonce at the moment, per nonce just sounds much faster to start use

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
dcastro: Hmm I think the benefit of `bits.Div64` is when you need to do a multiplication that may overflow followed by a division. E.g. you use `bits.Mul64` which, instead of returning a uint128, returns 2x uint64s. You then pass those 2x uint64s into `bits.Div64`.

In other words, `bits.Div64` is most useful when the numerator has 128 bits, which is not the case here. I think keeping `/` and `%` is better for readability.

heitor-lassarote: Ok, I don't have strong opinions about either of them.

akup:
> @akup what about Validation / Votes transactions in that case? Also free?
> 
> In general i think we must have finalization round to cost something but it's practically included in constant fee part
> 
> I personally don't see big difference between per tx vs per nonce at the moment, per nonce just sounds much faster to start use

I'm also thinking that finalization is already can be (virtually) included in constant part, so we can make it free.

In general, validation and votes can be non-free. Overall it is the user interest to validate. It could be free, but we have validations built-in in finalization logic (hosts should validate missed validations), in this design they should be free.
But honestly, I don't understand reasons of this design decision, to make finalization dependent on validations. Moreover for a developer that got invalid, non-validated inference, that was invalidated after finalization, just don't make sence, as this inferences maybe are already used to produce results. And getting information, like there was few hour ago a mistake will not make developer happy.
But this question is not related to this PR, I'm keeping thinking about whole validation logic and finalization.

I have another idea of completely other approach of handling this:
1. Make developer ability to add some coins to escrow. So it will never be frozen forever
2. If user runs out of escrow balance, give some time to finalize for user (finalization is still free), and if timeout, finalize by hosts. But it makes need of actions without user sequencing.
3. Keep finalization free

Maybe it is much better strategically, than handling all special cases when transaction should be free, as there are special cases in more then half of message types

akup:
> I personally don't see big difference between per tx vs per nonce at the moment, per nonce just sounds much faster to start use

According to this point. As many transactions can be mixed in one batch with one nonce increment, we can mix free and non free messages, just by making nonce that contains at least one non-free tx to be non-free.

In general, per-tx fee looks to me more precise

dcastro:
> I'm also thinking that finalization is already can be (virtually) included in constant part, so we can make it free.

@akup: @gmorgachev and I discussed this, we've agreed to make finalization nonces free, to avoid situations where a subnet is stuck and can't be settled.

I've just pushed a commit addressing this.

subnet/state/machine.go
```diff
@@ -233,6 +246,16 @@ func (sm *StateMachine) ApplyLocalBestEffort(nonce uint64, txs []*types.SubnetTx
 		applied = append(applied, tx)
 	}
 
+	// Charge per applied nonce only during the active phase.
+	if sm.state.Phase == types.PhaseActive {
```
akup, line 253: Just to clarify.
It switches to `PhaseActive` just before this line, when we run txs apply cycle if there is `MsgFinalizeRound` in `txs[]`

Maybe add a comment about it?

akup (approved):

akup:
@dcastro please rebase, and could be merged

subnet/state/machine.go
dcastro: Yup, that's it.
Added a comment explaining this:

```
	// Charge per applied nonce only during the active phase.
	// NOTE: During the finalization round, the `txs` slice will contain a [types.MsgFinalizeRound] message,
	// the call to [StateMachine.applyTx] above will transition the state machine's phase to [types.PhaseFinalizing],
	// and this block will be skipped.
```

Issue #966, opened 2026-03-27
Validation Eligibility and Accounting Consistency


akup:
# Validation Eligibility and Accounting Consistency

Work on Inferences and Validations on current version is now legacy. All future work will be concentrated on devnets (shardchains). This PR shouldn't be high priority or probably should be closed in nearest future.
It was useful to see issues that should be handled in devchains

## Summary

This branch fixes a cumulative protocol issue across validation lifecycle, validator eligibility, and claim accounting. Before these changes, a non-settled participant could still attempt validation/revalidation, multi-address participants could bypass intent of own-inference restrictions, already-decided inferences could be re-processed in edge cases, and claim accounting around invalidation/revalidation was vulnerable to expensive or inconsistent handling.

## Motivation

This matters now because revalidation is expensive, and stricter status guards (`VALIDATED` / `VOTING`) require strict eligibility enforcement to avoid unauthorized heavy paths. At the same time, claim checks must remain both correct and performant under epoch-scale load.

## Impact

- Affected components: `x/inference/keeper/msg_server_validation.go`, `x/inference/keeper/msg_server_claim_rewards.go`, revalidation collections in keeper, seed-based settlement checks.
- Who is impacted (developers, hosts): core protocol developers, validators/hosts, node operators, and participants whose rewards depend on correct claim accounting.
- Which metric is expected to improve and how much: unauthorized validation/revalidation acceptance is expected to drop to zero for post-rollout strict mode; claim-path overhead is reduced by avoiding per-inference `GetInference(...)` scans in the hot validation-credit path.

## Detailed description

- How issue impact was measured/confirmed:
  - Validation flow analysis showed eligibility was not always tied to settled validator evidence.
  - Multi-address threat analysis showed own-inference protection can be bypassed without eligibility binding.
  - Runtime behavior and tests confirm status-guard semantics and claim-path behavior.
- Links to evidence:
  - [PR #832: VALIDATED/VOTING status guard logic and discussion](https://github.com/gonka-ai/gonka/pull/832)

UPDATE:
  [PR #829: INVALIDATED-related claim accounting discussion](https://github.com/gonka-ai/gonka/pull/829/changes) was referenced but it was my mistake. It seams it should be handled in other way. See discussion on that PR

## Expected outcome

When done:

- Only settled, eligible validators can validate or revalidate.
- Unauthorized revalidation initiation is blocked.
- Already-decided inference states are protected from duplicate first-time processing.
[...]

PR #967, opened 2026-03-27
Validation Eligibility and Accounting Consistency

akup:
# PR Description: Validation Eligibility, Status Guards

## Important

Work on Inferences and Validations on current version is now legacy. All future work will be concentrated on devnets (shardchains). This PR shouldn't be high priority or probably should be closed in nearest future

## Summary

Fix validation/revalidation authorization and reward-accounting edge cases described at https://github.com/gonka-ai/gonka/issues/966

This PR consolidates validation-path fixes and checking that only eligable validators can vote, based on:

- [PR #832: add VALIDATED/VOTING status guards in validation handler](https://github.com/gonka-ai/gonka/pull/832)

It closes attacks reported at https://github.com/gonka-ai/gonka/pull/832 and possibility that participant having multiple addresses validates own-served inferences (from other owned address)

### 1) Validation handler guards (from #832)

- Enforced early no-op behavior for already-decided inference states (`VALIDATED`, `VOTING`) in normal validation flow.
- Kept revalidation-aware handling where needed by voting lifecycle.
- Prevents repeated first-pass validation from re-triggering accounting and proposal side effects.

### 2) Validator eligibility is enforced via on-chain seed settlement

- Added/used chain-level seed storage as settlement evidence.
- Validation/revalidation sender eligibility is checked against this settlement context.
- This prevents non-settled participants from validating or initiating revalidation.
- This also closes a multi-address bypass scenario where one participant could act through another owned address.

## Upgrade Rollout (2 Stages)

This change should be activated in two stages due to seed availability at upgrade time.

### Stage 1: Post-upgrade transition epoch

- Immediately after upgrade, chain-level participant seeds may not exist yet for the target epoch.
- During this transition, validator-eligibility check based on seed settlement cannot be enforced strictly for all messages.
- The current logic in `inference-chain/x/inference/keeper/msg_server_validation.go` (see the block around `GetParticipantEpochSeed(...)`) temporarily allows skipping strict `ShouldValidate` gating when seed is absent.
- A new epoch must be started and participant seeds collected on-chain.

### Stage 2: Strict enforcement after seed population

- Once seeds are being stored on-chain for the epoch, missing seed must become a hard failure.
- Validation/revalidation should return an error if participant seed is not present (cannot prove eligibility).
- Concretely, the temporary "skip when seed is not found" path in `msg_server_validation.go` should be removed and replaced with strict error handling (`ErrRandomSeedNotFound` flow).

## Test Notes

- Ran focused keeper tests for `Validation` and `ClaimRewards` flows after the changes.
- Existing status-guard scenarios and claim paths pass with the new revalidation tracking.


0xMayoor:
@akup 
line 89:  with the !msg.Revalidation gone the executor gets blocked from voting on their own revalidation dispute. VerifyInvalidation in the off-chain code (inference_validation.go:59-81) doesn't filter the executor out before sending, and voting.go:37 still sets the executor as proposer on the revalidation proposal. 

should the off-chain side be updated to skip the executor or is the exemption supposed to stay?

akup:
> @akup line 89: with the !msg.Revalidation gone the executor gets blocked from voting on their own revalidation dispute.
VerifyInvalidation in the off-chain code (inference_validation.go:59-81) doesn't filter the executor out before sending, and voting.go:37 still sets the executor as proposer on the revalidation proposal.
> 
> should the off-chain side be updated to skip the executor or is the exemption supposed to stay?

Maybe it's simpler to keep !msg.Revalidation and keep ability to vote.

Anyway I think this PR should be closed, because now all focus is moving to devshards. Moreover here sead is revealed, but current protocol version doesn't protect from sead early revealing, and some party can cheat with inferenceIds

@0xMayoor let's accept your PR https://github.com/gonka-ai/gonka/pull/832

PR #969, opened 2026-03-28
fix(dapi): use signal.NotifyContext for graceful shutdown

Mayveskii:
## Problem

Root context uses context.WithCancel which SIGTERM does not cancel. Config flush never runs on docker stop.

## Fix

Replace with signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM) so SIGTERM/SIGINT propagate through context tree.

## Impact: config reliably flushed on shutdown, zero change on happy path

### DIFF
--- a/decentralized-api/main.go
+++ b/decentralized-api/main.go
@@ -40,6 +40,8 @@ import (
 	"log"
 	"log/slog"
 	"os"
+	"os/signal"
+	"syscall"
 	"strconv"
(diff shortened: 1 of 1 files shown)


akup:
Tiny useful fix. Please rebase

Mayveskii:
Rebased as requested. Ready for review. @akup

PR #970, opened 2026-03-28
fix(keeper): use regexp instead of fmt.Sscanf to parse ErrInsufficientFunds in ClaimRewards

Mayveskii:
## Problem

parseBalanceError uses fmt.Sscanf with hardcoded denom names nicoin/ngonka. If SDK changes error format, parse silently returns 0,0 making balance diagnostics useless.

## Fix

Use compiled regexp matching amounts without hardcoded denom names.

## Impact: robust parsing regardless of denomination, pre-compiled pattern has zero overhead


x0152:
Closing - this is no longer an issue, the ClaimRewards path was reworked as part of other changes

PR #971, opened 2026-03-28
fix: propagate InjectParamsIntoContext error in StartInference and FinishInference

unameisfine:
Same pattern as #968 (Validation handler) — applied to StartInference and FinishInference.

## Problem

When `InjectParamsIntoContext` fails (e.g. corrupted params store), both handlers log a warning but continue executing with a broken context. Downstream calls to `GetDeveloperAccessParams` and `GetTransferAgentAccessParams` then fail silently, effectively bypassing access gates.

## Fix

Return error immediately instead of continuing, matching the pattern from #968:

```go
ctx, err := k.Keeper.InjectParamsIntoContext(sdk.UnwrapSDKContext(goCtx))
if err != nil {
    k.LogWarn("...", types.Inferences, "error", err)
    return nil, fmt.Errorf("cannot start/finish inference with stale params: %w", err)
}
```

## Tests

`TestStartInference_ReturnsError_WhenInjectParamsFails` and `TestFinishInference_ReturnsError_WhenInjectParamsFails` — corrupt the params store with invalid protobuf, then call the handler:

- **Without fix**: FAIL — handler returns nil error, continues with broken params, access gates bypassed
- **With fix**: PASS — handler returns `"cannot start/finish inference with stale params"` immediately


akup:
Same as https://github.com/gonka-ai/gonka/pull/968

@unameisfine I think tests here is overkill and just add unneeded code. Please clear the tests from the PR

unameisfine:
Done — removed both test files in 11d5d96. Only the two msg_server fixes remain.

akup:
> Done — removed both test files in [11d5d96](https://github.com/gonka-ai/gonka/commit/11d5d9635fd0e6e5ec7549d16f59c860e0fd5d23). Only the two msg_server fixes remain.

And please rebase

unameisfine:
Rebased on main (ec8f455).

unameisfine:

PR #973, opened 2026-03-30
fix: skip proof query when CreatedAtBlockHeight is 0 for old epochs

### LINKED ISSUE #983: Bug: GET /api/v1/epochs/{N}/participants returns 500 for past epochs (CreatedAtBlockHeight=0)
## Bug

`GET /api/v1/epochs/{N}/participants` returns **500 Internal Server Error** for past epochs. Current epoch works fine.

## Repro

```
GET http://node1.gonka.ai:8000/api/v1/epochs/215/participants
→ 500 Internal Server Error: height must be greater than 0, but got 0
```

Epoch 215 consistently reproduces this. Any past epoch where `CreatedAtBlockHeight` was not yet populated will fail.

## Root Cause

In `queryActiveParticipants` (`get_participants_handler.go`):

1. First query (no height) fetches `activeParticipants`
2. `blockHeight := activeParticipants.CreatedAtBlockHeight` — for old epochs this is **0** (field was not populated at storage time)
3. Second call `QueryByKeyWithOptions(..., height=0, prove=true)` — CometBFT rejects `height=0` with the above error

## Fix

Check if `blockHeight == 0` before the second query. If so, skip the proof query and return the first result directly, with a `Warn` log for observability.

Fix is implemented in PR #973.

mingles-agent:
Closes #983

## Problem

`GET /api/v1/epochs/{N}/participants` returns 500 Internal Server Error for past epochs:

```
height must be greater than 0, but got 0
```

Repro: http://node1.gonka.ai:8000/api/v1/epochs/215/participants

Current epoch works fine. Past epochs (where data was stored before the `CreatedAtBlockHeight` field was introduced) fail.

## Root Cause

In `queryActiveParticipants` (`get_participants_handler.go`):

1. First query (no height) fetches `activeParticipants`
2. `blockHeight := activeParticipants.CreatedAtBlockHeight` — for old epochs this is **0** (field not populated at the time of storage)
3. Second call `QueryByKeyWithOptions(..., height=0, prove=true)` — CometBFT rejects height=0

## Fix

Check if `blockHeight == 0` before the second query. If so, skip the proof query and return the first result directly. A `Warn` log is emitted for observability.

## Testing

Verified against epoch 215 (previously returned 500) and epoch 216 (current, unaffected).

### DIFF
--- a/decentralized-api/internal/server/public/get_participants_handler.go
+++ b/decentralized-api/internal/server/public/get_participants_handler.go
@@ -281,6 +281,15 @@ func queryActiveParticipants(rpcClient *rpcclient.HTTP, cdc *codec.ProtoCodec, e
 	// 2. The implemented proof system has a bug anyway and needs to be revisited
 
 	blockHeight := activeParticipants.CreatedAtBlockHeight
+
+	// If CreatedAtBlockHeight is 0 (old epochs before this field was populated),
+	// skip the proof query — CometBFT rejects height=0 with an error.
+	// Return the first query result directly instead.
+	if blockHeight == 0 {
+		logging.Warn("CreatedAtBlockHeight is 0 for epoch, skipping proof query", types.Participants)
+		return result, nil
+	}
+
 	result, err = cosmos_client.QueryByKeyWithOptions(rpcClient, "inference", dataKey, blockHeight, true)
 	if err != nil {
 		logging.Error("Failed to query active participant. Req 2", types.Participants, "error", err)


Doog-bot534 (approved):
## Review: fix: skip proof query when CreatedAtBlockHeight is 0 for old epochs

**Approve** ✅

Surgically correct fix. CometBFT rejects ABCI queries at height=0, so short-circuiting when `CreatedAtBlockHeight == 0` prevents runtime errors for legacy epochs that predate the field. Returning the first query result without the proof query is the right fallback.

### Strengths
- Correctly handles protobuf uint64 zero-value default as sentinel for old epochs
- Converts a hard failure into a graceful skip with appropriate Warn-level log
- Minimal regression risk since this path was already broken for old epochs

### Suggestions
1. Include the epoch index in the log message for traceability: `"epoch_index", epochIndex`
2. Add a comment noting this guard can be removed once all `CreatedAtBlockHeight == 0` epochs have aged out

bonujel:
@mingles-agent heads-up before this sits any longer — I reproduced #983 on current `main` (`4fa6be0`) and the failing call turns out not to be the one this patch guards. Full trace and a pasteable repro test: https://github.com/gonka-ai/gonka/issues/983#issuecomment-5164389648

Short version:

- `ABCIQuery(height=0, prove=true)` does **not** error. The SDK fork resolves height 0 to latest (`baseapp/abci.go:1256`, `isLatest := height == 0`; the proof guard is only `height == 1 && prove`). In the repro that call returns normally.
- The 500 comes from `GetValidatorSetByHeight(Height: CreatedAtBlockHeight)` → CometBFT `Validators(&0)` → `rpc/core/env.go:177`, which is where `height must be greater than 0, but got 0` is produced. That call is fatal.
- Independently, the handler has moved: `decentralized-api/internal/server/public/get_participants_handler.go` is down to 84 lines on `main` and no longer contains `queryActiveParticipants` at all. The logic now lives in `common/queryapi/epoch.go`.

So this needs redoing against the new location rather than a rebase. Not a criticism of the analysis — the old code did have a fatal zero-height `Block()` call right after the one you guarded, and the move to `common/queryapi` happened to fix that one while keeping the validator-set call.

The bug itself is still live, so the work is still worth having.

bonujel:
Correcting one line in my comment above: I wrote "the bug itself is still live, so the work is still worth having". That was not supported by what I had actually shown — my repro forces `CreatedAtBlockHeight = 0` through a stub, which demonstrates the path fails if reached, not that it is reachable.

On the public nodes, every past epoch I probed returns 404 rather than reaching that code at all; only the current epoch resolves, and its `created_at_block_height` is populated. Details and the caveat (I could not establish *why* past epochs are unretrievable, so an archive node might still hit it): https://github.com/gonka-ai/gonka/issues/983#issuecomment-5173348816

What I said about this PR is unaffected — the guard is on the ABCI query, which is not the call that errors, and `queryActiveParticipants` no longer exists on `main`. But I should not have framed it as urgent work. Sorry for the noise.

PR #974, opened 2026-03-30
feat: intra-epoch circuit breaker with reputation-adjusted executor selection

### LINKED ISSUE #975: Nodes with high miss rate continue receiving inference requests for the rest of the epoch
## Problem

Nodes with consistently high miss rates (wrong answers, timeouts) remain in the executor pool for the entire epoch. There is no mid-epoch mechanism to stop routing client inference requests to them.

Observed in testnet: nodes with 25%+ miss rates remain active for hundreds of blocks, causing client-visible failures.

## Solution in PR #974

**Circuit breaker state machine** per node: `ACTIVE → EXCLUDED → PROBE → ACTIVE`
- Exclusion: miss rate > 25% after ≥4 samples (governance-adjustable via `ValidationParams`)
- Recovery: after cooldown, node gets one probe slot; success → back to ACTIVE
- All state writes in `EndBlock` — query handlers read-only (Cosmos-safe)

**Reputation-adjusted selection**: stake weight scaled by reputation score, so well-performing nodes are preferred before hitting the exclusion threshold.

## Known issue (addressed in follow-up)

Same-block probe re-exclusion: when a probe succeeds, `UpdateCBStateForBlock` Pass 2 in the same block could immediately re-exclude the node due to stale miss-rate stats. Fixed in a separate commit — will be included in the new PR.

mingles-agent:
Closes #975

## Problem

When a node in the network starts malfunctioning — returning wrong answers, timing out, or degrading — it continues to receive inference requests for the **rest of the epoch**. There is no mechanism to remove it from the executor pool mid-epoch. Clients receive bad inference results with no recourse until the next epoch boundary.

## Solution

This PR introduces two complementary features:

### 1. Reputation-adjusted executor selection weight

Before this change, executor selection used raw stake weight. This PR adds a `CalculateSelectionWeight` function in `epochgroup` that scales stake weight by the node's reputation score (0–100), with a 1% floor for new nodes:

```
adjusted_weight = stake_weight × reputation / 100   (min: stake_weight / 100)
```

Reputation is already tracked per epoch — this change makes it influence who gets selected, so consistently well-performing nodes are preferred over degraded ones.

### 2. Intra-epoch circuit breaker with cooldown + probe recovery

A per-node circuit breaker that dynamically excludes failing nodes from the executor pool within an epoch, and recovers them once the issue is resolved.

#### State machine

```
ACTIVE ──(miss rate > threshold)──▶ EXCLUDED
EXCLUDED ──(cooldown expires)──▶ PROBE
PROBE ──(inference succeeds)──▶ ACTIVE
PROBE ──(inference fails/times out)──▶ EXCLUDED
```

#### Exclusion trigger

- Default threshold: **25% miss rate** with a minimum of **4 samples**
- Signal: `CurrentEpochStats.MissedRequests / (MissedRequests + InferenceCount)`
- These stats are already tracked per participant — no new on-chain data required

#### Recovery (probe)

After the cooldown period expires, the node is given **one test slot** (probe). If the inference completes successfully → node returns to the active pool. If it times out → node re-enters exclusion with a fresh cooldown.

#### Write path (Cosmos-safe)

All state transitions are written in **EndBlock** via `UpdateCBStateForBlock`, which runs two passes each block:
- **Pass 1**: scan active participants, exclude those exceeding the miss rate threshold
- **Pass 2**: scan excluded participants, promote those whose cooldown has expired to probe

The query handler `GetRandomExecutor` / `createHealthFilterFn` is **read-only**: it reads CB state and filters the candidate list, but makes no writes. This follows the Cosmos SDK constraint that query contexts must not mutate state.

Success/failure signals flow through existing write contexts:
- `RecordCBResult(success=true)` — called from `MsgServer.FinishInference` ✅
- `RecordCBResult(success=false)` — called from `EndBlock` timeout handler ✅
- `ClearAllCBState` — called from `EndBlock` on epoch boundary ✅

#### Governance parameters

Circuit breaker tuning parameters are exposed in `ValidationParams` proto and can be adjusted via on-chain governance without a software upgrade:

| Parameter | Default | Description |
|---|---|---|
| `cb_miss_rate_threshold` | 25 | Miss rate % to trigger exclusion |
| `cb_min_inferences` | 4 | Minimum samples before CB can trigger |
| `cb_cooldown_blocks` | 50 | Blocks to wait before probe attempt |

## Files changed

| File | Change |
|---|---|
| `epochgroup/epoch_group.go` | Add `CalculateSelectionWeight` |
| `epochgroup/random.go` | Use reputation-adjusted weights in selection |
| `keeper/circuit_breaker.go` | CB state machine: Get/Set/Exclude/Promote/Record/Clear |
| `keeper/query_get_random_executor.go` | Read-only health filter integrated into executor selection |
| `keeper/msg_server_finish_inference.go` | Wire `RecordCBResult(success=true)` on completion |
| `module/module.go` | Wire `UpdateCBStateForBlock` into EndBlock |
| `types/keys.go` | CB state store key prefix |
| `types/params.proto` + `params.pb.go` | CB governance params in `ValidationParams` |

## Tests

- `keeper/circuit_breaker_test.go` — unit tests for state machine transitions
- `keeper/circuit_breaker_endblock_test.go` — EndBlock pass 1/2 logic
- `epochgroup/epoch_group_test.go` — reputation-weighted selection distribution

## Addresses feedback from #943

Previous PR was closed because `createHealthFilterFn` wrote state inside a query handler. This has been fixed:
- All writes moved to EndBlock (`UpdateCBStateForBlock`)
- Query handler is now purely read-only
- Removed dependency on SPRT for CB trigger (uses miss rate stats directly)
- Weight naming corrected to reflect actual selection weight, not raw stake

### DIFF
--- a/inference-chain/proto/inference/inference/params.proto
+++ b/inference-chain/proto/inference/inference/params.proto
@@ -104,6 +104,12 @@ message ValidationParams {
   Decimal quick_failure_threshold = 23; // The threshold of probabilty of consecutive failures to cause a quick invalidation
   Decimal binom_test_p0 = 24; // The null hypothesis probability for the binomial test (default 0.10)
   bool claim_validation_enabled = 25; // Enables claim-time validation checks in MsgClaimRewards
+
+  // Circuit breaker parameters (governance-adjustable without a chain upgrade)
+  uint64 cb_miss_threshold_pct = 26;       // miss rate % (0–100) above which a node is excluded (default 25)
+  uint64 cb_min_samples = 27;              // minimum inferences before threshold applies (default 4)
+  int64  cb_initial_cooldown_blocks = 28;  // initial exclusion cooldown in blocks (~5 min at 5s/block; default 50)
+  int64  cb_max_cooldown_blocks = 29;      // max cooldown with exponential backoff (~50 min; default 500)
 }
 
 message PoCModelParams {
--- a/inference-chain/x/inference/epochgroup/epoch_group.go
+++ b/inference-chain/x/inference/epochgroup/epoch_group.go
@@ -52,6 +52,32 @@ func NewEpochMemberFromActiveParticipant(p *types.ActiveParticipant, reputation 
 	}
 }
 
+// CalculateSelectionWeight scales staking weight by reputation (0–100).
+//
+// Floor: 1% of stakeWeight ensures no complete exclusion at this layer —
+// SPRT handles final exclusion. New nodes (reputation 0) get 1% traffic
+// during ramp-up via the floor.
+func CalculateSelectionWeight(stakeWeight int64, reputation int64) int64 {
+	if stakeWeight <= 0 {
+		return 0
+	}
+	if reputation >= 100 {
+		return stakeWeight
+	}
+	if reputation <= 0 {
+		reputation = 1 // new nodes get 1% traffic during ramp-up
+	}
+	floor := stakeWeight / 100
+	if floor < 1 {
+		floor = 1
+	}
+	adjusted := stakeWeight * reputation / 100
+	if adjusted < floor {
+		return floor
+	}
+	return adjusted
+}
+
 // calculatePocParticipatingNodesWeight calculates the total weight of nodes participating in PoC
 func calculatePocParticipatingNodesWeight(mlNodes []*types.ModelMLNodes) int64 {
 	totalWeight := int64(0)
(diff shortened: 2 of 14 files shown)


x0152:
Closing this for now: circuit breaker only affects who GetRandomExecutor suggests, but on-chain StartInference can still send work to an excluded node, and even a successful probe can get immediately undone in the same block

PR #981, opened 2026-03-30
[P1] GIP: Transaction fees for spam prevention

### LINKED ISSUE #928: [P1] Open Questions: Block Gas Limits, Fees, Cost per Participant, and System TX Prioritization

0xgonka:
## Summary

Introduces consensus-level transaction fees to Gonka via a governance-controlled `FeeParams` (default: 10 ngonka/gas). Fees are enforced at both `CheckTx` and `DeliverTx` via a custom `TxFeeChecker`, with a `NetworkDutyFeeBypassDecorator` that exempts protocol-obligation messages (PoC, validations, inference, BLS DKG).

## Key design decisions

- **Two-component PoC fee** for sybil resistance: `MsgPoCV2StoreCommit` consumes `base_validation_gas` once per participant per epoch plus `gas_per_poc_count × delta`. Makes claiming high weight from fake participants economically prohibitive while keeping per-commit costs fair for legitimate hosts submitting many incremental updates.
- **Fee-exempt message set** audited for sybil resistance. Each exempt message has a non-fee rate limiter (timing window, duplicate check, epoch-scoping, or slot ownership). `MsgRequestThresholdSignature` intentionally NOT exempt.
- **Recursive `MsgExec` unpacking** in the bypass decorator prevents wrapping fee-required messages inside `x/authz` to avoid fees. Fails closed on any non-exempt inner message.

## Host onboarding story (fully automatic)

The v0.2.12 upgrade handler handles the entire transition for existing hosts:
1. Sets `FeeParams` to the chain default
2. **Auto-creates `x/feegrant` allowances** (100 GNK, 1 year) for every existing cold→warm ML ops authz pair — the DAPI's warm key can now pay fees from the cold account's balance without any host action
3. Migrates BLS params, clears training state, removes top_miner, adjusts legacy params

The DAPI self-configures at startup: it queries the chain for `FeeParams.MinGasPriceNgonka` and uses that value. No `config.env` changes required for existing hosts. Cosmovisor swaps the binary, DAPI restarts, everything Just Works.

For new hosts, the existing `grant-ml-ops-permissions` CLI command now issues both the authz grants and the `MsgGrantAllowance` in a single transaction. The command is idempotent — it skips the feegrant message if an allowance already exists (e.g., post-migration re-run).

[HOST_ONBOARDING.md](HOST_ONBOARDING.md) documents the full flow, upgrade procedure, troubleshooting, and a quick-reference card.

## What hosts need to do

- **Existing hosts**: ensure cold account has ≥100 GNK to cover the allowance spend limit. That's it. Everything else is automatic.
- **New hosts**: follow the existing quickstart; the standard `grant-ml-ops-permissions` step now sets up both authz and feegrant.

## Related proposal

Full design rationale in [proposals/transaction-fees/README.md](proposals/transaction-fees/README.md).

## Test plan

- [x] Unit tests for `NetworkDutyFeeBypassDecorator`, `GonkaFeeChecker`, `MsgExec` recursive unpacking, gas cap enforcement, priority preservation, and all exempt message types (individual `t.Run` subtests)
- [x] Unit tests for two-component PoC fee (first commit vs delta)
- [x] Integration tests verify pre-upgrade → governance enable → post-upgrade CLI rejection + acceptance paths
- [x] Integration test suite passes: 44-45/45 (matching `upgrade-v0.2.12` base branch pass rate; remaining 0-1 failures are pre-existing flaky tests unrelated to fees)
- [x] Manual QA on testnet: confirm auto-migration creates feegrant allowances for all existing pairs, DAPI auto-configures gas price, reward claims and hardware diffs succeed post-upgrade without host action


inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,218 @@
...
+
+// GonkaFeeChecker returns a TxFeeChecker that enforces a consensus-level minimum
+// gas price read from on-chain FeeParams. It respects the bypass flag set by
+// NetworkDutyFeeBypassDecorator. This checker runs inside DeductFeeDecorator
+// during both CheckTx and DeliverTx.
+func GonkaFeeChecker(inferenceKeeper *inferencemodulekeeper.Keeper) ante.TxFeeChecker {
+	return func(ctx sdk.Context, tx sdk.Tx) (sdk.Coins, int64, error) {
+		// If bypass flag is set, allow zero fees.
+		if IsNetworkDutyBypassed(ctx) {
+			return sdk.Coins{}, 0, nil
```
patimen: Is this intentionally setting priority to 0? Is that the correct decision? It will put these messages to the back of the priority queue in the mempool.

patimen: Yeah, your other notes indicate this is wrong. You should be returning ctx.Priority

inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,218 @@
...
+	const maxInt64 = int64(^uint64(0) >> 1)
+	divisor := maxInt64
+	if gas <= uint64(maxInt64) {
+		divisor = int64(gas)
+	}
+
+	var priority int64
+	for _, coin := range feeCoins {
+		gasPrice := coin.Amount.QuoRaw(divisor)
+		amt := gasPrice.Int64()
```
patimen, line 233: coin.Amount could in fact be greater than Int64 max, in fact massively greater, it's a BigInt. We need to check for overflow here as well.

inference-chain/x/inference/keeper/fee_params.go
```diff
@@ -0,0 +1,37 @@
...
+	fp, err := types.UnmarshalFeeParams(bz)
+	if err != nil {
+		k.LogError("Unable to unmarshal FeeParams, using defaults", types.System, "error", err)
+		return types.DefaultFeeParams()
+	}
+	return fp
+}
+
+// SetFeeParams stores the fee parameters in the KV store.
+func (k Keeper) SetFeeParams(ctx context.Context, fp types.FeeParams) error {
```
patimen: There is no call to this value anywhere outside of genesis or upgrade... that means it cannot be set by governance.
Is there a reason we are not putting this into the main params object, so we have a way of setting it via governance?

inference-chain/x/inference/keeper/fee_params.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+
+	"github.com/cosmos/cosmos-sdk/runtime"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// GetFeeParams returns the fee parameters from the KV store.
+// Returns defaults if not yet set.
+func (k Keeper) GetFeeParams(ctx context.Context) types.FeeParams {
+	store := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
```
patimen: If we DO decide to have this separate from the main Params, this should be done using a Cosmos SDK collection, not this manual method here.

0xgonka:
@patimen 

**1. Priority 0 on bypass** — `GonkaFeeChecker` was returning priority 0 for fee-exempt duty messages, pushing them to the back of the mempool queue. Fixed to return `ctx.Priority()` so the bypass decorator's priority (500k) is preserved. Added test.

**2. `coin.Amount.Int64()` overflow** — `getTxPriority` was calling `.Int64()` on a BigInt that could exceed int64 max. Added overflow clamp to `maxInt64` before the conversion.

**3. Not governance-settable** — `SetFeeParams` was only called from genesis/upgrade, meaning governance couldn't adjust fee params without a chain upgrade. Added `UpdateFeeParams()` on `msgServer` that's governance-gated via `CheckPermission`. Fee params can now be tuned via governance proposal.

**4. Manual KV store → Cosmos collections** — Replaced raw `store.Get/Set` with `collections.Item[FeeParams]`, matching the pattern used by every other keeper collection. Implemented a custom `ValueCodec` (JSON-based) since FeeParams isn't a proto type. Key format (`[]byte("p_fee_params")`) and value format (JSON) are identical to the old approach, so no migration concerns — and since both approaches were introduced in this PR, there's no existing on-chain data in either format.

inference-chain/x/inference/types/fee_params.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// GasPerPoCCount is additional gas consumed per unit of Count in MsgPoCV2StoreCommit.
+	// Charged on the delta (count increase) so total equals final_count * GasPerPoCCount.
+	GasPerPoCCount uint64 `json:"gas_per_poc_count"`
+}
+
+// DefaultFeeParams returns the default fee parameters.
+func DefaultFeeParams() FeeParams {
+	return FeeParams{
+		MinGasPriceNgonka: 10,
```
patimen: 10 ngonka is actually about 0.0000000056$

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -51,6 +52,15 @@ const (
 	idHeader   = "TX_ID"
 
 	maxBlockTimeDrift = 120 * time.Second
+
+	// MinGasPriceNgonka is the minimum gas price in ngonka, matching the on-chain
+	// FeeParams.MinGasPriceNgonka default. Used for both the single-tx path
+	// (cosmosclient.go WithGasPrices) and the batch-tx path (getSignedBytes).
+	MinGasPriceNgonka = 10
```
patimen: We should pull these from the on-chain values rather than hard coding.

deploy/join/docker-compose.mlnode.yml
```diff
@@ -1,7 +1,12 @@
 services:
   mlnode-308:
+<<<<<<< HEAD
```
patimen: Well, that's not what you want.

inference-chain/app/upgrades/v0_2_12/upgrades.go
```diff
@@ -100,3 +105,16 @@ func removeTopMiner(ctx context.Context, k keeper.Keeper) error {
 func clearTrainingState(ctx context.Context, k keeper.Keeper) error {
 	return k.ClearTrainingState(ctx)
 }
+
+func setFeeParams(ctx context.Context, k keeper.Keeper) error {
```
patimen, line 256: We don't have a separate "GetXParams" for any other type of params. 
Let's get rid of the extra layer and use GetParams() like all other sub-params

inference-chain/app/ante_fee_test.go
```diff
@@ -0,0 +1,328 @@
...
+		&inferencetypes.MsgInvalidateInference{},
+		&inferencetypes.MsgRevalidateInference{},
+		&inferencetypes.MsgMLNodeWeightDistribution{},
+		&blstypes.MsgSubmitDealerPart{},
+		&blstypes.MsgSubmitVerificationVector{},
+		&blstypes.MsgSubmitPartialSignature{},
+		&blstypes.MsgRequestThresholdSignature{},
+	}
+
+	for _, msg := range exemptMsgs {
```
patimen: Rather than a combined result for all exemptMsg, we should use t.Run to create (effectively) a test for each message.

deploy/join/docker-compose.mlnode.yml
0xgonka: updated, will see if it passes

patimen (commented):
Couple more things to fix.

0xgonka:
integration passing: https://github.com/gonka-ai/gonka/actions/runs/24288055626

inference-chain/app/upgrades/v0_2_12/upgrades.go
```diff
@@ -59,6 +93,78 @@ func CreateUpgradeHandler(
...
+) error {
+	type grantPair struct {
+		granter sdk.AccAddress
+		grantee sdk.AccAddress
+	}
+	seen := make(map[string]bool)
+	var pairs []grantPair
+
+	startInferenceMsgType := sdk.MsgTypeURL(&types.MsgStartInference{})
+	authzKeeper.IterateGrants(ctx, func(granterAddr, granteeAddr sdk.AccAddress, grant authz.Grant) bool {
```
patimen, line 133: Do we have an idea of how large this is on mainnet? We have to be careful not to take too long here.

inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,227 @@
...
+		if minGasPriceNgonka == 0 {
+			priority := getTxPriority(feeCoins, gas)
+			return feeCoins, priority, nil
+		}
+
+		// Calculate required fee using big-int math to avoid uint64 overflow.
+		requiredAmount := math.NewIntFromUint64(gas).Mul(math.NewIntFromUint64(minGasPriceNgonka))
+		requiredFee := sdk.NewCoin("ngonka", requiredAmount)
+
+		if !feeCoins.IsAnyGTE(sdk.NewCoins(requiredFee)) {
```
patimen: IsAnyGTE checks only the AMOUNT of the fee, not the denomination.
Wouldn't `feeCoins.AmountOf("ngonka").LT(requiredFee.Amount)` be safer?

inference-chain/x/inference/keeper/msg_server_poc_v2_commit.go
```diff
@@ -89,6 +98,29 @@ func (k msgServer) PoCV2StoreCommit(goCtx context.Context, msg *types.MsgPoCV2St
 		}
 	}
 
+	// Consume extra gas for sybil resistance (two-component fee).
+	// Base validation gas: charged once per participant per epoch (covers GPU validation cost).
+	// Count gas: charged on delta (so total equals final_count * gas_per_poc_count).
+	params, paramsErr := k.Keeper.GetParams(ctx)
```
patimen, line 104: As several agents point out, this is potentially expensive... 
But no ad-hoc solution makes sense. We need a cache for GetParams implemented reliably and globally, so let's use this for now.

testermint/src/main/kotlin/ApplicationCLI.kt
```diff
@@ -644,6 +644,29 @@ data class ApplicationCLI(
...
     }
 
+    fun sendTransactionWithFees(args: List<String>, fees: String, useColdAccount: Boolean = true): TxResponse {
+        val from = if (useColdAccount) this.getColdAccountName() else this.getWarmAccountName()
+        Logger.info("Sending transaction with fees=$fees")
+        val finalArgs = listOf("tx") + args + getTransactionArgsWithFees(from, fees)
+        return execAndParse(finalArgs, stdIn = passwordInjection)
+    }
+
+    private fun getTransactionArgsWithFees(from: String, fees: String) = listOf(
```
patimen: It'd be better to make this a base list that get's modified for Fees rather than duplicated lists (with `getTransactionArgs`.

inference-chain/app/ante_fee.go
```diff
@@ -0,0 +1,227 @@
...
+	if execMsg, ok := msg.(*authztypes.MsgExec); ok {
+		if ik == nil {
+			return false // fail closed
+		}
+		for _, innerMsg := range execMsg.Msgs {
+			var unwrapped sdk.Msg
+			if err := ik.Codec().UnpackAny(innerMsg, &unwrapped); err != nil {
+				return false // fail closed on unpack error
+			}
+			if !isNetworkDutyRecursive(unwrapped, ik) {
```
patimen: In what real world scenario do we have a recursive MsgExec? I would feel better if we simply checked one deep (and called isExemptMessage here instead)

decentralized-api/cosmosclient/cosmosclient.go
```diff
@@ -103,13 +103,33 @@ func updateKeyringIfNeeded(client *cosmosclient.Client, keyringDir string, confi
 	return nil
 }
 
+// queryChainMinGasPrice queries the inference module params for
+// FeeParams.MinGasPriceNgonka. Returns 0 if the chain has no FeeParams set
+// (pre-upgrade chain) or if the query fails. Callers should fall back to
+// a safe default (0 or the config value) on failure.
```
patimen: Is returning 0 the best default? Maybe we should return our intial value or some other fallback? 
Either that or we panic/return error. Because getting this wrong will amount to a fatal flaw for the client.

inference-chain/app/ante.go
```diff
@@ -207,7 +209,12 @@ func NewAnteHandler(options HandlerOptions) (sdk.AnteHandler, error) {
 			GasCap:          500000,    // safe cap for swap path; tune after measuring simulate
 			Priority:        1_000_000, // optional: ensure zero-fee txs aren't starved
 		},
-		ante.NewDeductFeeDecorator(options.AccountKeeper, options.BankKeeper, options.FeegrantKeeper, options.TxFeeChecker),
+		NetworkDutyFeeBypassDecorator{
+			InferenceKeeper: options.InferenceKeeper,
+			GasCap:          10_000_000, // cap for fee-exempt duty txs (must cover batch transactions)
```
patimen: 10 million is about 0.00001 GNK... is that enough?

patimen (approved):

0xgonka:
Addressed all 7 review comments:

| # | Comment | Fix |
|---|---|---|
| 1 | `IsAnyGTE` checks amount only, not denom | Replaced with `feeCoins.AmountOf("ngonka")` direct comparison so fees in other denoms can't satisfy an ngonka requirement |
| 2 | Recursive `MsgExec` unwrapping is overkill | Simplified to one level deep; nested `MsgExec` fails closed. Renamed `isNetworkDutyRecursive` → `isNetworkDuty` |
| 3 | `queryChainMinGasPrice` silently returns 0 on error | Now returns `(int64, error)`; DAPI startup aborts with an actionable error message on query failure. Pre-upgrade chains (nil `FeeParams`) still correctly return `(0, nil)` |
| 4 | 10M bypass gas cap may not be enough | Bumped to 20M with explanatory comment (2x `BatchGasLimit` headroom for future batch-size growth) |
| 5 | Duplicated `getTransactionArgs` / `getTransactionArgsWithFees` | Refactored `getTransactionArgsWithFees` to derive from `getTransactionArgs` by mutating the base list (swap `--gas-adjustment` out, patch `--gas` value, append `--fees`) |
| 6 | PoC commit `GetParams` is expensive | Acknowledged as TODO pending a global `GetParams` cache — no code change for now per your comment |
| 7 | How large is the feegrant migration on mainnet? | Added detailed scale/cost comment: ~100 hosts × ~20 grants per host = ~2k authz iterations + ~100 BasicAllowance writes. Well under a second even at 10x scale. |

All three modules build clean.

PR #984, opened 2026-03-31
feat: intra-epoch circuit breaker with reputation-adjusted executor selection

### LINKED ISSUE #975: Nodes with high miss rate continue receiving inference requests for the rest of the epoch

mingles-agent:
Closes #975

## Summary

Two complementary features to prevent malfunctioning nodes from continuing to receive client inference requests for the rest of an epoch.

### 1. Reputation-adjusted executor selection

Stake weight is scaled by reputation score at epoch start, so well-performing nodes are preferred before hitting the exclusion threshold. Nodes accumulate reputation from successful inferences and lose it from misses.

### 2. Intra-epoch circuit breaker

Per-node state machine: `ACTIVE → EXCLUDED → PROBE → ACTIVE`

- **Exclusion trigger**: miss rate > 25% after ≥4 samples (configurable via `ValidationParams`)
- **Recovery**: after cooldown period, node gets one probe slot; success restores it to ACTIVE, failure doubles the cooldown
- **Cosmos-safe**: all state writes in `EndBlock` only — query handlers are read-only

### 3. Probe re-exclusion fix

When a probe succeeds, `UpdateCBStateForBlock` Pass 2 running in the same block could immediately re-exclude the node using stale miss-rate stats. Fixed by adding `LastRestoredBlock` to `CBEntry`: Pass 2 skips nodes where `LastRestoredBlock == blockHeight` (one-block grace period).

### 4. Circuit breaker params in ValidationParams

CB thresholds (`MissRateThreshold`, `MinSamples`, `CooldownBlocks`, `ProbeInterval`) promoted to `ValidationParams` proto for on-chain governance adjustability.

## Files changed

- `inference-chain/x/inference/keeper/circuit_breaker.go` — CB state machine, UpdateCBStateForBlock, GetRandomExecutor
- `inference-chain/x/inference/keeper/circuit_breaker_endblock_test.go` — comprehensive tests
- `inference-chain/x/inference/types/` — CBEntry with LastRestoredBlock, ValidationParams with CB fields
- `inference-chain/x/inference/keeper/params.go` — CB param accessors

### DIFF
--- a/decentralized-api/internal/server/public/get_participants_handler.go
+++ b/decentralized-api/internal/server/public/get_participants_handler.go
@@ -281,6 +281,15 @@ func queryActiveParticipants(rpcClient *rpcclient.HTTP, cdc *codec.ProtoCodec, e
 	// 2. The implemented proof system has a bug anyway and needs to be revisited
 
 	blockHeight := activeParticipants.CreatedAtBlockHeight
+
+	// If CreatedAtBlockHeight is 0 (old epochs before this field was populated),
+	// skip the proof query — CometBFT rejects height=0 with an error.
+	// Return the first query result directly instead.
+	if blockHeight == 0 {
+		logging.Warn("CreatedAtBlockHeight is 0 for epoch, skipping proof query", types.Participants)
+		return result, nil
+	}
+
 	result, err = cosmos_client.QueryByKeyWithOptions(rpcClient, "inference", dataKey, blockHeight, true)
 	if err != nil {
 		logging.Error("Failed to query active participant. Req 2", types.Participants, "error", err)
--- a/docs/research/health-aware-executor-selection.md
+++ b/docs/research/health-aware-executor-selection.md
@@ -1,0 +1,96 @@
+# Research: Node Health-Aware Executor Selection
+
+**Issue:** #3  
+**Date:** 2026-03-24  
+**Status:** Complete — implementation tasks created (#4, #5, #6)
+
+---
+
+## Summary
+
+When a validator goes offline or degrades, it continues receiving inference requests via `GetRandomExecutor` (pure staking-weight random selection). This causes missed inferences (3.25% miss rate, 81k misses in epochs 161–191), validator penalties, and slow responses.
+
+## Key Findings
+
+### Miss Data Flow
+
+```
+Inference expires (EndBlock, module.go:319)
+  → executor.CurrentEpochStats.MissedRequests++
+  → UpdateParticipantStatus() → SPRT test (getInactiveStatus)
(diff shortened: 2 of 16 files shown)


x0152:
Core blocker from #974 is still present here. This pr is based on a wrong premise

Doog-bot534 (commented):
## Review: feat: intra-epoch circuit breaker with reputation-adjusted executor selection

**Approve with suggestions** ✅

Substantial, well-architected feature. The reputation-weighted selection (1% floor prevents starvation) and CB state machine (EXCLUDED → PROBE → HEALTHY) are sound. Excellent test coverage including probabilistic traffic distribution tests.

### Key strengths
- Same-block grace period (`ProbeRestored + LastRestoredBlock`) prevents immediate re-exclusion after probe
- Exponential backoff cooldowns prevent CB flapping
- 1% weight floor ensures zero-reputation nodes still get occasional traffic for recovery

### Potential issues

1. **JSON for CB state persistence:** `json.Marshal/Unmarshal` is deterministic for simple structs, but fragile compared to protobuf. If field ordering or float serialization diverges across Go versions, this could cause consensus divergence. **Recommend migrating to protobuf.**

2. **`ClearAllCBState` at epoch boundary:** Referenced in comments but not visible in this diff. Verify it's wired into the epoch transition handler.

3. **`selectRandomParticipant` fallback:** If `selectionWeights` is missing an address, it falls back to raw group weight, potentially creating inconsistency if `buildSelectionWeightsMap` silently skips a nil `ValidationWeight`.

4. **Governance param validation:** `cb_miss_threshold_pct` should be validated to stay in [1, 100].

x0152:
Closing for now - the core blocker from #974 is still present here

PR #986, opened 2026-03-31
[P0] Make handling of warm keys deterministic 

### LINKED ISSUE #955: [P0] Make handling of warm keys deterministic (implementation)
See #913 

Once research is finished and we agree on a decision, we'll update this issue with the implementation steps.

Brgndy25:
Implements #955


subnet/host/finality.go
```diff
@@ -0,0 +1,61 @@
...
+			}
+		}
+	}
+
+	threshold := twoThirdsWeight(group)
+	for f := latestNonce; f > 0; f-- {
+		if bitmapSlotWeight(confirmedBy[f], group) >= threshold {
+			return f
+		}
+	}
```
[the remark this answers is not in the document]
heitor-lassarote: Is `bitmapSlotWeight` still needed in this approach? Could we instead merge the logic of that function with the `for slotId := range sigs` loop to get some extra optimization?

subnet/gossip/interface.go
```diff
@@ -22,12 +22,12 @@ type GossipSig struct {
...
 // DiffFetcher retrieves diffs from a peer (backed by HTTPClient).
 type DiffFetcher interface {
-	GetDiffs(ctx context.Context, fromNonce, toNonce uint64) ([]types.Diff, error)
+	GetDiffs(ctx context.Context, fromNonce, toNonce uint64) ([]types.DiffRecord, error)
 }
 
 // StateUpdater applies recovered diffs and signs them (backed by Host).
 type StateUpdater interface {
-	ApplyRecoveredDiffs(ctx context.Context, diffs []types.Diff) ([]GossipSig, error)
+	ApplyRecoveredDiffs(ctx context.Context, diffs []types.DiffRecord) ([]GossipSig, error)
```
heitor-lassarote, line 30: Perhaps a documentation change to `DiffRecord` is in order, because right now it suggests it's only used for storage.

subnet/transport/server.go
```diff
@@ -646,8 +646,9 @@ func (s *Server) HandleGetDiffs(c echo.Context) error {
 
 	// Convert to JSON-friendly format.
 	type diffRecordJSON struct {
-		DiffJSON  `json:"diff"`
-		StateHash []byte `json:"state_hash"`
+		DiffJSON     `json:"diff"`
```
heitor-lassarote, line 649: Nitpick: align the `json:"diff"`

subnet/gossip/interface.go
heitor-lassarote: Or, possibly, should we create a new struct instead? I see an incomplete diff (no `Signatures`, no `CreatedAt`) is sent. So perhaps a full `DiffRecord` is not the correct data structure (as it was previously for storage).

subnet/host/finality.go
```diff
@@ -0,0 +1,61 @@
...
+	return total
+}
+
+// twoThirdsWeight returns ceil(2/3 * totalSlots).
+func twoThirdsWeight(group []types.SlotAssignment) uint32 {
+	total := uint32(len(group))
+	if total == 0 {
+		return 0
+	}
```
heitor-lassarote: Nitpick: I think we could remove this. If it's zero, we'll end up just computing (0*2 + 2) / 3 which is still 0. But likely Copilot's suggestion is better.

subnet/host/host.go
```diff
@@ -821,14 +821,34 @@ func (h *Host) AccumulateGossipSig(nonce uint64, stateHash, sig []byte, senderSl
 }
 
 // ApplyRecoveredDiffs applies diffs fetched during gossip recovery.
+// For each nonce, if F is the BFT-finalized nonce from local signatures
+// (see computeFinalizedNonce), nonces at or below F may trust WarmKeyDelta from
+// the wire without mainnet resolution; nonces above F fall back to ResolveWarmKey.
 // Returns GossipSig for each successfully applied nonce.
 func (h *Host) ApplyRecoveredDiffs(ctx context.Context, diffs []types.DiffRecord) ([]gossip.GossipSig, error) {
+	_ = ctx // interface hook; recovery is synchronous under h.mu today
```
heitor-lassarote, line 829: Not sure I understand, is this so Go doesn't complain about unused `ctx` or something?

subnet/host/finality.go
```diff
@@ -0,0 +1,61 @@
...
+		sigs, err := store.GetSignatures(escrowID, n)
+		if err != nil {
+			continue
+		}
+		for slotID := range sigs {
+			// This slot signed at n, confirming all nonces 1..n.
+			for prev := uint64(1); prev <= n; prev++ {
+				bm := confirmedBy[prev] // zero value if absent
+				bm.Set(slotID)
+				confirmedBy[prev] = bm
```
heitor-lassarote: Is the bitset mutated? Do we need to reassign it? Otherwise, perhaps just `confirmedBy[prev].Set(slotId)` is enough.

subnet/gossip/interface.go
Brgndy25: We use `DiffRecord` as the container for what the diff API returns (`Diff`, `StateHash`, `WarmKeyDelta`). `Signatures` / `CreatedAt` are storage-only, so they stay empty on fetch, same as `HTTPClient.GetDiffs`. Recovery only needs the wire fields.

subnet/host/finality.go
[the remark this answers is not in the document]
Brgndy25: bitmapSlotWeight tells us “how many group slots are set in this bitmap?” after the fact. The sigs loop only sets bits, it doesn’t compute that sum.

subnet/transport/server.go
Brgndy25: alligned, marking as resolved

subnet/host/finality.go
Brgndy25: [Removed](https://github.com/gonka-ai/gonka/pull/986/changes/a334443a7c5f5613e9a5b7ec56e3b3d616c6fb42)

subnet/host/host.go
Brgndy25: yes, leaving it open for to match the interface / future impls.

subnet/host/finality.go
Brgndy25: Set mutates the bitmap, but it has a pointer receiver, so you can’t call it on confirmedBy[prev], map values aren’t addressable. You have to copy out, Set on the local bm, then assign back so the map keeps the update.

subnet/gossip/interface.go
heitor-lassarote: I have no strong opinion about reusing it or not. I am aware that these fields will be empty, my comment is mostly that the domain for this datatype is storage and now we are giving it double duty. But I don't want to bother too much about it as it's a minor thing, so resolving this comment.

subnet/host/finality.go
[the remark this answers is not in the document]
heitor-lassarote: Ok, keeping things separate is better for maintenance too, so let's keep it as it is. Resolving this comment.

heitor-lassarote (approved):
LGTM

heitor-lassarote (approved):

subnet/host/finality.go
```diff
@@ -0,0 +1,57 @@
...
+func computeFinalizedNonce(store storage.Storage, escrowID string, latestNonce uint64, group []types.SlotAssignment) uint64 {
+	// confirmedBy[n] = bitmap of slots that have signed at nonce >= n.
+	// Bitmap128 is a value type (16 bytes); max group size is 128.
+	confirmedBy := make(map[uint64]types.Bitmap128)
+
+	for n := uint64(1); n <= latestNonce; n++ {
+		sigs, err := store.GetSignatures(escrowID, n)
+		if err != nil {
+			continue
+		}
```
akup: We should add log at error level here if `err != nil`

subnet/host/finality.go
```diff
@@ -0,0 +1,57 @@
...
+		for slotID := range sigs {
+			// This slot signed at n, confirming all nonces 1..n.
+			for prev := uint64(1); prev <= n; prev++ {
+				bm := confirmedBy[prev] // zero value if absent
+				bm.Set(slotID)
+				confirmedBy[prev] = bm
+			}
+		}
+	}
```
akup: Here is O(N^2) cycle when O(N) is intended to be.
We just don't need inner cycle 

```go
for prev := uint64(1); prev <= n; prev++
```

It could be:

```go
confirmedBy := make(map[uint64]types.Bitmap128)
// running = slots that have signed at any nonce >= current n.
var running types.Bitmap128
for n := latestNonce; n > 0; n-- {
	sigs, err := store.GetSignatures(escrowID, n)
	if err != nil {
		// keep previous running set; just skip adding new signers for this nonce
		confirmedBy[n] = running
		continue
	}
	// A slot that signed at n confirms all nonces <= n.
	// In reverse traversal, that means it belongs to running for n and all lower nonces.
	for slotID := range sigs {
		running.Set(slotID)
	}
	confirmedBy[n] = running
}
```

O(N) cycle

akup: Moreover we don't need next cycle, we can here add 

```go
...
confirmedBy[n] = running
if bitmapSlotWeight(confirmedBy[f], group) >= threshold {
  return n
}
```

subnet/host/host.go
```diff
@@ -821,16 +821,36 @@ func (h *Host) AccumulateGossipSig(nonce uint64, stateHash, sig []byte, senderSl
...
+	var latestNonce uint64
+	for _, rec := range diffs {
+		if rec.Nonce > latestNonce {
+			latestNonce = rec.Nonce
+		}
+	}
+
+	var finalized uint64
+	if h.store != nil {
+		finalized = computeFinalizedNonce(h.store, h.escrowID, latestNonce, h.group)
```
akup, line 843: We are calling here `computeFinalizedNonce` that is using stored values to calculate `finalized`

```go
sigs, err := store.GetSignatures(escrowID, n)
```

but it seams store is empty as peer signatures written after catch-up gossip exchange

Am I right or missed something? where do we fill the store?

subnet/host/host.go
```diff
@@ -821,16 +821,36 @@ func (h *Host) AccumulateGossipSig(nonce uint64, stateHash, sig []byte, senderSl
...
-	for _, diff := range diffs {
-		if err := h.applyAndPersist(diff); err != nil {
-			return sigs, fmt.Errorf("apply recovered diff nonce %d: %w", diff.Nonce, err)
+	for _, rec := range diffs {
+		if rec.Nonce <= finalized && len(rec.WarmKeyDelta) > 0 {
+			h.sm.InjectWarmKeys(rec.WarmKeyDelta)
+		}
+		if err := h.applyAndPersist(rec.Diff); err != nil {
+			return sigs, fmt.Errorf("apply recovered diff nonce %d: %w", rec.Nonce, err)
 		}
```
akup, line 864: Here is the possible attack from peer:
peer provides wrong warm key diffs and they are stored at `h.sm.InjectWarmKeys(rec.WarmKeyDelta)`

Then `h.applyAndPersist` fails

If attacker provides change in warm key where there was no change, we will not be able to catch up and recover at all even on retry as warm key is already broken.

We should make a transaction here and rollback if `h.applyAndPersist` fails

subnet/host/host.go
Brgndy25: `computeFinalizedNonce` runs before we `applyAndPersist` the recovered diffs. Peer sigs only go into storage via `AccumulateGossipSig`, which requires a stored diff for that nonce first, so for nonces in the recovery batch, peer sigs usually aren’t there yet at that line. Recovery also gossips our sigs after apply, so others’ sigs for that gap often arrive later. So finalized is a pre-recovery snapshot; that’s conservative for trusting `WarmKeyDelta` during catch-up.

subnet/host/finality.go
Brgndy25: [Fixed](https://github.com/gonka-ai/gonka/pull/986/changes/1b54432dd9886d03d0cb50de6da8e27a5ffed1ab)

subnet/host/finality.go
Brgndy25: [Fixed](https://github.com/gonka-ai/gonka/pull/986/changes/7b60578168aef60a000303f8264e404dfb79903a)

subnet/host/host.go
Brgndy25: [Fixed](https://github.com/gonka-ai/gonka/pull/986/changes/fbf49225d88b812f887bc93bdac8d5f2d9f1434e)

subnet/host/host.go
akup: I mean that at the next lines we inject warm keys only for finalized nonce:

```go
if rec.Nonce <= finalized && len(rec.WarmKeyDelta) > 0 {
	h.sm.InjectWarmKeys(rec.WarmKeyDelta)
}
```

So we have following situation:
1. We are trying to catch up
2. We got the diff and empty storage
3. We are trying to `computeFinalizedNonce`, but because of empty storage it is 0
4. We do not `InjectWarmKeys` as condition `rec.Nonce <= finalized` is never met
5. We fail on `applyAndPersist` if there was warm key changes


I think we need somehow get signatures at `computeFinalizedNonce`, maybe not from store ( `store.GetSignatures(escrowID, n)` ) but from diffs

tcharchian: @0xMayoor would you mind to help us here with the comment above?

0xMayoor: reproduced it and verified against r2/v0.2.14/main. @akup  is right, and the recovery tests only pass because recoverySigStore pre-seeds the sigs, so it's hidden in CI:
https://github.com/gonka-ai/gonka/blob/fbf49225d88b812f887bc93bdac8d5f2d9f1434e/subnet/host/host_recovery_test.go#L18-L24

agree on pulling the sigs from the diffs, and the inject-then-check-root shape is already there — user/recover.go does it for self-recovery:
https://github.com/gonka-ai/gonka/blob/fbf49225d88b812f887bc93bdac8d5f2d9f1434e/subnet/user/recover.go#L68-L81
just have to verify them (recover → group slot → signed root matches) and only trust the delta on 2/3 verified sigs, so a peer can't forge the quorum.


@Brgndy25 you still active?

tcharchian: @0xMayoor, would you mind implementing these changes?

tcharchian: @0xMayoor is willing to help, but we need to explain the task clearly: what exactly should be changed, whether it should be done on top of the current branch or adapted to the devshards codebase, and what the expected scope is.

a-kuprin: @0xMayoor we should implement it over fresh codebase: https://github.com/gonka-ai/gonka/pull/1366

We just need to finish this task but accurately and handling situation I've described in this thread.

we need somehow get signatures at computeFinalizedNonce, maybe not from store ( store.GetSignatures(escrowID, n) ) but from diffs

a-kuprin: @0xMayoor actually we can drop this branch and you are free to create a new one with same task but with fresh codebase

0xMayoor: @a-kuprin  looking at the recovery path on #1366, peer fetch comes back as bare `[]types.Diff` (`HandleGetDiffs` only ships `{diff, state_hash}`), so no sigs and no `WarmKeyDelta` in there. 
sigs are easy enough off `/signatures`, but the delta's the awkward bit. replay can't regenerate it (it's the resolver's seal-time output) and nothing serves it standalone, so it kind of has to ride the diff fetch right? 
or did you have the delta coming from somewhere else?

and on scope — gossip recovery isn't actually wired into devshardd yet, prod's on the store replay in `recoverSession`. is wiring it up part of this, or just get `ApplyRecoveredDiffs` right for when it's on?

a-kuprin: Actually I didn't dived in the task.

I was just reviewing it once and seen that it is not working properly and we never merged it to main.

The task is to find and propose good way to solve. It is up to you to analyze it and propose good solution for catching up. If you need you should add `WarmKeyDelta`

a-kuprin: We can discuss the proposed solution first here in the comments

0xMayoor: 1. carry the `WarmKeyDelta` and the per-nonce sigs on the recovery fetch.
the peer already has both in its `DiffRecord`, `HandleGetDiffs` just strips them today
2. verify those sigs against the root they carry, not `store.GetSignatures` 
that's what gets past the empty store
3. once a nonce clears 2/3 verified, inject the delta, apply, check the computed root matches the signed one
4. move the inject inside `applyCore`'s snapshot so a failed apply rolls it back too

the root folds the warm keys in, so a wrong delta can't reproduce a 2/3-signed root, nothing bad gets through.

two questions @a-kuprin :
catch up only to the finalized nonce and let gossip take the tip (can't trust warm keys below 2/3 anyway), or further? 
and sigs off one synced peer since it carries the 2/3?

0xMayoor: > 1. carry the `WarmKeyDelta` and the per-nonce sigs on the recovery fetch.
>    the peer already has both in its `DiffRecord`, `HandleGetDiffs` just strips them today
> 2. verify those sigs against the root they carry, not `store.GetSignatures`
>    that's what gets past the empty store
> 3. once a nonce clears 2/3 verified, inject the delta, apply, check the computed root matches the signed one
> 4. move the inject inside `applyCore`'s snapshot so a failed apply rolls it back too
> 
> the root folds the warm keys in, so a wrong delta can't reproduce a 2/3-signed root, nothing bad gets through.
> 
> two questions @a-kuprin : catch up only to the finalized nonce and let gossip take the tip (can't trust warm keys below 2/3 anyway), or further? and sigs off one synced peer since it carries the 2/3?

the approach doesn't hold: 
once an instance is on a warm key, it can't catch back up afterward. so it's not the right one,

a-kuprin: @0xMayoor 
I think we should catch up first below 2/3 you must not trust WarmKeyDelta from a peer.
But if further nonces have no warm-key delta we also can apply txs (till there is WarmKeyDelta) on catch up

And yes let's get sigs only from one peer

0xMayoor: I need some helpp with this one @a-kuprin 

correct me if I have it wrong but the 2/3-from-sigs check can't work once instances sign with warm keys right? we can't verify a warm sig without the binding it's supposed to establish, so it never reaches 2/3 on catch-up and stalls at the first warm-key nonce.

I only see two ways to trust a warm binding without the sigs:
resolve the grant at the diff's own height (deterministic, but diffs carry a timestamp not a height today, so it's a change)

or attest with cold keys (defeats the warm delegation)

am I missing something?

a-kuprin: btw. We can't rely even on 2/3 signed warm key delta, as normally gossiping is disabled. It means that when we are catching up from one host he has only WarmKeysDelta signed by itself...

I'm thinking. I will go back here

a-kuprin: I think we should have signed key rotation, but it should be another task.

Idea:
When host changes warm key, he should sign this message with an old one. And there should be some easy to use mechanics, like putting new key at key storage, and devshardd should automatically rotate and follow the protocol rules for key rotation. But the rotation message should be in the diff (possibly unified with other host message: tx is signed by new key, and new key in the tx is signed with the old one)

So we will be able to validate the warm key delta chain, and we will not need any other signatures.

What do you think?

0xMayoor: makes sense. does the rotation sig cover the escrow, or just the new key? warm keys aren't tied to one escrow, right? so if it's only the key, a rotation from one escrow replays into another.

a-kuprin: It's not escrow rotation.
It's a warm key rotation of a participant during one escrow.

The logic is:
we have created escrow and selected participants, every participant in devshard signs state root by itself (that are shared during finalization) and signs messages initiated by the host like ValidationVotes, ConfirmInferenceStart, FinishInference
Other participants can't trust this messages, they check the signature, but they are signed by participant warm key, that we can check for delegation on L1 chain. So when another host gets signed message it first checks if he knows this warm key, if it is a new warm key, it checks new key is delegated from participant of devshard and created `WarmKeysDelta` record

So the problem is that some warmkeys could be rotated (it is possible bot rare event) and we need the way to validate all messages signed by the participant even old if key was rotated multiple times.
I think if participant everytime issues some message that he rotated the key and it is signed by old key is enough.

So we are talking about one escrow and warmkeys rotation, not the escrows rotation

PR #991, opened 2026-04-01
[P0] Support multipart openai requests

### LINKED ISSUE #985: [P0] Bug: unsupported OpenAI type input for the inference requests
decentralized-api rejects valid Cursor/OpenAI chat requests where messages[].content is an array of text parts instead of a plain string. The request fails during request parsing on TA/executor because Message.Content is typed as string, producing 500 {"error":"json: cannot unmarshal array into Go struct field ... content of type string"} even though the payload is valid modern chat-completions format.

DimaOrekhovPS:


decentralized-api/internal/server/public/entities.go
```diff
@@ -31,7 +35,80 @@ type OpenAiRequest struct {
...
+}
+
+type MessageContent struct {
+	Text  *string
+	Parts []ContentPart
+}
+
+type ContentPart struct {
+	Type string `json:"type"`
+	Text string `json:"text,omitempty"`
```
akup: Maybe add here "image_url" field?

DimaOrekhovPS: Decided to go with a more general approach to preserve any unserialized fields

decentralized-api/internal/server/public/post_chat_handler.go
```diff
@@ -294,9 +294,10 @@ func (s *Server) handleTransferRequest(ctx echo.Context, request *ChatRequest) e
...
-	promptText := ""
-	for _, message := range request.OpenAiRequest.Messages {
-		promptText += message.Content + "\n"
+	promptText, ignoredParts := FlattenMessagesText(request.OpenAiRequest.Messages)
+	if ignoredParts > 0 {
+		logging.Info("Ignored non-text prompt parts while estimating prompt size", types.Inferences,
+			"ignored_parts", ignoredParts, "model", request.OpenAiRequest.Model)
 	}
 
 	promptTokenCount, err := s.getPromptTokenEstimation(promptText, request.OpenAiRequest.Model)
```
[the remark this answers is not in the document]
DimaOrekhovPS: We will fail before transaction creation anyway, not worth refactoring

x0152:
Why do we pass through image_url and other non-text parts if the current model only supports text? Shouldn't we return 400 for them for now, or align allowed types to model capabilities?

DimaOrekhovPS:
> Why do we pass through image_url and other non-text parts if the current model only supports text? Shouldn't we return 400 for them for now, or align allowed types to model capabilities?

I think it’s better to keep the API flexible for now so we don’t create unnecessary obstacles, for example around tool-calling workflows. It also doesn’t seem like a security issue to keep the API validation lenient here.

0xgonka (approved):

PR #1000, opened 2026-04-02
[#976] Distribute `devshard` rewards at the end of epoch

### LINKED ISSUE #976: [P0] `devshards`: Distribute `WorkCoins` at the end of the epoch
As described in https://github.com/gonka-ai/gonka/issues/914#issuecomment-4090483233, we want to:

* Distribute `WorkCoins` at the end of the epoch, instead of upon settlement.
* Take `devshards` stats into account when 
    * calculating punishments `WorkCoins`/`RewardCoins` (see `bitcoin_rewards.go`)
    * participant's inactivity status (see `status.go` -> `ComputeStatus`)

dcastro:
Distribute `WorkCoins` at the end of the epoch, instead of upon settlement.

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
+++ b/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"context"
 	"fmt"
+	"math"
 	"math/bits"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
@@ -87,23 +88,28 @@ func (k msgServer) SettleSubnetEscrow(goCtx context.Context, msg *types.MsgSettl
 			continue
 		}
 		paidValidators[addr] = true
+
+		// Track total payout for remainder/refund calculation.
 		nextTotalPayout, carry := bits.Add64(totalPayout, payout, 0)
 		if carry != 0 {
 			return nil, fmt.Errorf("total validator payout overflow")
 		}
 		totalPayout = nextTotalPayout
 
-		recipientAddr, err := sdk.AccAddressFromBech32(addr)
-		if err != nil {
-			return nil, fmt.Errorf("invalid validator address %s: %w", addr, err)
-		}
-		coins, err := types.GetCoins(int64(payout))
-		if err != nil {
-			return nil, fmt.Errorf("invalid payout amount: %w", err)
-		}
-		err = k.BankKeeper.SendCoinsFromModuleToAccount(goCtx, types.ModuleName, recipientAddr, coins, "subnet_escrow_payment")
-		if err != nil {
-			return nil, fmt.Errorf("failed to pay validator %s: %w", addr, err)
+		// Apply payout to participant balances (ledger-style, no bank transfer).
+		participant, found := k.GetParticipant(ctx, addr)
+		// If the subnet is settled on a different epoch, it's possible a subnet's host is no longer a participant on this epoch.
+		// In this case, we skip payment to this host
+		if found {
+			if payout > math.MaxInt64 {
+				return nil, fmt.Errorf("validator payout out of range for %s", addr)
+			}
+			if err := k.processParticipantPayment(ctx, &participant, int64(payout), "subnet_escrow_payment"); err != nil {
+				return nil, err
+			}
+			if err := k.SetParticipant(ctx, participant); err != nil {
+				return nil, fmt.Errorf("failed to update participant %s: %w", addr, err)
+			}
 		}
 	}
 
--- a/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow_test.go
+++ b/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow_test.go
@@ -28,6 +28,15 @@ func TestSettleSubnetEscrow_FeesSplitBySlotCount(t *testing.T) {
 	addrH2 := cosmosAddressFromDcrdKey(keyH2).String()
 	addrH3 := cosmosAddressFromDcrdKey(keyH3).String()
 
+	for _, addr := range []string{addrH1, addrH2, addrH3} {
+		err = k.SetParticipant(ctx, types.Participant{
+			Address: addr,
+			Index:   addr,
+			Status:  types.ParticipantStatus_ACTIVE,
+		})
+		require.NoError(t, err)
+	}
+
 	initialAmount := uint64(1_000)
 	fees := uint64(403)
 	expectedUserRefund := initialAmount - fees
@@ -53,16 +62,7 @@ func TestSettleSubnetEscrow_FeesSplitBySlotCount(t *testing.T) {
 	}
 	msg := buildSettlementTestData(t, escrow, []*dcrdsecp.PrivateKey{keyH1, keyH1, keyH2, keyH3}, hostStats, fees)
 
-	payouts := make(map[string]uint64)
-	mocks.BankKeeper.EXPECT().
-		SendCoinsFromModuleToAccount(gomock.Any(), types.ModuleName, gomock.Any(), gomock.Any(), gomock.Eq("subnet_escrow_payment")).
-		DoAndReturn(func(_ context.Context, _ string, recipient sdk.AccAddress, coins sdk.Coins, _ string) error {
-			require.Len(t, coins, 1)
-			payouts[recipient.String()] = coins[0].Amount.Uint64()
-			return nil
-		}).
-		Times(3)
-
+	mocks.BankKeeper.EXPECT().LogSubAccountTransaction(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
 	mocks.BankKeeper.EXPECT().
 		SendCoinsFromModuleToAccount(gomock.Any(), types.ModuleName, creator, gomock.Any(), gomock.Eq("subnet_escrow_refund")).
 		DoAndReturn(func(_ context.Context, _ string, _ sdk.AccAddress, coins sdk.Coins, _ string) error {
@@ -80,9 +80,23 @@ func TestSettleSubnetEscrow_FeesSplitBySlotCount(t *testing.T) {
 	//
 	// Remainder fees are distributed 1 coin per slot, starting from the first slot.
 	// H1 gets 2 remainder coins for its two slots, H2 gets 1 coin, and H3 gets 0 coins.
-	require.Equal(t, uint64(202), payouts[addrH1])
-	require.Equal(t, uint64(101), payouts[addrH2])
-	require.Equal(t, uint64(100), payouts[addrH3])
+	participantH1, found := k.GetParticipant(ctx, addrH1)
+	require.True(t, found)
+	require.Equal(t, int64(202), participantH1.CoinBalance)
+	require.NotNil(t, participantH1.CurrentEpochStats)
+	require.Equal(t, uint64(202), participantH1.CurrentEpochStats.EarnedCoins)
+
+	participantH2, found := k.GetParticipant(ctx, addrH2)
+	require.True(t, found)
+	require.Equal(t, int64(101), participantH2.CoinBalance)
+	require.NotNil(t, participantH2.CurrentEpochStats)
+	require.Equal(t, uint64(101), participantH2.CurrentEpochStats.EarnedCoins)
+
+	participantH3, found := k.GetParticipant(ctx, addrH3)
+	require.True(t, found)
+	require.Equal(t, int64(100), participantH3.CoinBalance)
+	require.NotNil(t, participantH3.CurrentEpochStats)
+	require.Equal(t, uint64(100), participantH3.CurrentEpochStats.EarnedCoins)
 }
 
 func TestSettleSubnetEscrow_HappyPath(t *testing.T) {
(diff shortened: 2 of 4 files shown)


inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
```diff
@@ -87,23 +88,27 @@ func (k msgServer) SettleSubnetEscrow(goCtx context.Context, msg *types.MsgSettl
...
 		totalPayout = nextTotalPayout
 
-		recipientAddr, err := sdk.AccAddressFromBech32(addr)
-		if err != nil {
-			return nil, fmt.Errorf("invalid validator address %s: %w", addr, err)
+		// Apply payout to participant balances (ledger-style, no bank transfer).
+		participant, found := k.GetParticipant(ctx, addr)
+		if !found {
+			return nil, fmt.Errorf("participant %s not found for payout", addr)
 		}
```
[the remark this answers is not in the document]
dcastro: Done

inference-chain/x/inference/keeper/msg_server_create_subnet_escrow_test.go
```diff
@@ -17,6 +17,8 @@ import (
 func setupSubnetEscrowTest(t testing.TB) (keeper.Keeper, types.MsgServer, sdk.Context, *keepertest.InferenceMocks) {
 	k, ctx, mock := keepertest.InferenceKeeperReturningMocks(t)
 	sdk.GetConfig().SetBech32PrefixForAccount("gonka", "gonka")
+	// Allow subaccount logging unless a test asserts specific log calls.
+	mock.BankKeeper.EXPECT().LogSubAccountTransaction(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
```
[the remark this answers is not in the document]
dcastro: Done

Brgndy25 (approved):

Brgndy25 (approved):

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
```diff
@@ -87,23 +88,28 @@ func (k msgServer) SettleSubnetEscrow(goCtx context.Context, msg *types.MsgSettl
...
-			return nil, fmt.Errorf("invalid payout amount: %w", err)
-		}
-		err = k.BankKeeper.SendCoinsFromModuleToAccount(goCtx, types.ModuleName, recipientAddr, coins, "subnet_escrow_payment")
-		if err != nil {
-			return nil, fmt.Errorf("failed to pay validator %s: %w", addr, err)
+		// Apply payout to participant balances (ledger-style, no bank transfer).
+		participant, found := k.GetParticipant(ctx, addr)
+		// If the subnet is settled on a different epoch, it's possible a subnet's host is no longer a participant on this epoch.
+		// In this case, we skip payment to this host
+		if found {
```
akup, line 103: I think we need to consider that epoch can change, and anyway participant should payed for inferences that he runned.
There is another thing, that we don't would like to pay fees to this participant for epochs where he wasn't active and there could be lot of issues that should be handled in dynamic shardchain development branch where we need to support adding/removing participants to subnet.

But anyway, It is better to handle the inference payout in this PR, and look for epochs where participant was active.

akup:
This PR defers subnet host payouts from immediate `BankKeeper` transfers to **internal participant ledger** fields (`CoinBalance` / `EarnedCoins`) until epoch claim. The review flags **two severe correctness issues**: an early return in `processParticipantPayment` that can skip **epoch stats initialization** and risk **nil-pointer panics** on consensus paths, and **settlement accounting** where `totalPayout` (and skipped `!found` hosts) can **mis-state refunds** and **trap value** in the module account versus internal obligations. Secondary items are **observability** for money-moving paths and **test / structure** cleanups in Go and Kotlin.

## Must fix

- **`ensureParticipantEpochStats` before zero payout** — If `processParticipantPayment` returns early when `payout == 0` without calling `ensureParticipantEpochStats`, `CurrentEpochStats` can stay **nil** and later logic may **panic** (`msg_server_start_inference.go` ~371–385). Initialize epoch stats **before** the zero-payout early return.
- **`totalPayout` only after successful participant updates** — Accumulating `totalPayout` **before** confirming the participant exists / payment applies **shrinks the creator refund** for amounts that were never credited, leaving value **stuck** (`msg_server_settle_subnet_escrow.go` ~92–113). Move accumulation **inside** the successful `found` / apply path (or equivalent invariant).
- **Defined outcome for `!found` payouts** — When a host is no longer a participant at settlement, skipped payouts must be **explicitly routed** (refund to escrow creator, community pool, documented burn, or tracked “orphan” ledger)—not silently dropped—so **module bank balance** and **sum of internal ledger obligations** cannot **diverge** permanently (same settlement file; aligns with skeptic review).

## Should fix

- **Logging on financial paths** — Log when payouts are **skipped** (`!found`) and when **`processParticipantPayment`** completes (balances / epoch stats updated), so operators and auditors can reconcile settlement (`msg_server_settle_subnet_escrow.go`, `msg_server_start_inference.go`).
- **Cosmos layout** — Consider moving **`processParticipantPayment`** from `msgServer` to **`Keeper`** for reuse and idiomatic state helpers (`msg_server_start_inference.go`).

## Nits

- **Settlement unit tests** — Reduce duplication (`setupActiveParticipant` helper), prefer **table-driven** balance checks, tighten bank mock expectations from `.AnyTimes()` to **`.Times(n)`**, avoid brittle fee math (`msg_server_settle_subnet_escrow_test.go`).
- **Kotlin e2e** — Replace **`!!`** with **`checkNotNull`** (or similar) and use clearer map assertions (e.g. AssertJ **`containsExactlyEntriesOf`**) in `SubnetTests.kt`.
- **Document helper constraints** — Note that **`processParticipantPayment`** is written for **non-negative** payouts only (not generic debits), if that remains true.

akup (changes requested):

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
akup: All other funds (non-inference related) should be either explicitly burned, refunded to the escrow creator, or sent to a community pool

dcastro: > I think we need to consider that epoch can change, and anyway participant should payed for inferences that he runned.

I guess if a subnet starts in epoch `X` and is settled on epoch `Y`, and a host `H` was active during `X` but not during `Y`, then we could fallback to making the payment to `H` immediately (i.e. using `SendCoinsFromModuleToAccount`) instead of being delayed (i.e. with `processParticipantPayment`). What do you think?

> There is another thing, that we don't would like to pay fees to this participant for epochs where he wasn't active 

Not sure what you meant here. Do you mean e.g:
* imagine a subnet that lasts over 2 epochs, `X` and `Y`
* some host `H` from that subnet ran some inference `I1` during `X` and another inference `I2` during `Y`
* meanwhile, in the main chain, that host `H` was active during `X` , but not during `Y`
* therefore `H` should be paid for inference `I1` but not for `I2`

Is that what you meant?

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
```diff
@@ -3,6 +3,7 @@ package keeper
 import (
```
dcastro, line 3: Starting a thread to reply to this feedback: https://github.com/gonka-ai/gonka/pull/1000#issuecomment-4189066087

> - **`ensureParticipantEpochStats` before zero payout** — If `processParticipantPayment` returns early when `payout == 0` without calling `ensureParticipantEpochStats`, `CurrentEpochStats` can stay **nil** and later logic may **panic** (`msg_server_start_inference.go` ~371–385). Initialize epoch stats **before** the zero-payout early return.

This is fine, it's not `ensureParticipantEpochStats`'s responsibility to make sure that `CurrentEpochStats` is initialized. No codepaths should assume that `CurrentEpochStats` is not nil anyway.

dcastro: > - **`totalPayout` only after successful participant updates** — Accumulating `totalPayout` **before** confirming the participant exists / payment applies **shrinks the creator refund** for amounts that were never credited, leaving value **stuck** (`msg_server_settle_subnet_escrow.go` ~92–113). Move accumulation **inside** the successful `found` / apply path (or equivalent invariant).
->  
> - **Defined outcome for `!found` payouts** — When a host is no longer a participant at settlement, skipped payouts must be **explicitly routed** (refund to escrow creator, community pool, documented burn, or tracked “orphan” ledger)—not silently dropped—so **module bank balance** and **sum of internal ledger obligations** cannot **diverge** permanently (same settlement file; aligns with skeptic review).

These 2 points are being discussed in this other thread: https://github.com/gonka-ai/gonka/pull/1000#discussion_r3036997712

dcastro: > - **Cosmos layout** — Consider moving **`processParticipantPayment`** from `msgServer` to **`Keeper`** for reuse and idiomatic state helpers (`msg_server_start_inference.go`).

Done

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
akup: It seams we need something to start from.

At first phase we can't handle multi-epoch devshards. What was discussed with @gmorgachev we should force finalization of a `devshard` when epoch ends. This finalization should be scheduled to next tasks.

But anyway it is possible that finalization is performed at epoch `X+1`, so we need to check if participant was active at epoch `X`, not current epoch but epoch when subnet escrow started.

So it means that we shouldn't handle for now any complex logic, but just look for participant from lists of participants of epoch when escrow started.

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
akup: > This is fine, it's not `ensureParticipantEpochStats`'s responsibility to make sure that `CurrentEpochStats` is initialized. No codepaths should assume that `CurrentEpochStats` is not nil anyway.

Yes it is not used. Noisy comment, sorry

inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
dcastro: > just look for participant from lists of participants of epoch when escrow started.

Can you please elaborate on this?

We need to look up the participant for 2 reasons:
1. to update their stats (inference count, validated, invalidated, missed) so that they're taken into account when calculating the punishments for the rewards at the end of the current epoch
2. to accumulate `EarnedCoins`, so they can be claimed at the end of the current epoch

---
Regarding point 1: There's no point in updating the stats for older epochs, since those rewards have already been distributed. In my opinion, if the subnet was started on a previous epoch, we have 2 options:
    * Option 1: don't update any stats
    * Option 2: if the participant is *still active* in the current epoch, update the stats for the *current* epoch. Otherwise, don't update the stats.

---

Regarding point 2: Correct me if I'm wrong, but I don't think we can accumulate `EarnedCoins` for previous epochs. `EarnedCoins` are settled at the end of the epoch and then claimed; but I don't think we can increment them after that stage.

So, in my opinion, if the subnet was started on a previous epoch, I think we should:
    * Option 1: send the coins directly to the host using `SendCoinsFromModuleToAccount`, and don't increment `EarnedCoins`.
    * Option 1: if the participant is *still active* in the current epoch, increment their `EarnedCoins` in the *current* epoch. Otherwise, send the coins directly using `SendCoinsFromModuleToAccount`

akup: You are right on both points.

We anyway cannot punish for previous epoch and so we don't need these stats.
There is a big thing that should be redesigned in network... and devshards. The missed inferences if they over limit should trigger event that is sent to `inference-chain` before finalization and settlement. But it is definetely out the scope of the PR.

But we need stats also for reputation calculations, and there `EpochPerformanceSummary` is used. I think here we need to update previous `EpochPerformanceSummary`.


> Option 1: if the participant is _still active_ in the current epoch, increment their `EarnedCoins` in the _current_ epoch. Otherwise, send the coins directly using `SendCoinsFromModuleToAccount`

It looks as the best option in current design.
But we shouldn't just send the coins directly to account. We first need to check if there was no missed confirmation PoCs, and also punishment for direct missed inference (inferences in old-fashion design).

We just need to check `EpochPerformanceSummary` to see if reward for epoch is `> 0`. If participant was punished his reward is 0, if not, it will be always `> 0`.

If reward wasn't claimed yet (also available at `EpochPerformanceSummary`) we can append revenue to `SettleAmount` and `EpochPerformanceSummary`, if it was already claimed we update only `EpochPerformanceSummary` and send coins directly to participant account

dcastro: > But we need stats also for reputation calculations, and there EpochPerformanceSummary is used. I think here we need to update previous EpochPerformanceSummary.

Good catch, I wasn't aware of `EpochPerformanceSummary`, thank you!

> We just need to check EpochPerformanceSummary to see if reward for epoch is > 0. If participant was punished his reward is 0, if not, it will be always > 0.

Good idea as well, agreed.

----

I implemented a simple approach, to minimize edge cases, in these 2 commits, to the branch for PR #1001:
* `[#976] Handle same-epoch and different-epoch settlements`
* `[#976] Skip payout for previous epochs if the host was punished`

To sum up:
* if it's a same-epoch settlement, treat it was normal onchain inferences:
    * increment `EarnedCoins`
    * aggregate into `CurrentEpochStats`
* if it's a cross-epoch settlement:
    * if the host was NOT punished in the previous epoch, send a direct bank transfer
    * aggregate into `EpochPerformanceSummary`

akup:
Very close logic is implemented and merged in https://github.com/gonka-ai/gonka/pull/1087

PR #1001, opened 2026-04-02
[#976] Aggregate `devshards` stats with `CurrentEpochStats`

### LINKED ISSUE #976: [P0] `devshards`: Distribute `WorkCoins` at the end of the epoch

dcastro:
In this PR: when a subnet is settled, we take the subnet's `HostStats` and merge them with the `participant.CurrentEpochStats`.

NOTE: `participant.CurrentEpochStats` keeps track of the total number of inferences performed (`InferenceCount`) and the number of successfully validated inferences (`Validated`). Subnets were _not_ keeping track of these 2 stats, so we had to alter the subnet to start doing that and include those stats in the settlement.

---

Since the subnet's stats are now merged into `participant.CurrentEpochStats`, they will automatically be taken into account when:

  * calculating punishments `WorkCoins`/`RewardCoins` (see `bitcoin_rewards.go`)
  * participant's inactivity status (see `status.go` -> `ComputeStatus`)


heitor-lassarote (approved):
LGTM, the only thing that I think would be nice in addition would be a Testermint test, but since validations are currently not working in E2E as expected, I think this may be skipped. In addition, testing some things such as missed requests will probably require some extra infra, which sounds out of scope for this. So I think, for now all is fine.

akup:
Here are new state fields,  and on update it will lead to immediate consensus splits and data corruption on live networks.

@gmorgachev How adding fields is supposed to be handled in subnets? It is not binded to block height of the mainnet

akup:
Threads subnet `InferenceCount` / `Validated` into L2 hashing and L1 settlement so subnet work feeds `CurrentEpochStats`. It has **hard consensus and economics risk** that is flagged: L2 hash changes without a coordinated rollout can split hosts, and **incremental `Validated` accounting** in `machine.go` is easy to get wrong (ordering, double application, underflow), which then flows into L1 rewards. **L1 settlement** may also mis-attribute stats across epochs, and **storage migrations / tests** need tightening before this is safe to ship.

## Must fix

- **Version or rollout gate for L2 hash changes** — Including `InferenceCount` and `Validated` in `subnet/state/hash.go` without a protocol/version boundary risks an immediate split between upgraded and non-upgraded hosts; coordinate upgrade or feature gating.

> This one point needs to be handled in some generic upgrade tooling (proposed here: https://github.com/gonka-ai/gonka/pull/1005#issuecomment-4189357931)

- **Idempotent validation outcome handling** — When vote thresholds fire, ensure `Validated` is not incremented again on **late or duplicate votes** after the record is already validated (`subnet/state/machine.go`).
- **Order-dependent running tallies** — The logic updates `sm.state.HostStats[rec.ExecutorSlot].Validated` via scattered increments and decrements based on the real-time stream of `MsgValidation` messages. 
If a malicious sequencer orders an `invalid` vote before a `valid` vote, the `rec.VotesInvalid == 0` condition fails, and the counter is not incremented. If the inference subsequently crosses the validation threshold, it appears another increment might occur in `applyValidationVote` (line 828). This spread of `++` and `--` logic across different handlers based on transient state (`rec.VotesValid == 0`) guarantees that any future changes to voting rules will introduce accounting bugs. 

## Should fix

- **Test isolation** — Stop mutating global **`sdk.GetConfig()`** in tests (parallel runs / flakiness); use `TestMain` or scoped setup (`msg_server_settle_subnet_escrow_test.go`).
- **Aggregation safety and failure behavior** — Aggregation of `InferenceCount` and `Validated` lacks overflow protection. **Fix:** Use `bits.Add32` and return an error if an overflow occurs.

## Nits

- **Factor duplicated settlement test setup** into a shared helper.
- **Tests re-deriving executor slot** (`inferenceID % len(hosts)`) — couple to the real assignment from state/records instead of duplicating internals (`machine_test.go`).
- **Redundant participant loads** vs data already in `SubnetHostEpochStatsMap` — simplify if one source of truth suffices.

akup (changes requested):
Should handle votes duplication, order-dependent `++`/`--`, overflows checks

dcastro:
> Should handle votes duplication, order-dependent ++/--, overflows checks

You mean in the updates of `HostStats.Validated` in `machine.go`?
Yeah, that is already being handled, there is a test in `TestApplyDiff_Validation_ValidatedCounterTransitions` that ensures duplicate calls and reordering of `MsgValidation` remains idempotent

akup (approved):

akup:
Very close logics is already implemented at merged https://github.com/gonka-ai/gonka/pull/1069

PR #1002, opened 2026-04-02
Remove top_miner code

patimen:
**1. What problem does this solve?**
This PR cleans up the codebase by removing the speculative `TopMiner` logic, which was never actually enabled on the live chain. It also cleans up the TopMiner collection, currently storing some 2000 entries in mainnet.
Note that in mainnet, no money was set aside or scheduled for disbursement in the top_miner program, so this change has no financial impact.
This also reduces potential attack vectors and vulnerabilities.

**2. How do you know it is a real problem?**
As noted in the PR context, the TopMiner program was never activated, yet it occupies space in the state machine, parameters, and API. In addition, a Certik audit has found a handful of issues with the TopMiner code, and removing it is the cleanest solution.

**3. How does this solve the problem?**
* **State Cleanup**: Introduces the `v0.2.12` upgrade handler which clears the `TopMiners` collection from the KVStore.
* **Code Removal**: Deprecates (marks as `reserved`) all related fields in the `.proto` files, removes the `TopMiner` gRPC/CLI query endpoints, stops minting these coins during `InitGenesis`, and removes the top miner registration hook from the end-of-PoC lifecycle.

**4. What risks does this introduce? How do we mitigate these risks?**
* **Risks**: Directly modifying chain state (clearing collections) during a live chain upgrade carries the risk of consensus failure or halting the chain if the balances are miscalculated or if the module lacks the correct permissions.
* **Mitigations**:  We will need to be sure to test this on test_net before release to main.

**5. How do you know this PR fixes the problem?**
The `CreateUpgradeHandler` function for `v0.2.12` demonstrably executes `k.TopMiners.Clear`. Furthermore, all references to the top miner generation in `onEndOfPoCValidationStage` have been removed, guaranteeing no new top miners or rewards will be generated going forward.


0xgonka:
Proto field reservations look correct, `TopMiners.Clear()` is safe when empty, and ConsensusVersion is bumped properly. Main issues:

**Must fix:**

- **Token burn is missing from the upgrade handler.** PR description says funds are "explicitly burned" but `removeTopMiner()` in `upgrades.go` only calls `k.TopMiners.Clear(ctx, nil)` and cleans up `tokenomicsData`. No `BurnCoins` call exists anywhere. The `Burner` permission is added to the module account in `app_config.go` but never used. Without the burn, 120M GONKA (12% of supply) sits permanently locked in the dead `TopRewardPoolAccName` module account. Need to add something like:

```go
rewardAddr := k.AccountKeeper.GetModuleAddress(types.TopRewardPoolAccName)
balances := k.BankView.GetAllBalances(ctx, rewardAddr)
if !balances.IsZero() {
    if err := k.BankKeeper.BurnCoins(ctx, types.TopRewardPoolAccName, balances); err != nil {
        return err
    }
}
```

- **`distrKeeper` parameter is accepted but never used.** It's passed to `CreateUpgradeHandler()` and `app.DistrKeeper` is wired in `upgrades.go:67`, but never referenced in the handler body. Either use it or remove it. Removing is cleaner since it also avoids a merge conflict with the tx-fees branch which has `(mm, configurator, k)` as the v0.2.12 handler signature.

**Should fix:**

- `genesis.proto` still imports `top_miner.proto` — unused after TopMiner list was removed from GenesisState. Clean up the import.
- `expected_keepers_mocks.go` removes the `SetDenomMetaData` mock, but `SetDenomMetaData` is still in the `BankKeeper` interface at `expected_keepers.go:33`. Is this intentional? Could break mock generation consistency.

**Consider:**

- `TopMiners` field in `keeper.go:66` and `TopRewardPoolAccName` / `TopMinerPrefix` in `keys.go` are retained — needed for the upgrade handler's `Clear()` call and module account registration. Worth adding a comment that these are retained only for migration and should be removed in a future version.

patimen:
In regards to the comments from @0xgonka :

I was incorrect about mainnet having funds allocated. It does not. I have updated the code and the description accordingly.

I have addressed the other issues in code.

0xgonka (approved):

PR #1009, opened 2026-04-03
Remove training

### LINKED ISSUE #468: Certik (finalization)

patimen:
When we introduce training it will doubtless use the devshards we are currently developing. The placeholder approach we already have is no longer needed, and has several security issues and presents a large surface area of attack for literally zero upside, as well as additional complexity.


tcharchian:
https://github.com/gonka-ai/gonka/pull/1002 needs review first

0xgonka:
The deletion looks thorough — DAPI, keeper, simulation, CLI all cleaned up. Proto field reservations in `params.proto`/`tokenomics_data.proto` are correct. One critical bug:

**CRITICAL: `training_state_cleanup.go:28` — unbounded iterator will delete module params and other non-training state**

```go
iterator := store.Iterator(keyPrefix, nil)
```

`store.Iterator(start, end)` with `end=nil` iterates from `start` to the **end of the entire module KV store**, not just keys with that prefix. The store is a raw `runtime.KVStoreAdapter`, not prefix-scoped.

The first prefix (`"TrainingTask/value/"`, byte 0x54) will sweep past all training keys, reach `p_inference` (byte 0x70, the module params key), and **delete it**. Since `clearTrainingState` runs before `adjustParameters` in the upgrade handler, `GetParams` will then return a zero/empty params struct, effectively wiping all module configuration.

Fix: use `sdk.KVStorePrefixIterator(store, keyPrefix)` instead of `store.Iterator(keyPrefix, nil)`:

```go
iterator := sdk.KVStorePrefixIterator(store, keyPrefix)
```

The unit test doesn't catch this because the in-memory test store only contains the seeded training keys — no `p_inference` key exists in the fixture.

**Should fix:**

- Generated files for deleted `network_node.proto` are still present (`network_node.pb.go`, pulsar files). Dead code — should be deleted.
- Consider adding comments in `tx.proto` and `query.proto` documenting the removed training RPCs to prevent accidental method name reuse.

**Looks correct:**

- `TopRewardPoolAccName` Burner permission addition is intentional (for top_miner burn)
- Training collection definitions retained in keeper — needed for `ClearTrainingState` during upgrade
- `key_training_task.go` retained — provides prefixes for cleanup
- `HardwareNodeStatus_TRAINING = 3` enum kept — removing proto3 enum values breaks wire compat
- `permissions.go` + `permissions_coverage_test.go` cleanup is correct and self-validating

0xgonka (approved):

PR #1012, opened 2026-04-04
fix(keeper): propagate SetPocValidationV2 storage error in SubmitPocValidationsV2

Mayveskii:
## Problem

`SubmitPocValidationsV2` used `LogWarn + continue` for **all** `SetPocValidationV2` errors, treating infrastructure-level storage failures identically to expected validation errors (e.g. invalid bech32 addresses).

A transient storage failure would:
- silently drop votes from the batch
- return `success` to the caller
- leave consensus state inconsistent with no observable signal

## Fix

Replace `LogWarn + continue` with `LogError + return error` for `SetPocValidationV2` failures.

**Why this is safe and does not break partial-success design:**

`HasPocValidationV2` (called earlier in the same loop iteration, line 116) already validates bech32 addresses for both `participantAddress` and `validatorAddress`. Any error reaching `SetPocValidationV2` is therefore definitively a **storage-layer failure**, not an address validation error.

Two distinct paths:
- **Invalid address** → caught by `HasPocValidationV2` → `LogWarn + continue`. Partial-success preserved.
- **Storage failure** → only reachable after address validation succeeds → `LogError + return nil, fmt.Errorf(...)`. Cosmos SDK rolls back the full transaction.

`TestSubmitPocValidationsV2_PartialSuccess` passes without modification — `"invalid_address"` is caught at `HasPocValidationV2`, never reaches `SetPocValidationV2`.

Adds `TestSubmitPocValidationsV2_StorageError_AddressPreValidation` documenting this invariant.

## Impact

- Storage failures are now observable: `LogError` surfaces in node monitoring, transaction rollback prevents partial writes
- Partial-success behavior for invalid/duplicate addresses unchanged (intentional design preserved)
- Consistent with error propagation pattern in `SubmitPocValidationV1` (single-item handler)

## Test

```
go test ./x/inference/keeper/... -run "TestSubmitPocValidationsV2"
--- PASS: TestSubmitPocValidationsV2_DuplicateSkipped
--- PASS: TestSubmitPocValidationsV2_PartialSuccess
--- PASS: TestSubmitPocValidationsV2_StorageError_AddressPreValidation
ok  github.com/productscience/inference/x/inference/keeper
```

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_poc_v2_test.go
+++ b/inference-chain/x/inference/keeper/msg_server_poc_v2_test.go
@@ -182,6 +182,31 @@ func TestSubmitPocValidationsV2_PartialSuccess(t *testing.T) {
 	require.True(t, exists2)
 }
 
+// TestSubmitPocValidationsV2_StorageError_AddressPreValidation documents the address
+// pre-validation invariant: HasPocValidationV2 validates bech32 addresses BEFORE
+// SetPocValidationV2 is called. Therefore any SetPocValidationV2 error is a storage failure.
+// This test verifies the invariant by showing SetPocValidationV2 with an invalid participant
+// address errors on bech32 decoding — and that this path is unreachable via the batch handler
+// (invalid addresses are already caught by HasPocValidationV2 at line 116).
+func TestSubmitPocValidationsV2_StorageError_AddressPreValidation(t *testing.T) {
+	k, ctx, _ := keepertest.InferenceKeeperReturningMocks(t)
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+
+	// Direct call: SetPocValidationV2 returns error for invalid participant bech32.
+	err := k.SetPocValidationV2(sdkCtx, types.PoCValidationV2{
+		ParticipantAddress:          "invalid_address",
+		ValidatorParticipantAddress: testutil.Validator,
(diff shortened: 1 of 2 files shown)


Doog-bot534 (approved):
## Review: fix(keeper): propagate SetPocValidationV2 storage error

**Approve** ✅

The core logic change is sound. The original code silently swallowed storage errors and continued processing the batch, which could leave consensus state partially written. Since Cosmos SDK rolls back the entire transaction on error, aborting is the correct behavior for infrastructure failures.

The invariant argument is well-reasoned: `HasPocValidationV2` validates bech32 addresses before `SetPocValidationV2` is reached, so input validation errors are unreachable at the `Set` stage.

### Suggestions

1. **Test gap:** The test only tests the direct `SetPocValidationV2` call with an invalid address. It does not test the new error propagation path in `SubmitPocValidationsV2` itself. A test injecting a mock store failure would be more valuable to verify the batch actually aborts and returns an error.

2. The `return nil, fmt.Errorf(...)` response returns a nil first value — verify callers/middleware handle nil response + non-nil error correctly. Standard Cosmos SDK convention, so likely fine.

**Verdict:** Correct directional fix. Ship it.

x0152:
Closing - this is no longer an issue, covered by design changes around PoC V2 error handling

PR #1013, opened 2026-04-04
fix(subnet): prevent fund loss in unsettled escrow distribution

### LINKED ISSUE #979: `devshards` escrow: fund loss on unsettled pruning + missing overflow guards in host stats aggregation
## Summary

Three related bugs in subnet escrow settlement and pruning code (v0.2.11):

### 1. Fund loss in unsettled escrow pruning (Medium)

`distributeUnsettledEscrow` (subnet_pruning.go) silently logs `SendCoinsFromModuleToAccount` failures but returns nil. The subnet Remover in `pruning.go` then deletes the escrow from state. Validators whose payments failed permanently lose their share — the funds remain locked in the inference module account with no way to recover.

**Remover also swallows errors**: Even if escrow or index deletion fails, the Remover returns nil (line 149), potentially leaving orphaned state entries while advancing the pruning cursor past them.

**Root cause**: Error-handling pattern treats pruning cleanup as "best-effort" but deletes the source-of-truth (escrow) regardless of outcome.

**Impact**: Permanent fund loss for validators in any unsettled escrow where at least one `SendCoins` call fails during pruning.

### 2. Inconsistent overflow protection in `AggregateSubnetHostStats` (Low-Medium)

`subnet_host_stats.go` — `Cost` (uint64) has an overflow check before addition, but `Missed`, `Invalid`, `RequiredValidations`, and `CompletedValidations` (all uint32) do not. The developer clearly intended overflow protection (present for Cost) but missed the other four fields.

### 3. Missing overflow check in `SettleSubnetEscrow` validator cost aggregation (Low)

`msg_server_settle_subnet_escrow.go:37` — `validatorCosts[addr] += hs.Cost` accumulates costs for validators with multiple slots without overflow protection. Same pattern that `AggregateSubnetHostStats` correctly guards for Cost.

## Affected files

- `inference-chain/x/inference/keeper/subnet_pruning.go` (distributeUnsettledEscrow)
- `inference-chain/x/inference/keeper/pruning.go` (GetSubnetPruner Remover)
- `inference-chain/x/inference/keeper/subnet_host_stats.go` (AggregateSubnetHostStats)
- `inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go` (SettleSubnetEscrow)

Mayveskii:
## Problem

`distributeUnsettledEscrow` sends coins to validators one-by-one. If validators 0..N-2 succeed but validator N-1 fails, the function returned an error and the escrow was preserved in state — but on the next pruning epoch, validators 0..N-2 would be paid again (double payment).

Additionally, the error from `SendCoinsFromModuleToAccount` was being swallowed: logged but the function returned `nil`, allowing `Remover` to proceed and delete the escrow unconditionally.

Note: `Pruner.Prune` uses `continue` on Remover errors (no Cosmos SDK tx rollback). The escrow survives because `Remover` exits before `SubnetEscrows.Remove()` is called.

Closes #979

## Fix

Wrap the entire payment loop in `sdk.CacheContext`. `commit()` is only called if **all** payments succeed. If any `SendCoins` fails:
- No coins are transferred (cache context discarded, no partial payment committed)
- Escrow remains in state for retry on the next pruning epoch
- No double-payment on subsequent epochs

Also propagate the error from `Remover` so `SubnetEscrows.Remove()` is never called when distribution fails.

## Impact

**Before:** partial payment committed on bank error → escrow deleted → irrecoverable fund loss. Or partial payment committed → escrow preserved → double-payment next epoch.

**After:** atomic all-or-nothing distribution. Any `SendCoins` failure rolls back all payments for that escrow. Consistent with bank-keeper error handling in `SettleSubnetEscrow` (`return nil, fmt.Errorf("failed to pay validator %s: %w", addr, err)`).

**Test added:** `TestDistributeUnsettledEscrow_BankError_NoPartialPayment` — mocks addr1 success + addr2 failure, verifies escrow remains in state after prune.

### DIFF
--- a/inference-chain/x/inference/keeper/devshard_pruning.go
+++ b/inference-chain/x/inference/keeper/devshard_pruning.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"context"
+	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/productscience/inference/x/inference/types"
@@ -12,6 +13,8 @@ const DevshardPruningMax = int64(100)
 
 // distributeUnsettledEscrow splits the escrowed funds equally among unique validators in the group.
 // Integer division remainder stays in the module account.
+// All payments are wrapped in a CacheContext: if any SendCoins fails, no payments are committed,
+// preventing partial distributions that would cause double-payment on the next pruning epoch.
 func (k Keeper) distributeUnsettledEscrow(ctx context.Context, escrow types.DevshardEscrow) error {
 	// Count unique addresses (first pass)
 	seen := make(map[string]bool)
@@ -32,6 +35,12 @@ func (k Keeper) distributeUnsettledEscrow(ctx context.Context, escrow types.Devs
 		return nil
 	}
 
+	// Wrap all payments in a cache context. If any SendCoins call fails, commit() is not
+	// called, so no coins are transferred. The escrow remains in state and will be retried
+	// on the next pruning epoch.
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	cacheCtx, commit := sdkCtx.CacheContext()
+
 	// Pay in slot order (deterministic iteration over escrow.Slots)
 	paid := make(map[string]bool)
 	for _, addr := range escrow.Slots {
@@ -50,12 +59,14 @@ func (k Keeper) distributeUnsettledEscrow(ctx context.Context, escrow types.Devs
 		if err != nil {
 			continue
 		}
-		err = k.BankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, recipient, coins, "devshard_escrow_unsettled_distribution")
+		err = k.BankKeeper.SendCoinsFromModuleToAccount(cacheCtx, types.ModuleName, recipient, coins, "devshard_escrow_unsettled_distribution")
 		if err != nil {
 			k.LogError("failed to distribute unsettled escrow funds", types.Pruning,
 				"escrow_id", escrow.Id, "address", addr, "error", err)
+			return fmt.Errorf("failed to send coins to %s for escrow %d: %w", addr, escrow.Id, err)
 		}
 	}
 
+	commit()
 	return nil
 }
--- a/inference-chain/x/inference/keeper/devshard_pruning_test.go
+++ b/inference-chain/x/inference/keeper/devshard_pruning_test.go
@@ -316,6 +316,7 @@ func TestDistributeUnsettledEscrow_BankError_NoPartialPayment(t *testing.T) {
 	// CacheContext ensures addr1 partial payment is also rolled back (not committed).
 	_, found := k.GetDevshardEscrow(ctx, id)
 	require.True(t, found, "escrow must remain in state when distribution fails")
+	t.Logf("SECURITY CHECK: escrow id=%d still exists after bank error (found=%v)", id, found)
 
 	// Balance check equivalent: verify addr1's payment was attempted with non-zero coins inside
 	// the CacheContext. Since commit() was never called (addr2 failed), addr1's balance is
--- a/inference-chain/x/inference/keeper/devshard_pruning_test.go
+++ b/inference-chain/x/inference/keeper/devshard_pruning_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"fmt"
 	"testing"
 
 	"cosmossdk.io/collections"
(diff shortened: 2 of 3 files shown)


tcharchian:
Hi @Mayveskii!

Before opening or updating a PR, contributors may use [gonka-ai/ai-reviewer](https://github.com/gonka-ai/ai-reviewer), a single-binary CLI for structured, repo-aware review. It runs multiple specialized review personas, such as security, correctness, architecture, and project-specific lenses, and supports primers for project context and waivers for explicit policy. It is intended to complement human review, not replace it, especially as the volume of change grows.

It is also recommended to review the full presentation before using the tool in practice.
- [Full presentation recording on Discord](https://discord.com/channels/1336477374442770503/1415622117629624362/1485979219711234058)
- [Supplemental video](https://www.youtube.com/watch?v=N4F74vd_pKQ)

Why contributors use it

- Helps surface cross-cutting risks early, including consensus, API, and operational edge cases, by using focused review prompts rather than one generic review pass.
- Review rules and project context can live in the repository, for example, under `.ai-review/<owner>/<repo>/` or in committed files with `ai_review` frontmatter, so feedback can stay aligned with how Gonka evolves in practice.
- Each run writes auditable artifacts, including reports, findings, and token or cost statistics, under `.ai-review/.../runs/...`, making results inspectable and easier to debug.
- `--dry-run` shows which personas and primers would run without calling model APIs, which is useful for validating configuration without incurring costs.

Disclaimer: AI output can be wrong or incomplete; maintainers retain final judgment on all merges. Do not paste secrets into prompts or reports.

Mayveskii:
>re

@tcharchian 
Hey, Tanya! I hope thath's not penalty comment. 
This ant orhter's my pr's were runned by John's ai-reviewer and it was such a crazy timing u wrote that... 
Pr's ok it seems like, i'm waiting for review by maintainers.

I drop an access for you and @patimen to my priv repo where been proved #859 #860 #878. There u can find an amateur research i've done. All the thing about the embryo repo is documented. I hope that u will find usefull set the quality mask by touching anything asspecialy - science. 


Ps: 
Sorry that some of the documentation is in RuLang format, I couldn't even finish it in time.

Doog-bot534 (approved):
## Review: fix(subnet): prevent fund loss in unsettled escrow distribution

**Approve** ✅

The CacheContext approach is the correct Cosmos SDK pattern for atomic multi-send operations, and the error propagation from `Remover` via `return err` in `pruning.go` ensures the escrow is never deleted after a failed distribution.

### Strengths
- Two-part fix addresses both the partial-payment problem (CacheContext) and the swallowed error (`return err`). Both are necessary — either alone leaves a gap.
- Test is well designed: mocks addr1 success + addr2 failure and verifies both escrow preservation and that addr1's payment was attempted inside the cache context.

### Minor observations

1. **\`continue\` on \`AccAddressFromBech32\` error (~line 52):** The existing \`continue\` on address parse failure is still there. If a slot has an invalid bech32 address, that slot is silently skipped, but \`commit()\` is still called and escrow is deleted. A validator with a corrupted address silently loses funds. Pre-existing, but worth noting since you're already in the function.

2. **No retry backoff:** The PR correctly notes "retried on the next pruning epoch." But if the bank error is permanent (e.g., module account drained), this will fail every epoch indefinitely. Consider whether a retry counter or escalation mechanism is warranted.

**Verdict:** The fix is correct and complete for the stated problem. Ship it.

Doog-bot534:
Sounds good, looking forward to the follow-ups on both points. Thanks for the quick turnaround.

Mayveskii:
@tcharchian Hi there !

It is about coup of months just freezed here, and the others my pr's...  Couple of them that is stiil freezed , look good. No feedback in 3 months. What's wrong with it ? I really can't understand...

tcharchian:
Hi @Mayveskii, try reaching out to community members to get feedback on your PRs.

Mayveskii:
@Doog-bot534 hi , bro!  ))

Here's Tanya asking about review & feedback by community members. What do u think about this pr and his lifetime in proto lifecycle ? Do we have better solution for this one or it's completely correct ? 

I'm confused about freezing my pr's and really don't get it what is problem to communicate and show where it's not correct.

tcharchian:
@Mayveskii, it is very nice that you are asking @Doog-bot534 for the opinion, but it is also important to get review not only from @Doog-bot534 person, but from other active community members whom the community trusts as well.

In general, not every PR automatically becomes part of an upgrade. Usually, only changes that have clear value and relevance for the protocol are included in the upgrade lifecycle.

So the question is not only whether the PR is technically correct, but also whether the community sees enough practical value in it at this stage, whether it solves a real current problem, and whether it is important enough to include in the upcoming protocol changes.

I understand that it can feel frustrating when PRs are paused or not moving forward quickly, but this does not mean there is no communication. It means we need proper review, alignment, and community confidence before moving changes into the protocol.

Mayveskii:
>re 

Such a pleasure your answers, @tcharchian ! Thanks. Actually what is frustrating -  my badness. Till it's freezed for 3 months , community members offer more intresting things then i did. So that upset me. The point I got , I should be more confident in the solutions I propose and less dependent on other people's opinions. By the way it won't hurt propose solutions that is not going invalid in 5 min, agreed completely with you. Thanks for providing the feedback, going review the whole contributing strategy for being more stable. Anyway it is invaluable experience u shared.

Have a good time guys, throw deep respect for every1 team member from me, guys you're best !  Bless you

tcharchian:
Hey @Mayveskii! Contributing to a live protocol is always a learning curve, and it is completely normal that some ideas become less relevant over time while the protocol lifecycle moves forward. 

Also, just as a reminder: Gonka is represented on the HackerOne platform as well https://gonka.ai/docs/report-vulnerability/. There, hackers have their own reputation/rating system. The higher the quality of the reports they submit, the higher reward they get. 

Another advantage is that reviewers inside the program also have their own KPIs, including response-time expectations, so in many cases you may be able to get a faster and more structured review there.

The only important thing is to read the program rules and scope very carefully before submitting anything, so that the report matches the expected format and requirements.

Mayveskii:
Heeey , @tcharchian  It is submitted now, as the other things. It wouldn't happen without that impulse u've done. Appreciate you and the team for giving that possibility to lurn curve. Let's run the next one 
    ;)

Mayveskii:
Hi , @gmorgachev You're the one I trust as a mentor. So, as u can see in comments  - in your view , is that I've done looks well now ? 

That thing - [Go Build and Test / Build and Test Devshard (pull_request)](https://github.com/gonka-ai/gonka/actions/runs/28863242611/job/85606831756?pr=1013)Failing .. 

A little confusing coupe of times

PR #1015, opened 2026-04-04
fix(subnet): add overflow guards for cost accumulation in settlement

### LINKED ISSUE #979: `devshards` escrow: fund loss on unsettled pruning + missing overflow guards in host stats aggregation

Mayveskii:
## Problem

`VerifySubnetSettlement` accumulates `totalCost += hs.Cost` (uint64) with no overflow guard, then checks `totalCost > escrow.Amount`.

Attack vector: a settler submits HostStats with `slot[0].Cost = math.MaxUint64` and `slot[1].Cost = 1`. `totalCost` wraps to `0`, bypassing the `escrow.Amount` bound check — an invalid settlement passes verification. With a wrapped `totalCost`, `remainder = escrow.Amount - totalCost` becomes nearly the full escrow amount, misallocating funds to the creator.

The same issue exists in `SettleSubnetEscrow` in two places:
- `validatorCosts[addr] += hs.Cost` (per-validator accumulation)
- `totalCost += cost` (aggregate accumulation)

Attack requires quorum of slot signers, but the vector is real and inconsistent with the `escrow.Amount ≤ MaxAmount` invariant established at creation.

Closes #979

## Fix

Add `math.MaxUint64` overflow guards before both accumulation sites (consistent with existing `Cost` guard in `subnet_host_stats.go`):

```go
if totalCost > math.MaxUint64-hs.Cost {
    return fmt.Errorf("total cost overflow: slot %d cost %d would exceed uint64", hs.SlotId, hs.Cost)
}
totalCost += hs.Cost
```

Two-layer defense: `VerifySubnetSettlement` (verification gate) and `SettleSubnetEscrow` (execution gate) both independently guard against overflow.

## Impact

**Before:** crafted HostStats could wrap `totalCost` → bypass escrow bound check → protocol fund misallocation to creator.

**After:** overflow in cost accumulation returns a hard error at both the verification and execution layers before any bank operations.

**Tests added:**
- `TestVerifySubnetSettlement_TotalCostOverflow` — uses `buildSettlementTestData + generateSubnetKeys(t, 2)` with valid signatures to reach the overflow check
- `TestSettleSubnetEscrow_ValidatorCostOverflow` — end-to-end through `MsgServer.SettleSubnetEscrow`

Run with: `CGO_CFLAGS='-D__BLST_PORTABLE__' go test ./x/inference/keeper/... -run TestVerifySubnetSettlement_TotalCostOverflow`

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
+++ b/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"math"
 	"context"
 	"fmt"
 
@@ -34,6 +35,9 @@ func (k msgServer) SettleSubnetEscrow(goCtx context.Context, msg *types.MsgSettl
 			return nil, fmt.Errorf("host_stats slot_id %d out of range", hs.SlotId)
 		}
 		addr := escrow.Slots[hs.SlotId]
+		if validatorCosts[addr] > math.MaxUint64-hs.Cost {
+			return nil, fmt.Errorf("validator cost overflow for slot %d addr %s", hs.SlotId, addr)
+		}
 		validatorCosts[addr] += hs.Cost
 	}
 
@@ -46,6 +50,9 @@ func (k msgServer) SettleSubnetEscrow(goCtx context.Context, msg *types.MsgSettl
 			continue
 		}
 		paidValidators[addr] = true
+		if totalCost > math.MaxUint64-cost {
+			return nil, fmt.Errorf("total cost overflow paying validator %s", addr)
+		}
 		totalCost += cost
 
 		recipientAddr, err := sdk.AccAddressFromBech32(addr)
--- a/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow_test.go
+++ b/inference-chain/x/inference/keeper/msg_server_settle_subnet_escrow_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"math"
 	"testing"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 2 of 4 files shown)


Doog-bot534 (approved):
## Review: fix(subnet): add overflow guards for cost accumulation in settlement

**Approve** ✅

Clean, minimal overflow guards at both verification and execution layers. The two-layer defense (VerifySubnetSettlement + SettleSubnetEscrow) is good practice.

### Strengths
- The guard pattern \`if x > math.MaxUint64 - y\` is the canonical Go idiom for unsigned overflow detection.
- Three separate guard sites cover all accumulation paths.
- Tests use real key generation and signature infrastructure, not just unit-level mocks.

### Concerns

1. **Per-validator overflow gap in Verify:** \`VerifySubnetSettlement\` only guards \`totalCost\` overflow, but \`SettleSubnetEscrow\` also guards per-validator \`validatorCosts[addr]\`. If a malicious settler assigns MaxUint64 cost across two slots belonging to the same validator, \`validatorCosts[addr]\` would wrap in Settle but not be caught in Verify. The execution layer catches it, but for consistency, consider adding the same per-validator guard in Verify.

2. **Test coverage for per-validator overflow:** A test where two slots map to the same address with costs that overflow the per-validator accumulator would strengthen coverage.

**Verdict:** Correct fix. The identified gap is mitigated by the execution layer guard.

x0152:
Closing - this is no longer an issue due to design changes in the devshard settlement path.

PR #1017, opened 2026-04-05
fix(keeper): add overflow guards in bitcoin supply-cap distribution loop

Mayveskii:
## Summary

`GetBitcoinSettleAmounts` proportional-reduction loop (approaching-supply-cap branch) had no overflow guard before `totalDistributed += reducedReward`. Additionally, `CalculateParticipantBitcoinRewards` had unguarded accumulation and underflow-prone subtraction in its `totalDistributed` logic.

### Root cause

**Supply-cap loop** (`GetBitcoinSettleAmounts`):

```go
// before fix — no overflow guard
reducedReward := uint64(decimal.NewFromUint64(amount.Settle.RewardCoins).
    Mul(remainingSupply).Div(originalDecimalAmount).IntPart())
amount.Settle.RewardCoins = reducedReward
totalDistributed += reducedReward   // ← no guard
```

`CalculateParticipantBitcoinRewards` has a defensive ceiling path (line 770):

```go
rewardCoins = ^uint64(0) // Max uint64 — when big.Int result does not fit uint64
```

If this `MaxUint64` value reaches `amount.Settle.RewardCoins` and then enters the proportional-reduction loop as `reducedReward`, `totalDistributed` silently wraps to near-zero, and `remainder = uint64(bitcoinResult.Amount) - totalDistributed` becomes enormous — effectively diverting all minted coins to governance while participants receive their unreduced amounts.

**Same pattern as PR #544 (25K GNK)**. Type incompatibility with the existing `saturatingAddUint64Max(int64, uint64)` helper makes it inapplicable to `uint64` accumulators.

**Gap 2 — `CalculateParticipantBitcoinRewards` itself**:
- `totalDistributed += rewardCoins` (line 772): same `MaxUint64` ceiling path triggers wrap
- `totalDistributed -= debt` (lines 781, 784): no underflow guard — subtraction can wrap if `totalDistributed < debt`

### Fix

**Supply-cap loop** — explicit `math.MaxUint64` guard before the addition:
```go
if reducedReward > 0 && totalDistributed > math.MaxUint64-reducedReward {
    amount.Settle.RewardCoins = 0
    overflowed = true
    continue
}
```
On overflow all remaining participants' rewards are zeroed (not just skipped), so no participant retains an un-reduced amount. The full remainder flows to governance.

**`CalculateParticipantBitcoinRewards`** — saturate via `bits.Add64` (already imported):
```go
_, overflow := bits.Add64(totalDistributed, rewardCoins, 0)
if overflow != 0 {
    totalDistributed = math.MaxUint64
} else {
    totalDistributed += rewardCoins
}
```

Underflow guards for both debt-recovery branches:
```go
if totalDistributed >= debt { totalDistributed -= debt } else { totalDistributed = 0 }
```

### Gap 1 — documented in comment

When the overflow guard fires on participant N, participants `0..N-1` have already received correctly reduced rewards. This is acceptable since the overflow path is theoretically unreachable with valid upstream data. A comment in the code explains why `break` is not used (would leave participants `N..end` with un-reduced `RewardCoins`).

## Test

`TestGetBitcoinSettleAmounts_SupplyCapOverflowGuard`:
- Exercises `GetBitcoinSettleAmounts` with the approaching-supply-cap branch active
- Verifies that the sum of all `RewardCoins` across participants does not exceed `bitcoinResult.Amount` (no overflow, no wrap)

## Severity

**MEDIUM** — mathematically unreachable with valid protocol data (upstream bounds prevent `big.Int.IsUint64() == false` in practice), but constitutes a defense-in-depth gap in the same class as the bounty-bearing PR #544.

### DIFF
--- a/inference-chain/x/inference/keeper/accountsettle.go
+++ b/inference-chain/x/inference/keeper/accountsettle.go
@@ -181,6 +181,15 @@ func (k *Keeper) SettleAccounts(ctx context.Context, currentEpochIndex uint64, p
 		return types.ErrNegativeRewardAmount
 	}
 	k.LogInfo("Bitcoin reward amount", types.Settle, "amount", bitcoinResult.Amount)
+	if bitcoinResult.SupplyCapOverflowed {
+		sdkCtx.EventManager().EmitEvent(sdk.NewEvent(
+			"bitcoin_supply_cap_overflow",
+			sdk.NewAttribute("epoch", fmt.Sprint(currentEpochIndex)),
+			sdk.NewAttribute("total_subsidy_paid", fmt.Sprint(settleParameters.TotalSubsidyPaid)),
+			sdk.NewAttribute("total_subsidy_supply", fmt.Sprint(settleParameters.TotalSubsidySupply)),
+			sdk.NewAttribute("reward_amount", fmt.Sprint(bitcoinResult.Amount)),
+		))
+	}
 	rewardAmount = bitcoinResult.Amount
 	governanceRewardAmount = bitcoinResult.GovernanceAmount
 
--- a/inference-chain/x/inference/keeper/bitcoin_rewards.go
+++ b/inference-chain/x/inference/keeper/bitcoin_rewards.go
@@ -111,11 +111,12 @@ func GetBitcoinSettleAmounts(
 					}
 					// This gives accurate response by not relying on a ratio before we need to
 					reducedReward := uint64(decimal.NewFromUint64(amount.Settle.RewardCoins).Mul(remainingSupply).Div(originalDecimalAmount).IntPart())
-					if reducedReward > 0 && totalDistributed > math.MaxUint64-reducedReward {
+					newTotal, overflow := checkedAddUint64(totalDistributed, reducedReward)
+					if overflow {
 						amount.Settle.RewardCoins = 0
 						overflowed = true
 						bitcoinResult.SupplyCapOverflowed = true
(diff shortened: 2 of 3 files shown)


Doog-bot534 (approved):
## Review: fix(keeper): add overflow guards in bitcoin supply-cap distribution loop

**Approve with suggestions** ✅

The most complex of the batch. The analysis is thorough and defense-in-depth framing is appropriate.

### Strengths
- The \`overflowed\` flag pattern in the supply-cap loop is correct: once overflow is detected, ALL remaining participants are zeroed, not just the triggering one.
- Using \`bits.Add64\` for saturation in \`CalculateParticipantBitcoinRewards\` is elegant.
- Underflow guards on \`totalDistributed -= debt\` address a real subtraction risk.

### Concerns

1. **Test does not exercise the overflow guard:** \`GetBitcoinSettleAmounts\` recalculates internally, so the manually injected \`MaxUint64\` is overwritten. The overflow path is not actually tested. To properly test, either (a) export the supply-cap reduction logic separately, or (b) mock \`CalculateParticipantBitcoinRewards\`.

2. **Silent zeroing:** The \`overflowed\` flag zeroes rewards but doesn't emit a log/event. For an event this anomalous, runtime observability matters.

3. **Saturation to MaxUint64:** This means \`remainder = fixedEpochReward - totalDistributed\` could underflow. Verify the downstream \`if remainder < 0\` guard handles \`totalDistributed > fixedEpochReward\` gracefully.

**Verdict:** Guards are well-placed. Recommend adding a unit test that directly exercises the overflow path.

PR #1035, opened 2026-04-08
fix(subnetctl): propagate fatal HTTP errors instead of waiting on timeout

### LINKED ISSUE #1019: subnetctl: inference error handling
## Summary

**Today `subnetctl` does not surface many host-side failures during inference.** When the HTTP call to the executor fails (for example **403 Forbidden** with `sender not in group`), the proxy treats that as “no response yet” and **waits through refusal / execution timeouts** before eventually failing with timeout-related errors (for example insufficient timeout votes). The original transport error is **not** returned to the API client immediately.

This behavior comes from `sendAndProcess` in:

- `subnet/cmd/subnetctl/proxy.go`

## What the code does today

Flow (simplified):

1. `SendOnly` → `transport.HTTPClient.Send` → on non-200 (including 403), the client returns **`(nil, error)`** (see `subnet/transport/client.go`).
2. `sendAndProcess` does:

   ```go
   resp, sendErr := p.session.SendOnly(ctx, prepared)
   if sendErr != nil && resp == nil {
       return false, 0, nil   // no error propagated to runInference
   }
   ```

3. `runInference` interprets that as “not finished, no receipt,” sleeps until **RefusalTimeout** or **ExecutionTimeout**, retries once, then may enter **timeout vote** collection.

So **403 / 401 / 5xx** from the host are **not** distinguished at the proxy layer today; send failures with `resp == nil` are **silent** until the timeout machinery runs.

## Why this is a problem

- **Misconfiguration** (wrong user key vs escrow `CreatorAddress`, wrong network, ACL bugs) surfaces as **minute-long waits** and confusing messages like “insufficient timeout votes,” instead of an immediate **4xx with a clear cause**.
- **Operators and clients** cannot distinguish “host is down” from “host rejected the request” without reading participant logs.

## Target behavior (to implement)

Errors from the executor (and related paths) should be split into two classes:

### 1. Non-retryable / fatal (interrupt the inference)

Fail **immediately** (or after a single idempotent sanity check) and return a clear error to the OpenAI-compatible client (e.g. HTTP 502/403 with message body).

Examples:

- **4xx from host** where retrying the same request will not help without changing client or chain state:
  - **403** — `sender not in group`, auth/signature mismatch, creator vs escrow mismatch.
  - **400** — malformed payload, validation errors.
- **401** if used for auth failures.

These should **not** block on RefusalTimeout / ExecutionTimeout.

### 2. Retryable (bounded retries, then fail)

Transient infrastructure or ordering issues:

- **Connection errors**, **timeouts**, **502/503/504** from a reverse proxy.
- **429** with backoff (if applicable).
- Brief **unavailability** where the same signed request might succeed shortly (optional: small retry budget with jitter).

Policy should be **explicit**: max attempts, backoff, and when to give up and return an error to the client.

#### After `MsgStartInference` succeeded, before `MsgFinishInference`

Once the session has **successfully advanced** so that **`MsgStartInference` is in effect** for this inference, but **`MsgFinishInference` has not yet appeared** in the mempool (executor still running or client is waiting on stream / follow-up sync):

- **5xx** and **network errors** on any later HTTP interaction with the same or peer hosts (status polls, catch-up, streaming chunk fetches, verifier traffic, etc.) should be treated as **retryable**, not as fatal client misconfiguration.
- Retries should stay within the existing **execution / refusal deadline** window where applicable, so the overall inference does not hang unbounded.

By contrast, a **403 / 401 / 400** on a request that clearly indicates **authorization or payload rejection** (e.g. first contact with wrong creator key) remains **non-retryable** unless the error semantics are known to be transient (rare).

### 3. Protocol continuation (keep current timeout path)

When the transport **succeeds** (HTTP 200) and the host returns a normal **`HostResponse`**, but the mempool still does not contain **`MsgFinishInference`** for this nonce, **`runInference`** should keep using **refusal / execution deadlines** and optional **timeout votes** as today.

That timeout path addresses **slow or stuck executors**, not **misclassified transport failures** — transport errors during this phase should still follow §2 (retryable 5xx / network) where possible before falling through to timeout.

## Implementation notes (for a future change)

- **Classify** `SendOnly` / `HTTPClient.Send` errors: parse status code when available (wrap errors with `errors.As` into a typed `HTTPStatusError` or similar).
- **In `sendAndProcess`**: on `sendErr != nil && resp == nil`, return **`(false, 0, sendErr)`** for fatal statuses, or implement a **retry loop** for retryable statuses before returning.
- **Phase-aware policy**: same status code may be **fatal** on the first “start inference” exchange and **retryable** once `MsgStartInference` is already committed for this nonce (see §2 above). Thread **inference phase** (pre-start vs post-start vs waiting-for-finish) into error handling.
- **Align** `subnet/cmd/subnetctl` and `subnet/testenv/cmd/subnetctl` (or share one `proxy` package to avoid drift).
- **Document** status mapping in this file once implemented.

## Related code

| Piece | Role |
|-------|------|
| `subnet/transport/client.go` | Returns error on `StatusCode != 200` |
| `subnet/cmd/subnetctl/proxy.go` → `sendAndProcess` | Swallows send failure when `resp == nil` |
| `subnet/user/user.go` → `SendOnly` | Thin wrapper over client `Send` |
| `subnet/testenv/cmd/subnetctl/proxy.go` | Same as production proxy |

unameisfine:
Closes #1019 (minimal fix — §1 of the issue).

## Problem

`sendAndProcess` in `subnet/cmd/subnetctl/proxy.go` swallowed any `SendOnly` error whose response was nil:

```go
resp, sendErr := p.session.SendOnly(ctx, prepared)
if sendErr != nil && resp == nil {
    return false, 0, nil   // fatal 403 is indistinguishable from a network blip
}
```

A **403 "sender not in group"** (the example from the issue — wrong creator key vs escrow ACL) was therefore indistinguishable from a transient network failure. `runInference` waited through `RefusalTimeout`, then `ExecutionTimeout`, and eventually returned a misleading `insufficient timeout votes` / `inference timed out: TIMEOUT_REASON_REFUSED` error.

Operators saw minute-long waits and confusing timeout messages instead of an immediate 4xx with the real cause.

## Fix

Classify HTTP transport errors by status code and fail fast on fatal 4xx responses. Retryable errors (5xx, 429, network) keep the existing deadline-based retry path so a transient blip still recovers.

### Scope

This PR implements **§1** of the issue (non-retryable fatal errors). The broader phase-aware retry policy (§2 post-start retries, §3 protocol continuation) is intentionally left for a follow-up — it is a larger design change that benefits from being reviewed on its own.

## Changes

- **`subnet/transport/errors.go`** (new): `HTTPStatusError` typed error with `StatusCode` / `Path` / `Body` and `IsFatal()` / `IsFatalHTTPError()` helpers. Fatal = 4xx except 429 (Too Many Requests is explicitly retryable).
- **`subnet/transport/client.go`**: `doPostRaw` returns `*HTTPStatusError` on non-200 instead of a plain `fmt.Errorf`, so callers can classify via `errors.As`.
- **`subnet/cmd/subnetctl/proxy.go`**: `sendAndProcess` now propagates fatal HTTP errors immediately with `host rejected inference: ...` wrapping. Retryable errors still return `(false, 0, nil)` to preserve the existing behavior.
- **Tests**:
  - `transport/errors_test.go`: unit tests for the fatal/retryable classification table and `errors.As` unwrapping.
  - `cmd/subnetctl/proxy_test.go`:
    - `killableClient` gains a `KillFatal()` method that makes `Send` return `*transport.HTTPStatusError{StatusCode: 403}`.
    - `TestRunInference_FatalHTTPErrorReturnsImmediately` is the red-green regression test for #1019.

## Red-green proof

**Without the fix** (stashed `proxy.go` change, ran the new test):

```
Error: "inference 1 timed out: TIMEOUT_REASON_REFUSED" does not contain "host rejected inference"
--- FAIL: TestRunInference_FatalHTTPErrorReturnsImmediately (1.00s)
```

The 1.00s is the `RefusalTimeout` the old code burned through before returning the misleading timeout error.

**With the fix**:

```
--- PASS: TestRunInference_FatalHTTPErrorReturnsImmediately (0.00s)
```

The test asserts:
- Error contains `"host rejected inference"` and `"403"`, and **not** `"timed out"` / `"insufficient votes"`.
- `transport.IsFatalHTTPError(err)` is true (error chain preserves the typed cause).
- `elapsed < 500ms` (would be 2s+ on the old path).
- Inference stays `StatusPending` — the timeout machinery is correctly bypassed.

## Test plan

- `go test ./cmd/subnetctl/ ./transport/` (all pass locally)
- Existing `TestRunInference_RefusalTimeout` unchanged — retryable errors still fall through to the timeout path.

### DIFF
--- a/subnet/cmd/subnetctl/proxy.go
+++ b/subnet/cmd/subnetctl/proxy.go
@@ -14,6 +14,7 @@ import (
 
 	"subnet/host"
 	"subnet/state"
+	"subnet/transport"
 	"subnet/types"
 	"subnet/user"
 )
@@ -190,9 +191,18 @@ func (p *Proxy) runInference(ctx context.Context, params user.InferenceParams, w
 // sendAndProcess sends the prepared inference and processes the response.
 // Returns finished=true when MsgFinishInference is in the host's mempool.
 // confirmedAt is the executor's receipt timestamp (0 if no receipt received).
+//
+// Fatal HTTP errors from the host (4xx other than 429) are propagated
+// immediately so the caller fails fast with a clear cause instead of
+// waiting through refusal/execution timeouts. Retryable errors (5xx,
+// 429, network failures) keep the existing behavior: no receipt is
+// returned and runInference falls through to its deadline-based retry.
 func (p *Proxy) sendAndProcess(ctx context.Context, prepared *user.PreparedInference, nonce uint64) (finished bool, confirmedAt int64, err error) {
 	resp, sendErr := p.session.SendOnly(ctx, prepared)
 	if sendErr != nil && resp == nil {
+		if transport.IsFatalHTTPError(sendErr) {
+			return false, 0, fmt.Errorf("host rejected inference: %w", sendErr)
+		}
 		return false, 0, nil
 	}
 
--- a/subnet/cmd/subnetctl/proxy_test.go
+++ b/subnet/cmd/subnetctl/proxy_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net/http"
 	"net/http/httptest"
 	"sync/atomic"
 	"testing"
@@ -18,6 +19,7 @@ import (
 	"subnet/signing"
 	"subnet/state"
 	"subnet/stub"
+	"subnet/transport"
 	"subnet/types"
 	"subnet/user"
 )
@@ -81,20 +83,31 @@ func TestHasMsgFinish(t *testing.T) {
 // --- Test infrastructure for proxy-level tests ---
 
 // killableClient wraps a HostClient. Kill/Revive toggle availability.
+// KillFatal simulates a non-retryable HTTP 4xx response from the host
+// (for example a 403 "sender not in group" from the auth layer).
 type killableClient struct {
 	inner  user.HostClient
 	killed atomic.Bool
+	fatal  atomic.Bool
 }
 
 func (c *killableClient) Send(ctx context.Context, req host.HostRequest) (*host.HostResponse, error) {
+	if c.fatal.Load() {
+		return nil, &transport.HTTPStatusError{
+			StatusCode: http.StatusForbidden,
+			Path:       "/sessions/escrow-proxy/chat/completions",
+			Body:       "sender not in group",
+		}
+	}
 	if c.killed.Load() {
 		return nil, fmt.Errorf("host killed")
 	}
 	return c.inner.Send(ctx, req)
 }
 
-func (c *killableClient) Kill()   { c.killed.Store(true) }
-func (c *killableClient) Revive() { c.killed.Store(false) }
+func (c *killableClient) Kill()      { c.killed.Store(true) }
+func (c *killableClient) Revive()    { c.killed.Store(false) }
+func (c *killableClient) KillFatal() { c.fatal.Store(true) }
 
 // verifierClient wraps killableClient and implements user.TimeoutVerifier.
 // This allows session.TimeoutVerifiers() to discover it.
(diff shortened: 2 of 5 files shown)


Doog-bot534 (approved):
## Review: fix(subnetctl): propagate fatal HTTP errors instead of waiting on timeout

**Strong Approve** ✅

The cleanest PR of the batch. Well-scoped, well-tested, and the abstraction (\`HTTPStatusError\` + \`IsFatal()\`) is reusable and properly layered.

### Strengths
- Proper structured error with \`errors.As\` support. The \`IsFatal()\` classification (4xx fatal except 429) matches HTTP semantics.
- Backward-compatible change — existing \`err.Error()\` callers see the same string format.
- Excellent test: asserts error message content, error chain unwrapping, timing (<500ms vs 2s+), AND state consistency (inference stays Pending).

### Minor suggestions

1. **Response body truncation:** \`doPostRaw\` reads the entire response body into the error via \`io.ReadAll\`. A malicious host could return a multi-MB error body. Consider truncating \`Body\` to ~4KB in \`HTTPStatusError\`.

2. **HTTP 408 Request Timeout:** Currently classified as fatal (4xx). It's technically retryable. Edge case — hosts probably don't return 408 — but worth considering for completeness.

**Verdict:** Ship it. Clean design, strong tests, well-scoped change.

gmorgachev:
I'd recommend focus on whole task / bigger issues. The overhead on review the atomic changes is too big in current dev flow

unameisfine:
Makes sense — will focus on bigger scoped work going forward. Keeping this open since it addresses a specific production scenario from #1019 (akup), but happy to close if the preference is to fold it into a larger change.

tcharchian:
This PR and https://github.com/gonka-ai/gonka/pull/1211 address the same underlying problem. They were combined, and one of the smaller features from them was included in this PR: https://github.com/gonka-ai/gonka/pull/1258
That said, I would still like to propose to the community that your contribution is also rewarded. Please email me tania.charchian @ productscience.ai, and we can discuss the process.

tcharchian:
After review, it turned out that this issue had already been solved, and in a more proper way, in the gateway that the other contributors were working on in parallel. So these changes are effectively dead and no longer relevant. We even had to clean them up. @a-kuprin, could you please share the link to the actual fix?

a-kuprin:
I think it's here:
https://github.com/gonka-ai/gonka/pull/1284

Gateway is the thing that handles errors and redirects user inference. It's purpose to make it stable, not return every failure to user

PR #1038, opened 2026-04-09
Versiond follow-up

DimaOrekhovPS:


deploy/join/docker-compose.yml
```diff
@@ -74,6 +76,7 @@ services:
     ports:
       - "9100:9100"
       - "127.0.0.1:9200:9200"
+      - "${NODE_MANAGER_GRPC_PORT:-9400}:${NODE_MANAGER_GRPC_PORT:-9400}"
```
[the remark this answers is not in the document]
DimaOrekhovPS: Ok, we're not exposing it at all. Only available inside the docker network

deploy/join/docker-compose.versiond.yml
```diff
@@ -0,0 +1,23 @@
...
+      - NODE_MANAGER_ADDR=api:9400
+      # Override a specific version binary for development:
+      # - VERSIOND_OVERRIDE_v0_2_11=/opt/local/subnet-dev
+      # Force versions not yet approved by governance:
+      # - VERSIOND_FORCE=v0.2.13-rc1
+    volumes:
+      - versiond-bin:/opt/versiond/bin
+      - versiond-data:/opt/versiond/data
+    ports:
+      - "${PORT:-8080}:8080"
```
DimaOrekhovPS, line 16: Conflicting port w proxy. Needs fixing

0xgonka (commented):
Follow-up to the initial versiond landing (#1003). Most of the 765-line diff is generated protobuf — actual hand-written changes are small: a `versiond-release` Make target, new compose file for production, testapp e2e fixture wiring node-manager gRPC, and minor `download.go` error-handling fixes. Versiond is the on-host version multiplexer that downloads/runs governance-approved subnet binaries and fronts them behind one HTTP port.

**Should fix:**

- **`PORT` env var collision risk in `deploy/join/docker-compose.versiond.yml:16`.** The new file uses `${PORT:-8080}:8080`, but `PORT` is an extremely common shell env var (Heroku conventions, Node.js defaults, CI runners). If an operator has `PORT` exported for any reason it silently overrides versiond's host port. The sister file `local-test-net/docker-compose.versiond.yml:16` already uses the more specific `VERSIOND_PORT` — rename the production file to match for consistency.

- **`docker-compose.versiond.yml:16` binds to all interfaces by default** (`"${PORT:-8080}:8080"` resolves to `0.0.0.0:8080`). Versiond fronts experimental/test subnet binaries — exposing it on the public internet by default means any subnet binary becomes reachable the moment governance approves it. Compare to `deploy/join/docker-compose.yml:78` where the admin port is correctly bound to `127.0.0.1:9200:9200`. Recommend `"127.0.0.1:${VERSIOND_PORT:-8080}:8080"` and require explicit opt-in for public exposure.

**Nits:**

- `deploy/join/docker-compose.versiond.yml:4` pins `versiond:0.2.11` by tag but not by `@sha256:...` digest — `bridge` in the same compose file uses digest pinning for supply-chain consistency.
- `deploy/join/docker-compose.versiond.yml:3` hardcodes `container_name: versiond` — won't prefix with `${KEY_NAME}` like other services, so multi-deployment hosts will collide.
- `versioned/e2e/testapp/main.go:56` uses `grpc.WithTransportCredentials(insecure.NewCredentials())`. Fine because this is the test fixture only, but worth a comment noting that real subnet binaries should use authenticated gRPC to the node manager so this code never gets copy-pasted into production.
- `testermint/src/test/kotlin/VersiondTests.kt:128-161` is a smoke test for "the wiring exists" but does NOT verify successful `AcquireMLNode` end-to-end — if some regression caused acquisition to always fail with a bogus error, the test would still pass (the assertion is conditional). Consider seeding a registered ML node in the fixture so the success path is exercised, plus a regression guard that `api:9400` isn't reachable from the host.
- Pre-existing `versioned/internal/download/download.go` hardening gaps adjacent to this PR's changes (worth a follow-up issue, not blocking):
  - No URL scheme allow-list — oracle could theoretically hand back `http://169.254.169.254/...` (SSRF to metadata endpoint)
  - No `io.LimitReader` on download size — zip-bomb risk if oracle is compromised
  - Default `http.Client` allows unrestricted redirects

**Looks correct:**

- **`download.go` error handling improvements** (both hunks at lines 59-61 and 91-94) are strict improvements — surface `Close()` errors so a partially flushed temp file whose sha256 still matches the in-memory TeeReader can't slip through. Plus cleanup `os.Remove(tmpPath)` on failure. Good.
- **`extractBinary` is not vulnerable to zip-slip** — uses `filepath.Base(f.Name)` and only writes via `AtomicWriteFile` into `destDir`, so a crafted entry name like `../../etc/cron.d/evil` is reduced to `evil` and only the entry whose basename matches `binaryName` is extracted.
- **Port 9400 (node-manager gRPC) is intentionally NOT in `deploy/join/docker-compose.yml`'s `ports:` block** — only reachable from inside the docker network. The commit message "Don't expose port 9400" confirms this was deliberate. Good.
- **`NODE_MANAGER_ADDR=api:9400` in versiond compose** correctly references the docker-network DNS name, not a host port.
- **New gRPC/protobuf dependencies are canonical** (`google.golang.org/grpc v1.80.0`, `google.golang.org/protobuf v1.36.11`) — no typosquatting or personal forks.
- **`Dockerfile.testapp-server` build context widening** is required because `main.go` now imports `pb "versioned/e2e/testapp/gen"`. Build still uses repo root as docker context (`Makefile:46`).
- **testapp's `/nodemanager-test` handler is only built into `local-test-net/Dockerfile.testapp-server`**, not the production `versioned/Dockerfile` (which builds `./cmd/versiond` only). Production versiond never bundles the e2e testapp.
- **`versiond-release` Make target mirrors other component release flows** — consistent with how `decentralized-api`, `inference-chain`, etc. ship.
- Testermint test correctly bumps `@Order(3)` → `@Order(4)` and uses the new ordering.

deploy/join/docker-compose.versiond.yml
gmorgachev: why expose? need to be integrated to proxy

gmorgachev:
@DimaOrekhovPS please stop expose port and let's merge as soon as you can

0xgonka (approved):

PR #1039, opened 2026-04-09
Multi Model POC

### LINKED ISSUE #728: [P0] Multimodel support
A detailed description will be provided by @gmorgachev later on.
https://github.com/gonka-ai/gonka/tree/gm/multi-models/proposals/multi-model-poc

gmorgachev:
## Motivation

The chain must support multiple models. Currently it can't because PoC is single-model.

PoC starts immediately for a security reason: it proves hardware is present and running the model right now. Confirmation PoC can interrupt inference at any point and must run on whatever hardware is already serving that model, so there is no window to swap models in between. If there were, an attacker could spin up hardware only for the PoC phase and shut it down afterward. That is why the current chain forces a single PoC model and has multi-model support disabled.

To support multiple models, this branch runs PoC for each model independently in separate model groups, starting immediately just like today. The security property stays the same: >= 2/3 of total network consensus weight must validate every host's PoC result. Since only hosts serving model X can directly validate model X, the branch introduces delegation so non-members can pass their validation power to a member of that group. Without delegation, any model group whose direct members hold less than 2/3 of total weight would never reach the acceptance threshold.

Consensus weight also determines bitcoin-style reward distribution, so the aggregation rule (how per-model PoC weight converts to chain-wide consensus weight) directly affects incentives. Inside each group, raw per-model PoC weight (before aggregation) drives inference routing and inference rewards.

## Solution

Each governance-approved model gets its own PoC group. PoC runs concurrently across all eligible groups within the same epoch. Each entry in `PocParams.models[]` holds `model_id`, `seq_len`, `stat_test`, `weight_scale_factor`, and `penalty_start_epoch`.

**Two separate weights.** `pocWeight(group_i, p)` is model-local: the number of nonces host `p` computed and got validated in group `i`. It drives inference routing inside the group. `consensusWeight(p) = sum(weight_scale_factor_i * pocWeight(group_i, p))` is the aggregated chain-wide weight that determines block signing power, governance voting power, PoC validation power, and bitcoin-style reward distribution.
**Eligibility.** A group is eligible for consensus weight when:
1. Governance-approved with a positive coefficient.
2. Members' consensus weight from the prior epoch's root `EpochGroupData` sums to >= `W_threshold` fraction of total prior-epoch network weight. Mid-epoch member removal is reflected: weight of a participant removed from the SDK group during epoch N-1 does not count toward N's eligibility.
3. At least `V_min` members with positive prior-epoch consensus weight (same source) pass PoC.

**Group cap.** An additional layer of protection in case validation within a group is compromised. Consensus weight from any non-initial group is capped at `cap_factor * sum(members' previous consensus weight from other eligible groups)`. If the raw total exceeds the cap, all member contributions in that group are scaled down proportionally. With `cap_factor = 1`, a group can contribute at most as much weight as its members already proved in other groups, which prevents any single non-initial group from exceeding ~50% of total network weight. The initial group is exempt. Delegation affects voting power but not the cap.

**Validation rule.** Host `p`'s PoC result in group `i` is accepted when `sum(votingPower of approvers) / totalNetworkWeight > 2/3`. If neither valid nor invalid votes reach 2/3, the guardian tiebreak rule from pre-multi-model PoC applies: the decision passes only if every voting guardian agrees unanimously. Hosts not in the group and not delegating effectively abstain, which counts against acceptance.

**Delegation.** A host not serving group `i` can delegate its consensus weight to a member of `i`, who then votes on the delegator's behalf. A group member's own delegation record for that same group is ignored: DIRECT membership takes precedence. Delegation is per-group, one target per `(model_id, delegator)`. Split delegation is not supported. Changes take effect from the next epoch.

**Participation modes.** For each governance-approved group, every host with consensus weight resolves to one of five modes at epoch formation.

1. DIRECT. The host submitted a PoC store commit for the model this epoch. DIRECT is not a transaction; the chain derives it from the commit's presence.
2. DELEGATE. The host stored a `MsgSetPoCDelegation` record targeting a group member. `delegation_share` of its weight transfers to the target.
3. REFUSE. The host stored a `MsgRefusePoCDelegation` record and pays `refusal_penalty`.
4. INTENT. The host stored a `MsgDeclarePoCIntent` record. Effective only for models not yet active; ignored for models that already have voting power in the current epoch. Its purpose is to surface bootstrap pre-eligibility before hosts commit hardware.
5. NONE. The host did none of the above and pays `no_participation_penalty`.

DIRECT is not exclusive with the tx-driven records. If a host stores a delegation record for group `i` and also submits a PoC store commit for `i`, it resolves as DIRECT and the delegation record is ignored. The three tx-driven records (DELEGATE, REFUSE, INTENT) are mutually exclusive per `(model_id, participant)` with last-write-wins; sending any one of them clears the other two.

Participation is not enforced at the tx layer. Nothing forces a host to choose; hosts that submit no commit and store no tx record resolve as NONE. Each model has a `penalty_start_epoch`, and before that epoch all penalties for that model are skipped, giving hosts time to prepare.

**Bootstrap period for new models.** When a governance-approved model is not yet active, the chain runs a bootstrap flow before PoC starts:
1. At `start_poc - deploy_window`, the chain snapshots INTENT and DELEGATE state for bootstrap candidate models.
2. It evaluates advisory pre-eligibility: governance approval, `W_threshold`, `V_min`, and whether >2/3 of total weight is reachable (direct intent + delegations). Results are emitted as events so operators can see if the model looks viable before committing hardware.
3. Between the snapshot and PoC start, hosts with INTENT deploy their hardware.
4. At validation start, DIRECT membership is determined by who actually submitted a store commit, not by who declared intent. A host that declared intent but did not submit PoC is not DIRECT.

Pre-eligibility is advisory. A group that fails pre-eligibility can still become eligible if enough hosts independently participate in PoC for it.

**Model lifecycle.**
1. Governance proposal approves model with coefficient, group created.
2. Pre-penalty phase: PoC runs but penalties disabled until `penalty_start_epoch`.
3. Penalty phase: full participation rules apply.

A group may or may not be eligible in any given epoch depending on whether it meets the conditions above.

### ActiveParticipant and EpochGroupData

`ActiveParticipant` now stores per-model delegation-resolved voting powers in a new `repeated ModelVotingPower voting_powers` field. `ActiveParticipants` is written once at epoch formation and stays immutable for the duration of that epoch. Because it cannot reflect mid-epoch changes, the same data is also propagated to `EpochGroupData`, which does reflect mid-epoch changes like member removal.

`EpochGroupData` is the mutable per-epoch state. A root group (`model_id = ""`) stores all members with their consensus weight. Each model subgroup (`model_id = "llama3"`) stores only members serving that model, with model-local fields:
- `ValidationWeight.weight` = sum of raw `PocWeight` for that model (no coefficient).
- `ValidationWeight.ml_nodes` = that model's node info.
- `ValidationWeight.voting_power` = delegation-resolved voting power for that model.

Root group entries have `voting_power = 0` and `ml_nodes = nil`.

Both stores exist because they serve different purposes. `ActiveParticipants` is the permanent historical record of each epoch's active set. `EpochGroupData` reflects the live state: when a host is invalidated mid-epoch, it is removed from the group, and `GetGroupMembers` excludes it from voting power snapshots and consensus weight calculations.

### New on-chain collections and upgrade

PoC commits, weight distributions, and validations gained `model_id` in their storage key. Adding a key component changes the codec, so the old KV prefixes (38, 39, 40) cannot be decoded with the new layout. New collections live under prefixes 58, 59, 60; no post-upgrade reader touches 38/39/40.

The three top-level PoC v2 messages batch all of a host's per-model state into a single transaction per PoC stage. `MsgPoCV2StoreCommit.entries[]` carries one `PoCV2CommitEntry{model_id, count, root_hash}` per model. `MsgMLNodeWeightDistribution.entries[]` carries one `MLNodeDistributionEntry{model_id, weights[]}` per model. `MsgSubmitPocValidationsV2.validations[]` carries one `PoCValidationEntryV2{participant, weight, model_id}` per (validated participant, model).

The `v0.2.12` upgrade handler:
1. Clears all entries under legacy prefixes 38/39/40 with raw store iteration.
2. Migrates singular `PocParams` fields (`model_id`, `seq_len`, `stat_test`, `weight_scale_factor`) into `PocParams.models[]` and sets `penalty_start_epoch = 0` on the resulting entry.
3. Initializes `DelegationParams` with zero-valued defaults (`w_threshold`, `v_min`, `cap_factor`, and all three penalty fractions are 0) and sets `initial_model_id` to the founding model.
4. Backfills `ActiveParticipant.VotingPowers` and subgroup `ValidationWeight.voting_power` for the current epoch (required because pre-upgrade data has these fields at zero).
5. Seeds new pruning state markers so the pruner does not walk empty historical ranges.

### DAPI artifact storage

DAPI stores PoC artifacts in local MMR-backed stores, one per `(stage, model_id)`. The directory layout is `<base>/<stage>/<url-encoded model_id>/`. Multiple model stores under the same stage are accessed concurrently during generation. Proof requests, proof signatures, callback routes (`/v2/poc-batches/:model_id/generated`, `/v2/poc-batches/:model_id/validated`), and artifact-state queries all include `model_id`.

### Delegation snapshots

Two separate snapshots, captured at different times, for different purposes:

**BootstrapDelegationSnapshot** is captured at `start_poc - deploy_window`. It stores delegations and intents for approved models that are not yet active. The chain uses it to evaluate advisory pre-eligibility and, later at validation start, to build voting powers for bootstrap models. It is captured early so operators see pre-eligibility events before committing hardware. The `deploy_window` is the runway for hosts that declared INTENT to provision MLNodes for the new model (on new hardware or on existing hardware) in time for PoC start.

**DelegationSnapshot** is captured at `poc_validation_start`. It stores delegations and refusals (no intents) for all approved models. The chain uses it at epoch formation to resolve participation modes (DIRECT/DELEGATE/REFUSE/NONE) and compute next-epoch voting powers.

For existing active models, validation-time voting powers come from `EpochGroupData` subgroups (already delegation-resolved at previous epoch formation). For bootstrap models, voting powers are computed fresh from consensus weights + bootstrap snapshot + actual store commits.

### Model resolution

Broker resolves which model a node generates PoC for via `resolvePoCModelForNode`:
1. If `EpochMLNodes` has exactly one entry, use that model.
2. If `EpochMLNodes` has multiple entries, skip (ambiguous).
3. If `EpochMLNodes` is empty (fresh node, no epoch assignment yet), pick the first governance-approved model from the node's configured models, alphabetically sorted. Sort is only for determinism; the rest of the pipeline still assumes one configured model per node, so this path is meaningful only in the single-model case.

### Slot sampling

In slot mode, the system computes how many of the global `validation_slots` belong to a model: `floor(modelVotingPower / totalNetworkWeight * validation_slots)`. The remaining global slots stay empty and behave as abstention. Approval still requires `>2/3` of the full global slot count. Slot assignment uses `votingPower` (delegation-resolved). When slot mode is disabled, approval uses model-local delegated voting power against total network weight.

### Weight formation

Two calculators handle different concerns. `PoCWeightCalculator` validates individual PoC results (>2/3 acceptance threshold, slot sampling, guardian tiebreak) and produces raw per-model `pocWeight`. It has no knowledge of cross-model aggregation, delegation economics, or eligibility. `DelegationWeightCalculator` operates after model assignment: it determines eligible groups, applies group caps, computes aggregated `consensusWeight`, resolves participation modes, and computes per-model voting powers. This separation keeps PoC validation independent from the multi-model policy layer.

The epoch formation pipeline in `onEndOfPoCValidationStage` (verified against code):

```
1. ComputeNewWeights
   PoCWeightCalculator validates PoC results per (participant, model).
   Preserved nodes keep their previous model bucket.
   Merge preserved + fresh PoC nodes per model.
   Output: activeParticipants with per-model MlNodes and raw pocWeight.

2. setModelsForParticipants
   Assigns each ML node to exactly one governance model.

3. DelegationWeightCalculator
   - EligibleGroups: filter by governance approval + W_threshold + V_min
   - ResolveGroupParticipation: DIRECT/DELEGATE/REFUSE/NONE per model
   - ComputeConsensusWeights: aggregate with coefficients, apply group caps

4. Delegation + bootstrap penalty accumulation
   Penalties from REFUSE, NONE, and bootstrap modes sum across models,
   capped at 1.0. DELEGATE transfers delegation_share to the target.
   Applied directly to participant weight.

5. AdjustWeightsByCollateral

6. ApplyEpochPowerCapping

7. computeAndSetVotingPowers
   Uses final post-adjustment weights + DelegationSnapshot + DIRECT membership.
   Writes to ActiveParticipant.VotingPowers.

8. AllocateMLNodesForPoC
   Per model, selects a fraction of nodes and sets POC_SLOT=true.
   These nodes continue serving inference during PoC.

9. SetActiveParticipants (immutable from this point)

10. addEpochMembers
    Propagates to EpochGroupData root group and model subgroups.
    addToModelGroups writes subgroup weight = sum of raw PocWeight (no coefficient).
    Subgroup ValidationWeight.voting_power = model-local delegation-resolved power.

11. Cleanup: delete consumed PoCRefusal and PoCDirectIntent entries.
```

The initial model is exempt from the group cap through `initial_model_id`. Confirmation PoC reuses the same model-aware validation snapshot and weight-calculation path.

## Open questions

Current implementation applies delegation penalties and delegation share directly to participant consensus weight. Should it affect only bitcoin-style rewards instead?

Each MLNode serves one model at a time. Fresh nodes with no epoch assignment resolve their model from `node.Models[0]` (first configured governance model). This is correct when each node has one model configured, but needs a cleaner resolution path if a node ever declares support for multiple models.

## TODO

- [ ] Cap for max delegation to host
- [ ] The upgrade handler leaves `DelegationParams` and every `penalty_start_epoch` at zero. Before release, set concrete values for `w_threshold`, `v_min`, `cap_factor`, the three penalty fractions, and per-model `penalty_start_epoch`.

### In the future

- [ ] The unregistered-model phase from `README.md` is not implemented. All PoC paths require the model to be governance-approved. If the unregistered-model flow is still required, it needs a separate implementation.


inference-chain/proto/inference/inference/params.proto
```diff
@@ -329,6 +340,19 @@ message TransferAgentAccessParams {
   repeated string allowed_transfer_addresses = 1;
 }
 
+// DelegationParams defines parameters for PoC delegation in the multi-model system.
+message DelegationParams {
+  option (gogoproto.equal) = true;
+  int64   deploy_window             = 1;  // blocks before start_poc for delegation snapshot
```
patimen: Formatted differently from other proto files.

inference-chain/x/inference/keeper/accountsettle.go
```diff
@@ -63,29 +64,30 @@ func CheckAndPunishForDowntime(total, missed, reward uint64, p0 *types.Decimal)
...
 
-// AggregateMLNodesFromModelSubgroups builds a map of participant addresses to their aggregated MLNodes
-// by collecting MLNode data from all model-specific EpochGroup subgroups for the given epoch.
-func (k *Keeper) AggregateMLNodesFromModelSubgroups(ctx context.Context, epochIndex uint64, validationWeights []*types.ValidationWeight) map[string][]*types.MLNodeInfo {
-	participantMLNodes := make(map[string][]*types.MLNodeInfo)
+// AggregateMLNodesFromModelSubgroups builds a map of participant addresses to their
+// per-model MLNodes by collecting MLNode data from all model-specific subgroups.
+// Model identity is preserved so callers can apply per-model coefficients.
+func (k *Keeper) AggregateMLNodesFromModelSubgroups(ctx context.Context, epochIndex uint64, validationWeights []*types.ValidationWeight) map[string]map[string][]*types.MLNodeInfo {
+	participantMLNodes := make(map[string]map[string][]*types.MLNodeInfo)
```
patimen, line 71: I know it's not caused directly by this PR, but given the sheer number of EpochGroups we should range this iteration to only matching by EpochIndex. They're already indexed that way, it should be fine.

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -41,126 +41,170 @@ func CalculateTimeNormalizationFactor(
...
 
-func (wc *WeightCalculator) getSortedParticipantKeys() []string {
-	var sortedKeys []string
+func (wc *PoCWeightCalculator) getSortedParticipantModelKeys() []types.PoCParticipantModelKey {
+	var sortedKeys []types.PoCParticipantModelKey
 	for key := range wc.StoreCommits {
 		sortedKeys = append(sortedKeys, key)
 	}
-	sort.Strings(sortedKeys)
+	sort.Slice(sortedKeys, func(i, j int) bool {
```
patimen: Prefer: 
```
slices.SortFunc(sortedKeys, func(a, b types.PoCParticipantModelKey) int {
    return cmp.Or(
        cmp.Compare(a.ParticipantAddress, b.ParticipantAddress),
        cmp.Compare(a.ModelID, b.ModelID),
    )
})
```

patimen: Clearer, more performant and safer

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -836,20 +885,27 @@ func (am AppModule) ComputeNewWeights(ctx context.Context, upcomingEpoch types.E
...
+	allowedCommits := make(map[types.PoCParticipantModelKey]types.PoCV2StoreCommit)
+	allowedDistributions := make(map[types.PoCParticipantModelKey]types.MLNodeWeightDistribution)
 
-	var sortedCommitKeys []string
+	var sortedCommitKeys []types.PoCParticipantModelKey
 	for key := range storeCommits {
 		sortedCommitKeys = append(sortedCommitKeys, key)
 	}
-	sort.Strings(sortedCommitKeys)
+	sort.Slice(sortedCommitKeys, func(i, j int) bool {
```
patimen: Prefer:
```
slices.SortFunc(sortedCommitKeys, func(a, b types.PoCParticipantModelKey) int {
    return cmp.Or(
        cmp.Compare(a.ParticipantAddress, b.ParticipantAddress),
        cmp.Compare(a.ModelID, b.ModelID),
    )
})
```

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -41,126 +41,170 @@ func CalculateTimeNormalizationFactor(
...
+			"participant", activeParticipant.Index,
+			"modelId", key.ModelID,
+			"weight", activeParticipant.Weight)
+	}
+
+	var participantAddresses []string
+	for participantAddress := range activeParticipantsMap {
+		participantAddresses = append(participantAddresses, participantAddress)
 	}
+	sort.Strings(participantAddresses)
```
patimen: `slices.Sort(participantAddress)`

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -536,64 +563,65 @@ func (am AppModule) GetPreviousEpochMLNodesWithInferenceAllocation(ctx context.C
...
-		for _, mlNode := range inferenceMLNodes {
-			if mlNode.NodeId == "" {
-				continue
+
+		// Sort model IDs for deterministic order
+		sortedModelIds := make([]string, 0, len(modelBuckets))
+		for modelId := range modelBuckets {
+			sortedModelIds = append(sortedModelIds, modelId)
+		}
+		sort.Strings(sortedModelIds)
```
patimen: `slices.Sort(sortedModelIds)`

inference-chain/x/inference/module/chainvalidation.go
```diff
@@ -661,47 +701,53 @@ func findParticipantByAddress(participants []*types.ActiveParticipant, address s
...
-		if mlNode.NodeId == "" {
-			continue
-		}
-		filteredMergedMLNodes = append(filteredMergedMLNodes, mlNode)
+	// Build sorted output
+	sortedModels := make([]string, 0, len(modelNodes))
+	for modelId := range modelNodes {
+		sortedModels = append(sortedModels, modelId)
 	}
+	sort.Strings(sortedModels)
```
patimen: `slices.Sort(sortedModels)`

inference-chain/x/inference/keeper/legacy_poc_v2_cleanup.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// ClearLegacyPoCv2Data removes all raw KV entries under the legacy PoC v2
+// prefixes (38, 39, 40). These collections changed key codec in v0.2.12 --
+// model_id was added to the key -- and were moved to new prefixes (58, 59, 60).
+// The old entries cannot be decoded with the new codec, so collections.Map.Clear
+// on the new collections is insufficient. This function uses raw store iteration
+// to delete the old bytes.
+func (k Keeper) ClearLegacyPoCv2Data(ctx context.Context) error {
+	store := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
```
patimen, line 18: OK, I see what you did here. I think we might be better off leaving the old collection values and clearing them with Clear? This feels like we've mixed in two different ways of dealing with the KVStore when we should only have 1.

patimen (commented):
Some initial comments for review

0xgonka (commented):
Major architectural change adding multi-model PoC with per-group consensus weight aggregation, delegation, eligibility checks, and penalty system. The design is coherent and the implementation is largely faithful to the spec, but there are several critical issues — most stemming from the all-zero default `DelegationParams`. Cannot ship as-is.

**Must fix:**

- **CRITICAL: `DefaultDelegationParams()` returns all zeros, silently disabling every multi-model safety check** (`x/inference/types/params.go:367-378` + `app/upgrades/v0_2_12/upgrades.go:160-166`). With `WThreshold=0`, `VMin=0`, `CapFactor=0`, and zero penalty fractions:
  - `MeetsWeightThreshold` always returns true (`delegation_weight_calculator.go:72`)
  - `MeetsMinHosts` always returns true (`:89`)
  - `ComputeGroupCap` returns -1 (uncapped) for **every** group, not just initial (`:247`)
  - All penalty accumulators are no-ops via `IsNoOp()`
  
  Net effect: any governance-approved model with a positive coefficient and any member is instantly eligible, uncapped, and unpenalized. The PR's TODO acknowledges this but defers to a post-upgrade governance proposal — that leaves a window where an attacker registers a model and grabs uncapped consensus weight before the proposal lands. **Must set non-zero defaults in the upgrade handler itself, or fail the upgrade if delegation params are zero.**

- **CRITICAL: `ComputeGroupCap` treats `CapFactor.IsZero()` as "uncapped" rather than "fully capped"** (`delegation_weight_calculator.go:247-249`). The spec says `cap = cap_factor * sum(other-group weight)` — if cap_factor=0 then cap=0 (no contribution). The current code inverts this. Combined with the zero default above, the safest "I haven't configured this" value produces the least safe behavior. Even outside the upgrade scenario, a governance proposal that mistakenly sets `cap_factor=0` silently uncaps every non-initial group. **Remove the `IsZero()` short-circuit; uncapped should be expressed via a large value, not zero.**

- **CRITICAL: `MsgPoCV2StoreCommit` and `MsgMLNodeWeightDistribution` use `NoPermission` with no blocklist gating** (`x/inference/keeper/permissions.go:110-111`, `msg_server_poc_v2_commit.go:16,126`). `SubmitPocValidationsV2` checks `IsPoCParticipantBlocked` (`msg_server_poc_validations_v2.go:38`) but the commit/distribution handlers don't. A blocked participant can still submit store commits, and there's no `ParticipantPermission` check either. Verify `MsgPoCV2StoreCommit.GetSigners()` returns `[Creator]` and add `IsPoCParticipantBlocked` + `ParticipantPermission` for parity with validation submission.

**Should fix:**

- **HIGH: Genesis-bootstrap `VMin` bypass uses chain-wide participant count, not initial-group member count** (`delegation_weight_calculator.go:180-190`). `isGenesisBootstrap` counts all participants with prior weight; once the network grows past `VMin` globally, the initial model can still drop below `VMin` direct members (e.g., members switched to other models) and the bypass stops protecting it. The chain can lose its cap-exempt anchor abruptly. The doc comment says "Once reached, VMin is enforced permanently" but no persistent flag implements stickiness. Either count initial-group members specifically, or make the bypass sticky with a persistent flag.

- **HIGH: `ComputeGroupCap` mixes current-epoch coefficient with N-1 raw weight** (`delegation_weight_calculator.go:262`): `thisGroupContrib := g.ConsensusKoeff.MulInt64(prevPocWeight)`. If governance changes `weight_scale_factor` between epochs, the N-1 contribution is miscomputed and the "other group weight" falls out of sync with what actually made it into `ConsensusWeights`. Store the N-1 coefficient in `PrevMemberPocWeights` or read it from historical params.

- **HIGH: `GetAllPoCDelegations` / `GetPoCDelegationsForParticipant` are O(n) full scans** (`keeper/poc_delegation.go:57-68, 90-106, 127-143`). They pull the entire collection into memory and filter. Currently called only from query handlers, but if they ever hit consensus paths it's a gas/DoS issue. Add a reverse index (by delegator) or pin to query-only with an explicit comment.

- **MEDIUM: `ComputeModelVotingPowers` and `ResolveGroupParticipation` use subtly different filtering** (`delegation_pipeline.go:644-655` vs `delegation_weight_calculator.go:210-226`). `ResolveGroupParticipation` requires `ConsensusWeights[target] > 0`; `ComputeModelVotingPowers` doesn't. Two code paths reading the same delegation snapshot can produce inconsistent decisions. Not exploitable today but future-fragile.

- **MEDIUM: `buildDelegationWeightCalculator` proceeds with empty delegations on snapshot-not-found** (`delegation_pipeline.go:37-41`). At epoch formation, missing a snapshot is consensus-relevant — chain proceeds as if no one delegated. Could be legitimate (very first epochs) or a bug. Either require snapshot past first epoch, or include epoch index in the log + add an event.

- **MEDIUM: Test coverage gaps:**
  - No test for "all hosts choose REFUSE" or "all hosts choose NONE" (chain liveness scenarios)
  - No test for delegation cycles across models (A delegates to B on model X, B delegates to A on model Y)
  - No integration test loading real `v0.2.11` KV state bytes — `TestMigrationSequence` uses in-memory fixtures only
  - No test for `AccumulateBootstrapPenalties` race with eligible transition

- **TODO blockers acknowledged in PR description must be addressed before release:**
  - Set concrete non-zero values for `w_threshold`, `v_min`, `cap_factor`, all three penalty fractions, and per-model `penalty_start_epoch` in the upgrade handler
  - Cap for max delegation to host

**Nits:**

- `initNewPruningState` silently skips seeding if no effective epoch exists (`upgrades.go:178-208`) — should error rather than silent skip; on first pruner call the markers are 0 and walk the entire epoch history.
- `ClearOtherDelegationState` ignores delete errors (`keeper/poc_delegation.go:187-197`) — three collections can become inconsistent silently.
- `captureDelegationSnapshot` logs but doesn't error on `SetDelegationSnapshot` failure (`delegation_pipeline.go:220-223`) — epoch formation proceeds with stale snapshot.
- `DelegationSnapshot` is a singleton with no epoch index — no explicit clear after consume; if next capture fails, old snapshot silently re-applies. Add explicit clear-on-consume.
- `setModelsForParticipants`: if a node serves multiple governance models, the alphabetically-first matching model claims it and others get nothing. Worth a warning event.
- `modelPenaltyStartEpochs` defaults missing entries to "penalties active immediately" (`delegation_weight_adjustment.go:116-122`); combined with the upgrade handler setting `PenaltyStartEpoch=0` for the migrated founding model, the founding model has no grace period once penalty fractions are set. Document or add a future-epoch default.
- `parseRegularDelegationSnapshot` / `parseBootstrapDelegationSnapshot` don't validate uniqueness per `(modelID, delegator)` — malformed snapshot silently picks last.
- `DelegationSnapshot` lacks `TotalNetworkWeight` while bootstrap snapshot has it — asymmetric, potential drift.

**Looks correct:**

- **Legacy KV cleanup is correctly bounded** via `prefix.NewStore(store, keyPrefix)` + `Iterator(nil, nil)` (`keeper/legacy_poc_v2_cleanup.go:24-25`). Despite the dangerous-looking pattern, the prefix store re-prefixes all subsequent reads, so iteration is bounded to that prefix only. Collect-then-delete pattern correctly avoids iterator invalidation.
- **DIRECT precedence is enforced** in both `ResolveGroupParticipation` (`delegation_weight_calculator.go:215`) and `ComputeModelVotingPowers` (`delegation_pipeline.go:651`) — a member's own delegation record for that group is correctly ignored.
- **Split delegation prevented** — collection key is `collections.Join(modelID, delegator)`, so one record per `(model, delegator)`.
- **Last-write-wins** across DELEGATE/REFUSE/INTENT is enforced via `ClearOtherDelegationState` after each handler.
- **Delegation does not bypass collateral/slashing** — `AdjustWeightsByCollateral` runs on `activeParticipants` *after* delegation transfers (`module.go:698`).
- **2/3 supermajority denominator is correct** (`chainvalidation.go:337`) — uses `wc.TotalNetworkWeight` (chain-wide), not per-group, with abstentions correctly counting against. Guardian tiebreak requires unanimous voting guardians.
- **Penalty 1.0 cap correctly enforced** in `PenaltyAccumulator.Apply:81-83`. Penalties apply to `originalWeight` (pre-collateral, pre-cap), so collateral/cap can't amplify them. Transfer delta clamped to `from.Weight` so participants can't go negative.
- **Snapshot timing is race-free** — captured deterministically by block height in `StartStage`, read at `onEndOfPoCValidationStage`, mid-epoch delegation changes don't affect current validation.
- **PoCV2 commit count monotonicity** correctly enforced per `(model, address)`; same-block double-submit rejected; multi-entry message keys are well-formed.
- **`backfillVotingPower` correctly assumes single-model pre-upgrade state** (`upgrades.go:233`) — `migrateParams` runs first so `Models[0]` is the migrated founding model.
- **Slot-mode and non-slot-mode validation paths** both use chain-wide weight as the denominator, consistent with the spec.
- **`PoCWeightCalculator` and `DelegationWeightCalculator` separation** is clean — PoC validation stays independent of multi-model policy layer.

Ryanchen911:
## Review Summary

Thanks for the work on this PR — the delegation/penalty flow is clearly laid out and the accumulator design is nice. I went through the delegation handler, the penalty accumulator, the mode resolver, and the v0.2.12 migration path, and I'd like to flag two correctness bugs and two hardening items before we merge.

**Must-fix (correctness)**

1. `SetPoCDelegation` accepts self-delegation, which lets a participant silently skip NONE/REFUSE penalties.
2. `PenaltyAccumulator.Apply` silently burns weight when a transfer target is missing from the active set.

**Should-fix (hardening / ops)**

3. `ComputeGroupVotingPowers` relies on `ModeDirect` being the Go zero value — correct today, but fragile.
4. `SetPoCDelegation` / `RefusePoCDelegation` / `DeclarePoCIntent` don't reject empty `model_id`.


Details below. 
---

### 1. Self-delegation is not rejected (abuse vector)

**File:** `inference-chain/x/inference/keeper/msg_server_poc_delegation.go` (L9–L37)

`SetPoCDelegation` validates the governance model and the target participant, but never checks `msg.DelegateTo != msg.Sender`. A participant that does not run the model can self-delegate and end up in `ModeDelegate` instead of `ModeNone` / `ModeRefuse`, which has two consequences:

- In `AccumulateDelegationPenalties`, the `ModeDelegate` branch skips both `no_participation_penalty` and `refusal_penalty`, so the participant dodges the penalty entirely.
- In `PenaltyAccumulator.Apply`, the pending transfer becomes `from.Weight -= delta; to.Weight += delta` with `from == to`, i.e. a no-op.
- In `ComputeGroupVotingPowers`, the self-delegator fails the `modes[target] == ModeDirect` check, so no voting power is awarded — which is the only reason this doesn't also become a voting-power exploit.

Net effect: any participant with positive N-1 weight can escape all multi-model participation penalties for any model by sending a single self-targeted `MsgSetPoCDelegation`. 

---

### 2. Weight is silently burned when a delegation target is missing

**File:** `inference-chain/x/inference/module/delegation_weight_adjustment.go` (~L97–L115, `PenaltyAccumulator.Apply` transfer loop)

```go
for _, t := range pa.transfers {
    from := weightIndex[t.from]
    if from == nil { continue }
    delta := t.rate.MulInt64(pa.originalWeight[t.from]).TruncateInt64()
    ...
    from.Weight -= delta
    if to := weightIndex[t.to]; to != nil {
        to.Weight += delta
    }
}
```

The `from == nil` branch is symmetric (skip entirely), but when only `to` is missing we still decrement `from.Weight` and drop `delta` on the floor. This breaks the stated "`delegation_share` transfers to the target" semantic and becomes an implicit burn whenever a target has been removed from the active participant set between the snapshot and application phases.

---

### 3. Fragile reliance on `ModeDirect` being the Go zero value

**File:** `inference-chain/x/inference/module/delegation_weight_calculator.go` (L9–L16 for the iota, L335–L360 for `ComputeGroupVotingPowers`)

`ParticipationMode` is declared as `ModeDirect = iota`, so `ModeDirect == 0`, which is also the value a Go map returns for a missing key. `ResolveGroupParticipation` skips any participant with `w <= 0`, so fresh DIRECT members (positive new `pocWeight`, zero N-1 weight) are absent from `modes` entirely. In `ComputeGroupVotingPowers`:

```go
for _, m := range g.Members {
    if modes[m] == ModeDirect { // absent key also matches
        votingPower[m] = finalWeights[m]
    }
}
...
if modes[target] == ModeDirect {
    votingPower[target] += finalWeights[delegator]
}
```

Today this happens to produce the right answer for fresh DIRECT members because the absent-key zero value coincides with the intent. But any future reorder of the iota constants, or the addition of a `ModeUnknown = iota` sentinel, will silently award voting power to non-DIRECT participants. 

---

### 4. Empty `model_id` is accepted by all three delegation handlers

**File:** `inference-chain/x/inference/keeper/msg_server_poc_delegation.go` (L9–L77, all three handlers)

`SetPoCDelegation`, `RefusePoCDelegation`, and `DeclarePoCIntent` all call `GetGovernanceModel(ctx, msg.ModelId)` directly without checking `msg.ModelId != ""`. The PR description mentions `model_id = ""` being used as the root-group sentinel in `EpochGroupData` storage keys, so an empty `ModelId` reaching this path is at best ambiguous and at worst a future collision waiting to happen. Cheap fix — validate `msg.ModelId != ""` at the top of each handler and return `ErrInvalidRequest`.

---

0xgonka (approved):

gmorgachev:
> ## Review Summary
> Thanks for the work on this PR — the delegation/penalty flow is clearly laid out and the accumulator design is nice. I went through the delegation handler, the penalty accumulator, the mode resolver, and the v0.2.12 migration path, and I'd like to flag two correctness bugs and two hardening items before we merge.
> 
> **Must-fix (correctness)**
> 
> 1. `SetPoCDelegation` accepts self-delegation, which lets a participant silently skip NONE/REFUSE penalties.
> 2. `PenaltyAccumulator.Apply` silently burns weight when a transfer target is missing from the active set.
> 
> **Should-fix (hardening / ops)**
> 
> 3. `ComputeGroupVotingPowers` relies on `ModeDirect` being the Go zero value — correct today, but fragile.
> 4. `SetPoCDelegation` / `RefusePoCDelegation` / `DeclarePoCIntent` don't reject empty `model_id`.
> 
> ## Details below.
> ### 1. Self-delegation is not rejected (abuse vector)
> **File:** `inference-chain/x/inference/keeper/msg_server_poc_delegation.go` (L9–L37)
> 
> `SetPoCDelegation` validates the governance model and the target participant, but never checks `msg.DelegateTo != msg.Sender`. A participant that does not run the model can self-delegate and end up in `ModeDelegate` instead of `ModeNone` / `ModeRefuse`, which has two consequences:
> 
> * In `AccumulateDelegationPenalties`, the `ModeDelegate` branch skips both `no_participation_penalty` and `refusal_penalty`, so the participant dodges the penalty entirely.
> * In `PenaltyAccumulator.Apply`, the pending transfer becomes `from.Weight -= delta; to.Weight += delta` with `from == to`, i.e. a no-op.
> * In `ComputeGroupVotingPowers`, the self-delegator fails the `modes[target] == ModeDirect` check, so no voting power is awarded — which is the only reason this doesn't also become a voting-power exploit.
> 
> Net effect: any participant with positive N-1 weight can escape all multi-model participation penalties for any model by sending a single self-targeted `MsgSetPoCDelegation`.
> 
> ### 2. Weight is silently burned when a delegation target is missing
> **File:** `inference-chain/x/inference/module/delegation_weight_adjustment.go` (~L97–L115, `PenaltyAccumulator.Apply` transfer loop)
> 
> ```go
> for _, t := range pa.transfers {
>     from := weightIndex[t.from]
>     if from == nil { continue }
>     delta := t.rate.MulInt64(pa.originalWeight[t.from]).TruncateInt64()
>     ...
>     from.Weight -= delta
>     if to := weightIndex[t.to]; to != nil {
>         to.Weight += delta
>     }
> }
> ```
> 
> The `from == nil` branch is symmetric (skip entirely), but when only `to` is missing we still decrement `from.Weight` and drop `delta` on the floor. This breaks the stated "`delegation_share` transfers to the target" semantic and becomes an implicit burn whenever a target has been removed from the active participant set between the snapshot and application phases.
> 
> ### 3. Fragile reliance on `ModeDirect` being the Go zero value
> **File:** `inference-chain/x/inference/module/delegation_weight_calculator.go` (L9–L16 for the iota, L335–L360 for `ComputeGroupVotingPowers`)
> 
> `ParticipationMode` is declared as `ModeDirect = iota`, so `ModeDirect == 0`, which is also the value a Go map returns for a missing key. `ResolveGroupParticipation` skips any participant with `w <= 0`, so fresh DIRECT members (positive new `pocWeight`, zero N-1 weight) are absent from `modes` entirely. In `ComputeGroupVotingPowers`:
> 
> ```go
> for _, m := range g.Members {
>     if modes[m] == ModeDirect { // absent key also matches
>         votingPower[m] = finalWeights[m]
>     }
> }
> ...
> if modes[target] == ModeDirect {
>     votingPower[target] += finalWeights[delegator]
> }
> ```
> 
> Today this happens to produce the right answer for fresh DIRECT members because the absent-key zero value coincides with the intent. But any future reorder of the iota constants, or the addition of a `ModeUnknown = iota` sentinel, will silently award voting power to non-DIRECT participants.
> 
> ### 4. Empty `model_id` is accepted by all three delegation handlers
> **File:** `inference-chain/x/inference/keeper/msg_server_poc_delegation.go` (L9–L77, all three handlers)
> 
> `SetPoCDelegation`, `RefusePoCDelegation`, and `DeclarePoCIntent` all call `GetGovernanceModel(ctx, msg.ModelId)` directly without checking `msg.ModelId != ""`. The PR description mentions `model_id = ""` being used as the root-group sentinel in `EpochGroupData` storage keys, so an empty `ModelId` reaching this path is at best ambiguous and at worst a future collision waiting to happen. Cheap fix — validate `msg.ModelId != ""` at the top of each handler and return `ErrInvalidRequest`.

Hi @Ryanchen911! Thanks for the review. Status of required fixes:

> 1. `SetPoCDelegation` accepts self-delegation, which lets a participant silently skip NONE/REFUSE penalties.

It's checked in `ValidateBasic` in `inference-chain/x/inference/types/message_poc_delegation.go`

> 2. `PenaltyAccumulator.Apply` silently burns weight when a transfer target is missing from the active set.
Added check. For punishment we still burn as it's not a transfer of weight

patimen (approved):
We should follow up with some message security issues (lacking ValidateBasic for the new message, we probably want Permissions to be stricter as well), but otherwise I think we're ok.

PR #1068, opened 2026-04-15
voting power caps

0xgonka:
## Summary

Adds two new governance-controlled concentration caps to protect validation integrity in the multi-model PoC system:

- **`max_model_voting_power_percentage`** — limits any single host's voting power within a single model group after delegation is resolved. Defends slot-based validation: a host that attracts heavy delegation in one model can't unilaterally push that model past its supermajority threshold.

- **`max_aggregated_voting_power_percentage`** — limits any single host's voting power summed across all eligible model groups. Catches hosts that sit just under the per-model cap in every model simultaneously.

Both default to zero (disabled) so current behavior is preserved until governance sets concrete values.

## Implementation

- Proto: two new `Decimal` fields on `DelegationParams` (fields 9 and 10).
- `capPerModelVotingPowers` runs inside the per-model loop in `computeAndSetVotingPowers`, using proportional redistribution with rounding-remainder conservation.
- `capAggregatedVotingPowers` runs after the per-model loop; scales a capped host's contribution proportionally across every model it participates in, preserving the relative per-model ratio, and redistributes freed voting power within each affected model group.
- Helpers are package-level functions taking a `votingPowerCapLogger` interface so they can be unit-tested without a full `AppModule`.
- Both helpers loop up to 32 times for convergence in degenerate inputs.

## Tests

9 unit tests covering: no-op cases, single-host groups, zero-cap disables, capping+redistribution, convergence under extreme inputs, proportional scaling across models.

## Test plan

- [x] Unit tests pass
- [x] Builds clean
- [ ] Integration suite (`testermint`)
- [ ] Concrete cap values chosen by governance before release

## AI Usage

This PR was generated by a Claude Opus 4.7 conversation that has reviewed all features in 0.2.12 and is also the same convo that originally implemented the tx-fees feature.


inference-chain/x/inference/types/params.go
```diff
@@ -374,6 +374,12 @@ func DefaultDelegationParams() *DelegationParams {
 		VMin:                   3,
 		CapFactor:              DecimalFromFloat(1),
 		InitialModelId:         "",
+		// Voting-power concentration caps are OFF by default. Governance
+		// must set concrete values via MsgUpdateParams after observing real
+		// network concentration. Zero on either field disables that cap;
+		// see computeAndSetVotingPowers for enforcement semantics.
+		MaxModelVotingPowerPercentage:      DecimalFromFloat(0),
```
patimen: Let's explicitly set these in the upgrade handler, even if we know it defaults to 0.

inference-chain/proto/inference/inference/params.proto
```diff
@@ -355,6 +355,17 @@ message DelegationParams {
   int64 v_min = 6; // min hosts with non-zero consensus weight
   Decimal cap_factor = 7; // max group weight as multiple of members' weight in other groups
   string initial_model_id = 8; // model exempt from group cap; set during migration to the founding model
+  // Max fraction of voting power any single host can hold within a single
+  // model group (after delegation is resolved). Prevents a host that
+  // attracts many delegations from unilaterally controlling validation
+  // within a model's slot allocation. Zero means disabled.
+  Decimal max_model_voting_power_percentage = 9;
```
patimen, line 362: We should add these settings to AppExport.kt to keep the Kotlin representation of parameters accurate.

inference-chain/x/inference/module/delegation_pipeline.go
```diff
@@ -691,14 +745,402 @@ func (am AppModule) computeAndSetVotingPowers(
...
+type votingPowerCapLogger interface {
+	LogInfo(msg string, subSystem types.SubSystem, keyvals ...interface{})
+	LogWarn(msg string, subSystem types.SubSystem, keyvals ...interface{})
+}
+
+// sumInt64Safe sums the values of a string-keyed int64 map, returning
+// (sum, true) on success or (0, false) if the sum would overflow int64.
+func sumInt64Safe(m map[string]int64) (int64, bool) {
+	var total int64
+	for _, v := range m {
```
patimen: I see, theoretically, that this should produce the same result regardless of the order we iterate (which will be random), I still think we should sort so we can have determinism in the loop in case we make changes here in the future (say, by adding a return or log or emitting events... etc). I'd rather err on caution.

inference-chain/x/inference/module/delegation_pipeline.go
```diff
@@ -691,14 +745,402 @@ func (am AppModule) computeAndSetVotingPowers(
...
+	logger votingPowerCapLogger,
+) {
+	if len(participantVP) < 2 {
+		return
+	}
+
+	// Compute per-host aggregated VP (with overflow safety).
+	hostTotals := make(map[string]int64, len(participantVP))
+	var totalVP int64
+	for addr, vps := range participantVP {
```
patimen: Again, let's err on caution and iterate over a sorted list of keys, not the map itself. At a minimum, our warning below would show different hosts across our chain, which could get confusing.

inference-chain/x/inference/module/delegation_pipeline.go
```diff
@@ -691,14 +745,402 @@ func (am AppModule) computeAndSetVotingPowers(
...
+			freed := mvp.VotingPower - newVP
+			if freed > 0 {
+				freedPerModel[mvp.ModelId] += freed
+			}
+			mvp.VotingPower = newVP
+		}
+		capped[overAddr] = true
+
+		// Redistribute freed VP within each affected model group.
+		for modelID, freed := range freedPerModel {
```
patimen: One last time. Iterate over sorted keys, not the raw map

patimen (changes requested):

gmorgachev:
@0xgonka could you elaborate what `max_aggregated_voting_power_percentage` is protecting us from?

gmorgachev:
@0xgonka also logic of re-distribution concerns me. if voting power gained from the the delegation, why we can move it  to another participant? essentially not-trusted participants gain more voting powers

```
	// rounding and burning changed total VP mid-pipeline, which has
	// knock-on effects in downstream reward and slot-sampling calculations.
```
could you elaborate what you mean? 


why wouldn't we also set all caps against totalConsensusWeight? just for simplicity (as we have requirements for > 2/3 consensusWeight is should not be a weaker approach)

patimen (approved):
My changes look good, but be sure the check with Gleb on his questions.
I ran the `ai-reviewer` tool using `gemini_standard` for most of this, but discarded several suggestions I didn't think warranted discussion. I did also look through the changes (broadly) myself.

inference-chain/proto/inference/inference/params.proto
```diff
@@ -355,6 +355,11 @@ message DelegationParams {
...
 }
```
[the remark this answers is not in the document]
DimaOrekhovPS: The PR description is outdated

inference-chain/x/inference/types/params.go
```diff
@@ -374,6 +374,11 @@ func DefaultDelegationParams() *DelegationParams {
 		VMin:                   3,
 		CapFactor:              DecimalFromFloat(1),
 		InitialModelId:         "",
+		// Per-model voting-power concentration cap is OFF by default.
+		// Governance must set a concrete value via MsgUpdateParams after
+		// observing real network concentration. Zero disables the cap; see
+		// computeAndSetVotingPowers for enforcement semantics.
+		MaxModelVotingPowerPercentage: DecimalFromFloat(0),
 	}
 }
```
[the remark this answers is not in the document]
DimaOrekhovPS: Sounds like a good suggestion

testermint/src/main/kotlin/data/AppExport.kt
```diff
@@ -104,6 +104,8 @@ data class DelegationParams(
     val capFactor: Decimal = Decimal(0, 0),
     @SerializedName("initial_model_id")
     val initialModelId: String = "",
+    @SerializedName("max_model_voting_power_percentage")
+    val maxModelVotingPowerPercentage: Decimal = Decimal(0, 0),
 )
```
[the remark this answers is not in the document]
DimaOrekhovPS: based on an outdated PR descriptio

DimaOrekhovPS (approved):

gmorgachev (approved):

PR #1069, opened 2026-04-15
Integrate Devshard settlement into Participant EpochStats

patimen:
# Integrate Devshard settlement into Participant EpochStats

Currently, when a Devshard escrow is settled, participants are paid directly via the bank module (`SendCoinsFromModuleToAccount`), and their performance statistics are only recorded in the isolated `DevshardHostEpochStatsMap`. Their work on Devshards does not update their main `Participant.CurrentEpochStats` (such as inference counts, missed requests, and earned coins), which are critical for accurate network performance tracking, rewards, and voting power in the Gonka primary chain. Additionally, there were no hard limits on the settlement `Nonce`, nor validations preventing a participant from reporting more missed or invalid inferences than they were actually assigned per slot. 

Looking at the previous implementation in `msg_server_settle_devshard_escrow.go`, the settlement loop simply invoked a direct bank transfer for every participant, completely bypassing the participant's `CoinBalance` and `CurrentEpochStats` state. Furthermore, `VerifyDevshardSettlement` lacked bounds checking on `msg.Nonce` and the `HostStats` struct fields, leaving the chain vulnerable to potentially fraudulent stat reporting (e.g., reporting 100 missed inferences when only 5 were assigned) or excessively large computational loops.

This PR overhauls the Devshard settlement pipeline to properly integrate with the core Participant state:
* **Targeted State Updates:** Introduces `AggregateDevshardHostStatsIntoCurrentEpochStats` to merge Devshard completion data (`Missed`, `Invalid`, `Completed`) directly into a participant's `CurrentEpochStats` (updating `InferenceCount`, `MissedRequests`, `InvalidatedInferences`, and `ValidatedInferences`).
* **Activity-Aware Payouts:** Modifies `SettleDevshardEscrow` to check if a participant is active in the current epoch.
    * *Active participants* receive payouts directly to their `CoinBalance` and `CurrentEpochStats.EarnedCoins` via a new `AddToCoinBalance` helper, and their epoch stats are aggregated.
    * *Inactive participants* (or settlements for the previous epoch) fallback to the legacy direct-to-bank payout mechanism (`payCoinsDirectly`) without touching `CurrentEpochStats`.
* **Strict Validations:** Adds `DevshardMaxNonce` (capped at 20,000) and ensures that reported `Missed` and `Invalid` requests logically cannot exceed the `assignedPerSlot` in `VerifyDevshardSettlement`.

* **Risk:** **Integer Overflows.** Directly mutating the participant's cumulative statistical counters could lead to integer overflow panics or exploits. 
    * *Mitigation:* All new arithmetic updating `CurrentEpochStats` and `CoinBalance` leverages `math/bits.Add64` to safely catch overflows and return standard errors instead of panicking.
* **Risk:** **State Corruption for Inactive Participants.** Modifying stats for participants who are no longer active could result in corrupted epoch boundaries.
    * *Mitigation:* The PR explicitly loads the `ActiveParticipantsSet` and gates the state modifications. If a participant is not active in the current epoch, their `CurrentEpochStats` are strictly ignored.
* **Risk:** **Testing Coverage Gaps.** Complex conditional branching around participant activity and epoch boundaries.
    * *Mitigation:* Extensive unit tests have been added to `msg_server_settle_devshard_escrow_test.go` targeting these exact branches (e.g., `TestSettleDevshardEscrow_CurrentEpochInactiveParticipantPaidImmediatelyWithoutParticipantStateChange`, `TestSettleDevshardEscrow_AggregatesParticipantStats`).

The PR includes comprehensive test suites that directly validate the new behavior:
1.  Tests explicitly verify that an active participant's `CurrentEpochStats.InferenceCount`, `EarnedCoins`, and related fields increment accurately after an escrow settlement.
2.  Tests verify that inactive participants successfully receive bank payouts without their `CurrentEpochStats` altering. 
3.  New constraint tests in `devshard_settlement_test.go` assert that transactions with nonces exceeding `DevshardMaxNonce` or impossible `Missed`/`Invalid` counts are actively rejected by the keeper.

---

## Other Notes
* **Refactoring:** A new helper method `AddToCoinBalance` was introduced in `msg_server_start_inference.go` to centralize coin balance and earnings updates. The `processInferencePayments` logic was refactored to use this helper, ensuring uniform overflow protection across the module.
* **State Migration Impacts:** Because `SettleDevshardEscrow` now actively modifies the `Participant` KV store object, care should be taken during subsequent upgrades if the `Participant` struct is extended further.

## AI Usage for producing this PR
- Primary model: GPT 5.4 (Medium thinking)
- Primary app: Codex on MacOS

I worked iteratively with Codex in the repo to shape the spec, pressure-test edge cases, and implement the keeper changes. I used it mostly as a pairing partner: I gave the direction, corrected course when the implementation or assumptions were off, and had it revise the code and tests until the behavior matched what we wanted. Along the way, I ran our local `ai-reviewer` multiple times, then brought in external review feedback as another input; I used Codex to analyze those comments, decide what to fix vs. push back on, and make the follow-up edits. I reviewed the resulting diffs and test coverage myself, then finalized and published the PR.

I did some hand changes here and there in the code as well, code is likely 90% AI generated.


0xgonka (commented):
Integrating devshard settlement into participant epoch stats with overflow-safe arithmetic and validation bounds. Well-structured PR with clear test coverage. A few issues to address:

**Must fix:**

- **[High] `payCoinsDirectly` truncates `uint64` to `int64` without overflow check.** `msg_server_settle_devshard_escrow.go:232`: `types.GetCoins(int64(payout))` — if `payout` exceeds `math.MaxInt64`, the `int64()` cast wraps negative and `GetCoins` either panics or creates negative coins. `AddToCoinBalance` (line 362 of `msg_server_start_inference.go`) correctly checks `payout > math.MaxInt64` before casting, but `payCoinsDirectly` skips this check entirely. An inactive participant receiving a large payout could trigger this. Fix: add the same `math.MaxInt64` check before the cast.

- **[Medium] `assignedPerSlot` integer division loses remainder.** `msg_server_settle_devshard_escrow.go:77`: `assignedPerSlot := msg.Nonce / totalSlots`. If `msg.Nonce = 19` and `totalSlots = 16`, `assignedPerSlot = 1`. But some slots actually served 2 nonces (the first 3 slots each get one extra from the remainder). The `AggregateDevshardHostStatsIntoCurrentEpochStats` call on line 193 passes this truncated `assignedPerSlot`, so a slot that legitimately handled 2 requests but reports `Missed = 0` will pass the bounds check (`0 <= 1`), but its `completed = 1` is wrong — it actually completed 2. Stats will undercount. Same issue exists in `devshard_settlement.go:132` where `VerifyDevshardSettlement` rejects `hs.Missed > assignedPerSlot` on line 140 — a slot that received the remainder nonce and missed it would be incorrectly rejected. Consider using `(msg.Nonce + totalSlots - 1) / totalSlots` (ceiling division) for the bounds check, or compute per-slot assignment accounting for the remainder.

**Should fix:**

- **`devshard_host_stats.go:45-46,51-52`: `Missed`, `Invalid`, `RequiredValidations`, `CompletedValidations` are `uint32` but aggregated without overflow checks.** `existing.Missed += slotStats.Missed` can silently wrap at `math.MaxUint32`. `Cost` gets an overflow check (line 47) but the other four fields don't. If a devshard processes many escrows in one epoch, these could theoretically overflow. Add `math.MaxUint32` checks for consistency, same pattern as line 47.

- **`msg_server_settle_devshard_escrow.go:160`: `escrow.Amount - totalPayout` assumes `totalPayout <= escrow.Amount`.** `VerifyDevshardSettlement` checks `totalDebit <= escrow.Amount` where `totalDebit = totalCost + fees`, but `totalPayout` in the settlement handler includes costs AND fees AND remainder-fee redistribution. If the remainder fee distribution pushes `totalPayout` above `escrow.Amount` (edge case with rounding), this underflows. Add an explicit `if totalPayout > escrow.Amount` guard before the subtraction.

**Nits:**

- `devshard_settlement.go:22-23`: `DevshardGroupSize = 16` and `DevshardQuorumSlots = 2*DevshardGroupSize/3 + 1` are compile-time constants, but `DevshardQuorumFor()` (line 32) computes the same formula dynamically. `DevshardQuorumSlots` is unused — consider removing it to avoid confusion about which is canonical.

- `msg_server_settle_devshard_escrow.go:86`: `int(hs.SlotId) >= len(escrow.Slots)` — `SlotId` is `uint32`, `len()` returns `int`. The explicit `int()` cast is unnecessary since Go handles this comparison correctly. Minor readability thing.

**Looks correct:**

- **Phase byte binding** (`DevshardSettlementPhase = 0x02`): prevents settling an escrow whose state hasn't been finalized. Active (0x00) and Finalizing (0x01) states produce a different hash and fail the state_root check. Critical for settlement integrity.

- **Warm key grant checking** via `HasWarmKeyGrant` using authz `MsgStartInference` as the sentinel grant type. This is the correct grant type to check — it uniquely identifies host ML ops grants and is consistent with the feegrant migration pattern.

- **Deterministic iteration** throughout: `slices.Sort(uniqueAddrs)`, sorted host stats by `SlotId`, deterministic proto marshal for hash computation. Settlement is consensus-critical and this ordering discipline is essential.

- **Overflow protection** via `bits.Add64` is applied consistently across the new code paths. The carry-check pattern is correct and used uniformly.

- **Active participant gating**: correctly loads `ActiveParticipantsSet` and only mutates `CurrentEpochStats` for active participants. Inactive participants fall back to direct bank transfer without touching epoch state, which is the safe default.

0xgonka (commented):
v2 of this PR addresses all three items flagged in the previous review (uint64→int64 cast guard, `assignedPerSlot` remainder accounting, `escrow.Amount - totalPayout` underflow). The new `devshardAssignedUpperBoundForSlot` correctly models devshard's nonce→slot assignment (`inference_id % group_size`), and the "remainder slot missed allowed" test confirms it. A couple of new items from the delta:

**Should fix:**

- **`devshardAssignedUpperBoundForSlot` gives slot 0 a different "first nonce" than all other slots, relying on an undocumented devshard convention.** `devshard_settlement.go:46-48` — `if slotID == 0 { firstAssignedNonce = slotCount }`. This looks surprising: by the general formula `firstAssignedNonce = slotID`, slot 0 would be 0. The code explicitly skips nonce 0 for slot 0 and starts slot 0 at `slotCount`. This is presumably because devshard reserves nonce 0 for something (initialization? the creator?) and begins user inferences at nonce 1, where `1 % slotCount == 1` lands on slot 1. If slot 0 only gets nonces `slotCount, 2*slotCount, …`, that means the **first `slotCount-1` nonces are never routed to slot 0** — is that an intentional devshard property, or a compensation for an off-by-one in the nonce counter?
   - The test `TestVerifyDevshardSettlement_NonRemainderSlotMissedRejected` confirms this with nonce=19, slotCount=16: slot 0 is expected to receive only one nonce (16), not two (16 and 0). But the mental model for why nonce 0 is excluded should be documented on the function, because the chain is now the authoritative validator of this convention.
   - If devshard ever changes its nonce counting (e.g. a future version starts at 0), this function silently misaligns. Add a doc comment describing the exact devshard-side contract it mirrors, with a pointer to the devshard source file that implements the sending side.

- **Payout/refund check order allows a narrow double-spend window.** `msg_server_settle_devshard_escrow.go:152-167` — the handler pays out every validator first (`AddToCoinBalance` for active, `payCoinsDirectly` for inactive), then checks `if totalPayout > escrow.Amount`. If this check fails, the error aborts the tx and the `SetParticipant`/`SetDevshardEscrow` writes at lines 211-217 never land — but the `AddToCoinBalance` mutations to in-memory `participantByAddr[addr]` at line 148 have already happened. The SDK's state isolation means unwritten mutations are discarded on error, so this is **not** a real double-spend — but any reviewer who doesn't know that gets suspicious. Move the `totalPayout > escrow.Amount` check **before** the payout loop (compute `totalPayout` from `validatorPayouts` aggregation first, guard, then pay). It's a single `for` loop summing `validatorPayouts` — cheaper than the current post-hoc check and easier to reason about.

**Nits:**

- **`payCoinsDirectly` is defined below `}` of `SettleDevshardEscrow` with no blank line.** `msg_server_settle_devshard_escrow.go:241` — cosmetic, but `gofmt` won't fix it. Add a blank line before `func (k Keeper) payCoinsDirectly(...)`.

- **Seen-set tracking in `firstForValidator` is awkward.** `msg_server_settle_devshard_escrow.go:184,198-200` — computed as `!seenValidators[addr]` before the `UpdateDevshardHostEpochStats` call, then `seenValidators[addr] = true` is set after the `if firstForValidator` branch. Reads fine but moving the set into the same `if firstForValidator { seenValidators[addr] = true }` branch (already there) and using `_, seen := seenValidators[addr]; firstForValidator := !seen` with the actual map-set after is more idiomatic.

- **`TestSettleDevshardEscrow_AggregatesParticipantStats` expected math is computed in the test comment rather than a helper.** `msg_server_settle_devshard_escrow_test.go:325-345` — for future readers, wrapping the `assignedPerSlot`/`completed`/`validated` math into a tiny helper (even just a local closure) would make the assertions self-documenting rather than requiring a reader to mentally re-derive them from the `nonce=20, slots=4` inputs.

**Looks correct:**

- **`devshardAssignedUpperBoundForSlot` correctly bounds per-slot assignment.** Formula `1 + (latestNonce - firstAssignedNonce) / slotCount` is right: for nonce=19, slotCount=16, slot=1 it gives `1 + (19-1)/16 = 2`, and slot=0 gives `1 + (19-16)/16 = 1`. Matches the tests.

- **Uint32 overflow checks added to all five `DevshardHostEpochStats` counters.** `devshard_host_stats.go:45-68` — every `+=` now has a matching `math.MaxUint32` pre-check, consistent with the existing `Cost` pattern at line 56. Addresses previous review.

- **`payCoinsDirectly` now checks `payout > math.MaxInt64` before the `int64` cast.** `msg_server_settle_devshard_escrow.go:243-245` — addresses the high-severity bug from the previous review.

- **`totalPayout > escrow.Amount` guard added** (even if placed late — see Should-fix above) at `msg_server_settle_devshard_escrow.go:163`. And the refund path checks `remainder > math.MaxInt64` at line 167. Both are belt-and-suspenders defenses that are correct.

- **Active-participant-set membership checked against `currentEpochIndex`, not `escrow.EpochIndex`.** `msg_server_settle_devshard_escrow.go:68` — correct. Settlement in the previous epoch (N-1) for an escrow whose epoch was also N-1 shouldn't update the current-epoch N stats; the inactive-fallback path handles this.

- **Deterministic iteration via `slices.Sort(uniqueAddrs)` and `slices.Sort(touchedAddrs)`** at lines 53 and 207 — settlement is consensus-critical and map iteration order would be nondeterministic.

- **`AddToCoinBalance` helper in `msg_server_start_inference.go:358-378`** with proper int64-additional-overflow checks. The previous review found `payCoinsDirectly` had truncation; this helper is the correct centralized version and is used by both inference-start and devshard-settlement code paths.

- **`participantByAddr` is looked up once up front, then mutated via pointers.** `msg_server_settle_devshard_escrow.go:55-73` — the approach of holding `*types.Participant` pointers and applying both the `AddToCoinBalance` and `AggregateDevshardHostStatsIntoCurrentEpochStats` mutations to the same in-memory struct before a single `SetParticipant` at line 209 is correct and efficient. No double-write, no race with `GetParticipant` re-reads.

- **Epoch gating: `currentEpochIndex < escrow.EpochIndex || currentEpochIndex > escrow.EpochIndex+1`** at `msg_server_settle_devshard_escrow.go:40-42` — correctly rejects settlements from epochs older than previous. Test `TestSettleDevshardEscrow_OlderThanPreviousEpochRejected` confirms.

patimen:
@0xgonka 
The slot-0 behavior is intentional and mirrors the existing devshard routing contract, not an off-by-one compensation. In [`devshard/user/user.go#L258`](https://github.com/gonka-ai/gonka/blob/e47da59a758e7c92e5efc2b75cdb89da4827d667/devshard/user/user.go#L258), diffs advance as `nonce := s.nonce + 1`, so real inference diffs start at nonce `1`, and in [`devshard/state/machine.go#L562`](https://github.com/gonka-ai/gonka/blob/e47da59a758e7c92e5efc2b75cdb89da4827d667/devshard/state/machine.go#L562), executor routing is `group[inference_id % len(group)]` with `inference_id == nonce`. That means slot `0` first receives work at nonce `slotCount`, while slots `1..slotCount-1` first receive work at their matching nonce, so the helper is matching current devshard semantics rather than introducing a new convention.

Regarding the "double spend":

I agree the current placement is harder to reason about than it needs to be, but I don’t think this is a real double-spend window. In Cosmos SDK, returning an error aborts the message and discards uncommitted keeper mutations, so the in-memory participant updates in [`msg_server_settle_devshard_escrow.go#L148`](https://github.com/gonka-ai/gonka/blob/e47da59a758e7c92e5efc2b75cdb89da4827d667/inference-chain/x/inference/keeper/msg_server_settle_devshard_escrow.go#L148) and the later `SetParticipant` / `SetDevshardEscrow` writes in [`msg_server_settle_devshard_escrow.go#L211-L217`](https://github.com/gonka-ai/gonka/blob/e47da59a758e7c92e5efc2b75cdb89da4827d667/inference-chain/x/inference/keeper/msg_server_settle_devshard_escrow.go#L211-L217) either all commit together or none do. That said, I agree moving the `totalPayout > escrow.Amount` guard ahead of the payout loop would make the flow easier to audit, so I’m happy to take that cleanup.

0xgonka (approved):

inference-chain/x/inference/keeper/devshard_host_stats.go
```diff
@@ -19,6 +20,20 @@ func (k Keeper) GetDevshardHostEpochStats(ctx context.Context, epochIndex uint64
 }
 
 func (k Keeper) AggregateDevshardHostStats(ctx context.Context, epochIndex uint64, participant sdk.AccAddress, slotStats types.DevshardSettlementHostStats) error {
+	return k.UpdateDevshardHostEpochStats(ctx, epochIndex, participant, slotStats, false)
+}
+
+func (k Keeper) IncrementDevshardHostEscrowCount(ctx context.Context, epochIndex uint64, participant sdk.AccAddress) error {
+	return k.UpdateDevshardHostEpochStats(ctx, epochIndex, participant, types.DevshardSettlementHostStats{}, true)
+}
```
gmorgachev, line 28: Seems like not used anymore?

gmorgachev (approved):

PR #1070, opened 2026-04-16
Fix gas estimation

0xgonka:
## Summary

Fixes a race in the BLS DKG dealing phase where later dealers in an epoch could be dropped from the signing group due to "out of gas" failures caused by linearly-growing write costs.

## The bug

`SubmitDealerPart` rewrote the entire `EpochBLSData` struct on every dealer submission. Since `EpochBLSData.DealerParts` accumulated inline as more dealers submitted, the Nth submission's write cost was roughly N × the first submission's cost. In an 8-participant group, the 8th dealer paid **~10.4M gas** vs **~187k** for the 1st — a 56× difference that matches the `WritePerByte` scaling.

This matters because the DAPI estimates transaction gas via simulation. Between simulation and block execution, more dealer parts could land on-chain, pushing the real write cost above the simulated estimate and failing later dealers with `out of gas`. Those dealers then:
1. missed `CalculateSlotsWithDealerParts`,
2. dropped below the `ITotalSlots/2` threshold in `TransitionToVerifyingPhase`,
3. got filtered out of the new epoch's signing group entirely.

This is what Maria observed on testnet — validators "dropping" after the upgrade were actually dealers whose `MsgSubmitDealerPart` failed in flight.

## The fix

Split `DealerParts` out of the inline struct and store each participant's dealer part under its own sub-key:

```
{DealerPartPrefix}{epochID:BE8}/{participantIdx:BE4}  →  DealerPartStorage
```

- **`msg_server_dealer.go`** — hot path now calls `SetDealerPart`, a single sub-key write. Cost is constant in the number of dealers already submitted. Every dealer in a round pays roughly the same gas regardless of submission order.
- **`SetEpochBLSData`** — zeroes `DealerParts` on a copy before persisting, so the base struct is constant-size for any consumer (phase transitions, dispute resolution).
- **`GetEpochBLSData`** — rehydrates `DealerParts` from sub-keys after unmarshaling the base. All existing consumers (`phase_transitions.go`, `dispute_resolution.go`, `msg_server_verifier.go`, `bls_crypto.go`) see the same slice shape they always have and work unchanged.
- **Backward compat** — if a base struct still has inline `DealerParts` (pre-upgrade writes), those serve as the baseline and any sub-key entries take precedence. No migration step required.
- **`DeleteDealerPartsForEpoch`** — helper for future teardown paths; not yet wired up since verifying / disputing phases still need to read dealer parts.

## Expected gas cost after fix

Every dealer pays ~187k gas (the base cost of a single dealer-part write) regardless of submission order. No more gas-estimation race, no more dropped dealers, no more failed DKG rounds caused by order-dependent gas cost.

## Files changed

- `inference-chain/x/bls/types/keys.go` — new `DealerPartKey`, `DealerPartEpochPrefix`
- `inference-chain/x/bls/keeper/dkg_initiation.go` — split `SetEpochBLSData`, rehydrated `GetEpochBLSData`, new `SetDealerPart` / `GetDealerPart` / `DeleteDealerPartsForEpoch`
- `inference-chain/x/bls/keeper/msg_server_dealer.go` — hot path writes single sub-key

## Test plan

- [x] Builds clean
- [x] `go vet` clean across `x/bls/...` and `x/inference/...`
- [ ] Existing `msg_server_dealer_test.go` suite passes in CI
- [ ] Integration: `BLSDKGSuccessTest` in testermint (exercises the 3-participant path end-to-end)
- [ ] Manual testnet verification that validators no longer drop during new-epoch DKG with >2 hosts

## AI Usage

This PR was generated by a Claude Opus 4.7 conversation that has reviewed all features in 0.2.12 and is also the same convo that originally implemented the tx-fees feature.


patimen (approved):
Approved.
I ran `ai-reviewer` using `gemini_standard` and suggested a few fixes (now done).
I also looked over dkg_initiation.go manually and decided we really needed to use the KVStore in a more standardized way (using prefix.Store) to reduce code duplication and possible errors.

All these have been addressed, looks good!

DimaOrekhovPS (approved):

PR #1071, opened 2026-04-16
hardening: propagate internal errors across inference/validation/pricing paths

Mayveskii:
StartInference, FinishInference, and Validation all swallowed InjectParamsIntoContext errors with LogWarn, continuing execution with a context missing params. This caused inference to run with wrong pricing, permissions, and epoch parameters.

Now returns error immediately, matching ClaimRewards pattern.


Doog-bot534 (approved):
## Review: fix: propagate InjectParamsIntoContext errors

**Approve** ✅

The original code logged the error but used the `ctx` value returned alongside it — classic Go anti-pattern. If `InjectParamsIntoContext` fails, downstream logic reads stale/missing params and silently misbehaves. Failing fast is correct.

**Minor concern:** If `InjectParamsIntoContext` has a transient failure mode (cache miss that self-heals), this turns recoverable into hard failure. Confirm it only errors on truly unrecoverable conditions.

**Suggestion:** Consider a module-specific sentinel error instead of bare `sdkerrors.Wrap` for programmatic distinction.

unameisfine:
Note: the StartInference and FinishInference changes here overlap with PR #971 (submitted Mar 28), which fixes the same InjectParamsIntoContext error propagation in those two handlers. The additional fix for msg_server_validation.go is new — might be worth scoping this PR to just that file to avoid merge conflicts.

Mayveskii:
>re 

It's ok now. Thx.

x0152:
@Mayveskii could I ask you to combine #1071, #1073, #1074, #1012, #968, and #1076 into a single PR? They are mostly hardening changes in different places, so one combined pr would be easier to review

Mayveskii:
> re

@x0152 , sure

Mayveskii:
>re

It's combined.

x0152:
Thanks. Can you close the duplicate prs?

Mayveskii:
>Re

 🤦  fell ill
Now it's ok!

patimen:
I'd like to revisit this a bit:
1 - We should remove the changes to Inferences. Inferences are now all but exclusively in Devshards.
2 - For POC validations... I don't feel this is well defined as to the desired behavior. If there is an error in HasPocValidationsV2 it means we either hit a storage error (we're in a really bad/edge case and likely everything is weird) or somehow an invalid ParticipantAddress was in the validation. Is abandoning all of the validations the correct behavior?

Mayveskii:
>re @patimen 

Hi, John. Thanks for the master view and seeing unnoticed problems by me. Now it looks like this new one is a  correct solve.

patimen:
One more thing: Could we not add assertions for logging at this time? We historically haven't added unit tests for pure logging concerns, and I don't see a particular good reason to start.. let's just assert on the error/continuation and leave the logging out of it?

Mayveskii:
>re 

It's clean now.

patimen (approved):
Now clean and minimal. We only escape out of the batch in the very unlikely issue of a storage issue.

patimen:
@Mayveskii Can you fix up the history here, force push to get the changes clean? Looks like changing base branch got things a bit mixed up.

Mayveskii:
>re

Looks able now
Is it ok @patimen  ?

PR #1072, opened 2026-04-16
fix: reject zero completion tokens in FinishInference

Mayveskii:
When CompletionTokenCount is 0, CalculateCost returns 0, causing:
- executor receives 0 payment
- escrow refunded in full to user
- financial inconsistency

Added validation at both msg_server and calculations level:
- msg_server: reject CompletionTokenCount == 0 early
- inference_state: return error instead of LogWarn

PromptTokenCount == 0 case preserved as LogWarn since streaming/interrupted inferences legitimately use the StartInference value as fallback.

### DIFF
--- a/inference-chain/x/inference/calculations/inference_state.go
+++ b/inference-chain/x/inference/calculations/inference_state.go
@@ -170,7 +170,7 @@ func ProcessFinishInference(
 		logger.LogWarn("PromptTokens is 0 when FinishInference is called!", types.Inferences, "inferenceId", currentInference.InferenceId)
 	}
 	if currentInference.CompletionTokenCount == 0 {
-		logger.LogWarn("CompletionTokens is 0 when FinishInference is called!", types.Inferences, "inferenceId", currentInference.InferenceId)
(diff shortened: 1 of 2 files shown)


Doog-bot534 (approved):
## Review: fix: reject zero completion tokens in FinishInference

**Approve** ✅

Zero completion tokens for a finished inference is semantically invalid. The existing code only warned, then passed zero into `CalculateCost`, corrupting reward accounting.

**Note:** The same condition is checked at both message-level (`failedFinish()`) and calculation-level (`sdkerrors.Wrapf`). The calc-level check is effectively dead code for normal flow — defense-in-depth, but worth a comment.

**Nit:** `PromptTokens == 0` still only warns. Consider hard rejection for consistency.

x0152:
Hi
Closing this PR: completion_token_count = 0 is currently valid in existing flows, so this change would break intended behavior

Feel free to reopen if you disagree

Mayveskii:
>re


Hi!
Thx for feedback,  i should checkout better untill.

PR #1073, opened 2026-04-16
fix: propagate errors in handleInferenceCompleted instead of swallowi…

Mayveskii:
…ng them

handleInferenceCompleted had two swallowed errors:
1. executor nil: LogWarn and skip participant updates, leaving inference stats inconsistent
2. effective epoch not found: LogWarn and default EpochId=0, EpochPocStartBlockHeight=0, corrupting epoch tracking

Both now return proper errors, matching ClaimRewards pattern. Function signature changed from void to error return.

This prevents inferences with nil executor or missing epoch from being enqueued for validation with corrupt state.

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_finish_inference.go
+++ b/inference-chain/x/inference/keeper/msg_server_finish_inference.go
@@ -144,7 +144,9 @@ func (k msgServer) FinishInference(goCtx context.Context, msg *types.MsgFinishIn
 		return failedFinish(ctx, err, msg), nil
 	}
 	if finalInference.IsCompleted() {
-		k.handleInferenceCompleted(ctx, finalInference, &executor)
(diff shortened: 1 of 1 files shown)


Doog-bot534 (approved):
## Review: fix: propagate errors in handleInferenceCompleted

**Approve** ✅ — Most impactful of the batch.

Three distinct swallowed failures: nil executor (skips stats silently), missing epoch (defaults to zero — corrupts state), enqueue failure (inference never validated). All now properly surfaced.

**Concern:** The epoch-not-found case previously defaulted to zero. If there's a legitimate race between epoch creation and inference completion (genesis block, transitions), this now rejects completions during that window. Verify no race condition exists.

**Suggestion:** Nil executor returns `ErrParticipantNotFound`, but the caller already loaded it upstream. If it's nil here, that's a code bug, not user error — consider a different error or clarifying comment.

x0152:
Closing as duplicate of #1071

PR #1074, opened 2026-04-16
fix: propagate storage errors in SubmitPocValidationsV2

Mayveskii:
HasPocValidationV2 and SetPocValidationV2 errors were swallowed with LogWarn+continue, causing:
1. HasPocValidationV2 error: potential duplicate vote if storage check fails silently
2. SetPocValidationV2 error: validator vote silently lost

Both now return errors immediately, preserving transactional integrity. Only business-level skips (duplicate validation) remain as LogWarn.


Doog-bot534 (commented):
## Review: fix: propagate storage errors in SubmitPocValidationsV2

**Approve with caveat** ✅

Storage errors are infrastructure-level — silent `continue` meant partial validations could be dropped while the tx appeared to succeed.

**Key question:** The original `continue` was arguably partial-success tolerance. With this change, one bad entry in a 100-validation batch poisons the whole tx. For blockchain determinism this is correct (partial writes get reverted), but confirm partial-success was **not** intentional.

**Suggestion:** If partial success was intentional, collect errors into the response while processing remaining validations.

PR #1075, opened 2026-04-16
fix: return errors from OverlapsWithPoC instead of swallowing them

Mayveskii:
CreatedAtBlockHeight == 0 in OverlapsWithPoC caused it to return false, allowing claim_rewards to skip PoC validation for inferences with missing block height data. This enables reward claims for inferences that should require PoC validation.

OverlapsWithPoC now returns (bool, error):
- nil inference details -> error
- CreatedAtBlockHeight == 0 -> error (unset field)
- CreatedAtBlockHeight < 0 -> error (negative)

Also added ErrInvalidValidationDetails error type.

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_claim_rewards.go
+++ b/inference-chain/x/inference/keeper/msg_server_claim_rewards.go
@@ -480,7 +480,11 @@ func (k msgServer) getMustBeValidatedInferences(ctx sdk.Context, msg *types.MsgC
 
 		totalWeight := modelTotalWeights[modelId]
 
-		if k.OverlapsWithPoC(&inference, epochContext) && !k.isActiveDuringPoC(&validatorPowerForModel) {
+		overlapsWithPoC, err := k.OverlapsWithPoC(&inference, epochContext)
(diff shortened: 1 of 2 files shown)


Doog-bot534 (commented):
## Review: fix: return errors from OverlapsWithPoC

**Approve with concern** ✅

Returning `false` on invalid input masked data integrity issues. Nil `inferenceDetails` or negative `CreatedAtBlockHeight` indicates a bug — silently treating as "no overlap" could cause incorrect reward claims.

**Critical concern:** If existing on-chain data has `CreatedAtBlockHeight == 0` (legacy records before the field existed), this breaks reward claims for epochs containing those records. **Needs data migration verification.**

**Nit:** If `CreatedAtBlockHeight` is `uint64`, the `< 0` check is unreachable. Verify the field type.

x0152:
Thanks for the pr! I'm closing this one
Not reproducible on v0.2.12, and this change adds stricter failure behavior than intended

Mayveskii:
>re 

The next  I'll drill deeper current.

PR #1076, opened 2026-04-16
fix: propagate storage errors in UpdateDynamicPricing instead of defa…

Mayveskii:
…ulting to zero

GetCachedModelCapacity error was swallowed with LogWarn+continue, leaving stale prices un-updated. GetModelLoadRollingAveragePerBlock error was swallowed with LogWarn+defaulting to zero load, causing incorrect utilization calculations (0% utilization) that would decrease prices incorrectly.

Both now return errors immediately, ensuring BeginBlocker retries on next block rather than silently corrupting pricing data.

Note: !found case (no load data yet) still defaults to zero — correct behavior for new/unused models.

### DIFF
--- a/inference-chain/x/inference/keeper/dynamic_pricing.go
+++ b/inference-chain/x/inference/keeper/dynamic_pricing.go
@@ -65,9 +65,7 @@ func (k *Keeper) UpdateDynamicPricing(ctx context.Context) error {
 		// Get cached capacity for this model
 		capacity, err := k.GetCachedModelCapacity(ctx, modelId)
 		if err != nil {
-			k.LogWarn("Failed to get cached capacity for model, skipping", types.Pricing,
-				"modelId", modelId, "error", err)
-			continue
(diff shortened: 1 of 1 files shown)


Doog-bot534 (commented):
## Review: fix: propagate storage errors in UpdateDynamicPricing

**Request changes** ⚠️

**Critical concern:** `UpdateDynamicPricing` appears to be called from BeginBlocker/EndBlocker context. Returning errors in block lifecycle functions can **halt the chain**. The original lenient behavior (default to zero) may have been intentional to prevent exactly this.

A transient storage hiccup would cause consensus failure across all validators. Consider:
1. Log at ERROR level with safe defaults (current behavior but louder)
2. Panic only after N consecutive failures
3. At minimum, verify the caller handles the error gracefully without chain halt

**Also:** Uses `fmt.Errorf` with `%w` instead of `sdkerrors.Wrapf` — inconsistent with the other PRs.

PR #1084, opened 2026-04-16
draft: Moveout devshardd

pixelplex:
## Core Idea
Extract the devshard host into a standalone service (`devshard/cmd/devshardd`) and build a `common` module of shared libraries. `decentralized-api` remains the core validator service and will reuse packages from `common` to build its own specific implementations.

`common/queryapi` provides a single set of read-only chain query HTTP handlers (OpenAPI-generated) that any service can mount into its Echo server

Inference validation in devshardd uses Postgres-backed validation leases. In the normal flow, a single devshardd instance owns an escrow session because requests are routed through the consistent-hashing LB and sessions are created lazily. The lease is therefore a reliability and deduplication guard: it lets validation retry after an instance fails mid-flow, and it prevents duplicate validation if routing/recovery ever causes more than one instance to process the same escrow session.

---

### `decentralized-api` — Core Validator Service (stays)
Owns everything that requires local ML node state or validator identity:
- Inference endpoint (`POST /v1/chat/completions`) — retry loop, ML node HTTP, SSE streaming
- Inference validation — `inference_finished` event, payload retrieval, `MsgValidation`
- Payload serving (`GET /v1/inference/payloads`) — for validators during regular inference validation
- NodeManager gRPC server (already implemented) — `AcquireMLNode` / `ReleaseMLNode` backed by broker
- PoC generation and artifact serving
- Training coordination (gRPC port 9003)
- Admin endpoints — ML node management, BLS, config, hardware
- Bridge queue processing
- Chain query endpoints (until migrated to use `common/queryapi`)

### `devshard` — Devshard Host Service (`devshard/cmd/devshardd`)
Runs the devshard protocol host for a validator. Connects to `decentralized-api`'s NodeManager gRPC for node locking. Uses `common/` for all concrete implementations.
- Devshard session endpoints (`/v1/devshard/sessions/:id/*`) via lazy routing
- Payload serving for devshard validators (`GET /v1/devshard/sessions/:id/payloads`)
- Chain bridge queries (via `common/chain`)
- Chain query endpoints via `common/queryapi`

### `common` — Shared Libraries (not a service)
A Go module imported by both `decentralized-api` and `devshard`. Not deployed — compiled into whatever service imports it.

```
common/
├── logging/           # structured logging over log/slog
├── completionapi/     # OpenAI request/response handling, SSE streaming
├── utils/             # misc helpers
├── storage/
│   ├── payloads/      # prompt/response payload storage (Postgres)
│   └── validationlease/ # validation lease storage (Postgres)
├── nodemanager/       # gRPC client + generated proto for NodeManager
├── chain/             # gRPC chain client — query clients for inference, BLS, restrictions, CometBFT
│   ├── tx/            # tx broadcast manager
│   └── events/        # chain event listener
├── validation/        # payload retrieval, signature verification, logit comparison
└── queryapi/          # read-only chain query HTTP handlers (OpenAPI-generated + impls)
                       #   wirable into any Echo server: gen.RegisterHandlers(e, queryapi.NewHandlers(client))
```

---

## Deployment Topology

```
clients
  │
  ▼
proxy
  │
  ├──► decentralized-api ──► NodeManager gRPC (same process, separate port)
  │         │                      │
  │         │                      ▼
  │         └──────────────────► ML nodes
  │
  └──► devshardd ──► decentralized-api NodeManager gRPC
              └──► inference chain (gRPC :9090)
```

Both services run on the same validator machine. Devshardd connects to dapi's NodeManager to acquire/release ML node locks.

---

## `devshardd` Wiring (current state)

Devshardd is wired from the local cosmos keyring and the chain gRPC/RPC endpoints. It builds:
- a chain identity/recorder for payload auth and signer address lookup
- a chain event bridge for escrow settlement events and current phase/epoch tracking
- a NodeManager gRPC client used by inference execution and validation
- Postgres payload and validation lease stores
- a validation engine wrapped with `LeaseValidator`
- a local SQLite-backed `HostManager` for devshard session state

On startup it recovers local sessions from SQLite, starts the validation retry loop, mounts queryapi plus devshard session routes, and starts the HTTP server.

### Key management
Devshardd reads directly from the cosmos keyring — no ignite client dependency. For the `file` backend the keyring password is injected via `KEYRING_PASSWORD` env var so the process runs non-interactively. The `devshard` key must be registered on-chain as a warm key grantee authorized for devshard operations. `MsgValidation` is a devshard message type — the devshard key covers it.

```
keyring/
  validator    ← dapi key — authorized for MsgStartInference, MsgFinishInference, etc.
  devshard     ← devshard key — authorized for devshard message types only
```

---

## Inference Validation Flow

Implemented for `devshardd` using `common/validation` for validation execution primitives and `common/storage/validationlease` for Postgres-backed retry/deduplication. Triggered when a finished inference requires validation.

In the current devshardd model, the LB should route each escrow to one owning instance. The validation lease is not the primary work distribution mechanism; it is a safety layer for retries and for rare duplicate-session cases.

Validation starts from the devshard host after an inference reaches the finished state and `ShouldValidate` selects the local validator slot for that inference. The validator first acquires a Postgres lease for `(escrow_id, inference_id)`. If the lease already exists, this instance skips the work.

After acquiring the lease, devshardd fetches the executor payloads from the executor's devshard session endpoint, verifies the executor signature, and checks that the payload hashes match the devshard state commitment. A hash mismatch produces an immediate invalid validation result and is not retried. Otherwise devshardd acquires an ML node through NodeManager, re-runs inference, releases the node, signs `MsgValidation`, adds it to the devshard host mempool, and marks the lease `submitted`.

```
inference requires validation
  │
  ▼
devshardd
  │
  ├─ ShouldValidate → bool
  │    (deterministic for the local validator slot and inferenceId)
  │
  ├─ if false → skip
  │
  ├─ postgre.acquireValidationLease(escrowId, inferenceId, epochId, instanceAddress)
  │    if conflict → validation already leased/submitted/skipped → skip
  │    (pending stale leases are handled by Validation Retry below)
  │
  ├─ GET executor's /v1/devshard/sessions/:escrowId/payloads?inference_id=...
  │    verify executor signature on response
  │    verify payload hashes match devshard state commitment
  │    if hash mismatch → immediate invalidation (no retry)
  │
  ├─ nodemanager.AcquireMLNode(model)
  ├─ re-run inference on ML node and process result
  ├─ nodemanager.ReleaseMLNode(lock_id, success)
  │
  ├─ sign MsgValidation with the devshard slot key
  ├─ add MsgValidation to the devshard host mempool
  └─ postgre.setValidationLeaseStatus(escrowId, inferenceId, submitted)
```

The `validation_leases` table is keyed by `(escrow_id, inference_id)` and stores the inference epoch, lease holder address, acquisition timestamp, and status (`pending`, `submitted`, or `skipped`).

### Validation Retry on Fail
If an instance crashes or fails before `MsgValidation` is added to the devshard host mempool and the lease is marked submitted, the lease remains `pending`. A background retry loop periodically scans active in-memory sessions for stale pending leases and retries them.

**Stale lease:** `status = 'pending' AND claimed_at < now() - validation_lease_ttl`

Config fields:
```
DEVSHARD_VALIDATION_LEASE_TTL        duration  // how long before an uncompleted lease is retried (default: 30m)
DEVSHARD_VALIDATION_RETRY_INTERVAL   duration  // how often to scan for stale leases (default: 5m)
```

On each retry tick, every devshardd instance scans only its active in-memory escrow sessions. For each session it atomically acquires stale pending leases one at a time with row locking, re-runs the validation flow, and then marks the lease `submitted` on success. The row-locking query prevents concurrent retry loops from acquiring the same stale lease.

```
every validation_retry_interval, for each escrowId in the instance's in-memory session map (HostManager.sessions):
  │
  └─ loop:
       ├─ postgre.acquireOneStaleLease(escrowId, newInstanceAddress)
       │    if no row returned → no more stale leases → break
       │
       ├─ if epoch/session no longer retryable → mark skipped → continue loop
       │
       └─ otherwise run validation flow
           ├─ success → status=submitted
           └─ retryable error → leave pending for next TTL window
```

If the inference is already stale for the current epoch, or the inference is no longer in a retryable finished state, the lease is marked `skipped`. Hash mismatches produce an immediate invalid validation result and are not retried. Other validation errors leave the lease `pending`; it can be retried again after the TTL.

The `submitted` status is set after `MsgValidation` is signed and added to the devshard host's in-memory mempool. Durable tx-hash tracking is not part of the current implementation.

## Validation Lease Cleanup
`validation_leases` rows are cleaned up by epoch. Cleanup runs on each new-block/epoch transition after reading the current epoch. Any instance can run it — it is idempotent and safe to run concurrently. Validation leases are only retained for the current epoch; older pending leases are no longer retryable and are dropped.

**Payload retention:** payload rows (`payload_storage`) keep the current epoch plus two previous epochs. On each new-block/epoch transition, devshardd drops `current_epoch - 3`.

---

### `decentralized-api` — endpoints that require keys or local state

**Inference:**
- `POST /v1/chat/completions` — inference execution, warm key for `MsgStartInference`/`MsgFinishInference`
- `GET  /v1/chat/completions` — inference record lookup
- `GET  /v1/inference/payloads` — payload serving for regular inference validation

**Chain writes (warm key required):**
- `POST /v1/training/tasks` — submits `MsgCreateTrainingTask`
- `POST /v1/participants` — submits `MsgSubmitNewUnfundedParticipant`
- `POST /v1/bridge/blocks` — submits `MsgBridgeExchange`

**Local state (in-memory, not portable):**
- `GET  /v1/bridge/status` — reads in-memory block queue

**Pricing:**
- `GET /v1/governance/pricing` - depends on old inference stats storage

**Admin, PoC, training gRPC, ML node callbacks:** (unchanged)

---

### `devshardd` — devshard session endpoints
- `POST /v1/devshard/sessions/:id/chat/completions`
- `POST /v1/devshard/sessions/:id/verify-timeout`
- `POST /v1/devshard/sessions/:id/challenge-receipt`
- `POST /v1/devshard/sessions/:id/gossip/nonce`
- `POST /v1/devshard/sessions/:id/gossip/txs`
- `GET  /v1/devshard/sessions/:id/diffs`
- `GET  /v1/devshard/sessions/:id/mempool`
- `GET  /v1/devshard/sessions/:id/signatures`
- `GET  /v1/devshard/sessions/:id/payloads`

---

### `common/queryapi` — pure chain reads, no keys, no local state. Can be used in decentralized-api or devshardd

**Participants, models, epochs:**
- `GET /v1/participants`, `GET /v1/participants/:address`
- `GET /v1/models`, `GET /v1/governance/models`, `GET /v1/governance/models-legacy`
- `GET /v1/epochs/:epoch`, `GET /v1/epochs/:epoch/participants`
- `GET /v1/status`

**Pricing, restrictions, versions:**
- `GET /v1/pricing`, `GET /v1/governance/pricing`
- `GET /v1/restrictions/status`, `GET /v1/restrictions/exemptions`
- `GET /v1/restrictions/exemptions/:id/usage/:account`
- `GET /v1/versions`

**BLS, bridge, PoC:**
- `GET /v1/bls/epoch/:id`, `GET /v1/bls/epochs/:id`, `GET /v1/bls/signatures/:request_id`
- `GET /v1/bridge/addresses`
- `GET /v1/poc-batches/:epoch`

**Training (read-only):**
- `GET /v1/training/tasks`, `GET /v1/training/tasks/:id`

**Stateless computation:**
- `POST /v1/verify-proof`
- `POST /v1/verify-block`
- `GET  /v1/debug/pubkey-to-addr/:pubkey`
- `GET  /v1/debug/verify/:height`

---


DimaOrekhovPS:
A question about Validation Retry on Fail.

What would happen if an mlnode returns an error like 400 to the devshard instance? Does the approach differentiate between retryable vs non-retryable errors?

In case of receiving 400 we shouldn't retry

tcharchian:
https://github.com/gonka-ai/gonka/pull/1084 was replaced with https://github.com/gonka-ai/gonka/pull/1165
cc @DimaOrekhovPS @akup @a-kuprin

PR #1087, opened 2026-04-17
Do not rollup when settle is from previous epoch

patimen:
1. Don't rollup if not in current epoch
2. Include separate memo for inference_finished vs devshard_settle in bankkeeper
3. Add a test for #1

## AI Usage
AI found the issues (ai-reviewer on `gemini_high`).
I fixed both manually.
AI wrote the test, I reviewed (Codex, GPT 4.5 Medium)


0xgonka (approved):
PR #1087 fixes a bug where devshard settlement from a previous epoch, when settled for participants still active in the current epoch, would incorrectly roll the payout into the active participant's current-epoch coin balance and stats. It also threads a caller-provided memo through `AddToCoinBalance` so subaccount logs distinguish `inference_finished` from `devshard_settle` transactions. Well-scoped, tight regression test covering exactly the previously-broken path.

**Should fix:**
- The condition `treatAsCurrentEpochSettle[addr] = active && escrow.EpochIndex == currentEpochIndex` at `msg_server_settle_devshard_escrow.go:76` has a redundant second clause. The earlier `if escrow.EpochIndex != currentEpochIndex { ...; continue }` already guarantees `escrow.EpochIndex == currentEpochIndex` is true at this point, so the final line simplifies to `treatAsCurrentEpochSettle[addr] = active`. Not a correctness issue — just dead belt-and-suspenders logic that makes the invariant harder to read.

**Looks correct:**
- `msg_server_settle_devshard_escrow.go:63-73`: Early `continue` when `escrow.EpochIndex != currentEpochIndex` correctly skips the `sdk.AccAddressFromBech32` + `ActiveParticipantsSet.Has` calls per participant. The address is re-validated inside the payout loop via the `sdk.AccAddressFromBech32(addr)` at the start of the per-validator block, so no validation is actually being skipped.
- `msg_server_settle_devshard_escrow.go:148-162`: Previous-epoch settlements correctly fall through to `payCoinsDirectly`, which routes via `BankKeeper.SendCoinsFromModuleToAccount` and does NOT touch `CoinBalance` or `CurrentEpochStats`. This is the correct semantic — the validator still gets paid their share, just out-of-band of the epoch-stats accounting.
- `msg_server_settle_devshard_escrow.go:195-211`: `UpdateDevshardHostEpochStats` still runs unconditionally (aggregates under `escrow.EpochIndex`, correct), while `AggregateDevshardHostStatsIntoCurrentEpochStats` is gated behind `treatAsCurrentEpochSettle[addr]`. This means the stats land under the right epoch but don't bleed into the current-epoch stats when settling a past escrow. Exactly the bug that was being fixed.
- `touchedParticipants` map (used to gate the `SetParticipant` loop) is only populated inside the `inCurrentEpoch` branches, so previous-epoch settlements correctly avoid writing participant state.
- The new test at `msg_server_settle_devshard_escrow_test.go:459-515` is the tight regression test: sets participants active in epoch 6, creates escrow at epoch 5, expects `SendCoinsFromModuleToAccount` with `devshard_escrow_payment` memo (direct-pay path). Under the pre-PR behavior, `IsActive` would have returned true, the AddToCoinBalance path would have been taken, and there'd be no BankKeeper call — so the gomock expectation would fail. Clean regression lock.
- Memo parameter: only two callers of `AddToCoinBalance` exist (`msg_server_start_inference.go:348`, `msg_server_settle_devshard_escrow.go:154`), both updated. The old hardcoded `"coin_balance_updated"` string had no downstream consumers relying on its exact value.

**Nits:**
- None blocking.

Overall: safe to merge once the redundant `&&` is cleaned up (or as-is — it's benign).

gmorgachev (approved):
I feel we need to refactor msg hadler. it's becoming quite complex
approve for 0.2.12

PR #1088, opened 2026-04-17
Fix gas scaling in group key validation signature submission

0xgonka:
## Summary

Fixes four more instances of the same N² write-per-byte gas-scaling bug that PR #1070 addressed for dealer parts. Each fix follows the same shape PR #1070 established: move the accumulating slice out of the main struct into per-entry sub-keys or a dedicated `collections.KeySet`, null out the inline field before the base write so the sync loops don't redundantly rewrite every entry on every confirmation, and migrate any in-flight pre-split state in the v0.2.12 upgrade handler.

The direct operational motivation is Maria's testnet trace (`MsgSubmitGroupKeyValidationSignature ... gasWanted: 10000000, gasUsed: 10240546: out of gas` in `WritePerByte`): the first commit here fixes the signing phase so validators stop getting dropped mid-DKG. The other four commits work back through the same pattern elsewhere in BLS and the bridge so turning on consensus fees in v0.2.12 doesn't expose the same race in other flows.

## Commits (one fix per commit, self-reviewed before push)

1. **`0c72a014a`** — `GroupKeyValidationState.PartialSignatures` → per-participant sub-keys at `bls_partial_sig/{new_epoch_id}/{participant_index}`. Direct fix for Maria's failing testnet tx.
2. **`3840df9f7`** — `EpochBLSData.VerificationSubmissions` → `epoch_bls_verification_submission/{epoch_id}/{participant_index}`.
3. **`dacacb1c9`** — `EpochBLSData.DealerComplaints` → `epoch_bls_dealer_complaint/{epoch_id}/{dealer_index}{complainer_index}`. Phase transition's filter now uses explicit `DeleteDealerComplaintsForEpoch` + re-sync so filtered-out complaints don't leak.
4. **`f0a77561c`** — `ThresholdSigningRequest.PartialSignatures` → `threshold_partial_sig/{request_id_len}{request_id}/{submitter}`. Retry paths now clear prior-attempt sub-keys before the new attempt begins.
5. **`e087f52d6`** — `BridgeTransaction.Validators` → dedicated `collections.KeySet[Quad[chainId, blockNumber, receiptIndex, validator]]`. `removeBridgeTransactionByID` clears the sub-keys before removing the base so cleanup doesn't leak.

## Key invariants (locked in by tests in each commit)

- **Prefix isolation** — every new sub-key prefix deliberately does not start with any existing base-state prefix, so prefix iterators over the base namespace never yield split-out entries. `TestGroupKeyValidationState_PrefixIsolation` and the other split tests pin these invariants; renames that re-collide will fail the tests.
- **Legacy rehydration** — every `Get*` call merges inline entries from the base (legacy, pre-split writes) with sub-key entries, deduplicated where relevant, so in-flight mid-upgrade state is observable. Deterministic iteration order (sorted sub-keys + slice walks, no map iteration) keeps consensus stable.
- **Hot path null-out** — every handler that used to call `Set*(ctx, struct)` with a populated slice now writes its own sub-key directly, nulls the slice, and then persists the base. Per-tx gas is constant in the number of prior contributors.
- **Legacy migration** — the v0.2.12 upgrade handler adds `migrateEpochBLSDataToSubKeys`, `migrateThresholdSigningRequestsToSubKeys`, and `migrateBridgeTransactionValidatorsToSubKeys`. Each one iterates all existing entries and re-Sets through the new sync-and-strip path. Idempotent; re-running is a no-op. Migrations run once in the upgrade block before any user tx, so the subsequent null-out-before-write pattern never drops legacy inline data.
- **BLS consensus determinism** — aggregation is commutative in partial-sig order, so the new sorted-by-key iteration order produces the same final signature as the prior append-order.

## Known gaps (worth flagging for reviewers)

- **`Inference.ValidatedBy`** ([msg_server_validation.go:160](inference-chain/x/inference/keeper/msg_server_validation.go#L160)) is the same pattern but materially lower impact: the Inference struct already gets rewritten on every validation for unrelated reasons (status, proposal details), and `ValidatedBy` adds only ~45 bytes per validator vs the Inference's own multi-KB payload. Flagged but not fixed in this PR — happy to do a follow-up if we want the consistency, but probably doesn't move the needle on the gas ceiling.
- **Dispute / phase-transition handlers** still call `SetEpochBLSData` with the rehydrated slice populated and pay the O(N) re-sync cost. These are cold paths (once per epoch at most), not per-participant. Defensible to leave, but a future cleanup could adopt the hot-path null-out pattern there too.
- **No end-to-end "upgrade a v0.2.11 snapshot" test.** The included tests exercise `Set/Get` invariants, legacy-merge behavior, delete/clear paths, and the split sync loops, but nothing runs the full upgrade handler against realistic pre-split chain state. We should replay the upgrade on a testnet snapshot before merging.

## Test plan

- [x] Full inference-chain `go test ./...` green across all commits.
- [x] Targeted tests for each split: strip-and-sync, legacy merge/override, delete/clear, O(1) duplicate check.
- [x] Existing BLS + inference test suites still pass after each commit in isolation.
- [ ] Replay the v0.2.12 upgrade on a testnet snapshot with in-flight DKG + bridge state to exercise all four migrations end-to-end.
- [ ] Verify a full 8-participant DKG + threshold-signing + bridge-confirmation round post-upgrade, watching for any "out of gas" in `WritePerByte`.

## AI usage

This PR was developed with Claude Code (Anthropic's CLI) acting as a pair-programming agent. Concretely:

- **Pattern triage** — fed the failing testnet tx log into the agent; it matched the "append-to-inline-slice + full-struct rewrite" pattern to the PR #1070 root cause and drove a targeted audit of the rest of the chain side looking for the same shape.
- **Design and implementation** — each of the five fixes was proposed (split layout, prefix naming, migration strategy, null-out-before-write pattern) before any code was written, and then drafted as a self-contained commit with matching tests.
- **Self-review pass per commit** — ran a `review-gonka`-style self-review on each fix before pushing, which caught a real prefix-collision bug (`group_validation_partial_sig/` overlapped `group_validation_`) and the "first post-upgrade write drops legacy inline data" bug, both fixed before the respective commits landed.
- **Second-pass audit for missed instances** — a follow-up audit flagged the `Inference.ValidatedBy` case listed under Known Gaps that I didn't touch in the initial scan.
- **Human oversight** — all split-vs-redistribute-vs-burn trade-offs, the call to leave the dispute/phase-transition handlers on the cold-path pattern, and the decision to ship `Inference.ValidatedBy` separately were human calls. The agent produced drafts and ran validations; commits, pushes, and merge decisions were human-approved.

No Claude co-author trailer is attached to the commits per project convention.


inference-chain/x/bls/keeper/group_validation.go
```diff
@@ -0,0 +1,307 @@
...
+// Backward compatibility: if the base struct has legacy inline
+// PartialSignatures (e.g. an in-flight validation round written by a
+// pre-split handler and not yet cleared), this function transparently
+// migrates them to sub-keys on first read and rewrites the base state
+// with PartialSignatures zeroed. The migration is idempotent and scoped
+// to the single validation round being read — it piggybacks on the hot
+// path rather than needing a chain-wide upgrade migration.
+//
+// Returns (nil, false, nil) if no state exists for the epoch.
+func (k Keeper) GetGroupKeyValidationState(ctx sdk.Context, newEpochID uint64) (*types.GroupKeyValidationState, bool, error) {
```
patimen, line 172: I am not a fan of having a Getter also do migration/setting. Is there any way we could separate this out instead?

inference-chain/app/upgrades/v0_2_12/upgrades.go
```diff
@@ -253,6 +285,93 @@ func adjustBLSParameters(ctx context.Context, blsKeeper blskeeper.Keeper) error
...
+// nulls them out before SetEpochBLSData to avoid the O(N) re-sync cost,
+// which would silently discard the legacy data. This migration runs once
+// in the upgrade block (before any user txs) to eliminate that risk.
+//
+// SetEpochBLSData itself does the work: its inline sync loops write any
+// populated entries to sub-keys, and the base is persisted with the split
+// fields zeroed. Each entry's re-Set is idempotent, so re-running the
+// migration would be a no-op.
+func migrateEpochBLSDataToSubKeys(ctx sdk.Context, blsKeeper blskeeper.Keeper) error {
+	all := blsKeeper.GetAllEpochBLSData(ctx)
```
patimen: All loaded at once... will there be enough memory? Maybe we should be iterating instead.

inference-chain/app/upgrades/v0_2_12/upgrades.go
```diff
@@ -253,6 +285,93 @@ func adjustBLSParameters(ctx context.Context, blsKeeper blskeeper.Keeper) error
...
+// Same rationale as migrateEpochBLSDataToSubKeys: the post-split
+// AddPartialSignature nulls out PartialSignatures before persisting the
+// base request, so legacy inline entries must be moved to sub-keys before
+// any post-upgrade tx can touch state.
+//
+// GetAllThresholdSigningRequests already rehydrates PartialSignatures
+// from sub-keys, so returning it through storeThresholdSigningRequest
+// below does the sync-and-strip pass in a single call.
+func migrateThresholdSigningRequestsToSubKeys(ctx sdk.Context, blsKeeper blskeeper.Keeper) error {
+	all := blsKeeper.GetAllThresholdSigningRequests(ctx)
```
patimen: Again, this seems like a risk to load the whole thing all at once.

inference-chain/app/upgrades/v0_2_12/upgrades.go
```diff
@@ -253,6 +285,93 @@ func adjustBLSParameters(ctx context.Context, blsKeeper blskeeper.Keeper) error
...
+	iter, err := k.BridgeTransactionsMap.Iterate(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("iterate bridge transactions for migration: %w", err)
+	}
+	values, err := iter.Values()
+	iter.Close()
+	if err != nil {
+		return fmt.Errorf("collect bridge transactions for migration: %w", err)
+	}
+	for i := range values {
```
patimen: Why get the values and then iterate over them instead of iterating over the iterator?

patimen (changes requested):
First set of changes... still reviewing but wanted to give you some to work on.

inference-chain/app/upgrades/v0_2_12/upgrades.go
gmorgachev: Seems like it's about 1 - 30MB of data max

inference-chain/app/upgrades/v0_2_12/upgrades.go
gmorgachev: Seems like under 6MB of data

inference-chain/x/inference/keeper/bridge_transaction.go
```diff
@@ -33,7 +67,47 @@ func (k Keeper) SetBridgeTransaction(ctx context.Context, tx *types.BridgeTransa
 	}
 }
 
-// GetBridgeTransactionByContent retrieves a bridge transaction by its content hash
+// AddBridgeTransactionValidator records that `validator` confirmed the
+// transaction identified by (chainId, blockNumber, receiptIndex). Cost is
+// constant in the number of prior confirmations, so every validator's tx
+// pays the same gas regardless of confirmation order.
+func (k Keeper) AddBridgeTransactionValidator(ctx context.Context, tx *types.BridgeTransaction, validator string) error {
+	return k.BridgeTransactionValidators.Set(ctx, collections.Join4(tx.ChainId, tx.BlockNumber, tx.ReceiptIndex, validator))
```
x0152: BridgeTransactionValidators keyed by (chainId, blockNumber, receiptIndex, validator), but the parent BridgeTransactionsMap is keyed by (chainId, blockNumber, contentHashPart)

inference-chain/x/inference/keeper/bridge_transaction.go
```diff
@@ -21,8 +30,33 @@ func (k Keeper) SetBridgeTransaction(ctx context.Context, tx *types.BridgeTransa
...
+			return
+		}
+	}
+
+	// Persist a stripped copy so the base value stays small regardless
+	// of how many validators have confirmed. We copy to avoid mutating
+	// the caller's struct.
+	stripped := *tx
+	stripped.Id = id
+	stripped.Validators = nil
```
x0152, line 59: Stripping Validators here is ok for storage, but query paths must load them from BridgeTransactionValidators otherwise BridgeTransaction and BridgeTransactions return empty validators

inference-chain/x/bls/keeper/genesis.go
x0152: seems missed here: ThresholdSigningRequests and GroupKeyValidationStates merge their split fields from sub-keys on export, but EpochBLSData doesn't

inference-chain/x/bls/keeper/threshold_signing.go
```diff
@@ -548,18 +623,144 @@ func (k Keeper) aggregatePartialSignatures(partialSigs []types.PartialSignature,
...
+		}
+	}
+
 	key := types.ThresholdSigningRequestKey(request.RequestId)
 	kvStore := k.storeService.OpenKVStore(ctx)
 
-	requestBytes, err := k.cdc.Marshal(request)
+	// Copy so we don't mutate the caller's request.
+	baseCopy := *request
+	baseCopy.PartialSignatures = nil
```
x0152, line 667: SigningStatus loads them, but SigningHistory still reads raw, so it returns empty PartialSignatures

inference-chain/app/upgrades/v0_2_12/upgrades.go
```diff
@@ -253,6 +296,159 @@ func adjustBLSParameters(ctx context.Context, blsKeeper blskeeper.Keeper) error
...
+		}
+		if len(tx.Validators) == 0 {
+			continue
+		}
+		toMigrate = append(toMigrate, tx)
+	}
+	iter.Close()
+	for i := range toMigrate {
+		tx := toMigrate[i]
+		k.SetBridgeTransaction(ctx, &tx)
```
patimen, line 447: We swallow any error here? Even though this failure is very unlikely, and l.ikely a local-only problem, failing here is cleaner rather than continuing and having a node in a different state than others.

inference-chain/x/bls/keeper/msg_server_group_validation.go
```diff
@@ -190,9 +177,20 @@ func (ms msgServer) SubmitGroupKeyValidationSignature(goCtx context.Context, msg
...
+	// existing sub-key entry.
+	if err := ms.SetGroupValidationPartialSignature(ctx, msg.NewEpochId, uint32(participantIndex), partialSignature); err != nil {
+		ms.Keeper.LogError("Failed to save partial signature", "new_epoch_id", msg.NewEpochId, "participant_index", participantIndex, "error", err.Error())
+		return nil, fmt.Errorf("failed to save partial signature: %w", err)
+	}
 
-	// Update slots covered
+	// Keep the in-memory view in sync so the threshold check and the
+	// aggregation path below see the newly-added signature.
+	validationState.PartialSignatures = append(validationState.PartialSignatures, *partialSignature)
```
patimen, line 193: But `SetGroupValidationPartialSignature` doesn't accept duplicates for the same participant, while this in-memory copy does. Is that going to cause problems? Definitely not logical.

inference-chain/x/inference/keeper/bridge_transaction.go
```diff
@@ -137,19 +258,42 @@ func (k Keeper) CleanupOldBridgeTransactions(ctx context.Context, chainId string
...
 
 func (k Keeper) removeBridgeTransactionByID(ctx context.Context, id string) error {
 	parts := strings.SplitN(id, "_", 3)
 	if len(parts) != 3 {
 		return fmt.Errorf("invalid bridge transaction id: %s", id)
 	}
+	// Clear any per-validator confirmation sub-keys before removing the
+	// base transaction so the split-out records don't linger in state.
+	rng := collections.NewSuperPrefixedQuadRange3[string, string, string, string](parts[0], parts[1], parts[2])
+	it, err := k.BridgeTransactionValidators.Iterate(ctx, rng)
```
patimen, line 284: There's no need to make a ranger, iterate and get the keys and then delete. Just do this:
`	k.BridgeTransactionValidators.Clear(ctx, rng)`

Far more idiomatic, simple and likely more performant.

inference-chain/x/inference/keeper/bridge_transaction.go
```diff
@@ -10,9 +10,18 @@ import (
...
+//
+// The Validators slice is split off into a per-validator KeySet so the
+// base BridgeTransaction value stays constant-size as more validators
+// confirm. Any inline Validators entries on the input tx are synced to
+// the KeySet here; the stored value is persisted with Validators zeroed.
+// Hot-path callers should add the new validator via
+// AddBridgeTransactionValidator directly and clear Validators on the
+// in-memory tx before calling SetBridgeTransaction to avoid redundant
+// re-sync writes on every confirmation.
 func (k Keeper) SetBridgeTransaction(ctx context.Context, tx *types.BridgeTransaction) {
```
patimen, line 23: Why aren't we returning the error?

patimen (approved):
I think there's some things to do in the comments, but not blockers.

PR #1089, opened 2026-04-18
Random Selection of Preserved MLNodes 

gmorgachev:
# Preserved Nodes

This note separates current implementation from the proposed change. It focuses on protocol behavior and only keeps API details that affect the design.

### Meaning of `timeslot_allocation`

`MLNodeInfo.timeslot_allocation` is defined in `inference-chain/proto/inference/inference/epoch_group_data.proto`.

- `PRE_POC_SLOT = 0`
- `POC_SLOT = 1`

Today `POC_SLOT=true` means the node is preserved for the whole epoch. In practice that means:

- it stays on inference during the epoch-start PoC
- it stays on inference during every confirmation PoC event in that epoch
- it contributes no fresh PoC proof during that epoch

`POC_SLOT=false` means the node is in the PoC-participating side for that epoch. It can be used in the epoch-start PoC and in any confirmation PoC event that happens in the inference phase.

One important precision: the chain uses `POC_SLOT` for weight accounting and epoch data, but the "stay on inference" behavior is realized mainly by the decentralized API broker. The chain does not use `POC_SLOT` as a hard reject rule for PoC submissions.

### Epoch timeline today

For a normal epoch, the flow is:

```text
[ epoch-start PoC x1 ] -> [ inference phase with 0..N confirmation PoC events ]
```

The confirmation count is not fixed. It can be zero because of random triggering, configuration, or because confirmation PoC is not active for early epochs.

With the current static epoch flag:

- `POC_SLOT=true` skips the epoch-start PoC and all confirmation events in that epoch
- `POC_SLOT=false` can participate in the epoch-start PoC and each confirmation event

### Weight handling today

Current weight handling uses the same split:

- preserved weight is the `POC_SLOT=true` slice
- confirmation weight covers the `POC_SLOT=false` slice
- effective reward weight is recomputed as preserved weight plus confirmation weight

Rollover also uses the same flag. `GetPreservedNodesByParticipant` and `GetPreviousEpochMLNodesWithInferenceAllocation` copy prior preserved nodes into the next epoch and reset their allocation to `[true, false]` for the new epoch.

### Allocation and stability

`POC_SLOT` is assigned in `model_assignment.go` when the next epoch is formed. The assigner limits preservation so enough weight remains available for generation, validation, and voting.

`ActiveParticipants` is written at the end of PoC validation in `module.go` and then treated as stable for that epoch. That is why the current design can carry one epoch-long preserved flag, but it is also why mid-epoch rewrites of `ActiveParticipants` are unattractive.

## Problem

The current design binds preservation to the whole epoch. That creates a predictable gap:

- a preserved node skips the only epoch-start PoC opportunity for that epoch
- the same node also skips every confirmation PoC event in that epoch
- if inference traffic is low, the hardware can go mostly unexercised between epoch boundaries

This makes hardware downgrade or partial-capacity substitution easier to plan. An operator knows in advance which boxes will avoid PoC for the entire epoch.

The asymmetry is simple:

- non-preserved nodes can be checked at epoch start and again during confirmation events
- preserved nodes wait until a later epoch boundary before they are forced back into PoC

### Core idea

Replace epoch-long preservation with episode-scoped preservation.

An episode is one PoC execution window:

- the epoch-start regular PoC
- one confirmation PoC event during the inference phase

At the start of each episode, the chain materializes one preserved snapshot for that episode only. The broker reads that snapshot for the active episode. The next episode gets a new snapshot.

This keeps the late-binding property we want: an operator cannot know far in advance whether a given node will be preserved for the next PoC window.

### Anchor point

The preserved snapshot should be fixed at the same boundary already used for generation start tracking:

- regular PoC: the boundary keyed by `upcomingEpoch.PocStartBlockHeight` in `module.go`
- confirmation PoC: the `GRACE_PERIOD -> GENERATION` transition keyed by `event.TriggerHeight` in `confirmation_poc.go`

That gives the design a deterministic episode boundary that already exists in protocol state.

### Candidate pool

The candidate pool for an episode draw should come from the current authoritative epoch-group state, not from raw `ActiveParticipants` alone.

For multi-model correctness, the primary source should be the current model subgroup `EpochGroupData.ValidationWeights` and `MlNodes` for the model that owns the episode. That matches how validation-time model state is already derived today.

The root epoch group still matters for participant consensus weights and network totals, but model-local preserved sampling should use subgroup state.

The pool should be derived from:

- current root epoch group for consensus-weight context
- current model subgroup `ValidationWeights` / `MlNodes`
- existing protocol exclusions

It should use the same live-member filtering principle already used by validation snapshots. In particular, members excluded from the effective validation snapshot should not be sampled for preserved status.

It should not read from future participant sets, draft merge state, any successor epoch data, or live `HardwareNodes` changes that happened after the current epoch state was formed.

Every anchor fires right before some PoC. At that anchor the sampler reads the current epoch group (the active participants at that block) and picks preserved nodes out of them. The sampler does not care whether the PoC is regular or confirmation.

The resulting snapshot is scoped to that single PoC. The next PoC gets its own anchor and its own snapshot.

Reward math reads `vw.ConfirmationWeight`, which already folds every episode's contribution via min-take. Nothing on the claim path needs to consult the snapshot.

### State model

Keep `ActiveParticipants` stable for the whole epoch.

Under the new design:

- `timeslot_allocation[1]` in `ActiveParticipants` stays `false` and is deprecated for scheduling
- preserved scheduling moves to a single snapshot slot, overwritten at each episode

Storage is one `collections.Item[PreservedNodesSnapshot]` in the keeper. At every episode anchor (regular PoC `upcomingEpoch.PocStartBlockHeight` or confirmation `event.TriggerHeight`) the chain samples a fresh snapshot and overwrites the slot. The payload carries `episode_anchor_height` so readers can tell which episode they are looking at, but nothing keys storage by it.

That singleton shape is enough because there is exactly one active episode at a time:

- the broker always wants the current episode's snapshot
- the sampler always writes the next episode's snapshot immediately before the PoC starts
- reward math reads `vw.ConfirmationWeight`, not the snapshot

Existing rollover helpers that currently infer preserved carry from `POC_SLOT=true` must stop using `timeslot_allocation[1]` for scheduling. Episode snapshots do not imply an epoch-to-epoch carry bit in `ActiveParticipants`.

### Minimal implementation shape

The intended implementation should stay simple and mostly local:

- reuse the current preserved-node allocation logic from `AllocateMLNodesForPoC` instead of introducing a new sampling rule
- add one singleton preserved-snapshot slot and overwrite it at each episode anchor
- write that snapshot at the same transition that already records generation start
- keep `ActiveParticipants` stable and keep `timeslot_allocation[1]` deprecated for this purpose
- make the broker read the current episode snapshot instead of static epoch-long `TimeslotAllocation[1]`
- update other hot-path readers that currently depend on `TimeslotAllocation[1]`, especially `GetRandomExecutor` and broker-side PoC availability checks
- keep existing command order and most existing PoC phase logic unchanged

This avoids a mid-epoch rewrite of `ActiveParticipants` and keeps the change concentrated in state generation, snapshot reads, and broker filtering.

### Broker behavior

The broker flow should stay structurally the same:

1. `ShouldBeOperational`
2. preserved for the active episode
3. otherwise PoC

The only change is step 2. Instead of reading static epoch-long `TimeslotAllocation[1]`, the broker reads the current episode snapshot.

Each confirmation PoC event gets a fresh preserved sample. There is no carry-over requirement from one confirmation event to the next.

### Weight behavior

The old model splits reward weight into two additive slices: preserved (`POC_SLOT=true`) plus confirmation-measured (`POC_SLOT=false`). That split is gone.

Under the new model, `ConfirmationWeight` starts at the participant's full coefficient-adjusted MLNode total — as if every node were measured successfully in a confirmation event. Each confirmation event lowers it via min-take:

```
ConfirmationWeight := min(ConfirmationWeight, preserved(event) + measured(event))
```

Where `preserved(event)` is the weight of nodes in the event's preserved snapshot, and `measured(event)` is the weight the event actually observed on non-preserved nodes. Honest operation keeps the min equal to the initial full weight; any missed/invalid readings pull it down.

This means reward math reads a single `vw.ConfirmationWeight` field that already folds every episode. There is no separate "preserved weight" term in settlement.

The selection algorithm itself is not a new design space: it reuses the preserved-node allocation logic and constraints from `model_assignment.go`, run at each episode anchor against the episode candidate pool. The snapshot for the next episode is allowed to produce a different split than the current one.

----

## AI Usage 
- Cursor + Opus 4.6 + Gemini Pro 3.1 for Planning 
- Cursor + ChatGPT 5.4 / Claude Code + Opus 4.7 for implementation 

Cross reviews with all we have


qdanik:
@gmorgachev The concept is looks good. Do we want to have an option to see preserved mlnodes history during 1-2 epoch for debug purpose? Correct me if I wrong, but I don't see that is possible now

gmorgachev:
> @gmorgachev The concept is looks good. Do we want to have an option to see preserved mlnodes history during 1-2 epoch for debug purpose? Correct me if I wrong, but I don't see that is possible now

Hi @qdanik ! 
I think for debug purpose it's good enough to use query data for state history (-H "X-Cosmos-Block-Height: <BLOCK>" if use `chain-api`)
Archive nodes provides all such data

inference-chain/x/inference/module/model_assignment.go
```diff
@@ -1125,7 +1081,9 @@ func (ma *ModelAssigner) sampleEligibleParticipantsWithHistory(
 		return []string{}
 	}
 
-	seed := fmt.Sprintf("filter_%d_%s_%s", upcomingEpoch.Index, allParticipantsHashStr, modelId)
+	// Episode anchor height is mixed into the seed so each PoC episode (regular and each
+	// confirmation event) produces an independent sample, preserving late-binding.
+	seed := fmt.Sprintf("filter_%d_%d_%s_%s", upcomingEpoch.Index, anchorHeight, allParticipantsHashStr, modelId)
```
patimen, line 1107: Not super efficient here, better to use the bytes... but, this is easy to understand.
This should only really be done once per PoC? Or is this being called somewhere hot?

inference-chain/x/inference/module/model_assignment.go
```diff
@@ -455,105 +455,101 @@ func (ma *ModelAssigner) setModelsForParticipants(ctx context.Context, participa
...
 		ma.LogInfo("PocSlotAllocation is nil or 0, using default 0.5", types.Allocation, "flow_context", FlowContext, "sub_flow_context", SubFlowContext, "step", "default_allocation")
 		allocationFraction = &types.Decimal{Value: 5, Exponent: -1}
 	}
 
+	rootData, found := ma.keeper.GetEpochGroupData(ctx, epoch.Index, "")
+	if !found {
+		return types.PreservedNodesSnapshot{EpisodeAnchorHeight: anchorHeight}, nil
+	}
+
+	sortedModelIds := append([]string(nil), rootData.SubGroupModels...)
```
patimen: We shuold use slices.Clone if that's the intent here (I assume it is).

decentralized-api/internal/event_listener/new_block_dispatcher.go
```diff
@@ -357,6 +357,10 @@ func (d *OnNewBlockDispatcher) handlePhaseTransitions(epochState chainphase.Epoc
 		logging.Error("Failed to update node with epoch data, skipping phase transitions.", types.Stages, "error", err)
 		return
 	}
+	if err := d.nodeBroker.EnsurePreservedMembershipCached(&epochState); err != nil {
+		logging.Error("Failed to update broker preserved membership, skipping phase transitions.", types.Stages, "error", err)
```
patimen: Are we sure we want to skip the entire transition if this one gRPC call fails? Is there anything we can continue doing? We would skip ALL the logic below for all of the different transitions.

patimen (commented):
I looked through all the perf implications... I think we're managing them well enough, and certainly this isn't going to make anything measurable worse.
I double checked the RNG determinism - looks good.
PreservedNodesSnapshot will start out empty... that was handled correctly.
My biggest concern is dropping all state transitions if there is a gRPC failure in getting the current state, even though that is unlikely to be a real problem. Maybe we want to log an error and continue instead?

inference-chain/x/inference/module/model_assignment.go
gmorgachev: Only once per PoC / CPoC per model

inference-chain/proto/inference/inference/preserved_nodes_snapshot.proto
```diff
@@ -0,0 +1,14 @@
...
+option go_package = "github.com/productscience/inference/x/inference/types";
+
+message PreservedNodesSnapshot {
+  int64 episode_anchor_height = 1;
+  repeated ModelPreservedNodes model_preserved_nodes = 2;
+}
+
+message ModelPreservedNodes {
+  string model_id = 1;
+  repeated string preserved_node_ids = 2;
```
DimaOrekhovPS: I may be missing something, but why aren't we including participant id? Node ids aren't unique accros all participants.

From my understanding it should be smth like:

```
message ModelPreservedNodes {
  string model_id = 1;
  repeated GlobalMlNodeId preserved_node_ids = 2;
}

message GlobalMlNodeId {
    string participant_id = 1;
    string mlnode_id = 2;
}
```

decentralized-api/internal/event_listener/new_block_dispatcher.go
gmorgachev: Agree, warning should be good enough

inference-chain/proto/inference/inference/preserved_nodes_snapshot.proto
gmorgachev: i think you're right. forgot about it, fixing

DimaOrekhovPS (approved):

PR #1096, opened 2026-04-21
Fix gas estimation 3

0xgonka:
## Summary

Follow-up to #1088. That PR introduced the sync-loop `Set*` pattern for four BLS/bridge state structs (EpochBLSData, GroupKeyValidationState, ThresholdSigningRequest, BridgeTransaction), but missed nine call sites where a handler fetched a rehydrated struct, modified one field, and called `Set*` with the split fields still populated. The sync loops then re-wrote every sub-key, costing O(N × per-entry-size) WritePerByte gas per call.

Maria's testnet-10 trace at block 5164: 4 of 8 participants' `MsgSubmitGroupKeyValidationSignature` failing with `out of gas in location: WritePerByte; gasWanted: 10000000, gasUsed: 10373936: out of gas`. The 4 that succeeded were those that fit under the signature threshold (~2M gas); the 4 that failed were the ones whose submissions crossed the threshold and hit the threshold-reached branch. This PR closes every remaining site.

### Production code — null split fields before `Set*` in every remaining hot path

- **`msg_server_group_validation.go`** (threshold-reached path) — null `DealerParts` / `VerificationSubmissions` / `DealerComplaints` before `SetEpochBLSData`. This is Maria's exact OOG site.
- **`msg_server_dispute.go`** (`RespondDealerComplaints`) — replaced the full `SetEpochBLSData` call with per-complaint `SetDealerComplaint`. The handler was only updating individual complaint entries; the full Set was triggering sync loops across DealerParts, VerificationSubmissions, and every complaint for what is effectively a single-entry update.
- **`phase_transitions.go`** (all four DKG transitions: `TransitionToVerifyingPhase`, `transitionFromVerifyingToDisputing`, `finalizeDisputingPhase`, `MarkDKGAsFailed`) — stash/null/restore pattern around `SetEpochBLSData`. Restore after Set so the subsequent `EmitTypedEvent` still carries the full state for downstream consumers. `transitionFromVerifyingToDisputing` intentionally preserves `DealerComplaints` populated because `DeleteDealerComplaintsForEpoch` just cleared the sub-keys and the filtered subset MUST be re-synced.
- **`threshold_signing.go`** (three sites) — `checkThresholdAndAggregate` nulls `PartialSignatures` before the terminal `storeThresholdSigningRequest`; `finalizeFailedThresholdSigningRequest` nulls at the top of the single failure funnel (covers both its own Set and the propagation to `maybeCloseRetryAfterFailedPostProcess`); `CancelThresholdSignature` nulls for status-only changes.

### Tests — pin the invariant so future regressions fire immediately

- **`x/bls/keeper/gas_regression_test.go`** — three gas-bounded tests, one per sync-loop Set*:
  - `TestSetEpochBLSData_GasBoundedWhenSplitFieldsNulled` — seeds N=16 realistic-sized dealer parts + verification submissions, calls `SetEpochBLSData` with split fields nulled, asserts gas < 500k.
  - `TestSetGroupKeyValidationState_GasBoundedWhenPartialSignaturesNulled` — same pattern for the exact function Maria's trace hit.
  - `TestStoreThresholdSigningRequest_GasBoundedWhenPartialSignaturesNulled` — covers all threshold-signing call sites.
- **`x/inference/keeper/bridge_transaction_gas_regression_test.go`** — `TestSetBridgeTransaction_GasBoundedWhenValidatorsNulled` covers `msg_server_bridge_exchange.go` paths.
- Each test has a companion sanity check asserting the populated-slice path costs at least 3–10× more gas, so if someone later removes the sync loop entirely (making both paths equal), the test fails loudly rather than silently passing.

### Infra — unblock local repro at N > 3

- **`local-test-net/node_payload_mock_server_join{3,4,5,6,7}.json`** — aligned with `join1/join2`'s map schema for `models`. The old array schema (`"models": ["model-id"]`) crashes the DAPI at startup with `json: cannot unmarshal array into Go struct field InferenceNodeConfig.models of type map[string]apiconfig.ModelConfig`. Went unnoticed because default `launch.sh` only spins up 2 join nodes, and the pre-split BLS test suite never exercised N > 3. Discovered while scaling `BLSDKGSuccessTest` locally to verify the fix; unrelated to the gas work itself but blocks anyone else from running testermint at N > 3.

## Verification

Locally scaled `BLSDKGSuccessTest` to N=8 participants (`setupLocalCluster(7, ..., reboot = true)`) — reproduces Maria's testnet scenario at the exact host count. On this branch:

- `BUILD SUCCESSFUL in 11m 13s`
- DKG reached `DKG_PHASE_SIGNED` at epoch 7 with 8 participants
- Zero `out of gas` / `WritePerByte` in any node log
- `[bls-verifier] Successfully processed DKG completion subsystem=BLS epochID=7 ... phase=DKG_PHASE_COMPLETED`

The scaled test itself is NOT committed (reverted before push); only the payload schema fix that unblocks it is.

## Test plan

- [x] Unit tests — gas regression suite passes locally on Go 1.24
- [x] Local N=8 integration — `BLSDKGSuccessTest` at 8 participants reaches SIGNED phase with zero OOG
- [ ] Testnet re-run at N≥8 — confirms Maria's 4-of-8 dropout is gone
- [ ] If dropouts still occur post-merge, investigate separately — Gleb and Maria flagged on Slack that multi-mlnode-per-api config complexity may be an orthogonal failure mode unrelated to gas


0xgonka:
running integration tests: https://github.com/gonka-ai/gonka/actions/runs/24738082425

inference-chain/x/bls/keeper/msg_server_dispute.go
```diff
@@ -77,10 +77,16 @@ func (ms msgServer) RespondDealerComplaints(ctx context.Context, msg *types.MsgR
 		complaint.ResponseShareBytes = response.ResponseShareBytes
 		complaint.ResponseOpeningMaterial = response.ResponseOpeningMaterial
 		epochBLSData.DealerComplaints[complaintIndex] = complaint
-	}
 
-	if err := ms.SetEpochBLSData(sdkCtx, epochBLSData); err != nil {
-		return nil, status.Error(codes.Internal, fmt.Sprintf("failed to store updated epoch %d BLS data: %v", msg.EpochId, err))
+		// Persist the updated complaint directly via its sub-key. We don't
```
patimen: Maybe a little less wordy on the comment here.

inference-chain/x/bls/keeper/phase_transitions.go
```diff
@@ -88,10 +88,23 @@ func (k Keeper) TransitionToVerifyingPhase(ctx sdk.Context, epochBLSData *types.
...
-		// Store updated epoch data
+		// Store updated epoch data. Null the split fields before Set so
+		// SetEpochBLSData's sync loops skip the O(N × per-entry-size)
+		// re-sync of data already in sub-keys — only DkgPhase and
+		// VerifyingPhaseDeadlineBlock are changing. Restore after Set so
+		// the event below still carries the full state.
+		dealerParts := epochBLSData.DealerParts
+		verSubs := epochBLSData.VerificationSubmissions
+		complaints := epochBLSData.DealerComplaints
+		epochBLSData.DealerParts = nil
```
patimen: It seems this "stash-nil-restore" pattern is repeated a lot... sure this should be a method? Otherwise we're relying on future changes to recognize the pattern.

patimen (commented):
I think we should rethink this abstraction, or at least rename/split the methods so the storage behavior is explicit.

Right now GetEpochBLSData returns a rehydrated “full” object, but SetEpochBLSData has hidden side effects: if DealerParts, VerificationSubmissions, or DealerComplaints are populated, it rewrites their separate substores before storing the stripped base record. That makes the hot path depend on callers remembering to nil fields before calling SetEpochBLSData, which is pretty unnatural and easy to regress.

Could we make this explicit instead? For example, keep/add something like SetEntireEpochBLSData or SyncAndSetEpochBLSData for genesis/import/migration/full replacement cases, and add a separate SetEpochBLSBaseData/UpdateEpochBLSBaseData for the common hot-path case where only phase/deadlines/signatures/base fields change. Then callers don’t need the nil-then-restore pattern to select which stores get written.

patimen (approved):
This should do it... nothing major. We might want to look at how we do PartialSignatures and mirror the separate API at some point, but it's not the essence of this PR so let's do it later.

PR #1099, opened 2026-04-22
Fixes

gmorgachev:


inference-chain/x/inference/module/delegation_pipeline.go
```diff
@@ -851,6 +853,77 @@ func (am AppModule) delegationAdjustmentParams(params types.Params) DelegationAd
...
+		parts = append(parts, fmt.Sprintf("%s:%d", vp.ModelId, vp.VotingPower))
+	}
+	return strings.Join(parts, ",")
+}
+
+// emitWeightPipelineLogs prints one line per group and one line per
+// participant so any final weight can be reproduced from the log alone.
+// The gap between after_penalty and final is the combined effect of
+// collateral adjustment and power-cap (both rarely active).
+func emitWeightPipelineLogs(
```
patimen, line 884: This seems like a lot of logging to add? Should these be Debug? I guess most operators are at Warn now, so this shouldn't have too much impact?
Maybe a check for LogLevel at the beginning and an immediate return of we're not at least Info?

patimen (approved):
Only one minor suggestion on logging.

inference-chain/x/inference/module/delegation_pipeline.go
gmorgachev: thought about it when added. it seems crucial data i think we first must get rid of everything else

PR #1100, opened 2026-04-22
fix: prevent uint64 wrap in settle amount sums 

### LINKED ISSUE #784: [P2] Possible underfunded issues

0xMayoor:
RewardCoins + WorkCoins gets summed as raw uint64 in three places. if the pair grows past MaxUint64 the sum wraps and you either skip a participant whose payout looks like 0, return a small wrapped total from `GetTotalCoins`, or have dapi recovery silently decide there are no unclaimed rewards. the int64 cap in `GetTotalCoins` doesn't help — wrap already happened in uint64 space.

producer side `bitcoin_rewards.go` writes `^uint64(0)` into RewardCoins when the big.Int division can't fit. unreachable under default params but the sentinel wires the wrap-trap into
the type.

fix: cap producer to MaxInt64 instead of the MaxUint64 sentinel, switch the three sum sites to `bits.Add64` and saturate on carry. `accountsettle.go` only skips when the sum is genuinely 0, never on overflow.

### DIFF
--- a/decentralized-api/internal/startup/reward_recovery.go
+++ b/decentralized-api/internal/startup/reward_recovery.go
@@ -8,6 +8,8 @@ import (
 	"decentralized-api/internal/seed"
 	"decentralized-api/internal/validation"
 	"decentralized-api/logging"
+	"math"
+	"math/bits"
 	"sync/atomic"
 	"time"
 
@@ -137,7 +139,11 @@ func (c *RewardRecoveryChecker) AutoRewardRecovery() {
 	}
 
 	settleAmount := settleAmountResp.SettleAmount
-	totalAmount := settleAmount.RewardCoins + settleAmount.WorkCoins
+	sum, carry := bits.Add64(settleAmount.RewardCoins, settleAmount.WorkCoins, 0)
+	totalAmount := sum
+	if carry != 0 {
+		totalAmount = math.MaxUint64
+	}
 	logging.Info("[AutoRewardRecovery] Found settle amount for participant", types.Claims,
 		"address", address,
 		"rewardCoins", settleAmount.RewardCoins,
@@ -146,7 +152,7 @@ func (c *RewardRecoveryChecker) AutoRewardRecovery() {
 		"epochIndex", settleAmount.EpochIndex)
 
 	// Check if we have unclaimed rewards (totalAmount > 0 indicates pending rewards)
-	if totalAmount <= 0 {
+	if totalAmount == 0 {
 		logging.Info("[AutoRewardRecovery] No unclaimed rewards found", types.Claims, "address", address, "totalAmount", totalAmount)
 		return
 	}
--- a/inference-chain/x/inference/keeper/accountsettle.go
+++ b/inference-chain/x/inference/keeper/accountsettle.go
@@ -3,6 +3,7 @@ package keeper
 import (
 	"context"
 	"fmt"
+	"math/bits"
 
 	"cosmossdk.io/log"
 	sdk "github.com/cosmos/cosmos-sdk/types"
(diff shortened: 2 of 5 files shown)


0xgonka (commented):
Narrow, targeted fix for a real uint64-wrap bug in the RewardCoins+WorkCoins sum path. The three sum sites + the producer-side sentinel are handled correctly and the semantics of "skip only when genuinely zero" is preserved. Looks ready to merge with a couple of minor suggestions.

**Must fix:**
- None.

**Should fix:**
- Test coverage is thin — [`settle_amount_extreme_test.go` L15-30](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/keeper/settle_amount_extreme_test.go#L15-L30) exercises `GetTotalCoins` only. Worth adding a case that (a) confirms `SettleAccounts` does *not* skip a participant whose `WorkCoins+RewardCoins` wraps to zero in naive uint64 arithmetic, since that was the originally-claimed behavioral hazard, and (b) confirms the producer cap in `bitcoin_rewards.go` writes `MaxInt64` rather than `^uint64(0)`. The function being covered is the shared one, but the two handler-side behaviors are what the PR body actually argues for.

**Nits:**
- [`bitcoin_rewards.go` L771-778](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/keeper/bitcoin_rewards.go#L771-L778): fine to keep, but consider switching the cap to a named constant (e.g. `maxSettleCoins = uint64(math.MaxInt64)`) and using it both here and in `settle_amount.go`. It makes the downstream invariant ("RewardCoins stored by the chain never exceeds MaxInt64") explicit and greppable.
- [`accountsettle.go` L268-270](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/keeper/accountsettle.go#L268-L270): the comment "a wrapping sum to 0 would skip a real payout" is good context. A one-liner pointing to `GetTotalCoins` for the companion saturation would help future readers see these as paired.
- [`reward_recovery.go` L155](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/decentralized-api/internal/startup/reward_recovery.go#L155): the old comment still says "totalAmount > 0 indicates pending rewards" but `totalAmount` is now `uint64`, so `> 0` and `== 0` are equivalent. Harmless; just stale wording.

**Looks correct:**
- Producer cap swap from `^uint64(0)` → `uint64(math.MaxInt64)` at [`bitcoin_rewards.go` L778](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/keeper/bitcoin_rewards.go#L778) is the right move. In addition to the documented "wrap trap," `^uint64(0)` also silently became `-1` through the downstream `int64(settleAmount.GetRewardCoins())` cast at [`msg_server_claim_rewards.go` L99](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/keeper/msg_server_claim_rewards.go#L99); `MaxInt64` casts cleanly. Free collateral win.
- `bits.Add64` + explicit `carry` handling in all three sum sites (`settle_amount.go`, `accountsettle.go`, `reward_recovery.go`) is the idiomatic Go pattern for this — cleaner than `if a > MaxUint64 - b`.
- Skip-logic at [`accountsettle.go` L269-273](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/keeper/accountsettle.go#L269-L273) correctly requires `paymentCarry == 0 && paymentSum == 0` — wrap-to-zero is no longer ambiguous with genuine-zero. The only case that now short-circuits is truly `(0, 0)`.
- `GetTotalCoins` at [`settle_amount.go` L9-14](https://github.com/gonka-ai/gonka/blob/f78104e5e08fd8001379158f48a6c1e0644d4b88/inference-chain/x/inference/types/settle_amount.go#L9-L14) handles both overflow branches (`carry != 0`) and the in-range-but-over-int64 case (`sum > MaxInt64`). Both reach the same `MaxInt64` saturation, which matches the implicit chain invariant that claim amounts fit in int64.
- PR body accurately describes the scope and the "unreachable under default params" qualifier on the producer-side cap — no overclaiming of severity.

tcharchian:
@0xgonka @0xMayoor, does this PR need to be rebased for the next release v0.2.14?

patimen (approved):
Also straightforward.

PR #1101, opened 2026-04-22
fix: widen ShouldValidate to uint64 to prevent silent validation skip

### LINKED ISSUE #646: Avoid truncation for large validation weights
#1101

### LINKED ISSUE #784: [P2] Possible underfunded issues

0xMayoor:
`getMustBeValidatedInferences` runs `safeUint32FromInt64` on `totalWeight`, validator weight, and executor weight before passing them to `calculations.ShouldValidate`. if the per-model summed weight crosses MaxUint32 (~4.29e9) the cast errors and the inference gets `continue`d — the claim still succeeds with an empty mustBeValidated, so the validator sidesteps validation duty without failing the message.

`ShouldValidate` itself only ever does decimal math on those values, the uint32 cap is type-choice not algorithmic. widen the signature to uint64 and let the sampling proceed at the real width. `decimal.NewFromUint64` instead of `decimal.NewFromInt(int64(x))` — the latter wraps negative once x crosses MaxInt64. dapi caller in `inference_validation.go` had the same uint32 truncation on uint64 fields, fixed there too.

added `safeUint64FromInt64` to keep negative-weight rejection at the call site (ValidationWeight.Weight is signed int64). regression test in keeper sets summed weight to 4.4e9 and asserts the inference IS sampled — was empty before.

ground for fix is the `MaxIndividualPowerPercentage = 0.25` per-host cap from #1068. it limits per-host concentration but multiple hosts can still aggregate the per-model totalWeight past MaxUint32 as the network grows.

### DIFF
--- a/decentralized-api/internal/validation/inference_validation.go
+++ b/decentralized-api/internal/validation/inference_validation.go
@@ -106,9 +106,9 @@ func (s *InferenceValidator) shouldValidateInference(
 	shouldValidate, message := calculations.ShouldValidate(
 		seed,
 		inferenceDetails,
-		uint32(inferenceDetails.TotalPower),
-		uint32(validatorPower),
-		uint32(inferenceDetails.ExecutorPower),
+		inferenceDetails.TotalPower,
+		validatorPower,
+		inferenceDetails.ExecutorPower,
 		validationParams,
 		false)
 
--- a/inference-chain/x/inference/calculations/should_validate.go
+++ b/inference-chain/x/inference/calculations/should_validate.go
@@ -14,9 +14,9 @@ import (
 func ShouldValidate(
 	seed int64,
 	inferenceDetails *types.InferenceValidationDetails,
-	totalPower uint32,
-	validatorPower uint32,
-	executorPower uint32,
+	totalPower uint64,
+	validatorPower uint64,
+	executorPower uint64,
 	validationParams *types.ValidationParams,
 	debug bool,
 ) (bool, string) {
@@ -31,7 +31,7 @@ func ShouldValidate(
 	// algebraic simplification/removal of temp variables
 	targetValidations := maxValidationAverage.Sub(rangeSize.Mul(executorReputation))
 	// 100% rep will be minValidationAverage, 0% rep will be maxValidationAverage
-	ourProbability := targetValidations.Mul(decimal.NewFromInt(int64(validatorPower))).Div(decimal.NewFromInt(int64(totalPower - executorPower)))
+	ourProbability := targetValidations.Mul(decimal.NewFromUint64(validatorPower)).Div(decimal.NewFromUint64(totalPower - executorPower))
 	if ourProbability.GreaterThan(one) {
 		ourProbability = one
 	}
--- a/inference-chain/x/inference/calculations/should_validate_test.go
+++ b/inference-chain/x/inference/calculations/should_validate_test.go
@@ -18,9 +18,9 @@ func TestShouldValidate(t *testing.T) {
 		name                 string
 		seed                 int64
 		inferenceDetails     *types.InferenceValidationDetails
-		totalPower           uint32
-		validatorPower       uint32
-		executorPower        uint32
+		totalPower           uint64
+		validatorPower       uint64
+		executorPower        uint64
 		expectedResult       bool
 		expectedProbability  float64
 		minValidationAverage float64
--- a/inference-chain/x/inference/keeper/claim_rewards_weight_overflow_test.go
+++ b/inference-chain/x/inference/keeper/claim_rewards_weight_overflow_test.go
@@ -1,0 +1,57 @@
+package keeper_test
+
+import (
+	"math"
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/productscience/inference/testutil"
+	"github.com/productscience/inference/x/inference/keeper"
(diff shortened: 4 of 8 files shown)


0xgonka (commented):
Widens the three weight inputs to `ShouldValidate` from uint32 to uint64 so a summed per-model weight above MaxUint32 no longer trips the `safeUint32FromInt64` guard and silently drops the inference from the claim's validation set. Clean, targeted fix; ships as a consensus change via the v0.2.12 upgrade.

**Must fix:**
- None.

**Should fix:**
- `safeUint64FromInt64` has no unit test — [`safecast_test.go`](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/inference-chain/x/inference/keeper/safecast_test.go) covers the other three casts with a table (zero / negative / MaxInt64 / etc). A one-line addition mirroring `TestSafeUint32FromInt64` would keep the file consistent and catches anyone re-introducing a negative-weight bug later.

**Nits:**
- After this PR, `safeUint32FromInt64` has no non-test callers (grep shows only `safecast_test.go:58`). Either note in a comment that it's retained for reuse or remove it — right now it looks live but isn't. Same situation for `safeUint32FromUint64`. Not blocking; just a source of "wait, who calls this?" later.
- [`claim_rewards_weight_overflow_test.go` L26](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/inference-chain/x/inference/keeper/claim_rewards_weight_overflow_test.go#L26) uses a single weight point (2.2e9 × 2). The assertion `require.NotEmpty(got)` relies on the default `ValidationParams` yielding a high enough `ourProbability` that `DeterministicFloat(42, "inf-weight-sum-above-uint32")` lands below. Fine today, but a future tuning of default `MinValidationAverage`/`MaxValidationAverage` could flip it without any real regression. Either pin the params explicitly in the test or assert on the absence-of-error + "didn't hit the negative-weight log" path instead of sampling outcome.

**Looks correct:**
- Widening is at the right layer. `InferenceValidationDetails.TotalPower` / `ExecutorPower` are already `uint64` in the proto ([`inference_validation_details.proto` L12,14](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/inference-chain/proto/inference/inference/inference_validation_details.proto#L12-L14)), and the keeper-side `ValidationWeight.Weight` summation produces `int64`. `safeUint64FromInt64` preserves the negative-weight rejection for signed inputs and drops the artificial MaxUint32 ceiling. Nothing else in the signature chain is narrower.
- `decimal.NewFromUint64(validatorPower)` / `decimal.NewFromUint64(totalPower - executorPower)` at [`should_validate.go` L34](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/inference-chain/x/inference/calculations/should_validate.go#L34) avoids the latent wrap that would have re-emerged post-widening: `decimal.NewFromInt(int64(x))` wraps negative at `x > MaxInt64`. `NewFromUint64` exists in `shopspring/decimal v1.4.0` (the pinned version), so no module bump needed.
- Subtraction `totalPower - executorPower` is guarded by the early `if totalPower <= executorPower` return at [`should_validate.go` L23](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/inference-chain/x/inference/calculations/should_validate.go#L23), so uint64 underflow is impossible by the time the `Div` runs. Same guard applied client-side at [`inference_validation.go` L101](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/decentralized-api/internal/validation/inference_validation.go#L101).
- DAPI caller cleanup at [`inference_validation.go` L106-114](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/decentralized-api/internal/validation/inference_validation.go#L106-L114) removes a *separate* truncation — `TotalPower` and `ExecutorPower` on the response type were already `uint64`, so the old `uint32(...)` was narrowing before passing. Fix removes that redundancy in lockstep with the chain-side widening.
- `claim_rewards_benchmark_test.go` keeps compiling unchanged because the literal args (`1000, 100, 50`) are untyped and adapt to the new uint64 parameter — no hidden caller break.
- Log message change from `"Weight overflow in validation sampling"` → `"Negative weight in validation sampling"` at [`msg_server_claim_rewards.go` L481-496](https://github.com/gonka-ai/gonka/blob/5b73abe1f1f1d5f20efe0010ff4818d86744bf34/inference-chain/x/inference/keeper/msg_server_claim_rewards.go#L481-L496) accurately reflects what `safeUint64FromInt64` now rejects (only the sign bit, not magnitude). Will make on-call faster at identifying the real cause.
- This is a consensus-level behavior change (pre-PR nodes would `continue` past inferences whose summed weight exceeds MaxUint32; post-PR nodes include them). Landing it on the `upgrade-v0.2.12` branch ensures the flip happens at a single upgrade height rather than a mixed-version fork.

tcharchian:
@0xgonka @0xMayoor, does this PR need to be rebased for the next release v0.2.14?

patimen (commented):
Looks straightforward to me.

patimen (approved):

PR #1106, opened 2026-04-23
fix: Certik audit fixes (GEB-61)

### LINKED ISSUE #1108: GEB-61

GLiberman:
From audit report:

The BLS snapshot includes one primary secp256k1 key plus every authz-derived warm key in AllowedSecp256K1PublicKeys, with no independent cap in the collection path. Dealer generation then emits one encrypted share per slot per allowed key, while chain-side shape validation requires that exact snapshot-derived count. At the same time, MsgSubmitDealerPart.ValidateBasic() enforces a hard limit of 16384 encrypted shares for each participant row.

This creates a hard liveness failure. If an active participant accumulates enough warm keys that slotCount * (1 + len(AllowedSecp256K1PublicKeys)) > 16384, every dealer is forced to build an oversized ciphertext array for that participant, and the dealer part becomes invalid before execution. Because all dealers must include shares for every participant, a single participant with excessive warm-key fanout can block dealer-part submission and stall the DKG round.

- Snapshot pruning in module.go (line 1513) prevents oversized warm-key fanout from entering the epoch snapshot.
- Keeper-side preflight in dkg_initiation.go (line 57) refuses DKG if the derived shape is still impossible.
- The shared bound in encrypted_shares_shape.go (line 26) keeps dealer-path validation aligned with the message limit.

### DIFF
--- a/inference-chain/testutil/keeper/expected_keepers_mocks.go
+++ b/inference-chain/testutil/keeper/expected_keepers_mocks.go
@@ -1298,6 +1298,21 @@ func (mr *MockBlsKeeperMockRecorder) GetEpochBLSData(ctx, epochID any) *gomock.C
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetEpochBLSData", reflect.TypeOf((*MockBlsKeeper)(nil).GetEpochBLSData), ctx, epochID)
 }
 
+// GetParams mocks base method.
+func (m *MockBlsKeeper) GetParams(ctx context.Context) (types4.Params, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "GetParams", ctx)
+	ret0, _ := ret[0].(types4.Params)
+	ret1, _ := ret[1].(error)
+	return ret0, ret1
+}
+
+// GetParams indicates an expected call of GetParams.
+func (mr *MockBlsKeeperMockRecorder) GetParams(ctx any) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetParams", reflect.TypeOf((*MockBlsKeeper)(nil).GetParams), ctx)
+}
+
 // GetSigningStatus mocks base method.
 func (m *MockBlsKeeper) GetSigningStatus(ctx types1.Context, requestID []byte) (*types4.ThresholdSigningRequest, error) {
 	m.ctrl.T.Helper()
--- a/inference-chain/x/bls/keeper/dkg_initiation.go
+++ b/inference-chain/x/bls/keeper/dkg_initiation.go
@@ -54,6 +54,21 @@ func (k Keeper) InitiateKeyGenerationForEpoch(ctx sdk.Context, epochID uint64, f
 		return fmt.Errorf("failed to assign slots: %w", err)
 	}
 
+	// Validate shape constraints before initiating
+	if len(blsParticipants) > types.MaxEncryptedSharesParticipantsCount {
+		return fmt.Errorf("number of participants %d exceeds maximum %d", len(blsParticipants), types.MaxEncryptedSharesParticipantsCount)
+	}
+	if tSlotsDegree+1 > uint32(types.MaxDealerPartCommitmentsCount) {
+		return fmt.Errorf("tSlotsDegree+1 (%d) exceeds maximum commitments count %d", tSlotsDegree+1, types.MaxDealerPartCommitmentsCount)
+	}
+
+	for _, p := range blsParticipants {
+		_, err := expectedEncryptedSharesCount(p)
+		if err != nil {
+			return fmt.Errorf("failed to validate encrypted share bounds for %s: %w", p.Address, err)
+		}
+	}
+
 	// Calculate phase deadlines
 	currentHeight := ctx.BlockHeight()
 	dealingPhaseDeadline := currentHeight + params.DealingPhaseDurationBlocks
--- a/inference-chain/x/bls/keeper/dkg_initiation_test.go
+++ b/inference-chain/x/bls/keeper/dkg_initiation_test.go
@@ -1,6 +1,7 @@
 package keeper_test
 
 import (
+	"fmt"
 	"testing"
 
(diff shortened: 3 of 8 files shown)


0xgonka (approved):
Three-layer fix for the GEB-61 GC liveness DoS: a participant accumulating enough authz warm keys could push `slotCount × (1 + len(AllowedSecp256K1PublicKeys))` past the 16384 per-participant share cap, making every dealer's `MsgSubmitDealerPart` invalid and stalling DKG. The fix prunes warm keys at snapshot time, gates DKG initiation on a shape preflight, and aligns the dealer-side bounds check with the existing message-level cap. Defense-in-depth is sound; the cap-derivation choice is conservative but correct.

**Must fix:**
- None.

**Should fix:**
- The `maxAdditionalKeys` cap is computed once globally as `MaxEncryptedSharesPerParticipantCount / ITotalSlots - 1` ([`module.go` L1520-1527](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/inference/module/module.go#L1520-L1527)) using **worst-case slot ownership** (one participant owning all slots). With production `ITotalSlots=1000` and max=16384, every participant is capped at 15 additional keys regardless of their actual slot count. A participant with 1 slot could safely have ~16k warm keys but is restricted to 15. This is a real correctness/UX tradeoff worth flagging — if hosts in production routinely register >15 warm keys, they will be silently truncated. Two options:
  1. Document the 15-warm-key practical limit prominently in host docs.
  2. Move the truncation into the BLS keeper after `AssignSlots`, so each participant gets `MaxEncryptedSharesPerParticipantCount / actualSlotCount - 1`. More invasive but more accurate.
  
  Either is fine — but operators need to know the number.

**Nits:**
- [`dkg_initiation.go` L61](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/bls/keeper/dkg_initiation.go#L61): `tSlotsDegree+1 > uint32(types.MaxDealerPartCommitmentsCount)` — `tSlotsDegree` is `uint32`, so `tSlotsDegree+1` wraps if `tSlotsDegree == math.MaxUint32`. In practice `tSlotsDegree` comes from `AssignSlots` and is bounded by `ITotalSlots` (uint32 too, but realistically ≤1000), so unreachable. Tighten with `uint64(tSlotsDegree)+1 > uint64(types.MaxDealerPartCommitmentsCount)` for paranoia.
- [`module.go` L1670-1675](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/inference/module/module.go#L1670-L1675): the WARN log says `"Pruning excess additional BLS participant public keys"` but doesn't mention the consequence — that the truncated warm keys won't be able to decrypt their share for this epoch. A line like `"truncated keys will not receive shares this epoch"` would make this oncall-friendly when a host complains their warm key isn't working.
- Test [`dkg_initiation_test.go` L412-435](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/bls/keeper/dkg_initiation_test.go#L412-L435) has stray trailing whitespace on L424; clean up if you re-touch.
- [`module.go` L1517](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/inference/module/module.go#L1517) logs an error and continues with `maxAdditionalKeys = 0` when `GetParams` fails. That means a transient param-fetch failure silently strips ALL warm keys for the epoch (every participant runs primary-only). Consider returning instead — if BLS params can't be read, DKG initiation is broken anyway and silently degrading misleads operators looking at why DKG completed with the wrong key set.

**Looks correct:**
- The choice to use `MaxEncryptedSharesPerParticipantCount` (the per-participant share cap from `MsgSubmitDealerPart.ValidateBasic`) as the alignment point at [`encrypted_shares_shape.go` L26-28](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/bls/keeper/encrypted_shares_shape.go#L26-L28) is exactly right — the dealer-side `expectedEncryptedSharesCount` and the chain-side message validation now share a constant, so the "expected count exceeds message cap" mismatch that caused GEB-61 cannot recur silently.
- Three-layer defense (snapshot prune → keeper preflight → shape check) is the right shape: the snapshot prune normally prevents the issue at the entry point; the keeper preflight catches it for any in-flight tx that bypassed the snapshot; the shape check is the last line in dealer validation. Each layer has a distinct trigger and could catch a regression independently.
- Constant renaming (`maxXxx` → `MaxXxx` at [`message_submit_dealer_part.go` L13-15](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/bls/types/message_submit_dealer_part.go#L13-L15)) is the minimum needed to share the bound across packages — no proto change, no behavior change. Good package-boundary discipline.
- Truncation at [`module.go` L1670-1676](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/inference/module/module.go#L1670-L1676) happens *after* the `slices.SortFunc` sort, so the kept set is deterministic across nodes — the chain wouldn't fork due to non-deterministic warm-key selection. This is the kind of detail that's easy to get wrong and was gotten right.
- New `GetParams` on the `BlsKeeper` interface ([`expected_keepers.go` L139](https://github.com/gonka-ai/gonka/blob/0ca35d50d02137df0ed0991ecd1b0a021d7b383e/inference-chain/x/inference/types/expected_keepers.go#L139)) is the minimum surface change needed — doesn't expose set/write methods to the caller.
- Three test cases cover the three trigger paths (oversize shape rejected, oversize commitments rejected, snapshot prunes correctly) — coverage matches the three fix layers.
- Targeting `upgrade-v0.2.12` ensures the new validation activates atomically across validators; no mixed-version DKG fork risk.

0xgonka:
running integration tests: https://github.com/gonka-ai/gonka/actions/runs/24906874781

PR #1112, opened 2026-04-24
fix: Certik audit fixes (GEB-29) + Selfvote

### LINKED ISSUE #1107: GEB-29

GLiberman:
Per Certik audit:
In the commit [a1119b819e0ced3383f0b272b7c04487d24d12f5](https://github.com/gonka-ai/gonka/commit/a1119b819e0ced3383f0b272b7c04487d24d12f5), the changes made to DetermineValidDealersWithConsensus() require a strict majority of total slots (totalSlots/2 + 1) to approve each dealer, but it always excludes the dealer’s own verification vote. As a result, any dealer controlling at least half of the total slots cannot mathematically reach quorum even if every non-self verifier approves and the dealer submitted valid parts.

This invalid result flows into transitionFromVerifyingToDisputing(), which sums the slots of candidateValidDealers and fails the epoch when they do not exceed half of total slots. In skewed slot distributions such as 50/50 or 60/40, an honest high-weight dealer can be excluded for arithmetic reasons alone, causing the DKG epoch to enter FAILED instead of progressing to DISPUTING.

We would narrow the fix recommendation a bit:

- Counting the dealer’s own vote is the cleaner fix. That aligns the approval threshold with the available voting weight and removes the impossible-to-satisfy case.

- Not lower the downstream candidateDealerSlots > totalSlots/2 rule by itself. That threshold intentional, because later threshold signing requires t+1 slots, and with default params that is also a strict majority: threshold_signing.go (line 563).

So: fix dealer validation to include self weight; keep the candidate/final dealer slot threshold unless there is a broader protocol change.

### DIFF
--- a/inference-chain/x/bls/keeper/phase_transitions.go
+++ b/inference-chain/x/bls/keeper/phase_transitions.go
@@ -392,13 +392,10 @@ func (k Keeper) DetermineValidDealersWithConsensus(epochBLSData *types.EpochBLSD
 			return nil, fmt.Errorf("invalid slot range for dealer %d in epoch %d", dealerIndex, epochBLSData.EpochId)
 		}
 		dealerOwnSlots := uint64(dealerParticipant.SlotEndIndex-dealerParticipant.SlotStartIndex) + 1
-		dealerOwnSlotsAtLeastHalf := totalSlots > 0 && dealerOwnSlots*2 >= totalSlots
-		maxPossibleNonSelfVotingSlots := uint64(0)
-		if dealerOwnSlots <= totalSlots {
-			maxPossibleNonSelfVotingSlots = totalSlots - dealerOwnSlots
-		}
 
-		var validVotingSlotsExcludingDealer uint64
+		var validVotingSlots uint64
+		// Implicitly, the dealer considers its own parts valid
+		validVotingSlots = dealerOwnSlots
 
 		for verifierIndex, verification := range epochBLSData.VerificationSubmissions {
 			if verification == nil || len(verification.DealerValidity) == 0 {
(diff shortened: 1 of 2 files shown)


DimaOrekhovPS (approved):

0xgonka (approved):
Targeted fix for the GEB-29 Certik finding: a dealer holding ≥50% of slots could never reach the strict-majority quorum because their own slots were excluded from the tally. The fix counts an implicit self-vote (the dealer's own slot count) against the per-dealer threshold, while leaving the downstream aggregate threshold in `transitionFromVerifyingToDisputing` and the `t+1` signing threshold untouched as intended. Looks correct and well-tested for the described case.

**Must fix:**
- None.

**Should fix:**
- None.

**Nits:**
- Comment-only inaccuracy at [`phase_transitions_test.go` L511-514](https://github.com/gonka-ai/gonka/blob/94c6018dee45b513150de455d4de4c84292afddf/inference-chain/x/bls/keeper/phase_transitions_test.go#L511-L514) (`_ShortVectorsCountAsNo`). The Dealer 0 walkthrough says "gets votes from verifier 0 (33)" — verifier 0 *is* dealer 0 and is skipped at the `verifierIndex == dealerIndex` guard, so the actual contribution is verifier 1's 33 + implicit self 33 = 66, not the 100 the comment claims. Dealer 2's "implicitly self-votes (33)" should also read 34 (last participant gets the remainder per `createTestEpochBLSData`). Math still works out for the assertion (`{true, true, true}` is correct), but the comments will mislead the next reader. No code change needed.

**Looks correct:**
- The implicit self-vote at [`phase_transitions.go` L396-398](https://github.com/gonka-ai/gonka/blob/94c6018dee45b513150de455d4de4c84292afddf/inference-chain/x/bls/keeper/phase_transitions.go#L396-L398) is bounded — `dealerOwnSlots` cannot exceed `totalSlots`, so no inflation of the dealer's effective weight. The dealer's own `VerificationSubmissions[dealerIndex]` is still skipped at L407-409, so this isn't double-counted on top of an explicit self-judgment.
- The `dealerSubmittedParts` gate at [`phase_transitions.go` L423-428](https://github.com/gonka-ai/gonka/blob/94c6018dee45b513150de455d4de4c84292afddf/inference-chain/x/bls/keeper/phase_transitions.go#L423-L428) still catches a dealer with high slot count but no submitted parts — the implicit self-vote alone cannot validate a non-submitter.
- Quorum stays at strict majority (`totalSlots/2 + 1`) — a dealer with exactly 50% slots (the boundary case in `_DealerOwnsExactlyHalfSlots`) still requires at least one peer vote, which preserves the security goal of the original threshold while removing the *impossibility* failure mode.
- Removal of the "Dealer cannot reach weighted quorum with self-vote excluded" warning log and the dead `dealerOwnSlotsAtLeastHalf` / `maxPossibleNonSelfVotingSlots` arithmetic is correct — those existed only because of the now-fixed exclusion.
- PR scope respects the author's stated boundary: `transitionFromVerifyingToDisputing`'s `candidateDealerSlots > totalSlots/2` check is intentionally untouched, which is right since the `t+1` signing threshold at `threshold_signing.go` is also a strict majority by default — relaxing one without the other would create asymmetry.
- Targeting `upgrade-v0.2.12` means all validators flip behavior at the same upgrade height; mixed-version forks aren't a concern.
- New `TestDetermineValidDealersWithConsensus_DealerOwnsExactlyHalfSlots` is exactly the GEB-29 regression case; pinning it prevents reintroduction.

x0152 (approved):

PR #1114, opened 2026-04-24
fix: Certik audit fixes (GEB-62)

### LINKED ISSUE #1109: GEB-62

x0152:
From audit report:

> Description
> 
> During the VERIFYING-to-DISPUTING transition, dealer candidacy is determined solely from raw verifier votes in phase_transitions.go and stored as CandidateValidDealers in phase_transitions.go. Complaints against dealers that do not make this vote-based candidate set are then filtered out and deleted in phase_transitions.go, while dealers can only respond to complaints that still exist in msg_server_dispute.go. The substantive truth check for a complaint is deferred until verifyDealerComplaintResponse in dispute_resolution.go and is only applied later during finalization in phase_transitions.go.
> 
> As a result, verifiers with sufficient slot weight can submit dealer_validity=false together with an admissible complaint selector, push an honest dealer below quorum before the dispute is adjudicated, and cause that dealer’s complaint to be discarded before the dealer has any chance to disprove it. Because the dealer’s candidacy is never recomputed after complaint resolution, false complaint-driven votes can remain decisive, removing honest dealers from the final set or causing DKG failure due to insufficient candidate dealer slots.
> 
> Recommendation
> 
> Do not treat vote-based dealer candidacy as final before complaint resolution. Preserve all admitted complaints through the disputing phase, allow dealers to respond even if they initially miss vote-based candidacy, and only finalize dealer eligibility after complaint adjudication.
> 
> In practice, this means moving complaint filtering and quorum/failure checks until after dispute resolution, or recomputing dealer validity once false complaints have been disproven.
> 

We no longer remove complaints before the dispute phase. After disputes are checked, we recalculate which dealers are valid, and only then make the final quorum decision

### DIFF
--- a/inference-chain/x/bls/keeper/dispute_resolution.go
+++ b/inference-chain/x/bls/keeper/dispute_resolution.go
@@ -19,53 +19,81 @@ const (
 	dkgShareBytesLen  = 32
 )
 
-// applyDealerComplaintOutcomes applies complaint outcomes to the candidate valid dealer set.
+// applyDealerComplaintOutcomes applies complaint outcomes to the provided dealer set
 // Current policy:
 // - missing/invalid dealer response => dealer fault (dealer removed)
-// - valid dealer response => false complaint (complainer removed if in candidate set)
+// - valid dealer response => false complaint (complainer removed)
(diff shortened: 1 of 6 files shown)


gmorgachev:
thanks @x0152 !
merged rebased version in #1123 
let's include bounty in next proposal

PR #1115, opened 2026-04-24
fix: Certik audit fixes (GEB-59, GEB-60)

### LINKED ISSUE #1110: GEB-60
independent of upgrade, to do before bridge deploy

### LINKED ISSUE #1111: GEB-59

x0152:
**From audit report (GEB-59):** 

> Description
> 
> instantiate() delegates initialization to cw20_base_contract::instantiate() and does not overwrite the CW2 contract metadata afterward, so deployed instances retain crates.io:cw20-base. However, migrate() only accepts wrapped-token.
> 
> image
> 
> As a result, wrapped-token contracts created through the normal flow fail their own upgrade check and revert during migration.
> Recommendation
> 
> After the delegated instantiate call, write the correct CW2 metadata with set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION).

Fixes migration of wrapped-token contracts. instantiate now overwrites the CW2 marker left by cw20-base with wrapped-token and migrate accepts both the canonical and legacy crates.io:cw20-base markers

**From audit report (GEB-60):** 

> Description
> 
> The bridge pauses withdraw() and mintWithSignature() with onlyNormalOperation, but it does not apply the same check to the auto-burn paths in transfer() and transferFrom() when to == address(this). So users and approved spenders can still burn WGNK and emit WGNKBurned while the contract is in ADMIN_CONTROL.
> 
> This bypasses the intended emergency-stop behavior for the reverse bridge. If off-chain relayers honor WGNKBurned during ADMIN_CONTROL, native asset releases from escrow may continue even though the bridge is supposed to be paused.
> 
> Additionally, submitGroupKey() is also not enforced with onlyNormalOperation.
> Recommendation
> 
> Require NORMAL_OPERATION for all bridge-consumable burn paths.

Closes emergency-stop bypasses in BridgeContract and the public submitGroupKey are now gated by NORMAL_OPERATION


GLiberman:
@gmorgachev i double checked - we didnt registered wrapped token contract id - so we're super safe with these changes

0xgonka (approved):
Two-part Certik audit fix targeting wrapped-token contract migration (GEB-59) and bridge emergency-stop completeness (GEB-60). Both fixes are tight and well-scoped: the wrapped-token migration adds a legacy `crates.io:cw20-base` path with a state-presence guard against arbitrary cw20-base hijacking, and the bridge gates `submitGroupKey` and the auto-burn paths in `transfer`/`transferFrom` on `NORMAL_OPERATION`.

**Must fix:**
- None.

**Should fix:**
- The legacy-marker migration guard at [`contract.rs` L330-337](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/inference-chain/contracts/wrapped-token/src/contract.rs#L330-L337) requires `CREATOR` and `BRIDGE_INFO` to exist before accepting a `crates.io:cw20-base` marker, which is the right defense, but the new test [`migrate_rejects_foreign_legacy_cw20_base_marker`](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/inference-chain/contracts/wrapped-token/src/contract.rs#L484) only verifies the negative case where neither key is set. Add a case where exactly one of `CREATOR` / `BRIDGE_INFO` is present (the other missing) — both ordering paths fail the same way, but pinning that explicitly prevents a future refactor from accidentally accepting partially-seeded state. One extra test, ~10 lines.

**Nits:**
- [`BridgeContract.sol` L294, L301](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/proposals/ethereum-bridge-contact/contracts/BridgeContract.sol#L294-L301) — `require(epochId > 1, "Epoch 1 must be set via Admin");` is duplicated. Pre-existing, but you're touching the surrounding lines so cheap to delete one. Same revert string both times.
- [`bridge_admin_control_guards.test.js` L34-37](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/proposals/ethereum-bridge-contact/test/bridge_admin_control_guards.test.js#L34-L37): the test relies on the auto-burn path reverting *before* the `_burn` call attempts to read balance. That's correct given current code (state check at L449-451 is the first statement inside the `if (to == address(this))` block), but if anyone reorders inside the override later the test would still pass with the wrong revert — consider matching `revertedWithCustomError` *and* asserting that the user's balance is unchanged after the call. Cheap robustness.
- [`contract.rs` L500](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/inference-chain/contracts/wrapped-token/src/contract.rs#L500): file no longer ends with a newline (per the diff `\ No newline at end of file`). Trivial, but flag-on-touch.

**Looks correct:**
- Inline state check in [`transfer`/`transferFrom`](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/proposals/ethereum-bridge-contact/contracts/BridgeContract.sol#L446-L477) (rather than slapping `onlyNormalOperation` on the whole function) is exactly right — only the `to == address(this)` auto-burn branch must be gated, regular ERC20 transfers between users should remain unaffected by bridge state. The new test `keeps regular ERC20 transfers unguarded by bridge state` pins this distinction; that test would catch a future blanket-modifier regression.
- Initial `epochMeta.currentState = ContractState.ADMIN_CONTROL` (constructor at [L200](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/proposals/ethereum-bridge-contact/contracts/BridgeContract.sol#L200)) means newly-deployed bridges are not burnable until explicit `setGroupKey` + `resetToNormalOperation` — so the GEB-60 fix is also "secure by default" at deploy time, not just after the first state transition.
- All three burn paths (`transfer` auto-burn, `transferFrom` auto-burn, `withdraw`) and the public `submitGroupKey` are now consistently `NORMAL_OPERATION`-gated. Searched for other `_burn` callers and there are none — this completes the emergency-stop coverage. `mintWithSignature` was already gated.
- `deps.branch()` at [`contract.rs` L73](https://github.com/gonka-ai/gonka/blob/e81af5cc22502634abfd1b2944bdb8004d04953a/inference-chain/contracts/wrapped-token/src/contract.rs#L73) is the idiomatic CosmWasm 2.x way to delegate to `cw20_base_contract::instantiate` and then continue using `deps` for the `set_contract_version` write — required because `DepsMut` doesn't `Copy`.
- The legacy migration's state-presence check is the right defense: a brand-new cw20-base contract instance that didn't go through gonka's instantiate path won't have `CREATOR` or `BRIDGE_INFO` in storage, so it can't migrate to wrapped-token by accident or by malice. Combined with WASM admin permissioning on `migrate`, the attack surface is closed.
- The hardhat plugin migration to `hardhat-toolbox-mocha-ethers` is required for `hre.network.getOrCreate()` (used in the new test); not a gratuitous churn.
- Targeting `upgrade-v0.2.12` ensures the wrapped-token CW2 fix lands together with the rest of the audit set so deployed instances pick up the marker rewrite during the upgrade window.

GLiberman (approved):

PR #1124, opened 2026-04-27
fix: stop stale PoC validation promptly

ouicate:
## Summary
- Replaces #827 with a clean branch based directly on upgrade-v0.2.12
- Cancels off-chain PoC validation when the validation phase ends or the PoC stage changes, and propagates cancellation into proof fetch / ML-node validation work.
- Keeps retry requeue handling cancellable and avoids treating exhausted transient failures as invalid participants.


patimen:
This still looks like the right idea to me. The underlying issue appears to still be live on current `main`, and this looks like a good narrow vehicle for fixing it. We should review the details carefully and get this into `v0.2.14`.

PR #1125, opened 2026-04-27
fix: bound inference validation work

ouicate:
## Summary
- Replaces #828 with a clean branch based directly on upgrade-v0.2.12
- Bounds recovery validation and sampled validation fan-out with a fixed worker pool.
- Adds a timeout-backed HTTP client for validator replay requests so stalled ML-node endpoints cannot block goroutines indefinitely.

### DIFF
--- a/decentralized-api/internal/validation/inference_validation.go
+++ b/decentralized-api/internal/validation/inference_validation.go
@@ -21,6 +21,7 @@ import (
 	"sort"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


patimen:
Closing this as stale for the current direction. The underlying operational concern is understandable, but this adds more complexity to the on-chain validation path at a point where we are moving away from that path and focusing on devshards instead. If a narrower piece still matters in the successor architecture, we can bring it back there.

PR #1126, opened 2026-04-27
Devshardd structure

pixelplex:
**Build system**

Point `devshardd-build` Makefile target at `devshard/Dockerfile` instead of `decentralized-api/Dockerfile`; update binary copy path from `decentralized-api/build/` to `devshard/build/`. Add `devshard/Dockerfile` — a dedicated, focused build for the standalone `devshardd` binary (and `devshardctl`) with proper musl/CGO setup for blst; no longer bundled inside the `decentralized-api` image. Remove `devshardd`/`devshardctl` build steps and their `COPY` directives from `decentralized-api/Dockerfile`; remove the now-unused `devshard` replace directive and direct dependency from `decentralized-api/go.mod`.

**devshardd binary structure**

Decompose the ~350-line `main.go` into focused files: `app.go` (startup wiring with a `closeStack` LIFO cleanup helper), `config.go` (env-var loading), `chain_events.go` (thin `chainEventBridge` glue struct wiring the listener, bridge and phase together), and `events/listener.go` + `events/events.go` (block/escrow event fan-out extracted from `common/chain/events/listener.go`). Remove the copy of the chain event listener from `common/chain/events` in favour of the new `devshard/cmd/devshardd/events` package; move the listener test alongside it. Delete the commented-out in-process HostManager wiring from `decentralized-api/main.go` and its dead `configParamsProvider` type.

**Postgres payload store**

Add `local-test-net/docker-compose.devshard-postgres.yml` — a minimal `postgres:16-alpine` service on the shared `chain-public` network for the devshardd payload store. Wire `PGHOST`/`PGDATABASE`/`PGUSER`/`PGPASSWORD` into the versiond compose file so the devshardd child process inherits the connection. Wire genesis-only Postgres compose extension in `DevshardStandaloneTests` (join1/join2 remain without Postgres). Expose gRPC on `0.0.0.0:9090` in `init-docker-genesis.sh` so devshardd can reach the chain gRPC endpoint from its container.

**Integration tests**

Fix existing devshard tests, change localnet deployment for the tests. Remove `DevshardTests` as it was using devshard from decentralized-api, which was removed. Add `devshardd bridge events and phases settle escrow lifecycle` — an end-to-end test that verifies the full escrow lifecycle driven by chain events rather than direct RPC.


devshard/cmd/devshardd/bridge/chain.go
```diff
@@ -44,33 +46,57 @@ func NewChainBridge(client *chain.Client, submitter Submitter) *ChainBridge {
...
+		info, err := b.GetEscrow(id)
+		if err != nil {
+			slog.Warn("chain events: failed to fetch escrow", "escrow_id", id, "error", err)
+			return
+		}
+		if err := b.OnEscrowCreated(*info); err != nil {
+			slog.Warn("chain events: escrow created handler failed", "escrow_id", id, "error", err)
+		}
+	})
+	// TODO: OnSettlementProposed?
```
DimaOrekhovPS, line 65: What's the plan for this TODO? Is it going to be addressed?

devshard/cmd/devshardd/bridge/chain.go
```diff
@@ -4,16 +4,18 @@ import (
...
 
 	"common/chain"
 	"devshard/bridge"
+	"devshard/cmd/devshardd/events"
 
 	inferencetypes "github.com/productscience/inference/x/inference/types"
 )
 
-const warmKeyMsgTypeGRPC = "/inference.inference.MsgSettleDevshardEscrow"
+const warmKeyMsgTypeGRPC = "/inference.inference.MsgStartInference"
```
DimaOrekhovPS, line 18: What's the reason behind this change?

devshard/cmd/devshardd/chain_events.go
```diff
@@ -0,0 +1,82 @@
...
+	eventListener.OnNewBlock(func(bctx context.Context, e events.NewBlockEvent) {
+		// TODO: should this be called for every block?
+		resp, err := chainClient.InferenceQueryClient().GetCurrentEpoch(bctx, &chaintypes.QueryGetCurrentEpochRequest{})
+		if err != nil {
+			slog.Warn("phase: failed to query current epoch", "block", e.BlockHeight, "error", err)
+			phase.SetBlockHeight(e.BlockHeight)
+			return
+		}
+		phase.Update(resp.Epoch, e.BlockHeight)
+	})
```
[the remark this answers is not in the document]
DimaOrekhovPS: I think we can keep the current implementation and optimize only if it becomes a problem

inference-chain/scripts/init-docker-genesis.sh
```diff
@@ -80,6 +80,8 @@ fi
 
 sed -Ei 's/^laddr = ".*:26657"$/laddr = "tcp:\/\/0\.0\.0\.0:26657"/g' \
   $STATE_DIR/config/config.toml
+sed -Ei 's/^address = ".*:9090"$/address = "0.0.0.0:9090"/g' \
```
DimaOrekhovPS, line 83: why was this change required?

testermint/src/test/kotlin/DevshardStandaloneTests.kt
```diff
@@ -392,6 +494,24 @@ class DevshardStandaloneTests : TestermintTest() {
...
+        waitForOverrideVersionedHealth(genesis)
+
+        logSection("Creating devshard escrow")
+        val escrowAmount = 7_000_000_000L
+        val txResponse = genesis.createDevshardEscrow(escrowAmount, modelId = devshardEscrowModel)
+        assertThat(txResponse.code).isEqualTo(0)
+        val escrowId = txResponse.getEscrowId() ?: 1L
+
+        logSection("Query devshardd mempool -- triggers lazy session creation")
+        val mempool = getVersionedDevshardMempool(genesis, escrowId)
```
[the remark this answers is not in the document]
DimaOrekhovPS: Valid comment, let's fix

devshard/cmd/devshardd/bridge/chain.go
kstdl: This is not used by the current devshardd flow. The TODO is intentional so the missing proposed-settlement event handling is visible if/when we add that flow later.

devshard/cmd/devshardd/bridge/chain.go
kstdl: This was mistakenly set to MsgSettleDevshardEscrow in the base branch. It should be MsgStartInference, matching the decentralized-api bridge behavior,

inference-chain/scripts/init-docker-genesis.sh
kstdl: I think this was needed for the local Docker testnet used by integration tests. The gRPC server was binding to localhost, so other containers could not reach the chain node over Docker networking. But I will verify this and revert this if it isn't needed

testermint/src/test/kotlin/DevshardStandaloneTests.kt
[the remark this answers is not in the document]
kstdl: fixed

DimaOrekhovPS (approved):

PR #1129, opened 2026-04-29
Fix fees again

### LINKED ISSUE #928: [P1] Open Questions: Block Gas Limits, Fees, Cost per Participant, and System TX Prioritization

0xgonka:
## Summary

DAPI was setting every batch's `gasWanted` to a constant `BatchGasLimit` (1B after #1120). Cosmos charges fees on `gasWanted`, not `gasUsed`, so every non-bypass tx was being billed at the worst-case PoC commit ceiling regardless of what it actually consumed. With `min_gas_price=10` re-enabled, this would burn ~1000 GNK/host/year on routine bookkeeping — incompatible with the 100 GNK/year/host promise.

This PR fixes that with two changes: dynamic per-batch `gasWanted` in DAPI, and bypass for two more routine messages on the chain.

## Why

Numbers from a 24-hour mainnet sample (~385K txs, pulled from a node Gleb gave me SSH on) showed the per-host yearly fee load if `min_gas_price` were re-enabled at 10 ngonka:

| sizing strategy | per-host/year (30 hosts) | per-host/year (129 hosts) |
|---|---:|---:|
| previous (constant `BatchGasLimit=1B`) | ~1,051 GNK | ~244 GNK |
| this PR (dynamic per-msg) | ~24 GNK | ~5 GNK |

The two messages added to the bypass list (`MsgSubmitHardwareDiff`, `MsgClaimRewards`) account for ~70% of routine non-bypass volume; both are protocol-driven host duties on a fixed schedule, not sybil vectors. `MsgPoCV2StoreCommit` deliberately stays non-exempt — its count-proportional fee in `FeeParams` is the actual sybil defense.

### DAPI: dynamic `gasWanted` per batch

- **New file [`gas_estimate.go`](decentralized-api/cosmosclient/tx_manager/gas_estimate.go)** with a per-message-type lookup table. Numbers sized at observed p99 × ~1.5 from the 24h mainnet sample. The two known linear-scaling messages mirror their on-chain `ConsumeGas` formula:
  - `MsgPoCV2StoreCommit`: `base + sum(entry.Count) × per_count`
  - `MsgMLNodeWeightDistribution`: `base + total_node_entries × per_node`
- **`estimateBatchGas(msgs, attempt)`** sums per-msg estimates plus a fixed tx-level overhead, doubles per retry attempt to escape OOG loops, capped at `BatchGasLimit`.
- **`applyGasAndFee(tx, gasWanted, minGasPrice)`** extracted from `getSignedBytes` so the gas/fee math is unit-testable without keyring scaffolding.
- **`broadcastMessage` / `BroadcastMessages`** now route through internal `*AtAttempt` helpers that thread the retry count from the queue (`tx.TxInfo.Attempts`) into the estimator. Public interface unchanged — `BroadcastMessages(id, msgs...)` just delegates with `attempt=0`.
- **`errors.go`**: `"out of gas"` added to `retryablePatterns` so under-estimated txs requeue (with bumped gas on the next attempt) instead of failing permanently.

### Chain: two more bypass entries

- **`isExemptMessageType`** in [`ante_fee.go`](inference-chain/app/ante_fee.go) now exempts `MsgSubmitHardwareDiff` and `MsgClaimRewards`. Both are scheduled host duties (HardwareDiff fires per-block on actual hardware change; ClaimRewards is once per epoch per host), neither is a sybil-attack surface.

After these additions, the only fee-paying routine message on the host hot path is `MsgPoCV2StoreCommit` (sized via its on-chain count formula) and `MsgSubmitSeed` (~7/host/day, ~2 GNK/host/year at min_gas_price=10).

### Tests

New [`gas_estimate_test.go`](decentralized-api/cosmosclient/tx_manager/gas_estimate_test.go) covers:
- Per-msg-type estimates pinned (silent retunes show up in the diff).
- PoCV2/MLNode formulas verified at multiple payload sizes.
- `estimateBatchGas`: tx overhead, retry multiplier, BatchGasLimit cap, empty-batch safety.
- `applyGasAndFee`: zero-fee path (v0.2.12 mainnet), non-zero-fee math, ceiling clamp.
- **Linter-style guard**: iterates `InferenceOperationKeyPerms` (the warm key's full authz permission list) and asserts every msg type has an explicit case in `lookupMsgGas`. Catches the case where someone adds a new msg to the perms list and forgets the gas estimate. Caught 5 real gaps during dev.

Updated [`ante_fee_test.go`](inference-chain/app/ante_fee_test.go) to reflect the new exempt set.

## Out of scope (separate work worth flagging)

- **Re-enabling `MinGasPriceNgonka`** stays a governance decision. This PR makes 100 GNK/year/host viable when that happens; it doesn't change `min_gas_price` itself.
- **`MsgSubmitHardwareDiff` fires every ~11 minutes per host** in mainnet data. Now bypassed so it doesn't cost fees, but suggests DAPI is sending diffs even when nothing changed. Worth a separate look.
- **`DefaultMLOpsFeeAllowance` (10 GNK in `permissions.go`) vs `MigratedFeeAllowance` (100 GNK in upgrade handler)** — 10x mismatch between onboarding and migration paths. Harmless at min_gas_price=0 but worth reconciling before fees re-enable.

## Test plan

- [ ] `make test` (or equivalent) on `decentralized-api` and `inference-chain` — locally verified all packages green (Go 1.24).
- [ ] Spin up a local cluster (or run testermint) and confirm batches show `gasWanted` matching the per-msg-type sum, not 1B.
- [ ] On testnet, verify warm-signed batches still succeed (no OOG, no feegrant errors). The bypass changes are consensus-affecting, so this should be tested in a multi-node setting before mainnet.
- [ ] Re-pull a fresh mainnet block sample after a few weeks of mainnet activity post-merge and tighten the table values; the lookup is one switch statement and easy to re-tune.

## Data and methodology

The numbers in `gas_estimate.go` come from a 24h mainnet pull (385K txs, 16K blocks) processed offline. Full report (per-type p50/p95/p99/max, linear-scaling correlation, multi-msg batch breakdown) at `/tmp/gonka-gas-analysis/report.md` on my workstation; can attach as a gist if useful for review.


decentralized-api/cosmosclient/tx_manager/gas_estimate.go
```diff
@@ -0,0 +1,244 @@
...
+		return gasDefaultEstimate, false
+	}
+}
+
+// estimateBatchGas returns the gasWanted for a batch of messages. The first
+// attempt (attempt=0) sums per-message estimates plus tx-level overhead.
+// Each subsequent attempt multiplies by gasRetryMultiplier to escape OOG
+// loops when the static estimate underestimates real gas. The result is
+// capped at BatchGasLimit so we never request more gas than the chain's
+// network-duty bypass cap can accommodate.
```
patimen: This whole file is mostly comment... Can we pare it down, or move it to somewhere that's less likely to obscure the actual code?

decentralized-api/cosmosclient/tx_manager/gas_estimate.go
```diff
@@ -0,0 +1,244 @@
...
+// fees regardless of gasWanted; the network-duty bypass also produces
+// zero-fee txs but at a different layer.
+func applyGasAndFee(tx client.TxBuilder, gasWanted uint64, minGasPriceNgonka int64) {
+	if gasWanted == 0 || gasWanted > BatchGasLimit {
+		gasWanted = BatchGasLimit
+	}
+	tx.SetGasLimit(gasWanted)
+	if minGasPriceNgonka > 0 {
+		feeAmount := math.NewIntFromUint64(gasWanted).MulRaw(minGasPriceNgonka)
+		tx.SetFeeAmount(sdk.NewCoins(sdk.NewCoin("ngonka", feeAmount)))
```
patimen: I'm sure we have an "ngonka" const value somewhere.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -571,8 +570,9 @@ func (m *manager) sendTxs() error {
 			}
 
 			if !tx.Sent {
-				logging.Debug("start broadcast batch async", types.Messages, "id", tx.TxInfo.Id)
-				resp, timeout, broadcastErr = m.BroadcastMessages(tx.TxInfo.Id, msgs...)
+				logging.Debug("start broadcast batch async", types.Messages, "id", tx.TxInfo.Id, "attempt", tx.TxInfo.Attempts)
+				// Pass attempt so OOG retries bump gasWanted (gas_estimate.go).
```
patimen: See, that's just a comment we do not need.

decentralized-api/cosmosclient/tx_manager/gas_estimate.go
```diff
@@ -0,0 +1,244 @@
...
+	// (better, future work) read FeeParams at DAPI startup and use those
+	// values dynamically with a fixed headroom multiplier.
+	gasPoCV2Base    = uint64(600_000) // base + sdk overhead + 50% headroom
+	gasPoCV2PerCount = uint64(150)    // 100 on-chain + 50% headroom
+
+	// MLNodeWeightDistribution linear in entries.
+	gasMLNodeBase    = uint64(100_000)
+	gasMLNodePerNode = uint64(50_000)
+
+	// Routine host duties (now bypass-exempt — see ante_fee.go isExemptMessageType).
```
patimen: Is this really how we're going to estimate gas? How will we adjust of the actual impact of one of these changes? 
I guess, since we have a retry that doubles the gas on failure, it will sort itself out... but we should probably look to have metrics of some kind about how often that is happening?

inference-chain/x/inference/keeper/msg_server_submit_hardware_diff.go
```diff
@@ -79,3 +93,19 @@ func (k msgServer) SubmitHardwareDiff(goCtx context.Context, msg *types.MsgSubmi
 
 	return &types.MsgSubmitHardwareDiffResponse{}, nil
 }
+
+// HardwareNodesUnchanged reports whether two sorted hardware-node lists are
+// element-wise identical. Used as a cheap idempotency check on
+// MsgSubmitHardwareDiff so DAPI's per-block diff submissions don't trigger
+// state writes when nothing actually changed. Exported for testing.
+func HardwareNodesUnchanged(after, before []*types.HardwareNode) bool {
```
patimen, line 103: This seems pretty shallow. Couldn't an attacker just vary the HardwareNode a bit? Like the Version... that's specifically informational only and isn't checked.

decentralized-api/cosmosclient/tx_manager/tx_manager.go
```diff
@@ -795,11 +795,19 @@ func (m *manager) GetJetStream() nats.JetStreamContext {
...
+// that knows about retry attempt. attempt=0 is the first try at the
+// per-msg-type estimate; subsequent attempts bump gasWanted to escape OOG
+// loops. See estimateBatchGas in gas_estimate.go.
+func (m *manager) broadcastMessagesAtAttempt(id string, attempt int, msgs []sdk.Msg) (*sdk.TxResponse, time.Time, error) {
 	if len(msgs) == 0 {
 		return nil, time.Time{}, nil
 	}
 	if len(msgs) == 1 {
-		return m.broadcastMessage(id, msgs[0])
+		return m.broadcastMessageAtAttempt(id, attempt, msgs[0])
```
patimen: I'm not sure why this is even needed... won't it more or less do the same thing as just continuing with one message? Can we refactor to just broadcastMessagesAtAttempt?

patimen (approved):
I'm most concerned about the potential misuse of MsgSubmitHardwareDiff by simply mutating Versions, and then a minor thought about how we're estimating gas.

0xgonka:
@patimen 

Pushed 39c80c518. Mapping to your comments:

1. HardwareDiff version mutation: no-op check now compares operational fields only (local_id, status, models, hardware, host, port). Version is explicitly excluded since the proto marks it informational. Test pins this. Documented as best-effort idempotency, not a rate limit.

2. Gas estimation observability: when a broadcast succeeds at attempt > 0, DAPI now logs a Warn with the new gasWanted. Surfaces in production when the static table needs re-tuning instead of relying on bug reports.

3. ngonka string literal: replaced with inferencetypes.BaseCoin.

4. Single-msg path refactor: broadcastMessageAtAttempt is now a 1-line wrapper around broadcastMessagesAtAttempt with a 1-element slice. -50 lines of near-duplicate.

5. Comment at sendTxs:574: removed.

6. Comments in gas_estimate.go: file header trimmed 22 → 8 lines, per-section comments dropped, kept only the load-bearing ones (FeeParams dependency on PoCV2 constants, authz wrap on txOverheadGas).

Net diff: -33 lines despite the new no-op helper. Build clean, full DAPI suite + chain ante/inference suites pass.

0xgonka:
integration tests running: https://github.com/gonka-ai/gonka/actions/runs/25219009743

PR #1132, opened 2026-04-29
fix: bound finish cost before invalidation refund

ouicate:
## Summary
This fixes a critical accounting bug where an executor could submit inflated finish-time token counts, persist an `ActualCost` far above the requester escrow, and then trigger `MsgInvalidateInference` to refund that unescrowed amount from the inference module account.

A requester/executor pair could start an inference with a very small escrow, finish it with token counts up to the hardcoded per-field maximum, and later receive an invalidation refund based on the inflated `ActualCost`. Because module-to-account transfers are allowed, the refund could drain materially more coins than were ever escrowed for that inference. Existing inflated inferences also needed protection before their invalidation path executes.

## Root Cause
`MsgFinishInference` only capped token counts against `MaxAllowedTokens`; it did not bind `CompletionTokenCount` to the start-time `MaxTokens`, and prompt token checks did not prevent finish-time values from feeding payment math. `ProcessFinishInference` then stored the raw calculated cost in `inference.ActualCost` even when the start-side escrow was lower. `refundInvalidatedInference` trusted that stored value directly for both the bank refund and executor debt.

## Fix
- Reject finish-after-start completions that exceed the start-time `MaxTokens`, using the default max for legacy zero-valued state.
- Reject prompt-token mismatches when both start and finish supplied nonzero prompt-token counts.
- Validate finish-before-start token counts when the matching start arrives, so out-of-order flows cannot bypass the start bounds.
- Cap persisted `ActualCost` to `EscrowAmount` whenever the start has already been processed.
- Cap invalidation refunds and executor debits to `min(ActualCost, EscrowAmount)` as defense in depth for already-persisted inflated inferences.

The exploit needs an untrusted finish value to become a real module-bank refund. After this change, finish token counts are checked against start-side bounds when that state exists, and the accounting value consumed by validator sharing and invalidation refunds cannot exceed the amount actually escrowed. The invalidation handler also independently caps refunds, so stale inflated state cannot drain the module account.

### DIFF
--- a/inference-chain/x/inference/calculations/inference_state.go
+++ b/inference-chain/x/inference/calculations/inference_state.go
@@ -69,6 +69,13 @@ func ProcessStartInference(
 			PerTokenPrice: existingPerTokenPrice,
 		}
 	}
+	maxTokens := getMaxTokens(startMessage)
+	if currentInference.FinishedProcessed() {
+		if err := validateFinishedTokenCountsAgainstStart(currentInference, startMessage, maxTokens); err != nil {
+			return nil, nil, err
+		}
+	}
+
 	// Works if FinishInference came before
 	currentInference.RequestTimestamp = startMessage.RequestTimestamp
 	currentInference.TransferredBy = startMessage.Creator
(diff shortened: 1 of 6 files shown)


patimen (approved):
`ActualCost` now consistently reflects the amount actually charged, and the added bounds/defense-in-depth on refund paths look correct.

PR #1133, opened 2026-04-29
fix: allow wrapped token migration from cw20-base marker

ouicate:
## Summary
This fixes a wrapped-token upgrade blocker: normally instantiated wrapped-token contracts store their cw2 identity as `crates.io:cw20-base`, but the wrapped-token `migrate()` entrypoint only accepted `wrapped-token`.

Governance-driven Wasm migration for existing wrapped tokens can fail before any upgrade logic runs. Because these contracts sit on the bridge mint/burn boundary, losing the normal migration path can leave production bridge token logic unpatchable during a future security or peg-integrity incident.

## Root Cause
`instantiate()` delegates to `cw20-base`, and `cw20-base` writes the cw2 contract name as `crates.io:cw20-base`. The wrapped-token migration guard then compared the stored cw2 contract name only against `wrapped-token`, rejecting legitimate deployed instances created by the standard instantiation flow.

## Fix
- Define the historical `crates.io:cw20-base` cw2 marker explicitly.
- Allow `migrate()` from either the canonical `wrapped-token` marker or the historical `crates.io:cw20-base` marker.
- Keep rejecting unrelated cw2 contract names.
- Always rewrite cw2 to the canonical `wrapped-token` name and current contract version after a successful migration.
- Add regression tests for fresh cw20-base-marked instances, first migration, subsequent canonical migration, and wrong-contract rejection.

A legitimate deployed wrapped token can now pass the migration guard even when its cw2 metadata still reflects the delegated `cw20-base` initializer. After the first successful migration, its cw2 metadata is canonicalized to `wrapped-token`, so future migrations use the expected wrapped-token identity.

### DIFF
--- a/inference-chain/contracts/wrapped-token/src/contract.rs
+++ b/inference-chain/contracts/wrapped-token/src/contract.rs
@@ -22,6 +22,7 @@ pub const CREATOR: Item<Addr> = Item::new("creator");
 
 const CONTRACT_NAME: &str = "wrapped-token";
 const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+const CW20_BASE_CONTRACT_NAME: &str = "crates.io:cw20-base";
 
 #[entry_point]
 pub fn instantiate(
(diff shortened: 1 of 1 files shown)


GLiberman:
Hey, the issue from this PR was already covered in PR #1115 - GEB50 which was found by Certik audit team

PR #1139, opened 2026-04-30
Default PROXY 0.2.12-post1

GLiberman:
Bug: nginx geo module does not expand variables in values - geo $whitelist_limit_key { default $binary_remote_addr; } sets the limit key to the literal string "$binary_remote_addr" for every request, causing all clients to share a single rate-limit bucket per zone.

Changes
**proxy/entrypoint.sh + proxy/sidecar/main.go**:

- Replace single-stage geo $whitelist_limit_key { default $binary_remote_addr; } with the canonical two-stage geo + map pattern so $binary_remote_addr is properly expanded per-request by map, giving each IP its own rate-limit bucket
- Same fix applied in sidecar's updateNginxConfig() for when validator whitelist sync is enabled

**deploy/join/docker-compose.yml**:

- Group env vars by category (ports, disable flags, exempt routes, blocked routes, connection limits, rate limits, timeouts, optional/SSL)
- Expose all proxy-configurable env vars (rate limits, burst, connection limits, timeouts, exempt/blocked routes) with overridable config.env defaults so node operators can tune per-deployment without editing compose 
- Add CHAIN_RPC_EXEMPT_ROUTES=status genesis default so high-traffic health/status checks bypass the standard chain RPC limiter
- Bump image to proxy:0.2.12-post1


0xgonka (approved):
Substantively correct, all the live-checkable claims pass:

- node4.gonka.ai is up (HTTP 200) and serves /v1/chat/completions ✓
- node4 does NOT expose /chain-api (404) — matches the doc note ✓
- devshard_escrow_params.allowed_creator_addresses is a real chain field with 13 entries
- The example's model (Qwen/Qwen3-235B-A22B-Instruct-2507-FP8) is in node4's /v1/models ✓
- Old transfer_agent_access_params still exists on chain (7 entries) but per the announcement those agents on node1/2/3 are disabled, so removing references is correct

Suggested polish before merging:

1. Add one explicit migration line for existing users. Anyone with an old SDK config has SOURCE_URL=http://node1.gonka.ai:8000 (or :2 :3). They need a clear "if you previously used node1/2/3 as SOURCE_URL, change it to https://node4.gonka.ai" callout — the current phrasing assumes the reader is starting from scratch. This is the actionable change for the existing user base.

2. "As of the latest mainnet update" will rot. Either pin to a version ("As of v0.2.12, ...") or just drop the temporal qualifier ("The recommended public inference gateway is https://node4.gonka.ai").

3. Admonition title "Public `devshard` creator node" reads ambiguously — "creator node" sounds like node4 creates devshard binaries. "Public inference gateway" or "Recommended inference endpoint" would be clearer for a quickstart audience. The body can still explain the devshard mechanism.

4. The block is still labeled !!! important. The original was warning about a limitation; the new content is mostly informational + a recommendation. !!! note or !!! info is a better fit.

5. Consider order: the conceptual paragraph ("Inference on Gonka is settled through a devshard-based billing flow...") is good context but heavy for a quickstart. People copy-pasting want the export SOURCE_URL=... line first, explanation second. I'd swap the order: action first, then explanation, then the optional inspect-the-allowlist curl.

6. Tiny: there are two trailing spaces after `https://node4.gonka.ai` on the recommended-gateway line. Probably an unintended markdown line-break trick. Drop them.

LGTM after 1 (migration line) and 2 (temporal phrasing). The other items are nice-to-have.

PR #1142, opened 2026-05-01
Reset ConsecutiveFaliures after it's used for invalidation

patimen:
# Reset ConsecutiveFaliures after it's used for invalidation

## Description of the PR

1.  **What problem does this solve?**
    If a Participant accumulates 5 invalid inferences in a row, they are marked as `INVALID` due to the `quick_failure_threshold` mechanism. However, when such a participant is later restored to `ACTIVE` status at the beginning of a new epoch, their `ConsecutiveInvalidInferences` counter was not being reset. This meant they would immediately fail again the first time their status was recalculated, leading to a loop of invalidation.

2.  **How do you know it is a real problem?**
    The problem is identified as a logical flaw in the participant status management. The fact that a participant, once restored to `ACTIVE`, would immediately become `INVALID` again due to an un-reset counter indicates a clear bug. The `testermint` integration test (`InvalidationTests.kt`) was specifically added to validate this scenario, providing end-to-end verification that the issue no longer occurs.

3.  **How does this solve the problem?**
    This PR solves the problem by implementing two key resets for the `ConsecutiveInvalidInferences` counter:
    *   **Immediate Reset on Invalidation:** When a participant is marked `INVALID` specifically due to `ConsecutiveFailures`, their `participant.ConsecutiveInvalidInferences` counter is immediately reset to `0` in `inference-chain/x/inference/keeper/participant_status.go` (lines 51-53).
    *   **Reset on Epoch Reactivation:** At the beginning of each epoch, when participants are moved from the `upcoming` to the `effective` group and set to `ACTIVE`, their `participant.ConsecutiveInvalidInferences` counter is also explicitly reset to `0` in `inference-chain/x/inference/module/module.go` (line 1261). This ensures that even participants who had accumulated some failures but didn't hit the `quick_failure_threshold` start the new epoch with a clean slate.

4.  **What risks does this introduce? How do we mitigate these risks?**
    The core changes are minimal and directly address the described bug.
    *   **Risk:** The direct modification of `participant.ConsecutiveInvalidInferences` in critical status update and epoch transition functions could, in theory, lead to incorrect participant state if not handled precisely.
    *   **Mitigation:** The changes are small and focused. The primary mitigation for these risks comes from the extensive testing added:
        *   Unit tests in `inference-chain/x/inference/keeper/participant_test.go` now assert that `ConsecutiveInvalidInferences` is correctly reset to `0` after invalidation.
        *   A new integration test scenario in `testermint/src/test/kotlin/InvalidationTests.kt` provides end-to-end validation, verifying that a participant, once restored to `ACTIVE` status after being invalidated, remains `ACTIVE` even after successfully serving fresh inference traffic. This testing ensures the fix works as intended across the system.

5.  **How do you know this PR fixes the problem?**
    This PR fixes the problem based on the following:
    *   **Unit Test Assertions:** The changes in `inference-chain/x/inference/keeper/participant_test.go` include new `require.Equal(t, int64(0), participant.ConsecutiveInvalidInferences)` assertions. These confirm that the `ConsecutiveInvalidInferences` counter is indeed reset to zero when a participant transitions to `INVALID` status, and also after failing the `quick_failure_threshold`.
    *   **Integration Test Validation:** The `testermint/src/test/kotlin/InvalidationTests.kt` file adds a new, comprehensive integration test. This test specifically simulates a participant being invalidated, then restored at a new epoch, and finally verifies that the restored participant successfully serves fresh inference traffic and *remains ACTIVE*. This end-to-end test confirms the full resolution of the problem in a live-like environment.

### DIFF
--- a/decentralized-api/cosmosclient/tx_manager/errors.go
+++ b/decentralized-api/cosmosclient/tx_manager/errors.go
@@ -67,6 +67,10 @@ var retryablePatterns = []string{
 
 	"unordered transaction has a timeout_timestamp that has already passed",
 	"unordered tx ttl exceeds",
+	// Out-of-gas: the per-msg-type estimate underestimated this batch's
+	// real consumption. Retry with a bumped gasWanted (estimateBatchGas
+	// applies a multiplier per attempt, see gas_estimate.go).
+	"out of gas",
 }
 
 // isRetryableRawLog checks if the raw log contains any retryable error patterns
--- a/decentralized-api/cosmosclient/tx_manager/gas_estimate.go
+++ b/decentralized-api/cosmosclient/tx_manager/gas_estimate.go
@@ -1,0 +1,186 @@
+package tx_manager
+
+import (
+	"cosmossdk.io/math"
+	"github.com/cosmos/cosmos-sdk/client"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	blstypes "github.com/productscience/inference/x/bls/types"
+	collateraltypes "github.com/productscience/inference/x/collateral/types"
+	inferencetypes "github.com/productscience/inference/x/inference/types"
+)
+
+// applyGasAndFee writes gasWanted (capped at BatchGasLimit) onto the tx
+// builder and computes the matching fee from minGasPriceNgonka. Extracted
+// for unit testing without keyring/signing setup.
+func applyGasAndFee(tx client.TxBuilder, gasWanted uint64, minGasPriceNgonka int64) {
+	if gasWanted == 0 || gasWanted > BatchGasLimit {
+		gasWanted = BatchGasLimit
+	}
+	tx.SetGasLimit(gasWanted)
+	if minGasPriceNgonka > 0 {
+		feeAmount := math.NewIntFromUint64(gasWanted).MulRaw(minGasPriceNgonka)
+		tx.SetFeeAmount(sdk.NewCoins(sdk.NewCoin(inferencetypes.BaseCoin, feeAmount)))
+	} else {
+		tx.SetFeeAmount(sdk.Coins{})
+	}
+}
+
+// Per-message gas estimates. Cosmos charges fees on gasWanted (not gasUsed),
+// so over-sizing inflates routine costs and under-sizing causes OOG.
+//
+// Numbers are p99 of observed gasUsed × ~1.5 from a 24h mainnet sample
+// (see /tmp/gonka-gas-analysis/report.md). The headroom expects ~1% of txs
+// to OOG; estimateBatchGas doubles per attempt to escape the retry loop.
+//
+// Two messages have linear-scaling formulas mirroring their on-chain
+// ConsumeGas: MsgPoCV2StoreCommit (per-Count) and MsgMLNodeWeightDistribution
+// (per-entry). Re-tune from a fresh sample after a handler change, a new
+// msg type, or a FeeParams.{base_validation_gas,gas_per_poc_count} change.
+const (
+	// Tx-level fixed cost: ante decorators + authz MsgExec unwrap. Mainnet
+	// hosts almost always run in authz mode, so this absorbs the wrap cost.
+	txOverheadGas = uint64(50_000)
+
+	// Doubled per retry attempt so OOG-on-underestimate eventually fits.
+	gasRetryMultiplier = 2.0
+
+	// Inference lifecycle (bypass-exempt).
+	gasStartInference  = uint64(250_000)
+	gasFinishInference = uint64(250_000)
+	gasValidation      = uint64(1_500_000) // outliers up to 2.1M observed
+
+	// PoC duty messages.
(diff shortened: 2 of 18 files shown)


0xgonka (approved):
Two-line fix for a real bug: a participant restored to ACTIVE after `ConsecutiveFailures`-induced INVALID would immediately re-invalidate because the counter wasn't reset. Patch resets it (a) right at the INVALID transition and (b) defensively on every epoch reactivation. Tests cover both the unit assertion and the end-to-end "restored participant stays active after serving fresh inference" scenario.

**Must fix:**
- None.

**Should fix:**
- **The epoch-reactivation reset at [`module.go:1261`](https://github.com/gonka-ai/gonka/blob/9868da7c23e6b90dc69c867aea8e6fc79e24a242/inference-chain/x/inference/module/module.go#L1261) changes the semantics of consecutive-failure detection.** Before this PR, `ConsecutiveInvalidInferences` was monotonic across the participant's lifetime (only reset by a successful validation at [`msg_server_validation.go:162`](https://github.com/gonka-ai/gonka/blob/9868da7c23e6b90dc69c867aea8e6fc79e24a242/inference-chain/x/inference/keeper/msg_server_validation.go#L162) or revalidation at [`msg_server_revalidate_inference.go:29`](https://github.com/gonka-ai/gonka/blob/9868da7c23e6b90dc69c867aea8e6fc79e24a242/inference-chain/x/inference/keeper/msg_server_revalidate_inference.go#L29)). After this PR, the counter resets every epoch for **every** participant — not just restored INVALID ones.

  That changes the quick-failure rule from *"5 consecutive failures, ever"* to *"5 consecutive failures within a single epoch"*. A participant that consistently fails (e.g.) 4 inferences per epoch and succeeds 1 will never trigger `ConsecutiveFailures` even though their failure rate is 80%. The statistical (`getInvalidationStatus`) path would still catch them, so it's not a complete regression in detection — but it is a deliberate weakening of the per-epoch-failure-budget side of the policy. The PR body acknowledges this as intentional ("start the new epoch with a clean slate"); just want to make sure that's the desired tradeoff and not an accidental side effect of fixing the restore-loop.

  If the goal is purely "unstick restored participants without changing other semantics", the narrower fix is to gate the reset on `participant.Status == INVALID` (i.e. only restored-from-INVALID participants get the counter wiped). Something like:

  ```go
  if participant.Status == types.ParticipantStatus_INVALID {
      participant.ConsecutiveInvalidInferences = 0
  }
  participant.Status = types.ParticipantStatus_ACTIVE
  ```

  That preserves the all-time monotonic semantics for already-active participants while still solving the restore loop.

**Nits:**
- The integration test at [`InvalidationTests.kt:73-89`](https://github.com/gonka-ai/gonka/blob/9868da7c23e6b90dc69c867aea8e6fc79e24a242/testermint/src/test/kotlin/InvalidationTests.kt#L73-L89) loops up to 20 attempts looking for an inference routed to the restored participant. With probabilistic routing this is generally fine, but if the network has many participants the restored one might not get picked within 20 windows. Not a correctness issue but a potential flake source — the assertNotNull message is clear, so debugging would be straightforward when it does flake.

**Looks correct:**
- The reset at [`participant_status.go:51-53`](https://github.com/gonka-ai/gonka/blob/9868da7c23e6b90dc69c867aea8e6fc79e24a242/inference-chain/x/inference/keeper/participant_status.go#L51-L53) is correctly gated on `reason == ConsecutiveFailures` rather than `newStatus == INVALID`. A participant going INVALID via `StatisticalInvalidations` keeps their counter intact, which is the right call — that path doesn't depend on the consecutive counter and the counter is informative for downstream analysis.
- Placement before the `originalStatus == newStatus` early return at L57 is intentional and harmless. The "already INVALID" case won't actually fire because [`status.go:55-57`](https://github.com/gonka-ai/gonka/blob/9868da7c23e6b90dc69c867aea8e6fc79e24a242/inference-chain/x/inference/calculations/status.go#L55-L57) short-circuits with `reason = AlreadySet`, not `ConsecutiveFailures`. So the reset only fires on the actual INVALID transition, not on subsequent re-checks of an already-INVALID participant.
- Unit tests assert the post-condition explicitly (`require.Equal(t, int64(0), participant.ConsecutiveInvalidInferences)`) at both transition test sites. The new assertion at `participant_test.go:168` and `:368` will catch any future refactor that drops the reset.
- The end-to-end test verifies the actual user-observable behavior (restored participant serves traffic AND stays ACTIVE), not just that the counter is 0. That's the right level for the integration coverage — unit tests pin the implementation detail, integration test pins the contract.
- The reset doesn't interact with any of the existing INVALID side effects (slashing, exclusion record, reputation reduction, epoch-group removal) — those are gated on the `originalStatus != INVALID && newStatus == INVALID` transition check at L65, all of which fire correctly because the counter is reset before that gate is evaluated and the gate doesn't depend on the counter.

patimen:
Was pulled into gm/microrelease

PR #1143, opened 2026-05-04
v0.2.13: microrelease

gmorgachev:
This PR prepares the v0.2.13 microrelease.

It fixes:

- confirmation PoC weight loss during new-model bootstrap.
- upstream `stop_reason` unmarshalling errors.
- devshard storage growth by adding pruning and Postgres support.
- automatic re-invalidation after a host comes back from invalid state.
- missing complaint-response authz grants for older DAPIs.
- devshard escrow param changes, including the old `20_000` nonce limit.
- Wasm keeper access for contract permission checks.
- devshard validation and recovery edge cases.

It also:

- adds the MiniMax-M2.7 governance model and PoC config
  (`PenaltyStartEpoch = 278`).
- adds addresses of several early miners and known brokers to the devshard
  creator allowlist.
- recalibrates `PocParams.Models[*].WeightScaleFactor` against the Qwen-on-B200
  reference after the vLLM 0.20.1 release, where Kimi ran too high on B* GPUs.
  Kimi `0.78` (Qwen-on-B200 + 10%, top-tier premium), MiniMax `0.3024`
  (= Qwen-on-B200).
- updates `Model.ValidationThreshold` from cross-version vLLM results: Qwen
  `0.940`, Kimi `0.900`, MiniMax `0.922`.
- reduces genesis guardian adjusted voting power to about 25% and sets the
  chain-wide governance quorum to `0.25`. With guardians not voting, this gives
  an effective 1/3 quorum among the remaining 75% of voting power
  (`0.25 / 0.75 = 0.334`).
- adds a guardian-signed transaction for emergency disabling and re-enabling
  devshard inference requests.

The upgrade also disables confirmation PoC for the rest of the upgrade epoch
so the new snapshot logic starts cleanly from the next epoch.

## models

- Adds MiniMax-M2.7 (`MiniMaxAI/MiniMax-M2.7`) as a governance model and as a
  PoC model config. `PenaltyStartEpoch` is hardcoded to `277` for the bootstrap
  activation epoch.
- Recalibrates `PocParams.Models[*].WeightScaleFactor` against the Qwen-on-B200
  reference after the vLLM 0.20.1 release. Kimi was too high on B* GPUs.
  Kimi = Qwen-on-B200 + 10% (top-tier premium), MiniMax = Qwen-on-B200:
  - Kimi (`moonshotai/Kimi-K2.6`): `0.78`
  - MiniMax (`MiniMaxAI/MiniMax-M2.7`): `0.3024`
- Updates `Model.ValidationThreshold` from cross-version vLLM results:
  - Qwen (`Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`): `0.940`
  - Kimi (`moonshotai/Kimi-K2.6`): `0.900`
  - MiniMax (`MiniMaxAI/MiniMax-M2.7`): `0.922`
- Adds `--enable-auto-tool-choice` to Kimi `ModelArgs` if missing.

## inference-chain

- Confirmation PoC used different model sets for measured weight, preserved
  weight, and reward rescaling. During new-model bootstrap, this could slash
  honest miners that served both an eligible model and a not-yet-eligible
  model. The fix stores one epoch snapshot of confirmable models and
  weight-scale factors, then uses that snapshot for all confirmation and
  reward-weight calculations.
- `ConsecutiveInvalidInferences` was not reset when a participant became
  ACTIVE again. One new bad inference could immediately invalidate them again.
  The counter is now reset on reactivation and upcoming promotion.
- DAPIs that joined before v0.2.12 did not have
  `MsgRespondDealerComplaints` in their cold-to-warm authz grants. The upgrade
  backfills that permission so they can respond to dealer complaints.
- Devshard settlement used a hardcoded `20_000` nonce limit. The limit is now
  `DevshardEscrowParams.MaxNonce`, and the v0.2.13 upgrade sets it to
  `1_000_000`. The upgrade also raises `MaxEscrowsPerEpoch` to `500_000`.
- The upgrade adds addresses of several early miners and known brokers to the
  devshard creator allowlist without removing existing allowed creator addresses.
- Genesis guardians previously held about 34% of adjusted voting power, which
  made quorum hard to reach when they did not vote. The upgrade reduces guardian
  power to about 25% via `GenesisOnlyParams.GenesisGuardianMultiplier = 0.33334`
  and sets the chain-wide governance quorum to `0.25`. Quorum is computed against
  total bonded power; with guardians not voting, this gives an effective 1/3
  quorum among the remaining 75% of voting power (`0.25 / 0.75 = 0.334`).
- The upgrade installs a grace-epoch entry for the current epoch with an
  extended `UpgradeProtectionWindow` (10000 blocks). Confirmation PoC triggers
  are skipped from the upgrade height through the end of the upgrade epoch, so
  the new snapshot logic only starts running from the next epoch. Reuses the
  v0.2.10 grace-epoch primitive.
- Wasm keeper access is resolved after app wiring, so contract permission checks
  work for bridge and liquidity-pool operations.
- Adds `MsgSetDevshardRequestsEnabled`, a guardian-signed transaction for
  emergency disabling and re-enabling devshard inference requests.

## decentralized-api

- Some OpenAI-compatible upstreams return numeric `stop_reason` values.
  `Choice.StopReason` now accepts any JSON type, so those responses no longer
  fail unmarshalling.
- Internal devshard storage migration no longer blocks dapi startup. Devshard
  routes stay unavailable until migration and recovery finish.

## devshard

- Devshard storage could grow forever because old escrow data stayed in one
  SQLite store. Storage is now epoch-scoped and prunes old epochs in the
  background, keeping the latest 3 epochs.
- Devshard needed a shared storage option for larger deployments. It can now
  use Postgres as the primary store, with SQLite kept as the local fallback.
- Postgres data is partitioned by `epoch_id` for sessions, diffs, and
  signatures, so pruning can drop old epoch data cleanly.
- State snapshots reduce recovery work for long-running sessions.
- Payload lookup is pinned to the escrow epoch with fallback for epoch-boundary
  and legacy epoch-0 requests.
- Current-epoch shard stats expose nonce, version, group, and per-host counters.

## bridge

- Bridge tooling handles Sepolia flags and converts Gonka BLS keys/signatures to
  the EIP-2537 format expected by the Ethereum contract.
- Adds scripts for GNK and wrapped-token bridge operations.


inference-chain/x/inference/types/weight.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func ConfirmationWeightOfModelNodes(modelNodes map[string][]*MLNodeInfo, scales []*ConfirmationWeightScale) int64 {
+	return confirmationWeight(modelNodes, scales)
+}
+
+func confirmationWeight(modelNodes map[string][]*MLNodeInfo, scales []*ConfirmationWeightScale) int64 {
+	coefficients := confirmationCoefficients(scales)
+	total := int64(0)
+	for modelID, nodes := range modelNodes {
```
patimen: I can see this is safe, for now, but unless we expect it to be a big perf hit we should STILL probably extract the keys for modelNodes and iterate over a sorted list of them instead of iterating over the map. Otherwise we are exposed inside this loop to possible consensus breaking changes.

inference-chain/x/inference/types/weight.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return confirmationWeight(modelNodes, scales)
+}
+
+func ConfirmationWeightOfModelNodes(modelNodes map[string][]*MLNodeInfo, scales []*ConfirmationWeightScale) int64 {
+	return confirmationWeight(modelNodes, scales)
+}
+
+func confirmationWeight(modelNodes map[string][]*MLNodeInfo, scales []*ConfirmationWeightScale) int64 {
+	coefficients := confirmationCoefficients(scales)
```
patimen: THis is all called in a loop, right? Shouldn't we be computing the coefficients along with scales (or instead of), so we only calculated it once instead of every time?

inference-chain/app/upgrades/v0_2_13/upgrades.go
```diff
@@ -0,0 +1,167 @@
...
+			"epoch", epochIndex)
+		return nil
+	}
+	params, err := k.GetParams(ctx)
+	if err != nil {
+		return err
+	}
+
+	confirmableModels := make(map[string]bool)
+	for _, groupData := range k.GetAllEpochGroupData(ctx) {
```
patimen, line 489: So we load all epoch group data into memory just to iterate and pick out one? 
We should use the iterator here... I get we don't have years fo EpochGroupData (and generally a max of 2 models), but it's a bad pattern. Ignore it if you must, but surely Cursor could fix this in a second.

patimen (commented):
Some initial concerns while I iterate a bit more on other things.

0xgonka (approved):
v0.2.13 microrelease: upgrade handler + a major refactor of the confirmation-weight system from per-call coefficient extraction to precomputed `ConfirmationWeightScale` entries on `EpochGroupData`. Bundles two unrelated fixes (PR #1142's consecutive-failure reset, devshard MaxNonce param) and an authz backfill for `MsgRespondDealerComplaints`. Generally well-shaped, but one semantic change worth confirming with the team and a couple of upgrade-handler concerns.

**Must fix:**
- None.

**Should fix:**
- **`ConfirmationWeightOfModelNodes` silently drops weight contributions from models not in `scales`** ([`weight.go:22-26`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/x/inference/types/weight.go#L22-L26)). Previously `ModelCoefficients(params.PocParams)` returned every governance-approved model; the new path only includes models marked "confirmable" (≥1 participant with `VotingPower > 0` per [`confirmation_weight_scales.go:17-24`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/x/inference/module/confirmation_weight_scales.go#L17-L24)). This is a real semantic change: a participant serving a model whose *other* members all have zero voting power contributes **zero** confirmation weight from that model under the new code, whereas the old code would have applied the model's coefficient regardless. If that's the intent (only count confirmation-weight for functional model groups) it's defensible — but the `continue` on missing key is silent. Worth either an explicit `LogDebug` when it fires, or a comment in the function explaining the design choice so a future reader doesn't assume it's a bug.

- **`GetAllEpochGroupData` in the upgrade handler is unbounded** ([`upgrades.go:100`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/app/upgrades/v0_2_13/upgrades.go#L100)). The keeper method iterates the full `EpochGroupDataMap` (every model × every epoch since genesis), then the handler filters in-memory by `groupData.EpochIndex == epochIndex`. On mainnet with many historical epochs this loads thousands of entries to find the few you actually need. Not a correctness issue — upgrades run once and there's plenty of memory — but worth a follow-up to use a prefix-scoped iterator (e.g. `EpochGroupDataMap.Iterate(ctx, collections.NewPrefixedPairRange[uint64, string](epochIndex))`) since this same access pattern will probably appear again. Doesn't have to land in this PR.

- **Conflict with PR #1142** ([`participant_status.go:51-53`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/x/inference/keeper/participant_status.go#L51-L53) and [`module.go:1269`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/x/inference/module/module.go#L1269)). These are byte-for-byte the same lines as the open PR #1142. Whichever lands first will make the other a no-op merge — but the PR descriptions don't cross-reference and the test additions partially duplicate. Recommend either pulling these specific lines out of #1143 (since #1142 is narrower and easier to review) or closing #1142 with a note that this PR supersedes. Otherwise they will merge-conflict.

**Nits:**
- The upgrade handler doesn't bump `InferenceModule.ConsensusVersion()` and adds no `RegisterMigration` entry. This is correct (the proto changes are pure additions, backward-compatible) but worth a one-line comment in `upgrades.go` saying "no consensus-version bump because all proto changes are field additions" — otherwise it looks like an oversight to the next reviewer.
- `IterateGrants` in `grantRespondDealerComplaintsAuthz` ([`upgrades.go:157-175`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/app/upgrades/v0_2_13/upgrades.go#L157-L175)) walks every authz grant on chain, then filters by `MsgStartInference` type URL. On the current network size (low hundreds of hosts × ~20 grants each) this is fine, but if the chain ever grows substantially this iteration becomes O(N grants). Not blocking; just worth knowing the cost is in the upgrade handler, not steady-state.
- `confirmationWeightScalesFromModels` at [`upgrades.go:222-226`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/app/upgrades/v0_2_13/upgrades.go#L222-L226) calls `coefficients[modelID].CloneOrOne()` — works because `coefficients[modelID]` returns nil for missing keys and `nil.CloneOrOne()` returns 1. Fine, but only if `CloneOrOne` is actually nil-safe; worth a quick re-read of that method to make sure the assumption holds for `*Decimal`.

**Looks correct:**
- `UpgradeName = "v0.2.13"` matches the PR title and (presumably) the governance proposal name. No ConsensusVersion bump is the right call — the proto additions are pure new fields, backward-compatible by protobuf semantics, and no existing data needs in-place rewriting.
- `backfillConfirmationWeightScales` correctly uses `vw.ConfirmationWeight > expected → cap to expected` ([`upgrades.go:128-130`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/app/upgrades/v0_2_13/upgrades.go#L128-L130)) — i.e. confirmation weight at upgrade can only go *down*, never up. That's the conservative direction and avoids any "free reward bump" via the migration.
- `grantRespondDealerComplaintsAuthz` correctly checks `existing != nil` before calling `SaveGrant` ([`upgrades.go:182-186`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/app/upgrades/v0_2_13/upgrades.go#L182-L186)) — `SaveGrant` would otherwise replace the existing expiration with the borrowed `MsgStartInference` expiration, which would be wrong if the host had already manually granted `MsgRespondDealerComplaints` with a different expiration.
- The upgrade handler's three migration steps are logically independent and idempotent: param setting overwrites cleanly, weight-scale backfill is a write-once-per-upgrade, authz backfill skips existing grants. Re-running (which shouldn't happen but is the right safety property) wouldn't double-apply.
- Proto changes are all pure additions (field 19 on `EpochGroupData`, field 8 on `DevshardEscrowParams`, new `ConfirmationWeightScale` message). No removals → no `reserved` markers needed. Backward compatible.
- `aggregation.go` deletion (-72 lines) and `epoch_group.go` -32 lines are clean — the agent verified no external references. The replacement pattern (centralized helpers in `types/weight.go`, scales precomputed at epoch end and stored in `EpochGroupData`) is more auditable than the previous "extract coefficients on every call" pattern. Easier to inspect what scaling factor is in effect for a given epoch.
- The `confirmation_poc.go` rewrite (+117/-95) is reorganization of phase-transition orchestration, not a change to consensus rules. Trigger window, deterministic randomness via block hash, and quorum logic are unchanged.
- The participant-status reset (same as #1142) is correctly gated on `reason == ConsecutiveFailures` rather than `newStatus == INVALID` — a participant going INVALID via `StatisticalInvalidations` keeps their counter intact, which is the right call since that path doesn't depend on the consecutive counter.
- Devshard `MaxNonce` is enforced at settlement time ([`devshard_settlement.go:38`](https://github.com/gonka-ai/gonka/blob/e59a07e0f68e683c265449364561c13df4780747/inference-chain/x/inference/keeper/devshard_settlement.go) and the new check in `msg_server_settle_devshard_escrow.go`), which is the right enforcement point — params nil-check before deref is also there.

inference-chain/x/inference/types/weight.go
gmorgachev: fixed

inference-chain/x/inference/types/weight.go
gmorgachev: sorted

gmorgachev:
> ConfirmationWeightOfModelNodes silently drops weight contributions from models not in scales

Yes, it's intended behaviour to not count non-eligible node in any punishments

> IterateGrants in grantRespondDealerComplaintsAuthz
quite cheap for now, less then second 

> confirmationWeightScalesFromModels a
yes, nil safe

blizko (approved):

devshard/storage/managed.go
```diff
@@ -0,0 +1,242 @@
...
+		return fmt.Errorf("%w: epoch %d below prune cursor %d", ErrEpochPruned, params.EpochID, m.prunedUpTo)
+	}
+	if err := m.inner.CreateSession(params); err != nil {
+		m.mu.Unlock()
+		return err
+	}
+	m.mu.Unlock()
+	m.observe(params.EpochID)
+	return nil
+}
```
x0152, line 190: It could be a bottleneck under load 

m.mu.Lock() is held across m.inner.CreateSession(params), so all session creations run one-by-one. Inner CreateSession is expensive for both db (SQLite file scan or Postgres network + tx), and PruneOnce uses the same lock

better solution:
```go
func (m *ManagedStorage) CreateSession(params CreateSessionParams) error {
    m.mu.RLock()
    pruned := m.prunedUpTo
    m.mu.RUnlock()
    if params.EpochID < pruned {
        return fmt.Errorf("%w: epoch %d below prune cursor %d", ErrEpochPruned, params.EpochID, pruned)
    }
    if err := m.inner.CreateSession(params); err != nil {
        return err
    }
    m.observe(params.EpochID)
    return nil
}
```

inference-chain/x/inference/module/confirmation_poc.go
```diff
@@ -455,33 +471,38 @@ func (am AppModule) evaluateConfirmation(
...
 	epochGroupData *types.EpochGroupData,
-	measured, preserved, notPreserved map[string]int64,
+	measured, preserved, totalExpected map[string]int64,
 ) (updated bool, ratios map[string]*types.Decimal) {
 	ratios = make(map[string]*types.Decimal, len(epochGroupData.ValidationWeights))
 	for i, vw := range epochGroupData.ValidationWeights {
 		addr := vw.MemberAddress
 		reading := preserved[addr] + measured[addr]
+		if totalExpected[addr] == 0 {
+			continue
```
x0152, line 481: It's not a bug right now, just a note: looks like this case was missed here (totalExpected == 0 path skips ratio update)

`ratios[addr] = types.DecimalFromDecimal(decimal.NewFromInt(1)) `

devshard/storage/postgres.go
```diff
@@ -0,0 +1,806 @@
...
+	_, err = s.pool.Exec(context.Background(),
+		`INSERT INTO devshard_snapshots (epoch_id, escrow_id, nonce, state_data, created_at)
+		 VALUES ($1, $2, $3, $4, $5)
+		 ON CONFLICT (epoch_id, escrow_id) DO UPDATE
+		 SET nonce = EXCLUDED.nonce, state_data = EXCLUDED.state_data, created_at = EXCLUDED.created_at
+		 WHERE devshard_snapshots.nonce <= EXCLUDED.nonce`,
+		epochID, escrowID, nonce, data, time.Now().Unix(),
+	)
+	return err
+}
```
x0152, line 686: This might also be a bottleneck: running CREATE TABLE IF NOT EXISTS ... PARTITION OF ... on every write is really expensive. Looks like ensurePartition was forgotten here, which already handles this with caching

decentralized-api/internal/devshard/manager.go
```diff
@@ -215,53 +383,263 @@ func (m *HostManager) recoverSession(escrowID string) error {
...
+		Nonce:           st.LatestNonce,
+		Version:         st.Version,
+		CachedAt:        now.Unix(),
+		CacheTTLSeconds: int64(statsCacheTTL / time.Second),
+		HostStats:       statsHostStatsFromState(st.HostStats),
+		Group:           append([]types.SlotAssignment(nil), st.Group...),
+	}
+
+	m.statsMu.Lock()
+	m.statsDetailsCache[escrowID] = statsShardDetailCache{response: resp, cached: now}
```
Ryanchen911, line 575: It seems that the statsDetailsCache does not have eviction, statsDetailsCache is keyed by escrowID and only ever inserted into here at line 567; seems like nothing in the file calls delete(m.statsDetailsCache, ...). The TTL check at line 537 only affects reads — an expired entry is bypassed but not removed.
-Once ManagedStorage prunes the escrow's epoch (retain=3), currentEpochActiveSession at line 544 returns ErrSessionNotFound.
-The function returns the error without touching the cache.
-Any escrowID whose epoch has been pruned is never hit, refreshed, or deleted again — its entry persists for the lifetime of the process.

devshard/storage/postgres.go
```diff
@@ -0,0 +1,806 @@
...
+	if err := rows.Err(); err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	for _, partition := range partitions {
+		if _, err := s.pool.Exec(ctx, fmt.Sprintf(`DROP TABLE IF EXISTS %s`, partition)); err != nil {
+			return fmt.Errorf("drop %s: %w", partition, err)
+		}
+	}
```
x0152, line 775: Should we clear in-memory index for pruned epochs before dropping partitions to avoid recreating pruned state via concurrent requests (same pattern seems present in SQLite prune too)?

devshard/storage/hybrid.go
```diff
@@ -0,0 +1,379 @@
...
+		pgErr = pg.pruneBefore(cutoff)
+	} else {
+		pgErr = fmt.Errorf("postgres backend unavailable for prune")
+	}
+	h.forgetPruned(func(ep uint64) bool { return ep < cutoff })
+	if pgErr != nil {
+		return pgErr
+	}
+	return sqliteErr
+}
```
x0152, line 355: sqliteErr gets lost when pgErr is set
same issue in pruneBefore and PruneEpoch

devshard/host/snapshot.go
akup, line 1: I think it is more efficient to store snapshots in protobuf, not json

decentralized-api/internal/devshard/manager.go
```diff
@@ -6,8 +6,10 @@ import (
 	"errors"
```
qdanik, line 7: I hope one day this god file will be splitted into small logical and covered by test

devshard/storage/factory.go
```diff
@@ -0,0 +1,36 @@
...
+//   - If PGHOST is set, use sticky hybrid storage: Postgres for new sessions
+//     when available, SQLite fallback while Postgres is down, and lazy retry.
+func NewStorage(ctx context.Context, sqliteDir string) (Storage, error) {
+	pgHost := os.Getenv("PGHOST")
+	if pgHost == "" {
+		slog.Info("devshard storage: PGHOST not set, using sqlite", "dir", sqliteDir)
+		return NewSQLite(sqliteDir)
+	}
+
+	sqlite, err := NewSQLite(sqliteDir)
```
akup, line 22: Here we always open connections to Sqlite even if postgres is available.
Also at `NewSQLite` we are creating 1 write connection pool, 10 read pools and attempt to create tables.
It would be better to do not pass `sqlite` to `NewHybridStorage` and open it only on real fallback

devshard/storage/factory.go
```diff
@@ -0,0 +1,36 @@
...
+	"os"
+	"time"
+)
+
+// NewStorage builds the canonical Storage for a host process.
+//
+//   - If PGHOST is unset, use SQLite under sqliteDir.
+//   - If PGHOST is set, use sticky hybrid storage: Postgres for new sessions
+//     when available, SQLite fallback while Postgres is down, and lazy retry.
+func NewStorage(ctx context.Context, sqliteDir string) (Storage, error) {
```
akup, line 15: The `NewSQLite` is very inefficient as it is not reusing pools. While Postgres is opened with `getOrConnectPostgres` and doesn't create pools on every `NewStorage` called per session.

`NewSQLite` is creating read and write pools and attempts to create tables on each call. This should be fixed and pools should be reused

devshard/storage/postgres.go
```diff
@@ -0,0 +1,806 @@
...
+	pool, err := pgxpool.New(ctx, "")
+	if err != nil {
+		return nil, fmt.Errorf("connect to postgres: %w", err)
+	}
+	if err := pool.Ping(ctx); err != nil {
+		pool.Close()
+		return nil, fmt.Errorf("ping postgres: %w", err)
+	}
+
+	if _, err := pool.Exec(ctx, pgCreateParents); err != nil {
```
akup, line 127: It's better to have a migration point on startup where we create parents or migrate them version to version, then calling it on each `NewPostgres` call. Same for `indexExisting`

devshard/storage/postgres.go
akup: I think it is better to have one migration point on devshardd startup where create or migrate all the tables and remove CREATE TABLE IF NOT EXISTS everywhere from hotpath

inference-chain/proto/inference/inference/params.proto
```diff
@@ -380,6 +380,7 @@ message DevshardEscrowParams {
   repeated string allowed_creator_addresses = 5;
   uint64 token_price = 6;
   repeated DevshardApprovedVersion approved_versions = 7;
+  uint32 max_nonce = 8;
```
akup, line 383: It is never read at devshardd, so `VerifyDevshardSettlement` can throw error but devshard will not cap nonces and it is possible that valid devshard will fail on settlement

devshard/storage/factory.go
akup: By design sqlite is using db per epoch, and a comment isn't correct.

But actually It is hard to estimate what is better for overall performance:
To create a lot of sqlite dbs (when there are a lot of escrows) with a total 11 connections in the pool for each db, but have ability to prune epoch by file delete. Or have one db file and prune by index lookup.

Overall pruning is happening once per epoch and could be optimized, and it seams it could be better to have better performance having less connections in a pool and one db file, but a little less efficient pruning

But when the postgres is available I still see no reason to create NewSQLite for the epoch.
Moreover by storage design doc it is stated that:

> Decision: When `PGHOST` is set, `HybridStorage` checks Postgres first when it is
available, then SQLite. Once an escrow is found or created in one backend, every
future session-keyed operation for that escrow uses the same backend.

So We should skip SQLite creation on each escrow when postgres is available

devshard/storage/managed.go
```diff
@@ -0,0 +1,242 @@
...
+	t := time.NewTicker(m.pruneInterval)
+	defer t.Stop()
+	for {
+		select {
+		case <-m.stop:
+			return
+		case <-t.C:
+			m.PruneOnce(context.Background())
+		}
+	}
```
akup, line 107: I think this pruning should be optimized a lot.

We need to prune only if epoch change. So even before touching dbs we can check if newEpoch != previousEpoch and this will optimize a lot and even using multiple db files (one per epoch) can be revised.

we can check on escrow start, what epoch it is (should be in escrow start message). And store in memory previousEpoch to compare on escrow start if it changed. And if `newEpoch != previousEpoch` call pruning asynchronously

decentralized-api/internal/devshard/shared_runtime.go
```diff
@@ -29,6 +29,11 @@ import (
 
 type MLRequestExecutor func(ctx context.Context, model string, body []byte) (*http.Response, error)
 
+const (
+	MLNodeHTTPTimeout   = 10 * time.Minute
```
blizko: Is this expected to be the E2E request limit? If so - likely needs to be higher

gmorgachev: I think you are right, here i bumped from 5min but historically we used 20 everywhere

devshard/storage/managed.go
gmorgachev: Agree, let's address in next devshard-only release

devshard/storage/postgres.go
gmorgachev: I think it also should be addressed in devshard-only upgrade, for simplicity. This version of `api` works on all nodes already, so it should be safe for now

devshard/storage/managed.go
gmorgachev: @x0152 maybe you can add in devshard-only release branch?

devshard/storage/postgres.go
gmorgachev: @x0152 maybe you can test and add in devshard-only release branch?

devshard/host/snapshot.go
gmorgachev: Agree... Let's address in devshard-only release too

inference-chain/proto/inference/inference/params.proto
gmorgachev: I think we should start control this on devshard side too

gmorgachev:
@akup @x0152 let's add devshard improvement separately, to not delay this microrelease
@akup has the branch where further changes are being merged for the next release

devshard/storage/postgres.go
akup: I think in-memory indexing is the SQLite feature, postgres don't need them

akup:
> @akup @x0152 let's add devshard improvement separately, to not delay this microrelease @akup has the branch where further changes are being merged for the next release

I think you are about this branch:
https://github.com/gonka-ai/gonka/pull/1162
But it is branch in the fork

But it is branch at fork, it would be nice to create branch based on 0.2.13 for devshard separate release. And pr to there there 
https://github.com/gonka-ai/gonka/pull/1164
https://github.com/gonka-ai/gonka/pull/1159
https://github.com/gonka-ai/gonka/pull/1157

and add there all db related fixes found at this review
https://github.com/gonka-ai/gonka/issues/1167

And also merge with https://github.com/gonka-ai/gonka/pull/1165

PR #1148, opened 2026-05-06
Multi model scheduling GIP

0xgonka:


unameisfine:
## Implementation-level review from the current codebase

Strong design — separating equilibrium (Layer 1) from price discovery (Layer 2) is the right decomposition. A few concrete gaps from the current code:

### 1. §4.1 clamp contradicts §5.4 retirement

§4.1: `effective_coeff = clamp(base × demand × novelty × status, min_band, max_band)`

§2.2 requires `0 < min_band ≤ base`, so the clamp prevents the coefficient from ever reaching 0. But §5.4 states RETIRED → `effective_coeff = 0`, and §4.4 has DEPRECATED decay 1.0 → 0.0. With `min_band > 0`, RETIRED models get paid `min_band` and decaying DEPRECATED models stop at `min_band`, not 0.

Either special-case the clamp for terminal states, multiply `min_band` by `status_factor`, or define `effective_min_band(status)`.

### 2. `scheduler_allocation_cap` is a no-op when bootstrapping with size-1 caps

§3.2: "MLNodes with a capability set of size 1 ... are not subject to scheduler reallocation."
§4.3: For LEARNING models, `scheduler_allocation_cap = 0.1` "limits the scheduler's exposure to unproven models even if their early demand is high."

The first hosts willing to bootstrap a new model will rationally declare a size-1 pin (the only honest signal of commitment). Size-1 pins bypass the scheduler, so the cap doesn't bind on them. The cap binds only on lazy hosts (size > 1), who by definition have weaker preference for the new model.

The safety cap is therefore inert when it would matter most. Either gate the cap at consensus-weight aggregation (not at scheduler allocation), or accept and document the limitation.

### 3. Build order: this GIP depends on multi-model PoC schema that doesn't exist yet

The proposal references `PoCModelConfig`, per-model `weight_scale_factor`, `penalty_start_epoch`, `BootstrapDelegationSnapshot`, `MODEL_STATUS_*`, `ConfirmationWeightScale`. Grepping `inference-chain/`:

- `PoCModelConfig` is not a proto type yet.
- `WeightScaleFactor` exists only as a **single global** value in `PocParams` ([params.pb.go:1031](https://github.com/gonka-ai/gonka/blob/main/inference-chain/x/inference/types/params.pb.go#L1031), consumed by [confirmation_poc.go:383](https://github.com/gonka-ai/gonka/blob/main/inference-chain/x/inference/module/confirmation_poc.go#L383) and [chainvalidation.go:906](https://github.com/gonka-ai/gonka/blob/main/inference-chain/x/inference/module/chainvalidation.go#L906)).
- `penalty_start_epoch`, `ConfirmationWeightScale`, model `status` enum: none present.

"Requires: Multi-Model PoC" is correct, but the upstream is also still in proposal stage. Phase 1 of this GIP can't ship before multi-model PoC's Phase 1, and a reviewer can't validate the extension points until the upstream proto contract is fixed. Worth pinning "Requires" to a specific commit of the upstream proposal.

### 4. `MLNodeCapability` overlaps with existing `HardwareNode.models`; migration is softer than spec implies

[`hardware_node.proto:12-19`](https://github.com/gonka-ai/gonka/blob/main/inference-chain/proto/inference/inference/hardware_node.proto#L12-L19) already has:

```proto
message HardwareNode {
  string local_id = 1;
  HardwareNodeStatus status = 2;
  repeated string models = 3;          // per-MLNode model list, today
  repeated Hardware hardware = 4;
  ...
}
```

So a per-MLNode `supported_models` list is not new. The new contributions are the canonical `gpu_class`, the paid `MsgSetMLNodeCapability` transaction, and the separate storage location.

§6 says "MLNodeCapability is NOT auto-backfilled" because `Hardware.type` strings can't be unambiguously mapped to a canonical `gpu_class`. But `HardwareNode.models` *can* be backfilled into `MLNodeCapability.supported_models` losslessly. Backfilling models but leaving `gpu_class` unset (with the documented "ineligible until declared" semantics) gives a softer migration path than empty-pool first epoch.

Also worth specifying the source-of-truth split: PoC validation today reads `HardwareNode.models`; the scheduler will read `MLNodeCapability`. Either deprecate the former or state which subsystem reads which.

### 5. `ENFORCED_MODEL_ID` overrides any operator-declared capability

[`enforced_model.go:38-50`](https://github.com/gonka-ai/gonka/blob/main/decentralized-api/broker/enforced_model.go#L38-L50) hard-overrides `node.Models` to a single enforced model unless `ENFORCED_MODEL_ID="disabled"`:

```go
node.Models = map[string]apiconfig.ModelConfig{
    modelId: {Args: args},
}
```

Default is `Qwen/Qwen3-235B-A22B-Instruct-2507-FP8`. Under current production posture, every operator's MLNodes effectively have `node.Models = {flagship}` regardless of any on-chain declaration. If `MLNodeCapability` is the new authoritative source and DAPIs apply enforced_model.go after reading it, capability sets of size > 1 are silently overridden to size 1.

§3.6 ("DAPI behavior") needs to address this: deprecate `ENFORCED_MODEL_ID` enforcement at the DAPI level (release-coordinated), or specify precedence (capability wins; env var becomes a default for undeclared nodes only).

### 6. Confirmation PoC currently uses one global `weight_scale_factor`

[`confirmation_poc.go:383`](https://github.com/gonka-ai/gonka/blob/main/inference-chain/x/inference/module/confirmation_poc.go#L383) reads one `LegacyDec` and propagates it through `evaluateConfirmation` → `updateConfirmationWeights{V1,V2}`. v0.2.13 is currently fixing a related bug ([#1143](https://github.com/gonka-ai/gonka/pull/1143) release notes: "Confirmation PoC used different model sets for measured weight, preserved weight, and reward rescaling. During new-model bootstrap, this could slash honest miners that served both an eligible model and a not-yet-eligible model.").

This GIP makes `effective_coeff` per-model **and** time-varying (within an epoch via `status_factor` decay). cPoC integration needs to specify:
- Which `effective_coeff` cPoC uses for an inference — per-model for the model the inference was on?
- For multi-epoch confirmations: read `effective_coeff` at submit time or at finalize time? Matters for decaying DEPRECATED models.
- The v0.2.13 snapshot mechanism ("stores one epoch snapshot of confirmable models and weight-scale factors") would need to extend to per-model `effective_coeff`.

PR #1157, opened 2026-05-10
`devshard` Keep upstream SSE alive through devshard_meta

akup:
# devshardctl: drain host SSE through `devshard_meta` on client cancel

## Problem

The devshard inference protocol relies on the **user (devshardctl) being the sole sequencer** of mempool transactions into signed user diffs. Host-to-host gossip is intentionally minimal (settlement-time only in the target deployment), so `MsgFinishInference`, `MsgValidation`, and friends only enter chain state when devshardctl puts them into a diff.

When a client streams `/v1/chat/completions` from devshardctl, the host emits SSE in this order:

1. `devshard_receipt` (executor receipt + state sig)
2. Model `data:` chunks, ending with the model `data: [DONE]`
3. **`devshard_meta`** — carries the host's mempool, including `MsgFinishInference` for this nonce

A naive OpenAI-style client closes the connection as soon as it sees `data: [DONE]`. Previously, devshardctl ran `runInference` with `r.Context()`, which canceled when the client disconnected. That cancel propagated up the call chain and aborted the **upstream** SSE read — *before* the host could deliver `devshard_meta`. Net effect on every early-close request:

- `ProcessResponse` saw an empty/partial mempool.
- `MsgFinishInference` never reached `session.pendingTxs`.
- The next signed user diff did not include Finish, so the inference stayed `Started` in the SM, escrow funds remained reserved, and downstream phases (Validation, Finalization, settlement) had nothing to act on.

Because gossip is not the recovery mechanism in this design, a client that always closes at `[DONE]` (whether by bug or as an attack) could systematically prevent inferences it sponsored from being settled — a liveness/integrity issue, not just noisy logs.

## Potential harm

If someone is using Gonka to provide inferences and hosts own devshardctl with authentication for clients, clients that drop connection can break devshardctl sequencing `MsgFinishInference`

## Fix

Decouple proxy upstream lifetime from `r.Context()` so the host SSE body is fully drained — including the trailing `devshard_meta` event — even when the OpenAI client disconnects mid-stream.

### Changes

- **New `cancelFlag`** (one-shot, nil-safe) carries "client gone" downward without canceling the upstream context.
- **`watchClientCancel`** triggers the flag from `r.Context().Done()`.
- **`proxyWriter`** (replaces `deferredWriter`) swallows writes after cancel but reports success so callers don't error mid-protocol.
- **`handleStreaming` / `handleNonStreaming`** run `runInference` with `context.Background()` and early-return when the flag fires.
- **`sendAndProcess`** applies `metaDrainTimeout` (5s) after the flag fires so a malicious upstream host cannot pin the proxy indefinitely.
- **`runInference`** skips attempt 2 + `writeStreamReset` when the client is gone (mitigation 2); `handleTimeout` still runs to record `MsgTimeoutInference` if the executor truly didn't finish.

### Tests

Unit coverage for `cancelFlag`, `proxyWriter`, and two new `runInference` scenarios:

- Client gone with executor down → skips retry, goes to timeout.
- Client gone on happy path → `MsgFinishInference` still reaches `pendingTxs`.

Existing tests updated for the new signature.

### DIFF
--- a/decentralized-api/cosmosclient/tx_manager/errors.go
+++ b/decentralized-api/cosmosclient/tx_manager/errors.go
@@ -67,6 +67,10 @@ var retryablePatterns = []string{
 
 	"unordered transaction has a timeout_timestamp that has already passed",
 	"unordered tx ttl exceeds",
+	// Out-of-gas: the per-msg-type estimate underestimated this batch's
+	// real consumption. Retry with a bumped gasWanted (estimateBatchGas
+	// applies a multiplier per attempt, see gas_estimate.go).
+	"out of gas",
 }
 
 // isRetryableRawLog checks if the raw log contains any retryable error patterns
--- a/decentralized-api/cosmosclient/tx_manager/gas_estimate.go
+++ b/decentralized-api/cosmosclient/tx_manager/gas_estimate.go
@@ -1,0 +1,186 @@
+package tx_manager
+
+import (
+	"cosmossdk.io/math"
+	"github.com/cosmos/cosmos-sdk/client"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	blstypes "github.com/productscience/inference/x/bls/types"
+	collateraltypes "github.com/productscience/inference/x/collateral/types"
+	inferencetypes "github.com/productscience/inference/x/inference/types"
+)
+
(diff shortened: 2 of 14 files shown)


akup:
Replaced by this PR from my another account while this account was blocked by github:

https://github.com/gonka-ai/gonka/pull/1235
It is merged with newer branch

PR #1159, opened 2026-05-11
`devshard` Finish gossip on missed diff

akup:
## Summary

- wait two full slot round-robin rotations before gossiping a locally produced `MsgFinishInference` that the user did not include in a diff
- make stale-finish gossip grace a fixed host policy (`2 * len(group)`) instead of a configurable option
- update recovery tests to cover the grace threshold and confirm peer-imported finishes are still not amplified


akup:
Replaced by this PR from my another account while this account was blocked by github:

https://github.com/gonka-ai/gonka/pull/1238
It is merged with newer branch

PR #1160, opened 2026-05-11
fees additional fixes

0xgonka:
Follow-up to #1129. Three independent fixes against the network-duty fee-bypass code path:

## 1. Correct `GasCap` comment + raise `Priority` (ante.go)

`NetworkDutyFeeBypassDecorator`'s `GasCap` is unchanged at **3,000,000,000**, but the inline comment was wrong: it cited `BatchGasLimit = 100M` in the DAPI's `tx_manager`. The actual constant is `BatchGasLimit = 1B` (decentralized-api/cosmosclient/tx_manager/tx_manager.go:58), so the 3B cap is 3× the batch limit (giving headroom for the largest legitimate batched PoC V2 / weight-distribution txs), not 30×. Comment updated to reflect reality.

`Priority` raised **500_000 → 10_000_000**. Reason: `LiquidityPoolFeeBypassDecorator` already uses `Priority=1_000_000` for zero-fee swap traffic. Network-duty txs (PoC, validation, BLS, weight distribution) are consensus-critical — if mempool fills with discretionary swap txs, duty txs must outrank them, not tie. 10× LP priority is the smallest gap that guarantees ordering under mempool pressure.

## 2. Add three missing message types to the fee-exempt allowlist (ante_fee.go)

The post-#1129 audit found three protocol-duty messages that pay fees today but shouldn't, because each is already rate-limited or access-controlled by other mechanisms:

| Message | Why exempt |
|---|---|
| `MsgSubmitSeed` | PoC seed-submission, throttled by `PocPeriodValidationDecorator` window like the other PoC duty messages already in the list. Omitting it was an oversight in #1129. |
| `MsgSettleDevshardEscrow` | Protocol-driven disbursement tx; allowlist-restricted via `EscrowAllowListPermission` (permissions.go:123) and capped by `DevshardEscrowParams.MaxEscrowsPerEpoch`. |
| `MsgRespondDealerComplaints` | BLS DKG protocol message; epoch-scoped, duplicate-checked, deadline-enforced like the other four BLS duty messages already exempt. |

Negative cases (kept paid) with explanation comments added to the source so this doesn't drift again:

- **`MsgPoCV2StoreCommit`** — intentionally NOT exempt because it carries a count-proportional sybil-defense gas charge (see `chargePoCV2StoreCommitGas` in `msg_server_poc_v2_commit.go`). Exempting it would defeat the sybil defense.
- **`MsgCreateDevshardEscrow`** — intentionally NOT exempt because it is user-driven (the escrow creator), not protocol-driven. The matching settlement is exempt; the creation is not.

## 3. Tests (ante_fee_test.go)

`TestNetworkDutyBypass_AllExemptMessages` and `TestIsExemptMessageType` both extended to cover the three newly-exempt messages and to lock in the negative-case rejection of `MsgPoCV2StoreCommit` and `MsgCreateDevshardEscrow`.

## Test plan

- [ ] `go test ./inference-chain/app/ -run TestNetworkDuty -v` passes
- [ ] `go test ./inference-chain/app/ -run TestIsExemptMessageType -v` passes
- [ ] Full CI (currently green: chain build/test, devshard, DAPI, lint)
- [ ] Devnet smoke: submit one of each newly-exempt tx with zero fee and confirm it lands
- [ ] Devnet smoke: submit a `MsgPoCV2StoreCommit` with zero fee and confirm it is REJECTED for insufficient fee (regression check on the negative case)

## Scope notes

- Base branch is `upgrade-v0.2.13`. This ships as part of whatever v0.2.13 re-attempt follows the prior governance failure — not against `main` directly.
- No proto changes, no migration, no state read/write changes. Pure ante-handler config + the allowlist set.


patimen (approved):
I think a good chunk of the comments in here belong in the PR Description. Otherwise seems fine

patimen (commented):
Sorry, meant to be just a comment review, not approve (yet)

PR #1162, opened 2026-05-12
`devshard` remove seed reveal and add temporary payload pruning

akup:
## Problem

The old flow still carried `SeedReveal`-related behavior and retained executor payloads for too long.

At the same time, after removing the reveal-based validation flow, we still do not have a precise protocol-level signal for when a finished inference definitely no longer needs validation. That means executor payloads can accumulate even when they are no longer practically useful.

This PR addresses both issues:
- removes the obsolete `SeedReveal` transaction flow
- reduces payload retention pressure with per-inference pruning

## Summary

- remove `SeedReveal` tx handling from the devshard flow
- keep session finalization deadline-driven instead of reveal-driven
- remove reveal/compliance-related state handling that is no longer needed
- add `DeleteInference` to payload storage backends and managed storage
- prune payloads immediately when an inference reaches a terminal outcome (`Validated`, `Invalidated`, `TimedOut`)
- add a temporary timeout-based pruning path for stale `Finished` inferences
- make validators skip quietly when the executor returns `404` for an already-pruned payload
- wire host prune events through the devshard manager into payload deletion
- add and update tests for reveal removal, pruning, and skip behavior

## Why timeout-based pruning is temporary

We still do not have a reliable way in the current protocol to know whether a `Finished` inference should still be validated later.

Because of that, this PR uses a conservative temporary heuristic for stale `Finished` inferences:
- wait for both a nonce-based grace window and a wall-clock timeout
- prune only if the inference is still `Finished`

This is a temporary solution until the new validation protocol is integrated and gives us a clearer lifecycle signal for validation and payload retention.

## Test plan

- `go test ./devshard/host/...`
- `go test ./devshard/...`
- `go test -short ./decentralized-api/payloadstorage/...`
- `go test -short ./decentralized-api/internal/validation/...`
- `go test -short ./decentralized-api/internal/server/public/...`

## Notes

`decentralized-api/internal/devshard` still has an existing upstream failing test:
`TestTokenCountValidation_AllowsOneTokenDrift`

That failure is unrelated to this PR.


gmorgachev:
@0xgonka could you review?

devshard/docs/storage-design.md
```diff
@@ -31,20 +31,7 @@ and route through a local `escrow_id -> epoch_id` index.
...
-escrow belong to that partition even if settlement happens after an epoch
-boundary.
-
-Consequence: If local storage sees the same escrow in two epochs, it is
-corruption. The code must return an error rather than choosing a side.
-
-Epoch `0`: the chain can set effective epoch index to `0`, and
-`MsgCreateDevshardEscrow` stores that value. Storage therefore treats epoch `0`
-as valid and does not use it as a missing-value sentinel.
+As
```
qdanik, line 34: Please, double check here. Looks like doc has been broken

PR #1094, opened 2026-04-20
feat: make devshard SessionConfig governance-configurable

unameisfine:
Closes #1028

## Problem

`SessionConfig` values (`RefusalTimeout=60`, `ExecutionTimeout=1200`, `ValidationRate=5000`) are hardcoded in `subnet/types/config.go`. Governance cannot tune timeout behavior or validation sampling rate without a code deploy, breaking the single-source-of-truth rule.

## Solution

Thread `RefusalTimeout`, `ExecutionTimeout`, and `ValidationRate` through the same governance -> escrow -> subnet pipeline that `TokenPrice` already uses.

## Changes

| Layer | File | Change |
|-------|------|--------|
| Proto | `params.proto` | +3 fields to `SubnetEscrowParams` (fields 7-9) |
| Proto | `subnet_escrow.proto` | +3 fields to `SubnetEscrow` (fields 9-11) |
| pb.go | `params.pb.go`, `subnet_escrow.pb.go` | Struct, getters, Marshal, Size, Unmarshal |
| Keeper | `msg_server_create_subnet_escrow.go` | Copy params to escrow at creation |
| Bridge | `interface.go`, `rest.go` | Parse + forward new fields |
| Subnet | `config.go` | New `SessionConfigFromEscrow()` factory |
| Subnet | `httpsession.go` | Use `SessionConfigFromEscrow` |

## Backward compatibility

All new fields default to zero. `SessionConfigFromEscrow` falls back to existing hardcoded defaults for zero values. Escrows created before this change work identically — no migration needed.

## Tests

- `subnet_escrow_session_config_test.go` — protobuf round-trip, zero-value backward compat, size encoding, getters (9 tests)
- `config_test.go` — `SessionConfigFromEscrow` full override, zero fallback, partial override, backward compat with `SessionConfigWithPrice` (7 tests)
- `rest_test.go` — REST response parsing with and without new fields (2 tests)

```
ok  subnet/types     0.003s  (7 tests)
ok  subnet/bridge    0.006s  (2 new + existing)
ok  x/inference/types 0.031s (9 tests)
```

PR #1164, opened 2026-05-13
feat: route all hardcoded SessionConfig values through DevshardEscrowParams (closes #1028 SessionConfig part)

### LINKED ISSUE #1028: `devshards` `SessionConfig` setting by governmant
Currently devshard `SessionConfig` has a lot of hardcoded values. They should be settable on new escrow start from mainnet, and should be configurable by governance.

For example https://github.com/gonka-ai/gonka/pull/1005 introduces `MaxInferencesPerSubnet` that is also used for checking at `‎inference-chain/x/inference/keeper/subnet_settlement.go` that is breaking single source of truth rule

unameisfine:
## Summary

Rebase + scope completion of #1094 against the v0.2.12 `subnet/` → `devshard/` rename. Closes the hardcoded-values part of #1028 by routing **all** session-config governance fields through `DevshardEscrowParams`.

Five new governance fields, captured per-escrow at creation time:

| Field | Default | Used in |
|---|---|---|
| `refusal_timeout` | 60 | `devshard/host` — wait before reason=refused timeout |
| `execution_timeout` | 1200 | `devshard/host` — wait before reason=execution timeout |
| `validation_rate` | 5000 (bps) | validation sampling |
| `create_devshard_fee` | 10_000 | `devshard/state` — one-time fee at session create |
| `fee_per_nonce` | 1_000 | `devshard/state` — per-nonce fee at diff apply |

Flow mirrors the existing `TokenPrice` plumbing from v0.2.12: `DevshardEscrowParams` → `DevshardEscrow` (captured at `MsgCreateDevshardEscrow`) → REST bridge → `devshard.types.SessionConfigFromEscrow`.

## Motivation

Per #1028 (CTO request, 2026-04-07):

> Currently devshard `SessionConfig` has a lot of hardcoded values. They should be settable on new escrow start from mainnet, and should be configurable by governance.

The issue also calls out the SST anti-pattern in #1005 (`MaxInferencesPerSubnet` duplicated in config + keeper). This PR addresses that by establishing the canonical **params → escrow → bridge → SessionConfig** path as the single source of truth for every session-config governance value.

## Backward compatibility — no upgrade handler / migration required

Zero values fall back to the compiled `DefaultSessionConfig` constants (same pattern `SessionConfigWithPrice` already used for `tokenPrice == 0` in v0.2.12). Existing chains and existing escrows keep behaving exactly as before until governance opts in.

## API changes

`SessionConfigFromEscrow` was introduced with a struct-based parameter (`EscrowSessionFields`) so future additions don't churn call-site signatures. `SessionConfigWithPrice` is retained as a deprecated wrapper.

## What's NOT included (intentional)

- **`VoteThreshold`** remains derived from `groupSize`. A configurable threshold below `groupSize/2` would weaken the liveness/security guarantees of the slot-majority check.
- **`penaltyValidationRate`** in `devshard/state/machine.go:19` is a separate penalty-policy parameter (rate for unrevealed seeds), not part of SessionConfig — happy to do this in a follow-up if you'd like the same treatment.

## Test plan

- [x] Protobuf round-trip for `DevshardEscrowParams` and `DevshardEscrow` with all 5 new fields
- [x] Zero-fallback to defaults; partial override; group-size-derived field preservation
- [x] REST bridge parsing for new fields + backward-compat test where they're missing from JSON
- [x] `DefaultDevshardEscrowParams()` populates all 5 new defaults
- [x] `DevshardEscrowParams.Validate()` rejects negative timeouts and `validation_rate > 10000`
- [x] All existing tests pass: `x/inference/{types,keeper}`, `app/upgrades/v0_2_12`, full `devshard/...` (14 packages), `decentralized-api/internal/devshard/...`, `cmd/inferenced/...`

cc: @gmorgachev @akup — this is a refresh of #1094 against the new devshard layout, expanded to cover the full hardcoded SessionConfig surface.

### DIFF
--- a/decentralized-api/internal/devshard/bridge.go
+++ b/decentralized-api/internal/devshard/bridge.go
@@ -43,13 +43,19 @@ func (b *ChainBridge) GetEscrow(escrowID string) (*bridge.EscrowInfo, error) {
 	}
 
 	return &bridge.EscrowInfo{
-		EscrowID:       escrowID,
-		Amount:         resp.Escrow.Amount,
-		CreatorAddress: resp.Escrow.Creator,
-		AppHash:        appHash,
-		Slots:          resp.Escrow.Slots,
-		TokenPrice:     resp.Escrow.TokenPrice,
-		EpochID:        resp.Escrow.EpochIndex,
+		EscrowID:          escrowID,
+		Amount:            resp.Escrow.Amount,
+		CreatorAddress:    resp.Escrow.Creator,
+		AppHash:           appHash,
+		Slots:             resp.Escrow.Slots,
+		TokenPrice:        resp.Escrow.TokenPrice,
+		EpochID:           resp.Escrow.EpochIndex,
+		MaxNonce:          resp.Escrow.MaxNonce,
+		RefusalTimeout:    resp.Escrow.RefusalTimeout,
+		ExecutionTimeout:  resp.Escrow.ExecutionTimeout,
+		ValidationRate:    resp.Escrow.ValidationRate,
+		CreateDevshardFee: resp.Escrow.CreateDevshardFee,
+		FeePerNonce:       resp.Escrow.FeePerNonce,
 	}, nil
 }
 
--- a/decentralized-api/internal/devshard/manager.go
+++ b/decentralized-api/internal/devshard/manager.go
@@ -240,7 +240,15 @@ func (m *HostManager) create(escrowID string) (*transport.Server, error) {
 
 	creatorAddr := escrow.CreatorAddress
 
-	config := types.SessionConfigWithPrice(len(group), escrow.TokenPrice)
+	config := types.SessionConfigFromEscrow(len(group), types.EscrowSessionFields{
+		TokenPrice:        escrow.TokenPrice,
+		RefusalTimeout:    escrow.RefusalTimeout,
+		ExecutionTimeout:  escrow.ExecutionTimeout,
+		ValidationRate:    escrow.ValidationRate,
+		CreateDevshardFee: escrow.CreateDevshardFee,
+		FeePerNonce:       escrow.FeePerNonce,
+		MaxNonce:          escrow.MaxNonce,
+	})
 
 	sm, err := state.NewStateMachine(escrowID, config, group, escrow.Amount, creatorAddr, m.verifier,
 		state.WithWarmKeyResolver(m.bridge.VerifyWarmKey),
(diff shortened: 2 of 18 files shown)


akup:
I've retargeted your pr to
[gonka-ai:devshard-0.2.13-v2](https://github.com/gonka-ai/gonka/tree/devshard-0.2.13-v2) where https://github.com/gonka-ai/gonka/pull/1162 should go

@unameisfine please take a look at https://github.com/gonka-ai/gonka/pull/1162
there was added SealGraceNonces parameter that we can also add to DevshardEscrowParams

Also `max_nonces` parameter was added on `inference-chain` level but is not affecting devshard. We should use this parameter to limit nonces on devshard.
Take a look at the comment https://github.com/gonka-ai/gonka/pull/1143#discussion_r3205419993

I think both this points should extend your PR, please add them

unameisfine:
@akup @gmorgachev — done, force-pushed the rebase:

**Rebase**: branch reset onto `upstream/devshard-0.2.13-v2` and the 5 governance fields renumbered to tags 10-14 (tag 9 reserved for `default_seal_grace_nonces` from #1162 so we don't collide).

**`MaxNonce` cap on devshardd** (per #1143#discussion_r3205419993):
- threaded `MaxNonce` through `DevshardEscrow` (tag 10), `bridge.EscrowInfo`, and `EscrowSessionFields` into `SessionConfig`
- both REST and gRPC bridges decode it; pre-v0.2.13 escrows decode to 0 → treated as no-cap for forward-compat
- `state.machine` (both `applyCore` and `ApplyLocalBestEffort`) reject diffs with `nonce > MaxNonce` returning `ErrInvalidNonce`
- two regression tests added: cap rejects `nonce > MaxNonce`; cap with `MaxNonce == 0` is disabled

**`SealGraceNonces` from #1162**: not added in this PR — #1162 already plumbs it end-to-end (params → bridge → SessionConfig with chain-side fallback). Reserving tag 9 in `DevshardEscrowParams` here so the two PRs don't collide on the wire. Happy to ship them in either order.

**API**: `SessionConfigFromEscrow` switched to an `EscrowSessionFields` struct so subsequent governance fields don't churn call-site signatures.

All tests pass: x/inference/{types,keeper}, app/upgrades/v0_2_{12,13}, all 17 devshard packages, decentralized-api/internal/devshard.

a-kuprin:
As Gleb said https://github.com/gonka-ai/gonka/pull/1094#issuecomment-4282830317 we need some "bigger strategy for config management rather then local fixes"
I'm refactoring parameters flow now, to make dapi main provider, as it already listens for the chain, and we don't need to pollute with params requests.

Will be back to your pr tomorrow

a-kuprin:
The parameters flow for devshard was changed at
https://github.com/gonka-ai/gonka/pull/1234
session-frozen params and life-params was introduced at
https://github.com/gonka-ai/gonka/pull/1258

Those larger changes already solve the problem intended to be solved by this PR

PR #1174, opened 2026-05-16
Implement request filters for chat messages and parameters

qdanik:
This pull request primarily adds new tests to improve coverage and validation of the `normalizeChatRequest` and related request filtering logic. The changes ensure that the request normalization correctly handles various edge cases, strips unsupported or invalid fields, and preserves important values such as large integers. Additionally, some unused code related to content normalization has been removed.

**Testing improvements:**

* Added tests to ensure that unsupported fields (like `json_schema` response formats, `guided_json`, `guided_choice`, and required tool choices) are correctly rejected by the request filter (`TestNormalizeChatRequestRejectsUnsupportedFields`).
* Added tests to verify that unsupported VLLM parameters are stripped from requests (`TestNormalizeChatRequestStripsUnsupportedVLLMParameters`), and that empty or whitespace-only bad words are removed from the `bad_words` field (`TestNormalizeChatRequestStripsEmptyBadWords`).
* Added tests to check that non-finite sampling values (like "nan", "inf", "-inf", "infinity") are removed from requests (`TestNormalizeChatRequestStripsNonFiniteSamplingValues`).
* Added tests to ensure that out-of-range values in the `logit_bias` field are stripped, and the field is omitted if all entries are invalid (`TestNormalizeChatRequestStripsOutOfRangeLogitBias`).
* Added a test to verify that large integer fields (such as a `seed` value greater than JavaScript's safe integer limit) are preserved accurately in the request body (`TestPrepareChatRequestBodyPreservesLargeIntegerFields`).

**Code cleanup:**

* Removed the unused `normalizeContent`, `normalizeEmptyContent`, and `rawMessageString` helper functions from `proxy.go`, as they are no longer needed.

**Other minor fixes:**

* Fixed a missing `func` keyword in a test definition (`TestNormalizeChatRequestStripsMinTokensWhenStopTokenIdsPresent`).

These changes collectively improve the reliability and correctness of request normalization and filtering, and ensure the codebase is easier to maintain and extend.


baychak (commented):
Strong architecture — the handler-based filter factory is extensible and well-tested, and stage-aware execution (`PreValidation` / `PostLimits`) is exactly right. The catalog covers all 9 crash / corruption vectors I've confirmed empirically on the Kimi-K2.6 + vLLM 0.20.0 stand (130+ probes on mlnode-001 — engine survived all of them with no GPU memory leak; every bug below is reproducible directly against raw vLLM).

**Two Important findings** flagged inline that I'd fix before merge:
- `allowed_token_ids` and `ignore_eos` left as passthrough — empirical evidence shows these cause silent semantic corruption / contract violations on 0.20.0.
- `requiredTextContentPart` hard-coded `len(part) != 2` — rejects valid OpenAI content parts that carry `cache_control` / `metadata`.

**One Nice-to-have** inline (`min_tokens` triple-registration pattern).

A few smaller items I held off inline-commenting:
- `validateOpenAICompatChatMessages` rebuilds the processor per call — cache as a package-level var.
- `normalizeContent` silently swallows JSON parse errors — could propagate.
- `ConditionalStripParameterHandler` is silently inert when `Predicate == nil` — would prefer a panic / error at construction.
- `SanitizeFloatParameterHandler` skips `Max` enforcement for string-encoded values (`"temperature": "5.0"` bypasses the 2.0 clamp).
- `ApplyModelOverrides` re-runs the full `PreValidation` pass — intentional but a few wasted iterations.
- The 54-entry catalog list could be grouped by `Support` type or have section comments separating vLLM extensions from OpenAI-standard fields.

**Test coverage gaps** worth a follow-up:
- `Force` semantics — does it always override the client value? (e.g. client sends `logprobs: false, top_logprobs: 1`).
- `Conditional` predicate true / false branches (current tests only cover one side for `min_tokens` + `stop_token_ids`).
- `tool_choice = "auto"` passthrough — only the `"required"` rejection branch is tested.
- `response_format.type = "text"` allowed branch.
- `min_tokens` clamp when only `max_completion_tokens` is set.

**Follow-up PR I have ready locally** (`mb/strict-whitelist-extras` on the `kaitakuai/gonka` fork) — to be opened after this merges:

- **Strict-whitelist mode** (`Strict bool` on the catalog; unknown top-level fields → 400). The catalog becomes the source of truth for the accepted API surface. Currently the framework iterates only over registered parameters, so any field not in the catalog is silently forwarded to vLLM — a gap given the decentralized adversary model.
- **`stream_options` sub-field sanitizer** — strip `continuous_usage_stats` (vLLM #9028 chunked-prefill usage-counter corruption) while preserving `include_usage`.
- **16 defense-in-depth strips** for fields with no Gonka use case: `prompt_embeds` (CVE-2025-62164 defense), `add_generation_prompt`, `continue_final_message`, `function_call`, `functions`, `echo`, `best_of`, `length_penalty`, `vllm_xargs`, `return_token_ids`, `cache_salt`, `priority`, `guided_decoding_backend`, `mm_processor_kwargs`, `audio`, `modalities`, `service_tier`, `kv_cache_dtype`.
- **32 new unit tests** covering strict mode, every new strip, and the `stream_options` whitelist; all existing tests still pass.

Happy to roll the two Important fixes into your branch if that gets you to merge faster — let me know.

devshard/cmd/devshardctl/request_filters_parameters.go
```diff
@@ -0,0 +1,543 @@
...
+		newPassthroughParameter("temperature", ParameterCategoryFloatRange),
+		newPassthroughParameter("top_p", ParameterCategoryFloatRange),
+		newPassthroughParameter("top_k", ParameterCategoryIntRange),
+		newPassthroughParameter("min_p", ParameterCategoryFloatRange),
+		newPassthroughParameter("repetition_penalty", ParameterCategoryFloatRange),
+		newPassthroughParameter("logit_bias", ParameterCategoryIntToFloat),
+		newPassthroughParameter("bad_words", ParameterCategoryStringList),
+		newPassthroughParameter("stop", ParameterCategoryStringList),
+		newPassthroughParameter("stop_token_ids", ParameterCategoryIntList),
+		newPassthroughParameter("allowed_token_ids", ParameterCategoryIntList),
```
baychak: **Important** — `allowed_token_ids` empirically corrupts output on Kimi-K2.6 + vLLM 0.20.0 even with valid inputs. Should be stripped, not passed through.

Confirmed on stand (raw vLLM, no gateway, single request):

| Trigger | Result |
|---|---|
| `allowed_token_ids: [NaN]` | **200 OK** — NaN slips past (vLLM does not NaN-guard array elements; mirrors the Bug F class we covered for `temperature` / `logit_bias` / etc.) |
| `allowed_token_ids: [1, 2, 3]` | **200 OK** with garbage output `'"#"#"'` — the small whitelist forces the model into a 3-token vocab → deterministic semantic corruption |
| `allowed_token_ids: [1, 1, 1, 1]` | **200 OK** → `'"""""'` |

vLLM Pydantic does catch type errors (`[-1]`, OOV id, empty array, non-int) with clean 400s — but `[NaN]` and small-valid-whitelist both reach the sampler.

Suggested fix:

```go
// Replace this passthrough with a strip; the field has no Gonka use case
// and is actively a corruption vector even with valid token ids.
newStrippedParameter("allowed_token_ids", ParameterCategoryIntList, RequestFilterStagePreValidation),
```

(Same pattern as `use_beam_search` / `truncate_prompt_tokens` strips below.)

devshard/cmd/devshardctl/request_filters_parameters.go
```diff
@@ -0,0 +1,543 @@
...
+		newPassthroughParameter("seed", ParameterCategoryIntRange),
+		newPassthroughParameter("ignore_eos", ParameterCategoryBool),
```
baychak: **Important** — `ignore_eos: true` is a known silent semantic violation when combined with `stop_token_ids` containing EOS: the model still stops at EOS even though the client asked it not to.

Upstream: [vllm-project/vllm#4589](https://github.com/vllm-project/vllm/issues/4589) — open since 0.4.1, no fix planned. Reproduces on 0.20.0.

For the validator use case where determinism and contract-honoring matter, this should be stripped:

```go
newStrippedParameter("ignore_eos", ParameterCategoryBool, RequestFilterStagePreValidation),
```

No Gonka client should need to set this directly — the only legitimate uses are evaluation harnesses, and those should opt out of going through the gateway.

devshard/cmd/devshardctl/request_filters_parameters.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+func defaultVLLMParameterCatalog() VLLMParameterCatalog {
+	return VLLMParameterCatalog{parameters: []VLLMParameter{
+		newPassthroughParameter("model", ParameterCategoryString),
+		newPassthroughParameter("stream", ParameterCategoryBool),
+		newPassthroughParameter("messages", ParameterCategoryObjectArray),
+		newPassthroughParameter("max_tokens", ParameterCategoryIntRange),
+		newPassthroughParameter("max_completion_tokens", ParameterCategoryIntRange),
+		newPassthroughParameter("min_tokens", ParameterCategoryIntRange),
```
baychak: **Nice** — `min_tokens` is registered three times in this catalog: passthrough here (385), conditional-strip at 412, sanitize at 434. Similar pattern for `bad_words` (393 + 417) and `tools` (403 + 423).

Because `Apply` iterates the entire slice and dispatches by stage, the passthrough no-op handlers are redundant — removing them would not change behavior, since the same name appears again with a real handler. They make the catalog ~longer to read for new contributors.

Two options:

1. **Dedupe** — drop the passthrough entry when a non-passthrough handler exists for the same name (cleanest).
2. **Comment** — keep them as "field exists for whitelist enrollment, real handler below" with a `// see line N` cross-reference.

Either is fine; flagging because today it's not obvious from reading the catalog top-to-bottom that the first `min_tokens` does anything beyond marking the field as known.

devshard/cmd/devshardctl/request_filters_messages.go
```diff
@@ -0,0 +1,416 @@
...
+			}
+		}
+		return nil
+	default:
+		return fmt.Errorf("must be a string or an array of typed content parts")
+	}
+}
+
+func requiredTextContentPart(part map[string]any, partIndex int) (string, error) {
+	if len(part) != 2 {
```
baychak: **Important** — this hard-coded length check rejects valid OpenAI content parts that carry additional fields.

OpenAI Chat Completions spec allows extra fields in content parts (e.g. `cache_control: {"type": "ephemeral"}` for prompt caching, plus future extensions). A request like:

```json
{"type": "text", "text": "hello", "cache_control": {"type": "ephemeral"}}
```

would be rejected by the gateway with 400 "must only include type and text" — even though it's a perfectly valid OpenAI request that vLLM would happily handle (or ignore the cache_control silently, which is fine).

The `partType != "text"` check below already protects against `image_url` / `input_text` / `input_audio` / other unsupported content types, which is the actual security goal. Suggestion: drop the length check entirely:

```go
func requiredTextContentPart(part map[string]any, partIndex int) (string, error) {
    partType, err := requiredNonEmptyString(part, "type")
    if err != nil {
        return "", fmt.Errorf("[%d].type: %w", partIndex, err)
    }
    if partType != "text" {
        return "", fmt.Errorf("[%d].type has unsupported value %q", partIndex, partType)
    }
    text, err := requiredNonEmptyString(part, "text")
    if err != nil {
        return "", fmt.Errorf("[%d].text: %w", partIndex, err)
    }
    return text, nil
}
```

baychak (commented):
Thanks for the fast turn-around — all four items from the first round are resolved, and the `Rules []ParameterRule` + `withRule` builder is a really nice answer to the multi-registration concern. The catalog reads top-to-bottom now and each field's full treatment is visible in one chain.

**Other resolved items:**
- `allowed_token_ids` and `ignore_eos` → `Strip` ✅
- `requiredTextContentPart` no longer rejects extra content-part fields ✅ (great that you also added `TestPrepareChatRequestBodyAllowsExtraFieldsOnTextContentParts` with the exact `cache_control` example)
- Doc comments on `RequestFilterStage*` / `ParameterSupport*` close the labeling gap I flagged in S-1
- Empirical bad_words matrix from our stand probing landed verbatim in `TestNormalizeChatRequestStripsWhitespaceOnlyBadWordsResearchCases` + `TestNormalizeChatRequestKeepsSafeBadWordsResearchCases` — that's exactly the safety net I wanted
- New `sanitizeRepetitionPenalty` + `MaxRepetitionPenalty = 2.0` closes a hole I didn't even raise (string-encoded values bypass the float clamp — same issue as N-6 in the previous review)

**Two new findings I noticed in v2** — flagged inline:
- `unsupportedChatParameterMessage` now produces double-quoted error messages (`feature '"json_object"' is temporarily unavailable`) — likely an oversight from the format-string change
- `sanitizeRepetitionPenalty` treats string-encoded values strictly (drop) while `SanitizeFloatParameterHandler` treats them leniently (drop only if non-finite). The asymmetry is surprising; either apply the strict treatment uniformly or document why `repetition_penalty` is special.

**Smaller items I'm not blocking on but worth tracking:**
- `withRule(stage, support, handler)` doesn't enforce `support` matches the handler type — `withRule(stage, ParameterSupportForce, StripParameterHandler{})` would compile cleanly. Type-safe variants like `withStripRule(stage)`, `withSanitizeRule(stage, handler)` would prevent that class of mismatch. Low priority.
- Unrelated dependency bumps in `decentralized-api/go.mod` + `devshard/go.mod` (prometheus_client, prometheus_common, go.yaml.in/yaml). Usually cleaner to split into a separate "chore: bump deps" PR, but not blocking.
- `validateOpenAICompatChatMessages` still rebuilds the processor per call (N-2 from previous review).
- `normalizeContent` still silently swallows JSON parse errors (N-3).
- `ConditionalStripParameterHandler` still silently inert when `Predicate == nil` (N-4).
- `ApplyModelOverrides` still re-runs the full PreValidation pass on the already-normalized body (N-7).

**Test gaps still open:**
- `Force` semantics — no test verifies that `top_logprobs: 1` from a client gets overridden to `5` (or that `logprobs: false` gets forced to `true`).
- Conditional predicate true / false branches for `min_tokens` + `stop_token_ids`.
- `tool_choice = "auto"` accepted path (only the `"required"` rejection branch is tested).
- `response_format.type = "text"` accepted branch.

**Our follow-up PR** (`mb/strict-whitelist-extras` on the `kaitakuai/gonka` fork) — still queued for after this merges:

- Strict-whitelist mode (`Strict bool` on the catalog → unknown top-level fields return 400 instead of being forwarded to vLLM as-is). The catalog becomes the source of truth for the accepted API surface.
- `stream_options` sub-field sanitizer (strip `continuous_usage_stats`, keep `include_usage` — vLLM #9028 counter corruption).
- 16 defense-in-depth strips for fields with no Gonka use case: `prompt_embeds` (CVE-2025-62164), `add_generation_prompt`, `continue_final_message`, `function_call`, `functions`, `echo`, `best_of`, `length_penalty`, `vllm_xargs`, `return_token_ids`, `cache_salt`, `priority`, `guided_decoding_backend`, `mm_processor_kwargs`, `audio`, `modalities`, `service_tier`, `kv_cache_dtype`.
- 3 OpenAI-standard passthroughs for strict-mode compatibility: `user`, `metadata`, `parallel_tool_calls`.
- 32 new unit tests covering strict mode, every new strip, and the `stream_options` whitelist.

With the two inline nits fixed, this looks good to me 👍

devshard/cmd/devshardctl/request_filters.go
```diff
@@ -239,39 +256,8 @@ func capOutputTokens(value uint64, explicitlySet bool, bypassLimit bool, limits
...
-	if value > MaxTemperature {
-		request["temperature"] = MaxTemperature
-	}
-}
-
-func forceValidationLogprobs(request map[string]any) {
-	request["logprobs"] = true
-	request["top_logprobs"] = 5
+func unsupportedChatParameterMessage(name string) string {
+	return fmt.Sprintf("feature '%q' is temporarily unavailable", name)
```
baychak: **Important** — minor regression introduced by the format change. `%q` already wraps its argument in double quotes, and the surrounding single quotes here produce double-quoted output.

Verified locally:

```go
fmt.Sprintf("parameter %q is unsupported", "json_object")         // → parameter "json_object" is unsupported   (old, clean)
fmt.Sprintf("feature '%q' is temporarily unavailable", "json_object") // → feature '"json_object"' is temporarily unavailable   (new, ugly)
```

Clients now see `feature '"json_object"' is temporarily unavailable` — tests still pass because they `Contains` the bare name, but the user-facing string is regressed. Drop one of the quotes:

```go
return fmt.Sprintf("feature %q is temporarily unavailable", name)
// or
return fmt.Sprintf("feature '%s' is temporarily unavailable", name)
```

devshard/cmd/devshardctl/request_filters_parameters.go
```diff
@@ -0,0 +1,567 @@
...
+		return nil
+	}
+	if _, exists := structuredOutputs["regex"]; exists {
+		return badChatRequest("%s", unsupportedChatParameterMessage("structured_outputs.regex"))
+	}
+	return nil
+}
+
+// repetition_penalty is kept numeric-only because stringified values were observed to produce garbage upstream behavior.
+func sanitizeRepetitionPenalty(ctx *RequestFilterContext, parameter VLLMParameter) error {
```
baychak: **Nice** — this is meaningfully stricter than the `SanitizeFloatParameterHandler` used for `top_p` / `top_k` / `min_p` / `temperature`. Here string-encoded values are unconditionally dropped; in `SanitizeFloatParameterHandler` they're dropped only if they parse as non-finite (`"nan"`, `"inf"`, etc.) — anything else like `"5.0"` slips through and even bypasses the float `Max` clamp.

The comment hints at the rationale ("stringified values were observed to produce garbage upstream behavior"), but the same logic should plausibly apply to the other numeric knobs we sanitize — a client sending `"temperature": "5.0"` is equally suspicious.

Two options:

1. Apply the strict-string treatment uniformly: extend `SanitizeFloatParameterHandler` so any string value (not just non-finite ones) is dropped. Then `sanitizeRepetitionPenalty` would just be `SanitizeFloatParameterHandler{StripNonFinite: true, StripAnyString: true, Max: floatPointer(MaxRepetitionPenalty)}` and we'd retire the custom handler.
2. Keep the custom handler but expand the comment with what specifically you saw on the stand for `repetition_penalty` (model emits, vLLM serializer hiccup, etc.) so future readers know why this field is special.

Low priority — works correctly today, just flagging the inconsistency.

baychak:
v3 LGTM 🎉 Both v2 nits resolved cleanly.

Our follow-up PR (`mb/strict-whitelist-extras` on the `kaitakuai` fork) — strict whitelist mode + `stream_options` sub-field sanitizer + 16 defense-in-depth strips — is committed locally and queued to open after this merges.

Ship it 🚢

libermans (approved):

PR #1177, opened 2026-05-16
feat(devshard): add OpenAI observability fields (user/metadata/parallel_tool_calls/stream_options)

baychak:
## Summary

After #1174's strict-whitelist gate and #1180's `paramvalidators` framework, the catalog rejects four OpenAI Chat Completions standard fields that every OpenAI-built client sends: `user`, `metadata`, `parallel_tool_calls`, `stream_options`. This PR moves them into the supported set as bounded passthroughs / sanitizers, using the same `paramvalidators` framework #1180 introduced.

These four fields are **observability/tracking**, not security-relevant — no inference-side semantics on the vLLM upstream, no DoS vector, no parser invoked. Treating them with the same strict-reject as `response_format.json_schema.schema` was over-application of the principle: it broke every legitimate OpenAI-compatible client (official SDK with `user=...`, LangChain `metadata={...}`, agents with `parallel_tool_calls`, any streaming client asking for final-chunk token usage via `stream_options.include_usage`) without preventing any attack.

### Two new validators

- **`paramvalidators.StreamOptionsValidator`** — sub-field whitelist. Only `include_usage` survives. `continuous_usage_stats` is stripped (vLLM-project/vllm#9028 — per-chunk usage counter is wrong under chunked prefill). Any other / future sub-field is also stripped. If the object empties out, the field is dropped so the upstream never receives `{}`. The validator mutates the live document via `Raw()` — the existing in-place sanitize pattern.
- **`paramvalidators.MetadataValidator`** — OpenAI-documented bounds: ≤16 keys × 64-char keys × 512-char string values. Non-string values rejected. Sentinel errors (`ErrMetadataShape`, `ErrMetadataKeyCount`, `ErrMetadataKey`, `ErrMetadataValue`) match #1180's error-style.

### Four catalog entries

Registered alongside the other passthrough fields, before the `logprobs` force rules:

```go
newParameter("user"),
newParameter("parallel_tool_calls"),
newParameter("metadata").
    withRule(RequestFilterStagePreValidation, DocumentValidatorHandler{
        Validator: paramvalidators.MetadataValidator{},
    }),
newParameter("stream_options").
    withRule(RequestFilterStagePreValidation, DocumentValidatorHandler{
        Validator: paramvalidators.StreamOptionsValidator{},
    }),
```

### Spec doc update

`devshard/docs/supported-params.md` moves the four fields from the "Unsupported parameters" table to the "Supported parameters" table, with per-field rationale documenting why these are OpenAI-standard observability fields and not provider-specific. The "🚫 Won't add" note is updated. `tool_choice="required"` rejection rationale stays separate from `parallel_tool_calls` (different fields, different reasons).

## Tests

- `paramvalidators/stream_options_test.go` (8 cases): wrapper shape accepts/rejects, strips `continuous_usage_stats`, strips unknown sub-field, drops empty object, drops empty input.
- `paramvalidators/metadata_test.go` (15+ cases): wrapper shape, key count, key length, value type, value length boundaries; custom-limits respect.
- `request_filters_test.go` (4 cases): pipeline-level end-to-end coverage that all four fields survive the catalog gate, sanitizer mutates the forwarded body correctly, metadata bound enforcement reaches the pipeline error path.

## Why this categorization is right

OpenAI's own spec, Kilo Gateway, OpenRouter, Together AI, Azure OpenAI, and all three reference agent clients (OpenClaw, Hermes, Kilo) accept these four fields. Gonka rejecting them would be the outlier:

| Gateway | `user` | `metadata` | `parallel_tool_calls` | `stream_options` |
|---|---|---|---|---|
| OpenAI direct | accept | accept | accept | accept |
| Azure OpenAI | accept | accept | accept | accept |
| OpenRouter | accept | accept | accept | accept |
| Together AI | accept | accept | accept | accept |
| Kilo Gateway | accept | — | — | auto-injects `{include_usage: true}` |
| Gonka (this PR) | passthrough | bounded passthrough | passthrough | sanitized passthrough |

Security stays intact — #1180's bounds on `response_format`, `tools[].function.parameters`, `chat_template_kwargs`, body-level depth pre-scan, length caps, `pattern` regex compile-check, CVE coverage — all unchanged.

## Risk

Low. Two new validators (~130 LOC) following the existing `paramvalidators` pattern, four catalog entries, ~70 LOC of pipeline tests. No existing handler modified. `metadata` enforces strictly tighter bounds than the body-level cap, so it cannot make a previously-rejected request succeed.

## Test plan

- [x] `go test ./cmd/devshardctl/paramvalidators/` — green (new validators).
- [x] `go test ./cmd/devshardctl/...` — full suite green (~294s).
- [ ] Post-deploy probe against the real devshardctl binary on the Kimi-K2.6 stand: confirm `stream_options.include_usage` reaches the upstream and elicits a final-chunk `usage` object; confirm `user` / `metadata` / `parallel_tool_calls` are forwarded; confirm `continuous_usage_stats` is stripped before vLLM sees it.

### DIFF
--- a/devshard/cmd/devshardctl/request_filters_parameters.go
+++ b/devshard/cmd/devshardctl/request_filters_parameters.go
@@ -593,6 +593,32 @@ func defaultVLLMParameterCatalog() VLLMParameterCatalog {
 					MaxPatternLen: 512,
 				},
 			}),
+		// OpenAI Chat Completions standard observability fields. No inference-side
+		// semantics on the vLLM upstream; clients send them for end-user tracking,
+		// distributed tracing, agent control, and streaming token accounting.
+		// `user`: type-checked and byte-capped at the gateway boundary so a non-string
+		// payload (number, object, …) and an over-long string are caught early instead
+		// of being forwarded as a no-op carrier under the 10 MiB body cap.
+		newParameter("user").
+			withRule(RequestFilterStagePreValidation, DocumentValidatorHandler{
+				Validator: paramvalidators.UserValidator{},
+			}),
+		newParameter("parallel_tool_calls"),
+		// metadata: OpenAI bounds it to 16 keys × 64-char keys × 512-char string values;
+		// we enforce the same bounds at the gateway boundary as a free defensive cap.
+		newParameter("metadata").
+			withRule(RequestFilterStagePreValidation, DocumentValidatorHandler{
+				Validator: paramvalidators.MetadataValidator{},
+			}),
+		// stream_options: sub-field whitelist. Only `include_usage` survives;
+		// `continuous_usage_stats` is stripped to neutralize vLLM-project/vllm#9028
+		// (per-chunk usage counter is wrong under chunked prefill), and any other /
+		// future sub-field is default-stripped. If nothing remains the field is dropped
+		// so the upstream does not receive an empty `{}` object.
+		newParameter("stream_options").
+			withRule(RequestFilterStagePreValidation, DocumentValidatorHandler{
+				Validator: paramvalidators.StreamOptionsValidator{},
+			}),
 		newParameter("logprobs").
 			withRule(RequestFilterStagePostLimits, ForceLiteralParameterHandler{Value: true}),
 		newParameter("top_logprobs").
--- a/devshard/cmd/devshardctl/paramvalidators/metadata.go
+++ b/devshard/cmd/devshardctl/paramvalidators/metadata.go
@@ -1,0 +1,77 @@
+package paramvalidators
+
+import (
+	"errors"
+	"fmt"
+)
+
+// MetadataValidator enforces the OpenAI Chat Completions `metadata` contract: an optional
+// object with at most 16 string-keyed entries, keys up to 64 characters, values that are
+// strings up to 512 characters. The field has no inference-side meaning for vLLM (it is
+// ignored upstream) — clients use it for distributed tracing (LangSmith, OpenAI tracing,
+// W3C trace-context propagation) and A/B test tagging. Validating at the gateway boundary
+// keeps the same bounds the OpenAI API itself enforces, so the field is bounded surface
+// rather than an unbounded passthrough.
+//
+// Zero-valued limit fields fall back to the OpenAI-documented defaults.
+type MetadataValidator struct {
+	MaxKeys     int
+	MaxKeyLen   int
+	MaxValueLen int
+}
+
+// ErrMetadataShape covers the wrapper-level rejection: metadata must be a JSON object.
+// ErrMetadataKeyCount, ErrMetadataKey, ErrMetadataValue mark the OpenAI-spec bound
+// violations.
+var (
+	ErrMetadataShape    = errors.New("metadata: invalid wrapper shape")
+	ErrMetadataKeyCount = errors.New("metadata: key count exceeded")
+	ErrMetadataKey      = errors.New("metadata: key invalid")
+	ErrMetadataValue    = errors.New("metadata: value invalid")
+)
+
+// OpenAI's documented bounds — see https://platform.openai.com/docs/api-reference/chat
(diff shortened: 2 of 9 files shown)


devshard/cmd/devshardctl/request_filters_parameters.go
```diff
@@ -461,10 +475,69 @@ func defaultVLLMParameterCatalog() VLLMParameterCatalog {
...
+		// kept as passthrough so legitimate clients are not broken by strict mode.
+		newParameter("user", ParameterCategoryString),
+		newParameter("metadata", ParameterCategoryObject),
+		newParameter("parallel_tool_calls", ParameterCategoryBool),
+		// stream_options: whitelist sub-fields. include_usage is preserved; continuous_usage_stats
+		// is stripped to avoid the chunked-prefill usage-counter bug (vLLM #9028).
+		newParameter("stream_options", ParameterCategoryObject).
+			withRule(RequestFilterStagePreValidation, ParameterSupportSanitize, CustomParameterHandler{ApplyFunc: sanitizeStreamOptions}),
+		// CVE-2025-62164 defense in depth (fixed in vLLM 0.11.1).
+		newParameter("prompt_embeds", ParameterCategoryAny).
```
qdanik: all of added strip parameters are not necessary right now. They are will throw an error to the client. VLLMParameterCatalog already works like whitelist. We should an implement validators for the new parameter with edge case coverage

baychak:
## Empirical validation against vLLM 0.20.0 (Kimi-K2.6) — refreshed for the trimmed scope

After #1174 merged its strict whitelist + reject contract directly, this PR was rebased and reduced to just the additions that did not overlap with the merged base: `stream_options` sanitizer + three OpenAI passthrough entries. Routed the new behaviour through a thin filter+forward proxy against a real vLLM upstream.

**`stream_options` sanitizer (vLLM-project/vllm#9028)**

Identical payload `{"stream_options":{"include_usage":true,"continuous_usage_stats":true}}`, streaming:

- raw vLLM: **12** `usage` objects in the SSE stream (one per chunk under chunked prefill — the buggy counter behaviour)
- through the pipeline: **1** `usage` object (final chunk only — matches the OpenAI contract)

A payload with an unknown sub-field `{"include_usage":true,"continuous_usage_stats":true,"future_field":99}` is captured downstream as `{"stream_options":{"include_usage":true}}` — both stripped, only the whitelisted sub-field survives.

**Three OpenAI passthroughs**

`{"user":"u-42","metadata":{"trace_id":"abc"},"parallel_tool_calls":true}` is forwarded verbatim — confirmed by tee-ing the upstream request through a capturing stub. Without these entries the unknown-key gate would 400 legitimate OpenAI clients.

**No regressions for #1174's behaviour**

- `{"unknown_field_xyz":42}` → 400 `feature "unknown_field_xyz" is temporarily unavailable` (#1174's gate)
- `{"response_format":{"type":"json_schema","json_schema":{…}}}` (the schema-bomb shape seen against production node4 — 573 cloudlogging records) → 400 `feature "response_format" is temporarily unavailable`. Bonus: after #1174 dropped its explicit `response_format` reject in favour of the catalog-level unknown-key gate, the schema-bomb is now caught by the gate rather than by a specific rule — same outcome, slightly stricter.

---

_Replaces my earlier validation comment, which described the original scope (strict mode + 18 defense-in-depth strips) before this PR was rebased onto the merged #1174._

baychak:
@qdanik — could you take a look when you have a moment? This PR builds on top of your #1174 (now merged), adding four OpenAI Chat Completions fields that the catalog's unknown-key gate currently rejects: `user`, `metadata`, `parallel_tool_calls`, and `stream_options` (the last one with a sub-field sanitizer for the vLLM-project/vllm#9028 issue).

qdanik (approved):

libermans (approved):

PR #1179, opened 2026-05-17
Preserved Guardians

gmorgachev:


x0152:
Shouldn't participantVotingPowers be stored per model to stay symmetric with PoC validation logic?

inference-chain/x/inference/module/model_assignment.go
```diff
@@ -477,17 +480,42 @@ func (ma *ModelAssigner) SamplePreservedForEpisode(
...
 	slices.Sort(sortedModelIds)
 
 	currentEpochData := NewEpochMLNodeData()
 	previousEpochData := NewEpochMLNodeData()
-	totalCurrentEpochWeight := int64(0)
+	totalNetworkWeight := int64(0)
+	for _, vw := range rootData.ValidationWeights {
+		if vw != nil {
+			totalNetworkWeight += vw.Weight
+		}
```
x0152: should it use only live members to match preserved selection?

gmorgachev:
> Shouldn't participantVotingPowers be stored per model to stay symmetric with PoC validation logic?

Fixed this one

inference-chain/x/inference/module/model_assignment.go
gmorgachev: agree. in root group they removed as members but not from `ValidationWeights`

sam0x17 (approved):
Fixes the preserved-nodes sampler so its non-voting cap and 75% threshold use the same per-model delegation-resolved VotingPower / root-group network Weight that PoC validation already uses, instead of coefficient-adjusted weight sums. Also excludes genesis guardians from the preserved set (so they always remain voting) and filters subgroup members against the live SDK-group set, matching the symmetric filtering in `getEffectiveValidationBaseState`. Targeted at the same `dl/poc-voting-265-epoch-issue` family of bugs from the epoch-265/266 weight drop; the unit-alignment fix is correct and the new code paths are coherent.

**Addressed in follow-up commit (please verify):**
- Per-model redundant `GetCurrentEpochGroup` call hoisted out of the model loop. The original per-model helper has been replaced with a single bulk helper that resolves the current epoch group once and returns subgroup data + live-member sets keyed by modelId — see https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/keeper/power.go#L67-L104. `SamplePreservedForEpisode` now does one keeper round-trip across all models instead of N, and the error path is logged once.
- Test coverage added for the two new code paths:
  - `TestSamplePreservedForEpisode_ExcludesGuardians` — sets a valid `gonkavaloper` guardian address whose converted acc-bech32 matches a participant `Index`, runs the sampler, asserts that participant is absent from the snapshot while non-guardian participants survive.
  - `TestSamplePreservedForEpisode_FiltersDeadSubGroupMembers` — populates `liveSubGroupOverrides` to exclude one member that is present in `ValidationWeights`, asserts that "dead" member is not preserved.

Both tests would fail if the corresponding guard were removed; they're real negative-coverage, not just exercise-the-line tests.

**Nits:**
- The PR title "Preserved Guardians" reads as "guardians get preserved" but the diff does the opposite — guardians are explicitly *excluded* from the preserved (non-voting) bucket at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L676-L678 so they always remain in the voting set. Consider renaming to "Always-voting guardians" or "Exclude guardians from preserved-nodes set"; the current title invites a misread by reviewers and by future archaeologists tracing the epoch-265/266 fix family.

**Looks correct:**
- Unit consistency for the non-voting cap: `totalNetworkWeight` is summed from root-group `ValidationWeight.Weight` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L500-L505, and `participantVP` is read from subgroup `ValidationWeight.VotingPower` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L534-L538. Both originate from the same `member.Weight` / `member.VotingPowers[modelId]` source at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/epochgroup/epoch_group.go#L182-L189, so comparing accumulated per-model VP against 34% of root weight is unit-consistent. The function comment at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L641-L644 accurately describes what changed and why — the prior `totalCappedWeight` (coefficient-adjusted sum across models) was the wrong yardstick for a cap that's supposed to mirror PoC validation quorum.
- `eligibleParticipantModelWeight` and `participantModelWeight` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L696-L709 are both raw `PocWeight` sums for the same participant/model — no coefficient applied to either side. The "did we consume all this participant's nodes?" check is correctly unit-consistent because both sides cancel out coefficients.
- `liveSubSet != nil` defensive guard at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L519 is necessary: on the error path, `currentSubData` is zero-value and the outer `len(currentSubData.ValidationWeights) > 0` check short-circuits before the inner loop runs, so the nil-set case is unreachable in practice — but the explicit nil-check makes that safe even if a future refactor changes the outer guard.
- Guardian-address handling at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L483-L496: guardians are stored on-chain as valoper bech32 strings (per https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/keeper/params.go#L112-L113) but the subgroup `MemberAddress` is the AccAddress, so the `OperatorAddressToAccAddress` conversion is the right normalization. A malformed guardian address logs a warning and skips that entry instead of poisoning the whole set — correct fallback.
- The new `getParticipantWeightsForThreshold` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L944-L961 and `calculatePerParticipantThreshold` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L972-L992 both use `sortedKeys` and only iterate `epochData.data`, which is single-model when called via `ForModel(modelId)` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L109-L119. No cross-model leak, consensus-deterministic iteration order.

0xgonka (approved):
Fixes the preserved-nodes sampler so its non-voting cap and 75% threshold use the same per-model delegation-resolved VotingPower / root-group network Weight that PoC validation already uses, instead of coefficient-adjusted weight sums. Also excludes genesis guardians from the preserved set (so they always remain voting) and filters subgroup members against the live SDK-group set, matching the symmetric filtering in `getEffectiveValidationBaseState`. Targeted at the same `dl/poc-voting-265-epoch-issue` family of bugs from the epoch-265/266 weight drop; the unit-alignment fix is correct and the new code paths are coherent.

**Addressed in follow-up commit (please verify):**
- Per-model redundant `GetCurrentEpochGroup` call hoisted out of the model loop. The original per-model helper has been replaced with a single bulk helper that resolves the current epoch group once and returns subgroup data + live-member sets keyed by modelId — see https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/keeper/power.go#L67-L104. `SamplePreservedForEpisode` now does one keeper round-trip across all models instead of N, and the error path is logged once.
- Test coverage added for the two new code paths:
  - `TestSamplePreservedForEpisode_ExcludesGuardians` — sets a valid `gonkavaloper` guardian address whose converted acc-bech32 matches a participant `Index`, runs the sampler, asserts that participant is absent from the snapshot while non-guardian participants survive.
  - `TestSamplePreservedForEpisode_FiltersDeadSubGroupMembers` — populates `liveSubGroupOverrides` to exclude one member that is present in `ValidationWeights`, asserts that "dead" member is not preserved.

Both tests would fail if the corresponding guard were removed; they're real negative-coverage, not just exercise-the-line tests.

**Nits:**
- The PR title "Preserved Guardians" reads as "guardians get preserved" but the diff does the opposite — guardians are explicitly *excluded* from the preserved (non-voting) bucket at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L676-L678 so they always remain in the voting set. Consider renaming to "Always-voting guardians" or "Exclude guardians from preserved-nodes set"; the current title invites a misread by reviewers and by future archaeologists tracing the epoch-265/266 fix family.

**Looks correct:**
- Unit consistency for the non-voting cap: `totalNetworkWeight` is summed from root-group `ValidationWeight.Weight` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L500-L505, and `participantVP` is read from subgroup `ValidationWeight.VotingPower` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L534-L538. Both originate from the same `member.Weight` / `member.VotingPowers[modelId]` source at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/epochgroup/epoch_group.go#L182-L189, so comparing accumulated per-model VP against 34% of root weight is unit-consistent. The function comment at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L641-L644 accurately describes what changed and why — the prior `totalCappedWeight` (coefficient-adjusted sum across models) was the wrong yardstick for a cap that's supposed to mirror PoC validation quorum.
- `eligibleParticipantModelWeight` and `participantModelWeight` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L696-L709 are both raw `PocWeight` sums for the same participant/model — no coefficient applied to either side. The "did we consume all this participant's nodes?" check is correctly unit-consistent because both sides cancel out coefficients.
- `liveSubSet != nil` defensive guard at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L519 is necessary: on the error path, `currentSubData` is zero-value and the outer `len(currentSubData.ValidationWeights) > 0` check short-circuits before the inner loop runs, so the nil-set case is unreachable in practice — but the explicit nil-check makes that safe even if a future refactor changes the outer guard.
- Guardian-address handling at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L483-L496: guardians are stored on-chain as valoper bech32 strings (per https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/keeper/params.go#L112-L113) but the subgroup `MemberAddress` is the AccAddress, so the `OperatorAddressToAccAddress` conversion is the right normalization. A malformed guardian address logs a warning and skips that entry instead of poisoning the whole set — correct fallback.
- The new `getParticipantWeightsForThreshold` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L944-L961 and `calculatePerParticipantThreshold` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L972-L992 both use `sortedKeys` and only iterate `epochData.data`, which is single-model when called via `ForModel(modelId)` at https://github.com/gonka-ai/gonka/blob/dl/poc-voting-265-epoch-issue/inference-chain/x/inference/module/model_assignment.go#L109-L119. No cross-model leak, consensus-deterministic iteration order.

d-bogdan-engenious:
@0xgonka Tested. Test cases (practical):

1. Guardian excluded from preserved snapshot — [PASS] (adding join-1 to guardian_addresses flipped preserved from {join-1:[node2]} → empty; true positive test)

2. Operator→acc bech32 conversion — [PASS] (6 distinct conversion paths observed)

3. Mid-epoch dead-member filter (liveSubSet) — [PASS] (via offline replay) 

4. asymmetric VP-vs-PoCWeight cap (delegate-driven VP shift) — [NOT ABLE TO RUN] - MsgDelegate disabled, sdk code 18 - "Delegate is disabled". That seems intentional, not a TC bug: on Gonka a validator's power comes from PoC weight via SetComputeValidators each epoch, not from delegated stake, so plain delegation is turned off. The restriction lives in the cosmos-sdk fork ([github.com/gonka-ai/cosmos-sdk](http://github.com/gonka-ai/cosmos-sdk) v0.53.3-ps17).  Happy path not runnable as written above.

5. 75% threshold uses live-root total — [PASS] (threshold tracks root-based target, which is different from model-local sum)

6. Multi-epoch consensus determinism — [PASS] (3 consecutive epochs, stable preserved snapshot)

Supporting checks:

1. Cap formula = liveRootTotalWeight × 0.34 (new), not coeff × PoC (old) — [PASS] (exact arithmetic match in all sampler logs)

2. root.Weight = Σ validator.tokens / subgroup VP = tokens — [PASS]

3. Bad-guardian-input sweep — [PASS] (malformed entries WARN-logged, no crash; bad config rejected at DeliverTx)

4. canAllocateParticipantNode cap-saturation reject — [NOT TESTED MANUALLY] (code present, path unreachable in the testnet-3 with 3-validator single-mlnode config). Triggering it requires a participant whose voting power exceeds the 34% cap while fully preserved. On the testnet that's not achievable: (a) the 3 validators carry near-equal VP (~33% each), so none naturally exceeds 34%; (b) attempts to synthesize a high-VP participant with mocks failed because fake PoC is rejected by cross-validation (validated_weight=-1 → invalid majority → zero weight), so a mock can never reach the VP needed. 

FINDINGS — none crucial, all discretionary for devs:

1. [LOW] network_maturity_threshold = 0 is silently coerced to 2,000,000 and omitted from the query. Cosmetic/UX confusion only, no consensus impact. Optional: reject 0 at validate.

2. [LOW] guardian_addresses aren't validated at submit (no bech32/dedup). A typo just surfaces as a WARN ~25 min later; doesn't break anything. Optional: add the AccAddressFromBech32 or any other address conversion/validation to Params.Validate() (e.g., same pattern already used for the allow-list). 

3. [INFO / by-design] single-mlnode participant is never preserved. Confirmed intentional — code comment "Option B: 0% eligible, 100% voting." Based on the code examination - not a bug, introduced intentionally by dev team. Optional: document the tradeoff so single-node operators aren't surprised.

PR #1197, opened 2026-05-19
fix: validate Ethereum destination on bridge withdrawal

ouicate:
## Summary
This fixes a validation gap where `MsgRequestBridgeWithdrawal` accepted malformed `DestinationAddress` values at `ValidateBasic` and only failed later during BLS signature preparation. Users paid gas for transactions that could never succeed, while `MsgRequestBridgeMint` already rejected the same malformed addresses up front.

Related deposit-side fixes: #1196.

## Root Cause
`MsgRequestBridgeWithdrawal.ValidateBasic` checked `DestinationAddress` for non-empty length only. `DestinationBridgeAddress` was validated with `isValidEthereumAddress`, but the user's ultimate Ethereum recipient was not. The handler called `ethereumAddressToBytes(recipient)` in `prepareBridgeWithdrawalSignatureData`, which failed after the message was already included in a block.

## Fix
- Apply `isValidEthereumAddress` to `DestinationAddress` in `MsgRequestBridgeWithdrawal.ValidateBasic`, matching `MsgRequestBridgeMint`.
- Add unit tests rejecting short/invalid destination addresses.

Malformed withdrawal destinations are rejected before broadcast, eliminating wasted gas and inconsistent UX between symmetric bridge mint and withdrawal flows. Execution-time failures from invalid recipient encoding on the withdrawal path are no longer reachable through `ValidateBasic`-accepted messages.

## Test plan
- [ ] `go test ./x/inference/types/... -run 'Withdrawal|BridgeWithdrawal'`
- [ ] Submit `MsgRequestBridgeWithdrawal` with `DestinationAddress = "0x12345"` and confirm `ValidateBasic` error before execution

### DIFF
--- a/inference-chain/x/inference/types/message_request_bridge_withdrawal.go
+++ b/inference-chain/x/inference/types/message_request_bridge_withdrawal.go
@@ -41,9 +41,12 @@ func (msg *MsgRequestBridgeWithdrawal) ValidateBasic() error {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "amount must be positive")
 	}
 
-	// Validate destination address is not empty (Ethereum address format not validated here)
+	// Validate destination address is not empty
 	if len(msg.DestinationAddress) == 0 {
 		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "destination address cannot be empty")
+	}
+	if !isValidEthereumAddress(msg.DestinationAddress) {
+		return errorsmod.Wrap(sdkerrors.ErrInvalidRequest, "destination address must be a valid Ethereum address")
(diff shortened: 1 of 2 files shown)


GLiberman:
Your fork is based around x.10 release while main is already on x.12 version. 
And it looks like PR description only covers one of commits in this PR

ouicate:
Thanks, fixed @GLiberman.

I rebased both branches onto current `gonka-ai:main` (`v0.2.12`) and cleaned the PR split:

- #1196 now contains only the deposit canonicalization/global completion dedup fix.
- This PR now contains only the withdrawal destination validation fix.

PR #1208, opened 2026-05-20
`devshard` No seed reveal and state cleanup from inferences

a-kuprin:
Drops the devshard commit–reveal seed round: MsgRevealSeed stays on the wire as a deprecated no-op so old diffs decode, but hosts no longer track RevealedSeeds or recompute compliance from revealed seeds. Finalizing → settlement is driven only by the finalize deadline. Local validation sampling (DeriveSeed / ShouldValidate) is unchanged.

It is ok as seed revealing was used to check completed validfations, but this checks never run for devshard. And validation protocol is subject to change, and future versions are expected to do not have seed reveal stage at all.

Hosts evict finished inferences from RAM and state, as earlier state was polluted by all inferences devshard ever served). Keep a small durable inference index in session storage for late-validation edge cases, and delete prompt/response payloads once an inference is past the validation window (PruneEventSink → PayloadStorage.DeleteInference). State hashing gains sealed composition and related tests; chain params add seal grace for settlement timing.

Design is documented in `devshard/docs/inference-lifecycle.md`.

Pruning logic is:
```
apply diff → persist diff
     ↓
processPruneEventsLocked  (Tier A / Tier C rules)
     ↓
emitPruneLocked (per inference)
     ├─ SealInference        → RAM + sealed index  (sync, same mutex)
     └─ OnInferencePrunable  → payload delete      (async queue)
```


0xgonka (commented):
Drops the devshard commit-reveal seed round (kept as deprecated no-op for wire compat), adds inference pruning with two-tier emission (terminal flip + stale-finished gates), introduces sealed-accumulator state composition (gated off via `useV2StateRootComposition = false`), and adds a chain-side `DevshardEscrowParams.DefaultSealGraceNonces` knob with an upgrade-handler backfill. Most of the new code is well-scoped, but the seal/prune machinery interacts with the v1 hash composition in a way that looks like a real settlement bug under the production flag setting.

**Must fix:**
- [Severity: settlement-breaking in v1 mode] In production `useV2StateRootComposition = false`, but `SealInference` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/seal.go#L223-L255 unconditionally `delete(sm.state.Inferences, id)` (line 253), while leaving `committedEntries[id]` intact. The v1 branch of `computeStateRootLocked` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/seal.go#L83-L95 reads `sm.committedEntries`, so intra-session signatures sign over a root that *includes* sealed entries. But `BuildSettlement` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/settlement.go#L25-L46 (v1 branch, line 32) computes `RestHash = ComputeRestHash(st.Balance, st.Inferences, st.WarmKeys)` from `st.Inferences` — which is missing the sealed entries. Chain verification at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/inference-chain/x/inference/keeper/devshard_settlement.go#L113-L129 reconstructs `state_root` from `msg.RestHash` and checks signatures against it; the signatures were collected over a different `state_root` (the committedEntries-based one), so recovery returns addresses that aren't in the group and the quorum check at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/settlement.go#L122-L126 fails. This is reachable in production because `manager.go` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/decentralized-api/internal/devshard/manager.go#L134-L138 wires `pruneSink` whenever `payloadStore != nil`, and `emitPruneLocked` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/host/host.go#L536-L557 calls `SealInference` unconditionally on terminal/stale-finished. Fix options: (a) guard `delete(sm.state.Inferences, id)` on `effectiveV2Composition()` so v1 keeps the entry in the live map (cheapest, preserves intent that v1 hash is unchanged by sealing — already what `TestSealInference_PreservesRootAndBlocksDuplicateID` asserts intra-host, but settlement isn't asserted); (b) plumb a v1-aware `BuildSettlement` that sources from `committedEntries` so its RestHash matches the signed root. Either way, add a `TestSeal_v1_SettlementRoundTrip` that calls `SealInference` and then `BuildSettlement(SnapshotState(), ...)` and asserts the reconstructed `state_root` matches `sm.ComputeStateRoot()` at that nonce.

**Should fix:**
- v1 inference-hash refactor at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/hash.go#L176-L218 changes `computeInferencesHash` from `deterministicMarshal.Marshal(&InferencesMapProto{Entries:...})` to a manual `protowire.AppendTag(1, BytesType) + AppendVarint(len) + entry` loop. The byte equivalence depends on `marshalInferenceEntry` producing the same bytes that the wrapper proto's repeated-field encoding would emit, in id-sorted order. The PR's regression coverage at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/hash_test.go#L131-L155 (`TestStateRoot_V1_Untouched`) checks `ComputeStateRoot` against a manual reconstruction via `ComputeRestHash` — but both sides go through the same new code, so a silent regression in entry assembly would pass. Add a golden-vector test pinning the v1 inference-hash bytes for a specific record set (e.g. hex-encoded 32 bytes), generated once against `release/v0.2.13` HEAD, so any future tweak to `marshalInferenceEntry` or the protowire buffer assembly fails fast.
- `processPruneEventsLocked` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/host/host.go#L441-L478 only emits terminal prunes when the diff contains one of `{FinishInference, TimeoutInference, Validation, ValidationVote}`. If the state machine has any path where an inference flips to a terminal status from a different tx (or via a multi-step apply that doesn't surface in the txs list of the same diff), the terminal prune never fires — the Tier C ledger then carries it until the stale-finished gates clear. Worth a one-shot audit of every site that writes `rec.Status = StatusValidated|StatusInvalidated|StatusTimedOut` in `devshard/state/machine.go` to confirm the four tx types above are the only entry points.

**Nits:**
- Deprecation comments on `MsgRevealSeed` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/proto/devshard/v1/diff.proto#L24 and https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/proto/devshard/v1/tx.proto#L64 say "retained for wire compatibility; reveal-seed txs are ignored at runtime" but don't pin the earliest release where the field can be reclaimed. Add a "do not remove until ≥v0.2.X" marker so a future cleanup doesn't break decode of in-flight or replayed diffs from older binaries.

**Looks correct:**
- Bounded prune queue with drop-on-full + RWMutex send-vs-close coordination at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/decentralized-api/internal/devshard/prune_sink.go#L107-L122. Senders take RLock and check `closed` before sending; shutdown takes the write lock to flip `closed` and `close(queue)` atomically. Send-on-closed-channel is not reachable.
- Shutdown drain semantics at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/decentralized-api/internal/devshard/prune_sink.go#L158-L187: workerCtx is separate from the shutdown ctx so closing the queue doesn't immediately cancel in-flight deletes; the per-delete `pruneDeleteTimeout = 5s` bounds drain time, and the outer `pruneShutdownTimeout = 15s` caps the total. workerCancel only fires after the drain races out, then a final `wg.Wait()` keeps workers from outliving payload-store close.
- DefaultSealGraceNonces backfill at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/inference-chain/app/upgrades/v0_2_13/upgrades.go#L101-L122 reads existing GroupSize before deriving (falling back to `DefaultDevshardGroupSize` when zero) and only overwrites when the persisted value is 0. Covered by both `TestSetDevshardEscrowParams_BackfillsDefaultSealGraceNonces` and `_PreservesExistingDefaultSealGraceNonces` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/inference-chain/app/upgrades/v0_2_13/upgrades_test.go#L16-L52.
- MsgRevealSeed wire-compat: field number 7 of the `tx` oneof in `DevshardTx` is preserved (https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/proto/devshard/v1/diff.proto#L25), and the `MsgRevealSeed` message struct itself is preserved (https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/proto/devshard/v1/tx.proto#L65-L69). Old diffs decode; new emitters don't produce. No `reserved` is required because nothing is being deleted.
- `VerifyDevshardSettlement` signature widening at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/inference-chain/x/inference/keeper/devshard_settlement.go#L80-L106 now takes the whole `*DevshardEscrowParams` with a mandatory nil check (line 82) and threads through the approved-versions allowlist via `validateDevshardSettlementVersionApproved` at lines 31-44. Empty allowlist is permissive (intentional for devnet/tests); non-empty rejects non-matching version tags before any signature work.
- `FoldSealedAccumulator` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/hash.go#L119-L137 commits to `seal_nonce_be || inf_id_be || committed_entry` in a domain-separated chain. Order-sensitive but the only call sites (`drainLiveIntoSealedAccLocked` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/devshard/state/seal.go#L181-L221 and `SealInference` itself, one id at a time under the state mutex) preserve a deterministic id-ordered fold.
- Snapshot recovery in `manager.go` at https://github.com/gonka-ai/gonka/blob/akuprin/no-seed-reveal/decentralized-api/internal/devshard/manager.go#L410-L425 restores `state → committedEntries → sealedNonces` in the right order: `RestoreState` rebuilds `committedEntries` from `state.Inferences` (rough), then `RestoreCommittedEntries` overwrites with the snapshot's authoritative entries (precise). Followed by `RebuildSealedInferenceIndex` which truncates and re-inserts the per-escrow durable sealed-inference index. This is the correct ordering — flipping the two restore calls would lose the post-seal committed entries.

a-kuprin:
@0xgonka I've forgot to clear `useV2StateRootComposition` that was used only for development purposes.
Now I've removed it, and it closes "[Severity: settlement-breaking in v1 mode] ..." and "v1 inference-hash refactor ..."

Also I've found that governance parameters for configuring stale nonces and time was passed, but never been used and already fixed last concern `processPruneEventsLocked`, that inferences wasn't pruned from terminal state.

I think it is clean now

PR #1226, opened 2026-05-22
MiniMax-M2.7 route — per-model dispatch + tool-message shape + reasoning_split

qdanik:
## Summary

Adds the **MiniMax-M2.7** (`MiniMaxAI/MiniMax-M2.7`) chat-completions route — a per-model branch off the universal contract. Lets clients running the M2.7 model line hit the gateway with the model-specific message shape, native extensions, and `extra_body` keys that vLLM's `minimax_m2` parser expects.

## What changes for clients

**Newly supported on this route:**
- `extra_body.reasoning_split` — when `true`, vLLM emits reasoning as a separate `reasoning_details[]` array instead of inline `<think>...</think>` in `content`.
- `messages[].reasoning_details` — assistant-turn reasoning history; **pass-through** (stripping these breaks M2.7 interleaved-thinking continuity).
- Inline `<think>...</think>` in assistant `content` — **pass-through verbatim**; the validator does not strip them from prior turns.
- MiniMax-shaped `role:"tool"` messages: `content: [{name, type:"text", text}]` array, no `tool_call_id`. Per-entry caps: ≤16 entries × `name` ≤64 B × `text` ≤64 KiB; closed allow-list of keys.

**Silent-stripped on this route** (universal/Kimi/Qwen behavior diverges): `thinking`, `thinking_token_budget`, `enable_thinking`, `tools[].function.strict`. MiniMax has no `chat_template_kwargs` toggle for thinking — it's structural to the chat template, not configurable per request.

**Rejected on this route** (HTTP 400): `partial`, `web_search` / `enable_search` / `search_kwargs`, `mask_sensitive_info` — MiniMax-platform-only fields that vLLM does not implement; fail-loud beats silent-strip. Also `structured_outputs` — MiniMax's OpenAI-compatible API doesn't declare it; forwarding would lean on parser paths with known fragility. Mirrors the Kimi-K2.6 policy.

## Architecture

- **Per-model dispatch** at the message-validator layer — universal validators stay default; M2.7 plugs in route-specific overrides (tool-message shape, content normalizer, assistant-turn pass-through).
- **Validator packages extracted** — message-shape rules moved into a dedicated `messagevalidators` package; parameter-rule handlers consolidated under `paramvalidators` and wired into the catalog through a single adapter. The previous mix of in-file handlers and ad-hoc helpers is gone; each rule is its own type with its own tests.
- **JSON numeric helpers consolidated** in the top-level `devshard` package (`JSONNumericUint64` + new `JSONNumericFloat64`). Removes the parallel `jsonutil` package introduced during the refactor — single source of truth, no duplicated coercion logic.

## Docs

- New per-model contract doc following the structure of `kimi-k2.6.md` / `qwen3-235b-a22b-instruct-2507.md`: model facts, deployment requirements, parameter overrides, native extensions, rejected keys, structured-outputs status, known model-side bugs.
- Adds the `MiniMax-N` and `SGLang-N` namespaces to the references index with sourced vendor docs, vLLM upstream issues, and CVE links backing each accept/reject/strip decision.
- Deprecates the legacy `devshard/docs/supported-params.md` in favor of `docs/chat-api/README.md` as the canonical contract reference. User-facing HTTP 400 error message URL updated to the new location.

## Why this matters

M2.7 is the first model on the gateway whose tool-message wire shape and thinking semantics diverge meaningfully from both the OpenAI baseline and the existing Kimi/Qwen routes. Without per-model dispatch the choices were: (a) loosen the universal validators (breaks safety guarantees for other routes), or (b) reject valid M2.7 traffic. The per-model split keeps universal behavior tight while letting M2.7 ship with the wire format its upstream parser actually emits.

The validator-package split is a side effect of needing route-specific behavior — but it also makes the rules independently testable and removes a longstanding "one giant file" pain point in `request_filters_*`.

## Test plan

- [x] Real-traffic smoke against staging M2.7 host: tool-call round-trip (assistant turn → tool message → assistant turn), with and without `reasoning_split=true`.
- [x] Verify HTTP 400 for each rejected key (`partial`, `web_search`, `structured_outputs`, etc.) with the new error URL pointing to `docs/chat-api/README.md`.
- [x] Confirm `<think>` blocks in prior assistant turns survive round-trip unchanged.
- [x] Regression: Kimi-K2.6 and Qwen3 routes unchanged (universal contract untouched).
- [x] `go test ./...` across the devshard module.


a-kuprin (approved):
LGTM. Just few styling nits, you are free to skip:

1. `ValidateNonEmptyContent` takes the specific field value (`any`), while `ValidateRequiredContent` and `ValidateAssistantContent` take the parent message (`map[string]any`). While the latter two need the map to check for field existence, the naming doesn't clearly distinguish between "validating a value" and "validating a field within a map."
For better readability, consider renaming functions that take maps to include `Field` in the name (e.g., `ValidateRequiredContentField` or `ValidateRequiredField`).
2. Local test helpers like `floatPtr` (in `handlers_test.go`) and `asMap` (in `normalizers_test.go`) are defined locally within different test files. Consolidating these and other similar utilities into a shared `testutil` package would improve discoverability and reduce duplication across the hierarchy.

Issue #1205, opened 2026-05-20
chain-halt: `markValidatorForDeletion` jailed branch races CometBFT validator-update lag → slashing fails with ErrNoValidatorFound


vitaly-andr:
## Summary

In the [`gonka-ai/cosmos-sdk`](https://github.com/gonka-ai/cosmos-sdk) fork, the jailed branch of `markValidatorForDeletion` (`x/staking/keeper/compute.go:559-563`) immediately deletes the validator record **and** the `ValidatorByConsAddr` index via `deleteValidatorInternal` (`x/staking/keeper/val_state_change.go:78-126`). When `SetComputeValidators` runs with a set that excludes a still-jailed validator, this branch fires. CometBFT, however, continues to include the deleted validator in `LastCommit` for `ValidatorUpdateDelay` blocks (≈ 2 effective blocks: `header.Height + 1 + 1` per cometbft `state/execution.go`'s "next next height" rule). On the next block, `slashing.BeginBlocker` (`x/slashing/abci.go:24-29`) walks `LastCommit` votes and calls `HandleValidatorSignature → GetValidatorByConsAddr` (`x/slashing/keeper/infractions.go:26-30`). The lookup returns a bare `stakingtypes.ErrNoValidatorFound` → `BeginBlocker` returns an error → consensus halts.

## Reproduction (chain halt)

Deterministic halt under the `inference-chain` simulation harness with seed 99 and `IS_TEST_NET=true` (selects the post-`ValidatorIndexFixHeight=658087` production path of `SetComputeValidators` — the same code mainnet runs):

```
block finalization failed: validator does not exist
```

Both top-level sim tests reproduce this halt deterministically on the current upstream code:
- `TestFullSimulation_x_Inference_Integrated`
- `TestFullAppSimulation`

## Root cause

`x/staking/keeper/compute.go:559-563`:
```go
// Jailed validators can't be added to power index, so delete them immediately
if validator.Jailed {
    logger.Info("deleting jailed validator immediately", "operator", validator.OperatorAddress)
    return k.deleteValidatorInternal(ctx, validator, valAddr)
}
```

[...]

PR #1182, opened 2026-05-18
feat(sim): Cosmos SDK simulation runnable + first-wave x/inference ops (#982)

vitaly-andr:
## Summary

First milestone of #982 — Cosmos SDK simulation that is both **runnable** and
**exercises real `x/inference` logic**. The issue's "Proposed Next Step"
defines this milestone as exactly that pair: *"simulator run ergonomics, a
smoke run path, first-wave x/inference operation support, repeated seeded
runs"*.

**Supersedes #1156.** #1156 was opened as a standalone Phase-1 PR under an
initially proposed PR-A/B/C split; Phase 2 came together quickly, and since
the issue itself bundles ergonomics + first-wave ops into one milestone,
combining them is one self-contained, reviewable PR instead of a stacked
cross-fork pair — and gives a reviewer something that actually hammers chain
logic, not plumbing alone. Phase 3 (second-wave ops — `InvalidateInference` /
`RevalidateInference` / training / admin flows — plus operation-weight tuning,
custom invariants, store decoders, parameter-edge fuzzing) still follows as a
separate PR.

Built fresh from `main`, not on @hleb-albau's #995 (closed). The Make-target
naming and `fixBankGenesisState` helper are ported from #995, with credit.

## Phase 1 — make simulation runnable

- simsx integration in `app/sim_test.go` / `sim_bench_test.go`; `*App`
  satisfies `simsx.SimulationApp`.
- Make targets `node-sim-smoke-test` (50×20) and `node-sim-full-test`
  (500×200); root Makefile delegates to `inference-chain/Makefile`.
- `disabledOpsSimModule` wrapper — nil `WeightedOperations` for staking /
  distribution / wasmd while preserving `GenerateGenesisState`.
- `fixBankGenesisState` — recomputes Supply, adds Gonka denom metadata so
  InitGenesis succeeds with staking ops disabled (ported from #995).
- Upstream tests reimplemented in simsx idioms: `TestAppImportExport_Postrun`
  and `TestAppSimulationAfterImport_Postrun` (both `t.Skip`'d pending the fork
  bonded-pool fix proposed in #1153); `TestAppStateDeterminism`.
- SDK config split (`InitSDKConfig` / `SealSDKConfig`); `docs/simulation.md`;
  `workflow_dispatch` smoke CI.

## Phase 2 — first-wave x/inference ops

- The 5 operations named in the issue, registered via `HasWeightedOperationsX`:
[...]

PR #1228, opened 2026-05-22
feat(sim): Cosmos SDK simulation + first-wave + V2 PoC chain (Phase 1+2+3 of #982)

### LINKED ISSUE #982: Enable simulation and fuzz testing for inference-chain
# Proposal: Enable Cosmos SDK Simulation and Fuzz Testing for `inference-chain`

## Summary

`inference-chain` should make Cosmos SDK simulation and fuzz-style state exploration a near-term engineering priority.

For a Cosmos chain, this kind of testing is not just “more tests.” It is one of the main ways to pressure-test the chain as a state machine. The simulator repeatedly generates and executes randomized operations, feeds the application unexpected or adversarial sequences of state transitions, and helps expose invalid assumptions that are easy to miss in hand-written tests. Its purpose is to hammer chain state, explore weird edges, and discover cases where the application accepts bad transitions, enters inconsistent state, fails determinism, or breaks import/export assumptions.

Today, `inference-chain` has the app-level simulator wiring in place, but it does not yet have meaningful custom-module simulation coverage. As a result, one of the most valuable testing tools in the Cosmos SDK stack is largely unused for the logic that matters most.

This proposal recommends a focused effort to:

1. treat Cosmos SDK simulation as a chain-hardening priority,
2. make simulator runs easy to execute locally and in CI,
3. implement real simulation coverage for `x/inference` first,
4. expand coverage to other custom modules where it creates real value,
5. use repeated seeded runs to find edge cases, invalid states, and hidden assumptions.

## What Cosmos SDK Simulation and Fuzz Testing Is For

Cosmos SDK simulation exists to explore the behavior of a blockchain application under broad, randomized, and often ugly execution conditions.

Rather than proving one expected path, simulation asks a harder question: what happens when the chain is subjected to long sequences of randomized operations, odd ordering, unusual account selections, unexpected timing, and state transitions we would not think to hand-author in tests?

The purpose is to:

- generate large amounts of unusual but structurally valid activity,
- feed the application edge-case inputs and bad sequences of state transitions,
- stress interactions between modules over time,
- explore unusual parameter configurations and boundary values,
- reveal invalid assumptions about ordering, balances, epochs, or participant state,
- detect import/export and determinism issues,
- surface bugs that only appear after many steps of evolving chain state.

For a complex Cosmos application, this is one of the closest things we have to adversarial state-machine testing inside the normal SDK toolchain.

## Problem

`inference-chain` contains complex custom stateful logic, especially in `x/inference`. That logic spans multi-step workflows, participant lifecycle, epoch-based coordination, balances, rewards, validation, and cross-module interactions. These are exactly the kinds of features where bugs often emerge not from one message in isolation, but from the sequence and combination of many valid actions over time.

Unit tests and integration tests remain necessary, but they are not designed to broadly search for strange emergent behavior. They usually validate known scenarios. Simulation and fuzz-style state exploration exist to find the scenarios we did not already think of.

Without meaningful simulator coverage, we are more likely to miss:

- invalid state transitions that only show up after long chains of operations,
- combinations of valid messages that produce broken or inconsistent state,
- subtle cross-module regressions,
- bugs that only appear under unusual parameter values or combinations,
- edge cases around randomized accounts, ordering, and sequencing,
- failures in import/export or non-determinism under broader state exploration,
- assumptions in custom business logic that hold in example-based tests but not under sustained random execution.

The problem is not that the chain has no simulation support at all. The problem is that simulation is not yet doing the job it is supposed to do for the custom logic of the chain.

## Why This Should Be a Priority

This is a high-leverage hardening investment.

- It improves confidence in the correctness of the chain as a state machine, not just the correctness of isolated handlers.
- It helps us find cases where the chain accepts or produces invalid state under complex execution histories.
- It provides a repeatable way to stress workflows that are awkward to cover comprehensively with unit tests.
- It surfaces higher-order bugs earlier, before they become upgrade issues or production incidents.
- It creates a foundation for long-running automated and agent-assisted hardening work, since repeated seeded runs and simulator implementation are highly parallelizable.

In short, simulation matters because the chain’s biggest risks are not only “does this message work,” but also “what happens after many steps of valid and semi-adversarial state evolution.”

## Current State

`inference-chain` already has the base app-level simulator hooks in place:

- the app constructs a `SimulationManager`,
- app-level simulation tests exist for import/export, post-import simulation, and determinism,
- the standard Cosmos SDK simulator entrypoints are present.

However, the custom-module layer is still mostly scaffolding.

- `x/inference` registers many weighted operations, but the individual simulator implementations are currently stubs that return `NoOpMsg`,
- several other custom modules include simulation files but empty `WeightedOperations`,
- randomized genesis support is minimal,
- parameter-space exploration is minimal, especially for unusual runtime parameters and boundary combinations,
- store decoders are not implemented for the custom modules,
- simulator runs are not part of the normal developer or CI workflow.

The practical result is that simulation can be “enabled” without meaningfully hammering the custom business logic of the chain.

## Parameter Surface and Testing Strategy

One especially important area for this initiative is parameter variation.

For `inference-chain`, there are two parameter categories that should be treated differently:

- genesis-only parameters, which are fixed at genesis and cannot be changed later through governance,
- standard runtime parameters, which can be modified after chain start.

The primary fuzzing target in this proposal is the runtime parameter surface.

For `inference-chain`, risk does not only come from message execution under normal settings. It also comes from the space of possible runtime parameter values and combinations the chain may operate under. Some issues only appear when parameters are pushed toward edge values or when multiple parameters interact in surprising ways.

Genesis-only parameters should generally remain at the actual chain values during simulation and fuzz runs. They represent fixed operating assumptions of the network, and varying them freely would often reduce realism and produce lower-signal results. They should only be changed intentionally when there is a specific reason to believe that a genesis-only edge is needed to reach a meaningful correctness or security condition.

This matters because runtime parameter combinations can create correctness, safety, and vulnerability issues:

- parameter combinations may make previously safe logic unsafe,
- boundary values may expose arithmetic, sequencing, or liveness problems,
- modules may behave correctly under common settings but fail under extreme or merely less common ones.

As a result, simulation and fuzz-style testing should not only randomize operations. It should also intentionally explore runtime parameter edges and dangerous combinations where doing so produces meaningful signal.

This does not always mean forcing parameter changes through governance inside the simulation itself. In many cases, we can explore more of the runtime parameter state space by starting a simulation run with alternate initial values for parameters that are normally mutable at runtime. That gives us broader coverage without requiring every test path to first simulate a governance vote.

The intended strategy is:

- use the real genesis-only parameters by default,
- vary standard mutable parameters aggressively,
- allow alternate initial values for mutable parameters as a testing shortcut,
- use governance-driven parameter changes inside simulation only when the governance transition itself is the behavior under test.

## Proposal

We should adopt a phased plan that first makes simulation runnable and useful, then expands coverage in a disciplined way.

### Phase 1: Make Simulation Operational

Goal: make it easy for developers and CI to run a small simulation smoke test and a larger seeded run.

This phase should include:

- a documented command or Make target for a short simulator smoke run,
- a documented command for a longer seeded run,
- clear guidance on required flags and expected runtime,
- optional CI coverage for a short simulation run so the path does not silently rot.

This phase is about reliability and ergonomics, not depth of coverage.

### Phase 2: Implement Meaningful `x/inference` Simulation

Goal: make the simulator actually exercise the most important chain logic.

This should start with a deliberately chosen subset of operations that represent real workflows instead of trying to implement every message at once.

Recommended first-wave operations:

- `SubmitNewParticipant`
- `StartInference`
- `FinishInference`
- `Validation`
- `ClaimRewards`

Possible second-wave operations:

- `InvalidateInference`
- `RevalidateInference`
- training-task flows,
- selected admin or allow-list flows where they are central to system correctness.

This phase should also include helper logic for selecting valid accounts and valid preconditions so the simulator creates realistic messages rather than mostly generating invalid transactions.

### Phase 3: Improve Simulation Quality

Goal: move from “it runs” to “it finds real issues.”

This phase should include:

- tuning operation weights based on realistic workflow frequency and bug-finding value,
- using realistic genesis defaults by default, with targeted alternate initial states only where they improve coverage of mutable runtime parameter behavior,
- exploring parameter boundaries and parameter combinations more aggressively,
- deciding which mutable parameters should be varied by alternate initial values versus by in-simulation governance flows,
- implementing store decoders for clearer simulation diff output,
- adding invariants or strengthening existing ones where simulator findings justify it,
- running repeated seeded simulations and triaging recurring failures.

### Phase 4: Expand to Other Custom Modules Selectively

Goal: cover the rest of the chain where simulation adds meaningful value.

Not every module needs the same level of simulator depth. Some modules may justify only a small number of operations; others may not need much beyond genesis and invariants. The point is to make conscious choices rather than leaving default Starport scaffolding in place indefinitely.

## What Needs To Be Done

The work can be broken into concrete deliverables:

1. Simulator execution

- add local run targets,
- document short and long run modes,
- verify simulator runs complete in current developer environments.

2. `x/inference` operation implementation

- replace `NoOpMsg` stubs with real message generation and execution,
- build helper utilities for account and state selection,
- ensure operations honor message preconditions,
- keep the first set of operations intentionally small and high-value.

3. Debuggability and confidence

- add store decoders for custom modules,
- improve failure output and seed capture,
- define a strategy for runtime parameter variation,
- document that genesis-only parameters remain fixed by default, with targeted exceptions only where justified,
- define a catalog of mutable parameters that should be varied directly at simulation start for faster coverage,
- add targeted automated coverage for parameter edge cases and dangerous combinations,
- add guidance for replaying failing seeds locally.

4. CI and maintenance

- run a short smoke simulation in CI or another regular automation path,
- define ownership so simulator coverage grows with module changes,
- treat new major message flows as candidates for simulator support.

## Benefits

If we do this well, we should get several concrete benefits:

- higher confidence in custom state-machine correctness,
- earlier detection of regressions in long workflows,
- better confidence that unusual runtime parameter settings do not create unsafe or inconsistent chain behavior,
- broader runtime-parameter coverage without requiring every scenario to simulate governance first,
- better validation of import/export and determinism assumptions,
- stronger upgrade safety,
- a reusable framework for agent-assisted hardening and repeated stress runs,
- fewer blind spots in the chain’s most complex business logic.

## Scope and Non-Goals

This proposal does not recommend trying to implement every simulator operation immediately.

Non-goals for the initial effort:

- full coverage for every message in every module,
- perfect random genesis modeling on day one,
- using simulator implementation to change product semantics,
- treating simulator success as a replacement for unit, integration, or property tests.

The right first milestone is meaningful coverage for the most important workflows, not exhaustive completion.

## Risks and Tradeoffs

There are real tradeoffs to manage:

- poorly designed simulation operations can generate noise instead of signal,
- implementing too many operations too early can create maintenance burden,
- naive parameter randomization can create mostly invalid or low-signal runs if it is not constrained thoughtfully,
- varying genesis-only parameters too freely can pull testing away from the real chain assumptions and reduce signal,
- random exploration without good precondition logic may mostly hit invalid tx paths,
- long-running simulation in CI can become expensive if not scoped carefully.

These are manageable risks, and they argue for a phased rollout rather than against the effort itself.

## Success Criteria

We should consider this initiative successful when:

- a developer can easily run a short simulation smoke test locally,
- CI or another regular automation path exercises that smoke path,
- `x/inference` has a first meaningful set of real simulation operations,
- parameter-edge testing covers runtime params in a deliberate way,
- genesis-only parameters remain fixed by default, with exceptions only where they are intentionally justified,
- seeded simulation failures are reproducible and diagnosable,
- simulation results begin surfacing actionable bugs, invariants, or hardening work.

## Suggested Priority

This should be treated as a high-value engineering hardening initiative, with `x/inference` as the first target.

It does not need to block all other work, but it should be prioritized above lower-leverage test scaffolding because it directly improves confidence in the correctness of the chain’s stateful core.

## Proposed Next Step

Approve a first milestone focused on:

- simulator run ergonomics,
- a smoke run path,
- first-wave `x/inference` operation support,
- repeated seeded runs and triage.

That would give the project a practical starting point without overcommitting to a large one-shot implementation.

vitaly-andr:
Closes #982

## Summary

Combined Phase 1 + Phase 2 + Phase 3 of #982 — full local-verification milestone.

- **Phase 1**: simsx migration of `inference-chain` simulation app. `HasWeightedOperationsX` registry, disabled-incompatible upstream sim ops (staking/distribution/wasm), `make sim-smoke-test`/`sim-full-test`, postRunActions for ImportExport / SimulationAfterImport, `fixBankGenesisState`. Pinned `-GenesisTime` for reproducibility.
- **Phase 2**: First-wave x/inference real ops via simsx — `SubmitNewParticipant`, `StartInference`, `FinishInference`, `Validation`, `ClaimRewards`. Real secp256k1 dev/TA signatures (`signing.go`), real model registration (`SimModelIDs`), `EnsureModelsInEpochGroup`. Failing-validation branch through real `x/group.SubmitProposal` + revalidation votes (`MsgRevalidationVoteFactory`).
- **Phase 3**: V2 PoC chain factory family — `MsgPoCV2StoreCommit`, `MsgMLNodeWeightDistribution`, `MsgSubmitPocValidationsV2`, `MsgSubmitSeedFactory` — implementing cosmos-sdk simsx `HasFutureOpsRegistry`. State-driven (target, model) selection in F2/F3 with reservoir sampling. Picker age-filter + model-weight cache filter in `MsgValidation`/`MsgRevalidationVote`. Bootstrap goes through real production paths (`EpochGroup.AddMember` + `BuildEpochDataTransientCache` + `Staking.SetComputeValidators`).

## Verification (canonical Docker, seed=99, GenesisTime=1700000000)

- `make sim-smoke-test` (50×20): PASS, lifecycle `total=75 startProcessed=72 finishProcessed=62 validated=50 proposed=4`.
- `make sim-full-test` (500×200) against `gonka-ai/cosmos-sdk` carrying the fixes from #1205 and #14: PASS, height=501, opsCount=60763, app-hash `d7a20cf044230b56189194968872515be0d059d0a0401203f67995e87870b9d8`.
- Cherry-picked onto `gm/microrelease` (v0.2.13 staging) + same patched cosmos: PASS, height=501, opsCount=53892, app-hash `8108a42faf20e550bec3ec748330c0453e58d238d156afb199a3db87a1baccf6`. One adapter line: `NewEpochMemberFromActiveParticipant` 3-arg signature change.

## Surfaced fork-side findings (already filed, NOT in this PR)

This simulation surfaced two distinct `gonka-ai/cosmos-sdk` bugs end-to-end. Both filed separately:

1. **`markValidatorForDeletion` jailed-branch race** vs CometBFT validator-update lag — issue gonka-ai/gonka#1205. Reproduces as sim-full halt around block ~140-147 depending on RNG path.
2. **GON-191 stale consensus-key reassignment** — gonka-ai/cosmos-sdk PR #14 by `@0xgonka`. Independently needed for validator re-registration with reused cons-keys.

Isolated test with only PR #14 applied (no #1205 patch) reproduces the same b147 halt — confirming the two fixes are orthogonal and both required for clean sim-full liveness. With both applied, sim-full completes 500 blocks deterministically.

## Closes / Supersedes

- Supersedes #1182 (closed with forward-pointer to this PR).

## Test plan

- [ ] `make sim-smoke-test` PASS in canonical Docker
- [ ] `make sim-full-test` PASS in canonical Docker (against patched cosmos-sdk per above)
- [ ] Adversarial review via codex:rescue — APPROVE (1 FIX folded, 1 COURSE-CORRECT folded)
- [ ] `gonka-ai/ai-reviewer` (claude_max_proxy) — 0 must-fix, 1 major (folded), 3 nits (folded)
- [ ] V2 PoC chain advances ≥12 epoch rotations (`currentEpoch=12` observed at end-of-run)

cc @patimen

---

## Update — Phase 3 quality added (commit `983fcaac4`)

This PR now also carries the Phase 3 hardening/quality work (4th commit):
custom invariants via `RegisterInvariants` (bank-backs-positive-balance
solvency, no-stuck-voting, effective-epoch-fresh, active-invalidations-ref-live),
store decoders, aggressive parameter-boundary fuzzing, and three real
secondary-op factories (`SubmitUnitOfComputePriceProposal`, `SubmitHardwareDiff`,
`SubmitNewUnfundedParticipant`) replacing NoOp stubs.

### Prerequisites — sim-full is green only with all of these applied

The full 500×200 run surfaced, and depends on, these production fixes — each
filed as its own focused PR (kept out of this PR to preserve scope):

- **gonka-ai/gonka#1275** — `expireInferences` refunds the client escrow for
  `VOTING` inferences stranded on an x/group quorum miss (reported in #1265, fixed by #1275).
- **gonka-ai/gonka#1276** — `revalidateInferenceVote` membership precheck: skip
  the x/group votes when the voter is not an epoch-group member (reported in #1269, fixed by #1276).
- **gonka-ai/cosmos-sdk#16** — `markValidatorForDeletion` hardening (jailed-delete
  race vs CometBFT lag + non-jailed `DelegatorShares` divide-by-zero); this is the
  PR for the #1205 issue noted above, and is complementary to GON-191
  (gonka-ai/cosmos-sdk#14).

Verified combined (canonical Docker, seed=99, `GenesisTime=1700000000`, against
cosmos-sdk#16): **`sim-full` 500×200 PASS, height=501, opsCount=38727, app-hash
`f76df5cdfe03161ff114e8c926c719bed0e27d60e5548a883b79ef9826372b82`** — this
supersedes the earlier app-hash above (that run predates the quality work and the
fixes). Unit tests (`keeper` / `simulation` / `module`) pass standalone on stock
`v0.53.3-ps17`.

### Open design question (not a blocker)

The post-run `bank-backs` solvency invariant holds in the canonical 500×200 run
(transient intra-epoch under-backing heals via debt-recovery by settlement). The
asymmetric refund-debit drift it surfaces under shorter / aggressive runs is a
design question, filed as **gonka-ai/gonka#1273** with reproducer
`TestRefundInvalidation_LeavesAccountingDrift` — deliberately not fixed here,
pending maintainer guidance on intended semantics.

### DIFF
--- a/.github/workflows/simulation.yml
+++ b/.github/workflows/simulation.yml
@@ -1,0 +1,48 @@
+name: Simulation Smoke (manual)
+
+# Manual trigger only; simulation runs add ~5-15 min to CI cost.
+# Maintainers may add `pull_request` / `push` triggers when sim coverage is
+# considered worth gating on. See docs/simulation.md for context.
+#
+# To activate auto-triggers, replace the `workflow_dispatch:` block below with:
+#
+#   pull_request:
+#     branches:
+#       - '**'
+#   push:
+#     branches:
+#       - main
+#
+# Or keep manual-only and trigger via the Actions tab / `gh workflow run`.
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  sim-smoke:
+    runs-on: ubuntu-24.04
+    name: Run sim-smoke-test
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.24.2'
+
+      - name: Cache Go modules
+        uses: actions/cache@v4
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('inference-chain/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - name: Run sim-smoke-test
+        working-directory: inference-chain
+        run: make sim-smoke-test
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: release decentralized-api-release inference-chain-release tmkms-release proxy-release proxy-ssl-release bridge-release versiond-release check-docker build-testermint run-blockchain-tests test-blockchain local-build api-local-build node-local-build api-test node-test mock-server-build-docker proxy-build-docker proxy-ssl-build-docker bridge-build-docker run-bls-tests devshardctl-build devshardd-build versiond-build-docker testapp-server-build-docker
+.PHONY: release decentralized-api-release inference-chain-release tmkms-release proxy-release proxy-ssl-release bridge-release versiond-release check-docker build-testermint run-blockchain-tests test-blockchain local-build api-local-build node-local-build api-test node-test mock-server-build-docker proxy-build-docker proxy-ssl-build-docker bridge-build-docker run-bls-tests devshardctl-build devshardd-build versiond-build-docker testapp-server-build-docker node-sim-smoke-test node-sim-full-test
 
 VERSION ?= $(shell git describe --always)
 DEVSHARD_VERSION ?= dev
@@ -21,6 +21,13 @@ api-build-docker:
 
 node-build-docker:
 	@make -C inference-chain build-docker SET_LATEST=1 $(if $(GENESIS_OVERRIDES_FILE),GENESIS_OVERRIDES_FILE=$(GENESIS_OVERRIDES_FILE),)
+
+# Simulation dispatchers (issue #982 Phase 1)
+node-sim-smoke-test:
+	@make -C inference-chain sim-smoke-test
+
+node-sim-full-test:
+	@make -C inference-chain sim-full-test
 
 mock-server-build-docker:
 	@echo "Building mock-server JAR file..."
--- a/docs/simulation.md
+++ b/docs/simulation.md
@@ -10,22 +10,23 @@ its phased implementation.
 From repo root:
 
 ```bash
-make sim-smoke-test   # 50 blocks × 20 ops, ~15 min
-make sim-full-test    # 500 blocks × 200 ops, ~120 min (Phase 2+ recommended)
+make sim-smoke-test   # 50 blocks × 20 ops, ~10 s
+make sim-full-test    # 500 blocks × 200 ops, ~35 s (Phase 2+ recommended)
 ```
 
-Both targets enter the `inference-chain` module and run `go test -run
-TestFullAppSimulation` with the `sims` build tag. The `*_Postrun` tests
-(see Test inventory) are not invoked by these targets; they call `t.Skip`
-with the reason and only run when invoked directly via
-`go test -tags sims ./app/...` without a `-run` filter.
+Both targets enter the `inference-chain` module and run
+`TestFullAppSimulation` with the `sims` build tag; `sim-smoke-test` also
+runs `TestFullSimulation_x_Inference_Integrated` (per-op state assertions).
+The `*_Postrun` tests (see Test inventory) are not invoked by these
+targets; they call `t.Skip` with the reason and only run when invoked
+directly via `go test -tags sims ./app/...` without a `-run` filter.
 
 ## Run modes
 
 | Target | Blocks × Ops | Seeds | Wall-clock | When to use |
 |---|---|---|---|---|
-| `sim-smoke-test` | 50 × 20 | 1 (`-Seed=99`) | ~15 min | Pre-PR sanity check; CI via [`.github/workflows/simulation.yml`](../.github/workflows/simulation.yml) (`workflow_dispatch` only, not a required PR check) |
-| `sim-full-test` | 500 × 200 | 1 (`-Seed=99`) | ~120 min | After Phase 2 real ops land. Informative only when ops are not NoOpMsg stubs |
+| `sim-smoke-test` | 50 × 20 | 1 (`-Seed=99`) | ~10 s | Pre-PR sanity check; CI via [`.github/workflows/simulation.yml`](../.github/workflows/simulation.yml) (`workflow_dispatch` only, not a required PR check) |
+| `sim-full-test` | 500 × 200 | 1 (`-Seed=99`) | ~35 s | After Phase 2 real ops land. Informative only when ops are not NoOpMsg stubs |
 | direct `go test` (with `-Seed=N`) | per `-NumBlocks` / `-BlockSize` flags | 1 (the supplied `-Seed`) | depends | Reproducing a specific seed, debugging |
 | direct `go test` (no `-Seed`) | per flags | simsx `defaultSeeds` (37 seeds, parallel subtests) | depends | Broad coverage sweep |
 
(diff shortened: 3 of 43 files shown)


vitaly-andr:
Closing this one. The work is still in progress, it just does not belong on this branch any
more.

The harness is being carried onto the current release branch, and part of what this PR
contains no longer applies, since the inference lifecycle it generates traffic for was removed
from the chain. It will come back as a new pull request against the release branch rather than
as an update to this one.

#982 stays open, and so does the work on it. I will write up what changes and what stays when
the new pull request is up.

PR #1232, opened 2026-05-22
feat(api): optional signed agent request envelope

aeoess:
## Summary

Reference implementation for the envelope schema discussed in [#1230](https://github.com/gonka-ai/gonka/discussions/1230). Follows up on @gmorgachev's broker-side, request-layer placement suggestion in [#1185](https://github.com/gonka-ai/gonka/discussions/1185).

The middleware sits on `/v1/chat/completions` and `/v1/completions`, verifies a signed agent request envelope when the headers are present, and preserves today's behavior when they are absent.

The envelope cryptographically binds an Ed25519 agent identity to an existing Gonka Developer's Cosmos secp256k1 key via ADR-036, enforces request-layer scope (model allowlist, expiry, audience, chain_id), and emits a structured attribution event after inference submission. The existing Developer/TA/Executor signing chain, AuthKey replay protection, `x/authz` delegation, escrow logic, and `MsgFinishInference` are untouched.

## Problem

Gonka's existing identity model treats a Developer as a single Cosmos secp256k1 key. When an automated agent calls inference under a Developer today, the three available paths each have a real cost:

1. Use the Developer's cold key directly. Security risk.
2. Use a Cosmos `x/authz` warm key. Works for signing, but `RequestedBy` resolves to the grantee address; the specific agent identity is invisible on chain.
3. Wrap the inference API with a custom service that holds the warm key. Most common today; the wrapping service carries trust, and the agent identity is still invisible.

`x/authz` is also message-type scoped, not attribute scoped. A Developer cannot grant a warm key the right to run inference only for model X, only for purpose Y, only until Tuesday. A custom `InferenceAuthorization` type in `x/authz` would address that, but it's a chain change requiring governance.

A third gap, unrelated to the above: there is no on-chain field for the beneficiary when the payer (Developer) and the entity the work is done for differ. Existing on-chain attribution maps to payer only.

## Solution

A new middleware reads an optional `X-Agent-Passport` header. When present, it:

1. Verifies the envelope's `principal_sig` against the Developer's Cosmos secp256k1 key via **ADR-036**, the same arbitrary-data path Keplr `signArbitrary` uses. Cross-validated by an independent `@cosmjs/amino` reference vector in tests.
2. Confirms the embedded `principal_pubkey` derives to the stated `principal_address`. No chain query needed, which eliminates first-tx-pubkey and DoS surfaces.
3. Verifies a per-request `X-Agent-Sig` over a domain-separated, length-prefixed payload, binding `chain_id`, HTTP method, full request URI (path + query string), the JCS-canonicalized envelope, and the request body.
4. Checks scope: `audience` equals `"gonka"`, `chain_id` matches, `not_before` and `expires_at` bound the window, `expires_at - issued_at` is within `MaxEnvelopeTTL`, `issued_at` is not more than 30s in the future (clock-skew bound), and the request's model is in `allowed_models` if the field is present.
5. Validates `beneficiary` is well-formed: rejects control chars, bad UTF-8, and length over 255 bytes; no silent transformation.
6. Attaches a verified `Context` to the Echo request with `RequestBound = false`.

After the existing `readRequest` parses `RequesterAddress`, the handler confirms `principal_address == RequesterAddress`, or resolves via `x/authz` (Grantee to Granter) using Gonka's existing `AuthzCache.GetPubKeyForSigner` directly. On success, it sets `RequestBound = true`. After `s.recorder.StartInference` succeeds, the attribution event is enqueued through a bounded async emitter.

## Composition with existing Cosmos patterns

This adds no new root of trust. It composes on top of primitives Gonka already uses:

- **ADR-036** (Arbitrary Message Signature Specification) for the Developer's signature on the passport.
- **`x/authz`** for warm-key resolution. Reuses `AuthzCache.GetPubKeyForSigner(granter, signer, msgType)` directly; no new resolver layer.
- **Ed25519** from the Go standard library for the agent's per-request signature.
- **RFC 8785 JCS** for canonicalization. Vendored, scoped honestly to the envelope schema: no JSON numbers, no full Unicode surrogate handling, and unknown fields are rejected.

## Hardening invariants

All non-optional. Each has a corresponding test.

| Concern | Mitigation |
|---|---|
| Body OOM | `http.MaxBytesReader` at `MaxBodySize` (default 10 MiB, capped at 1 GiB) |
| Cross-chain replay | `chain_id` in envelope + `chain_id` in agent-sig payload (defense in depth) |
| Cross-endpoint replay | Method + full request URI (path + query) in agent-sig payload |
| Body substitution | Body bytes in agent-sig payload |
| Field-boundary substitution | Length-prefix domain separation, TLS 1.3 transcript pattern |
| Envelope tampering | Strict decode (`DisallowUnknownFields`, no trailing content), duplicate-key rejection, principal_sig over full JCS envelope |
| Algorithm confusion | Strict `*secp256k1.PubKey` type assertion; multi-sig rejected explicitly |
| Stolen passport blast radius | Bounded `MaxEnvelopeTTL` (default 1h, configurable) |
| Future-dated `issued_at` | 30s clock-skew tolerance; envelopes further in the future rejected |
| Log injection (beneficiary) | Reject on control chars, bad UTF-8, length over 255 bytes; no transformation |
| Logger backpressure | Bounded channel + background drain + drop-on-full + counter |
| Logger panic | Drain goroutine recovers; sink panic cannot crash the node |
| Misconfiguration | Layer enabled with empty `chain_id` panics at startup, not at first request |
| Required fields | `validateEnvelopeRequired` rejects missing fields as 400, not as downstream errors |
| Body parse failure | 400 when an envelope is present and the body is malformed |

## Tests

`go test -race ./internal/agentenvelope/... ./internal/server/middleware/...` clean under `GOTOOLCHAIN=go1.24.2`.

- ~80 unit tests in `internal/agentenvelope/` (one per named error class, plus happy paths, JCS Appendix B vectors, cosmjs cross-check).
- 13 HTTP integration tests in `internal/server/middleware/` covering passthrough (no headers), disabled-by-config, full happy path, every named-error rejection, and the request-URI-with-query-string binding.
- Independent ADR-036 cross-check via `@cosmjs/amino` in `signing_cosmjs_test.go`, using the amino path Keplr `signArbitrary` uses. This cross-check gives confidence that the sign bytes match the Keplr path.
- Two benchmarks: absent-header overhead, full-verify warm-authz overhead.

## Components affected

Only `decentralized-api/`. No changes to `inference-chain/`, broker, devshard, event listener, validation, cosmosclient, ML node, consensus, ante handlers, escrow, settlement, tokenomics, or `MsgFinishInference`.

New code lives in `decentralized-api/internal/agentenvelope/`. The middleware is mounted in `server.go` (~20 lines). Handler changes in `post_chat_handler.go` total ~17 lines: principal-match or `x/authz` Grantee resolution, then `RequestBound = true`, then attribution enqueue after `StartInference`. New config struct in `apiconfig/` with bounded defaults.

A `examples/aps-client/` reference client and a `docs/specs/aps-agent-envelope.md` spec doc ship with the change.

## What this PR explicitly does NOT do

Listing exhaustively because reviewers historically check each surface; flagging them upfront so review can focus where it matters.

1. Replace any existing signing layer (Developer, TA, Executor)
2. Modify Cosmos `x/authz` semantics, `authzcache`, AuthKey replay protection, or `ValidateTimestamp`
3. Modify escrow, settlement, pricing, tokenomics, consensus, ante handlers, ML node, PoC, or validation
4. Modify `MsgStartInference` or `MsgFinishInference`
5. Implement scoped delegation chains (Phase 2)
6. Implement passport revocation beyond short TTL (Phase 2)
7. Implement multi-sig principal support (Phase 2)
8. Implement devshard daemon integration (Phase 2)
9. Verify beneficiary identity (it's a signed claim, not a verified identity)
10. Require any inference request to identify an agent (header is optional)
11. Change request body size limits or any other operational constraint
12. Add new external dependencies (Ed25519 is Go standard library; cosmos-sdk and Echo are already in the dep tree)
13. Add chain query latency when `principal_address == RequesterAddress`
14. Claim adoption of APS as a Gonka-wide standard

## Future work explicitly out of scope

Each of these gets a separate Discussion if interest exists:

- Extending `MsgFinishInference` with optional `AgentId` and `Beneficiary` fields (chain change, governance)
- Scoped delegation chains (parent agent narrows scope, hands to sub-agent)
- Multi-hop attribution receipts
- Devshard daemon integration (different code path, different signing semantics)
- CosmWasm helper contracts for on-chain enforcement
- Multi-sig principal support
- Revocation mechanism beyond short TTL

## Disclosure

I maintain APS (Agent Passport System, [agent-passport.org](https://agent-passport.org)), an Apache 2.0 open protocol for AI agent identity and scoped delegation. The envelope format in this PR is APS-compatible. The verifier is fully isolated in `internal/agentenvelope/`, so the wire format can be swapped without re-architecting anything else. Adopting it Gonka-side does not require Gonka to adopt APS network-wide.

The Future work items above would all get separate Discussions if Gonka maintainers wanted to pursue them; none are assumed by this PR.

## Reviewer notes

@gmorgachev, this follows up on the broker-side, request-layer placement direction from your reply in [#1185](https://github.com/gonka-ai/gonka/discussions/1185). The middleware sits in front of `/v1/chat/completions` and `/v1/completions`; no changes anywhere else in the request path.

Happy to split the PR into stacked commits if a smaller review shape would be easier. As-is, the diff is ~2,700 new lines, mostly in one new package, and ~25 modified lines across two existing files; the commits are already logically separated by layer: foundation, verifier, middleware, tests, docs, and review-pass refinements.

Ready for review when you are.

### DIFF
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -48,6 +48,11 @@ const (
 	DefaultAgentEnvelopeMaxTTLSeconds       int64 = 3600
 	DefaultAgentEnvelopeMaxBodySize         int64 = 10 * 1024 * 1024
 	DefaultAgentEnvelopeAttributionQueueCap       = 1024
+	// MaxAgentEnvelopeBodySize is the hard upper bound for max_body_size. A
+	// configured value outside (0, MaxAgentEnvelopeBodySize] falls back to the
+	// default. The bound keeps the signed-payload length prefixes well within
+	// uint32 range.
+	MaxAgentEnvelopeBodySize int64 = 1 << 30
 )
 
 type NatsServerConfig struct {
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -53,6 +53,14 @@ const (
 	// default. The bound keeps the signed-payload length prefixes well within
 	// uint32 range.
 	MaxAgentEnvelopeBodySize int64 = 1 << 30
+	// MaxAgentEnvelopeTTLSeconds is the hard upper bound for max_ttl_seconds.
+	// A value outside (0, MaxAgentEnvelopeTTLSeconds] falls back to the
+	// default. The bound prevents a negative or overflowing time.Duration.
+	MaxAgentEnvelopeTTLSeconds int64 = 7 * 24 * 3600
+	// MaxAgentEnvelopeQueueCap is the hard upper bound for
+	// attribution_queue_cap. A value outside (0, MaxAgentEnvelopeQueueCap]
+	// falls back to the default, bounding the buffered-channel allocation.
+	MaxAgentEnvelopeQueueCap = 1 << 20
 )
 
 type NatsServerConfig struct {
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -27,8 +27,28 @@ type Config struct {
 	BandwidthParams          BandwidthParamsCache     `koanf:"bandwidth_params" json:"bandwidth_params"`
 	PoCParams                PoCParamsCache           `koanf:"poc_params" json:"poc_params"`
 	TransferAgentAccessCache TransferAgentAccessCache `koanf:"-" json:"-"` // not persisted, synced from chain
-	DevshardVersionsCache      DevshardVersionsCache      `koanf:"-" json:"-"` // not persisted, synced from chain
+	DevshardVersionsCache    DevshardVersionsCache    `koanf:"-" json:"-"` // not persisted, synced from chain
+	AgentEnvelope            AgentEnvelopeConfig      `koanf:"agent_envelope" json:"agent_envelope"`
 }
+
+// AgentEnvelopeConfig configures the optional signed agent request envelope
+// verifier. The envelope layer is off unless Enabled is set. ChainID must be
+// set to this network's chain ID when the layer is enabled.
+type AgentEnvelopeConfig struct {
+	Enabled             bool   `koanf:"enabled" json:"enabled"`
+	ChainID             string `koanf:"chain_id" json:"chain_id"`
+	MaxTTLSeconds       int64  `koanf:"max_ttl_seconds" json:"max_ttl_seconds"`
+	MaxBodySize         int64  `koanf:"max_body_size" json:"max_body_size"`
+	AttributionQueueCap int    `koanf:"attribution_queue_cap" json:"attribution_queue_cap"`
+}
+
+// Default agent-envelope settings, applied by GetAgentEnvelopeConfig when a
+// field is left at its zero value.
+const (
+	DefaultAgentEnvelopeMaxTTLSeconds       int64 = 3600
+	DefaultAgentEnvelopeMaxBodySize         int64 = 10 * 1024 * 1024
+	DefaultAgentEnvelopeAttributionQueueCap       = 1024
+)
 
 type NatsServerConfig struct {
 	Host                  string `koanf:"host" json:"host"`
--- a/decentralized-api/apiconfig/config_manager.go
+++ b/decentralized-api/apiconfig/config_manager.go
@@ -151,7 +151,7 @@ func (cm *ConfigManager) GetAgentEnvelopeConfig() AgentEnvelopeConfig {
 	if cfg.MaxTTLSeconds == 0 {
 		cfg.MaxTTLSeconds = DefaultAgentEnvelopeMaxTTLSeconds
 	}
-	if cfg.MaxBodySize == 0 {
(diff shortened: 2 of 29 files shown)


gmorgachev:
@aeoess 

1.`/v1/chat/completion` is effectively deprecated way of billing. all inference goes to desvshard based path
2. could you elaborate why it should be on hosts side? i think independent brokers, who creates escrow's gateway (`dl/devshards-gateway-to-main`), can handle it. we can avoid adding this to protocol at all, no?

aeoess:
it seemed like the most protocol neutral hook - no protocol change, no escrow change. Just verify the envelope before the existing inference path continues.

should I close this PR and move it into a sidecar against `dl/devshards-gateway-to-main`?

aeoess:
closing. continuing on #1230.

PR #1202, opened 2026-05-19
feat(devshard): default thinking_token_budget for Kimi-K2.6 reasoning split

clanster:
## Summary

Kimi-K2.6 is a reasoning model — it emits `<think>...</think>` alongside the visible content, and **both count toward `max_tokens`**. With the current `RequestMaxTokensCap = 4_096` (intentionally low because the per-request connection-time budget is also bounded), the model regularly spends the entire 4k on reasoning and the visible content never starts: `finish_reason=length`, `content=""`.

Moonshot's own platform docs recommend `max_tokens ≥ 16_000` for exactly this reason ([platform.kimi.ai](https://platform.kimi.ai/docs/guide/use-kimi-k2-thinking-model)). We cannot lift the cap to 16k today, so instead this PR whitelists vLLM's `thinking_token_budget` parameter through the catalog and injects a per-model default for Kimi-K2.6 — it caps the reasoning portion separately and forces the model to enter the visible-content phase before `max_tokens` runs out.

## The bug, observed live

Live probe against the production INT4 image (`ghcr.io/kaitakuai/mlnode-b200-kimi-k2-6:0.2.13-vllm0.20.0-q.int4-k2`, digest `43943c0b`), `temperature=1.0` per Moonshot recommendation, three reasoning prompts × four scenarios in parallel:

| prompt | scenario | finish | tokens | reasoning chars | content chars |
|---|---|---|---:|---:|---:|
| math | current_bug (no budget) | **length** | 4096 | 9975 | **0** |
| math | this-PR default (budget=2048) | stop | 2870 | 4400 | **1960** |
| math | client override (budget=500) | stop | 2234 | 1034 | 3723 |
| math | thinking:disabled | **length** | 4096 | 10659 | **0** |
| code | current_bug (no budget) | **length** | 4096 | 13468 | **0** |
| code | this-PR default (budget=2048) | stop | 2696 | 6394 | **2555** |
| code | client override (budget=500) | stop | 1163 | 1930 | 2616 |
| code | thinking:disabled | **length** | 4096 | 12030 | **0** |
| logic | current_bug (no budget) | stop | 937 | 2257 | 1378 |
| logic | this-PR default (budget=2048) | stop | 732 | 1494 | 1442 |
| logic | client override (budget=500) | stop | 604 | 1187 | 1202 |
| logic | thinking:disabled | stop | 578 | 1002 | 1249 |

[...]

PR #1233, opened 2026-05-23
feat(devshard): exempt Kimi-K2.6 reasoning-burn empties from host quarantine

qdanik:
## Summary

Kimi-K2.6 hosts were being quarantined for what OpenAI and Anthropic both document as an **expected reasoning-model outcome**: at small `max_tokens`, the model consumes the entire budget inside `<think>...</think>` and vLLM strips `</think>` as a special token, leaving `content=null` with `finish_reason=length`. This PR closes that bleed on two layers — response-side classification and request-side shaping — and consolidates the Kimi token-budget rules behind two paramvalidators.

## The bug, observed live on Kimi-K2.6

The dominant signature is the same probe shape:

```json
{
  "model": "moonshotai/Kimi-K2.6",
  "max_tokens": 100, "min_tokens": 100, "thinking_token_budget": 50,
  "stream": true, "temperature": 0, "logprobs": true, "top_logprobs": 5
}
```

Per-capture attempt breakdown showed one host emitting all output via `delta.reasoning` (winner) while the other host(s) in the fan-out returned 479 bytes of SSE protocol with zero content/reasoning chunks — penalized. The losing hosts weren't broken; the model just rolled a different reasoning trajectory under FP8/attention-backend nondeterminism between parallel instances. After three such "strikes" they entered a 30-minute quarantine.

## Industry precedent

Both OpenAI and Anthropic explicitly document empty content with `finish_reason=length`/`stop_reason=max_tokens` as a valid model outcome at reasoning-heavy budgets:

> "The incomplete status might occur before any visible output tokens are produced, meaning you could incur costs for input and reasoning tokens without receiving a visible response."
> — [OpenAI Reasoning Guide](https://platform.openai.com/docs/guides/reasoning)

> "If max_tokens is hit during thinking, visible content may be truncated or absent."
> — [Anthropic — Handling stop reasons](https://docs.anthropic.com/en/docs/build-with-claude/handling-stop-reasons)

OpenRouter, similarly, [does not fall back](https://openrouter.ai/docs/api/reference/errors-and-debugging) to another provider on empty content — it passes the response through as success. Our gateway was the outlier.

### 1. Response-side classification (`redundancy.go` + `participant_limiter.go`)

The SSE writer now extracts `usage.completion_tokens` from each chunk into a new `inflight.usageComplTokens` field. A new predicate `isModelBurnEmpty` separates two cases that previously collapsed into "empty stream":

| Pattern | Classification | Action |
|---|---|---|
| `contentChunks == 0` AND `usageComplTokens == 0` | **Host fault** | `ObserveEmptyStream` — counts toward `emptyStreamStreak` |
| `contentChunks == 0` AND `usageComplTokens > 0` | **Model burn** | `ObserveModelBurnEmpty` — telemetry-only, no streak, no quarantine |

The penalty switch in `recordSample` routes the two cases separately.

### 2. Request-side shaping (`paramvalidators/`)

Two new validators replace eight catalog rules:

- **`KimiMaxTokensFloorValidator`** — floors `max_tokens` and `max_completion_tokens` to 16 (one validator, both fields). Was two separate `ModelScopedParameterHandler{Handler: MinUintParameterHandler{Min: 16}}` entries.

- **`KimiThinkingTokenBudgetValidator`** — supersedes `ThinkingTokenBudgetDefaultsValidator` and four follow-up handler rules:
  1. `max_tokens < 256` → force `thinking_token_budget = 0` (bypass thinking — any thinking phase at this budget starves content)
  2. Else if `thinking_token_budget` absent → default to `max_tokens / 2`
  3. Cap at 96 000 (Moonshot HLE/AIME budget)
  4. Clamp to `max_tokens − 64` so visible content always has headroom after `</think>`

| Behavior | Before | After |
|---|---|---|
| `max_tokens=100, ttb=50` (the probe shape) | passed through, model burns | **ttb forced to 0**, content emits |
| `max_tokens=4096, ttb=10000` | clamped to 4096 | clamped to **4032** (content headroom) |
| `max_tokens=4096`, ttb absent | default 2048 | default 2048 (unchanged) |
| `max_tokens=256`, ttb absent | default 128 | default 128 (unchanged, at force-zero boundary) |

### Performance

The SSE writer's hot path runs the usage parser per chunk. Added a `bytes.Contains` fast-skip on the `"usage"` key marker (vLLM emits the usage object only in the final chunk of a stream):

```
BenchmarkSseChunkUsageCompletionTokens/no_usage_fast_skip-12   12697225   80.71 ns/op    0 B/op   0 allocs/op
BenchmarkSseChunkUsageCompletionTokens/with_usage_full_parse-12   896880   1363  ns/op  392 B/op  10 allocs/op
```

**17× faster on the 99% path** (every non-final chunk), zero allocations.

## Test plan

- [x] `go test ./...` from `devshard/` — all 18 packages green (`cmd/devshardctl` 16s, `paramvalidators` 0.8s).
- [x] `BenchmarkSseChunkUsageCompletionTokens` confirms the fast-skip path is zero-alloc.

## Stacking

After PR #1227 (the `max_tokens >= 16` floor) — this PR generalizes that fix into one Kimi-aware validator and adds the classification half of the story. No dependency on other open PRs.

## References

- OpenAI — [Reasoning Guide](https://platform.openai.com/docs/guides/reasoning)
- Anthropic — [Handling stop reasons](https://docs.anthropic.com/en/docs/build-with-claude/handling-stop-reasons)
- OpenRouter — [Errors & Debugging](https://openrouter.ai/docs/api/reference/errors-and-debugging) (model-side errors do not trigger provider fallback)
- vLLM — `kimi_k2` reasoning parser at `vllm/reasoning/kimi_k2_reasoning_parser.py`
- Prior art: #1202 (`thinking_token_budget` defaulter), #1227 (`max_tokens` floor)


baychak (approved):
## Approved ✅

Reviewed code and tests, ran the full devshard suite locally — all green.

### What's strong

- **Two-layer fix is the right shape.** Response-side `isModelBurnEmpty` cleanly composes on top of `isEmptyStreamAttempt`, inheriting probe / no-receipt / error-source exclusions for free. Request-side consolidates 5 catalog rules into 2 single-source-of-truth validators.
- **Independent confirmation of the probe-burn pattern.** Verified against a separate captured-requests bundle (2026-05-22) — same `max_tokens=100, min_tokens=100, thinking_token_budget=50` shape dominates the empty-stream captures, and the "479 bytes of SSE protocol with zero content/reasoning chunks" signature is a tight match. The classification fix will eliminate a large chunk of false-positive host quarantines.
- **Performance fast-skip via `bytes.Contains(sseUsageKeyMarker)`** is the right shape for the hot path: vLLM emits `usage` only in the terminal chunk, so 99%+ of calls return after the marker check with zero allocs.
- **Test coverage locks every behavior.** `TestNormalizeChatRequestKimiThinkingTokenBudgetClosesProbeBurnPattern` literally reproduces the probe shape and pins the regression; `TestParticipantRequestLimiterObserveModelBurnEmptyIsTelemetryOnly` / `NeverQuarantines` lock the streak-bypass property.
- **Industry-precedent framing** (OpenAI/Anthropic docs, OpenRouter no-fallback behavior) is the right framing — this isn't a bug to silently fix, it's a documented model outcome the gateway was over-penalizing.

### Minor nits (non-blocking)

1. **Doc the override semantics in `KimiThinkingTokenBudgetValidator` header.** Current comment doesn't make explicit that force-zero *overrides client-provided ttb*, which is the load-bearing behavior closing the probe-burn case. One-line addition like `(overrides any client-set ttb — see ProbeBurnPattern test)` would help the next reader.

2. **Inline comment in the noop list for `max_completion_tokens`.** Not obvious that the field was demoted from its own catalog entry because `KimiMaxTokensFloorValidator` handles both fields in a single pass (operating on the full `vctx.Document`). A 1-line comment in `newParameters([]string{..., "max_completion_tokens"})` would prevent a future reader from re-introducing a per-field rule.

3. **Field alignment** in `KimiThinkingTokenBudgetValidator` struct — `ForceZeroBelowMaxTokens` is off by one space. `gofmt` should fix.

4. **Optional**: wire `BenchmarkSseChunkUsageCompletionTokens` into a `-run=^$ -bench=Sse` CI step so the 17× claim is regression-locked.

None of these block merge.

LGTM 🚀

qdanik:
Merged into [upgrade-v0.2.14](https://github.com/gonka-ai/gonka/tree/upgrade-v0.2.14) branch
PR: https://github.com/gonka-ai/gonka/pull/1434

PR #1234, opened 2026-05-23
`davshard` runtime config long-poll, lean Testermint e2e, epoch-driven prune

### LINKED ISSUE #1028: `devshards` `SessionConfig` setting by governmant

a-kuprin:
# Summary

Stops devshard hosts from polling chain governance params on a timer. **dapi** owns chain subscription and serves a **`GetRuntimeConfig`** long-poll RPC on NodeManager; **devshardd** consumes snapshots via `devshard/runtimeconfig`. Embedded `/v1/devshard` keeps in-process `ConfigManager` access. Storage retention prunes old epoch DBs **only on epoch change**, not every 30s.

Also includes a small **ManagedStorage** read-lock fix from [PR #1143 discussion](https://github.com/gonka-ai/gonka/pull/1143#discussion_r3200794751).

It was initially for fix findings during 0.2.13 review. But config management refactoring make this PR large.
We still need to add migrations for postgreSQL and sqlite, from db version-to-version. Remove CREATE TABLE IF NOT EXIST from hot paths, and do not create sqlite database on escrow start if we have postgres in our deployed stack.

Also the intention is to remove "embedded devshard inside dapi" in next release ([see](https://github.com/gonka-ai/gonka/pull/1084)), and we move DevshardAvailibilityGate from decentralized-api to devshards.

## Commits in this PR

| Commit | Summary |
|--------|---------|
| [`38c6805bf`](https://github.com/gonka-ai/gonka/commit/38c6805bf) | ManagedStorage: `RWMutex` on `CreateSession` hot path |
| [`26e9ddbc6`](https://github.com/gonka-ai/gonka/commit/26e9ddbc6) | Runtime config long-poll, devshardd/embedded wiring, Testermint + local infra |
| [`735ff3150`](https://github.com/gonka-ai/gonka/commit/735ff3150) | Prune devshard DBs on epoch change only (remove 30s ticker) |

**Reviewers can check each of these commits separately**

All earlier commits are aligning the branch with 0.2.13 release

## Test plan

- [ ] `go test ./decentralized-api/apiconfig/... ./decentralized-api/nodemanager/... ./decentralized-api/cmd/devshardd/...`
- [ ] `go test ./devshard/storage/... ./devshard/runtimeconfig/...`
- [ ] `cd testermint && ./gradlew :test --tests RuntimeConfigTests -DexcludeTags=unstable,exclude`
- [ ] `cd testermint && ./gradlew :test --tests DevsharddRuntimeConfigTests -DexcludeTags=unstable,exclude` (requires `make devshardd-build`)
- [ ] Existing devshard / versiond suites as needed (`DevshardTests`, `VersiondTests`)

## Docs

- [`params-dataflow.md`](./params-dataflow.md) — short data-flow overview 

## Notes on e2e tests

While logic is correct one e2e test fails (test for DevshardAvailibilityGate). The reason is that `dapi` doesn't handle 503 http errors from `devshardd`. And this is solved at https://github.com/gonka-ai/gonka/pull/1211

# PR commits changelog

Branch **`ak/devshard-0.2.13-optimizations`**. Only commits **after** `e915ccd67` (merge `akuprin/no-seed-reveal`).

Use with [`pr-description.md`](./pr-description.md) for the GitHub PR body.

---

## `38c6805bf` — Lock use optimization

**Subject:** Lock use optimization (per [gonka#1143 discussion](https://github.com/gonka-ai/gonka/pull/1143#discussion_r3200794751), @x0152)

**Scope:** 1 file — `devshard/storage/managed.go`

### Major changes

- Replace `sync.Mutex` with `sync.RWMutex` for the prune-cursor guard on `ManagedStorage`.
- `CreateSession` takes a **read** lock while checking `prunedUpTo` and creating the session, so concurrent reads are not blocked by other session creates (prune still uses write lock in `PruneOnce`).

### Why

Hot-path session creation should not serialize on an exclusive mutex when only reading the prune cursor.

---

## `26e9ddbc6` — Runtime config via dapi long-poll; lean Testermint e2e

**Subject:** `feat(devshard): runtime config via dapi long-poll; lean Testermint e2e`

**Scope:** Large cross-cutting change (~90+ files). Main themes below.

### dapi — runtime config server

- **`apiconfig`:** `RuntimeConfig` snapshot, `params_block_height`, `RuntimeConfigNotifier`, `ApplyRuntimeConfigBlockIfChanged` (notify on param or epoch change).
- **`nodemanager`:** `GetRuntimeConfig` gRPC with long-poll (`max_wait`, `unchanged` short-circuit); unit tests for long-poll, back-compat, max-wait cap.
- **Event listener:** Updates runtime cache from block dispatcher; removes devshard `AvailabilityTracker` wiring from dispatcher path.
- **Embedded devshard:** `ConfigManagerAvailability`, `ConfigManagerDefaults` for bind-time grace; `configParamsProvider` for logprobs/epoch.

### devshardd — runtime config client

- **`devshard/runtimeconfig`:** Long-poll gRPC client, `OnEpochChange`, backoff, tests + test gRPC server.
- **`cmd/devshardd`:** Drops periodic `chainParamsProvider` / `QueryParams` loop; mandatory `runtimeconfig` via `params_provider.go`.
- **`RegisterEpochPrune`:** `OnEpochChange` → `ManagedStorage.PruneOnce` (initial wiring; ticker removed in next commit).

### Proto consolidation

- Single source: `devshard/nodemanager/nodemanager.proto` → `devshard/nodemanager/gen`.
- Remove duplicate proto/generate under `decentralized-api/nodemanager` and `devshard/mlnode`.

### Bind / bridge

- **`devshard/bridge`:** `DevshardDefaults` / grace from runtime snapshot; `BuildGroupFromEscrow` — one `GetEscrow` per bind.
- **`internal/devshard/bridge.go`:** `NewChainBridgeWithDefaults`, runtime snapshot for grace defaults.

### Testermint & local runs

- **`RuntimeConfigTests.kt`:** dapi NodeManager long-poll matrix (no versiond).
- **`DevsharddRuntimeConfigTests.kt`:** versiond + host `devshardd`, genesis-only (`joinCount=0`), compose overlay.
- **`DockerGroup.kt`:** Two-phase genesis boot for versiond overlay; deferred pair discovery.
- **`docker-compose.versiond.yml`:** `NODE_MANAGER_ADDR`, keyring env for devshardd child.
- **Apple Silicon:** `scripts/blst-portable.*`, `BLST_PORTABLE` in Docker/Makefiles; versiond `linux/amd64` notes in docs.
- **`Makefile`:** `devshardd-build`, portable BLST wiring.

### Docs (this commit)

- [`params-dataflow.md`](./params-dataflow.md)
- [`testermint-infrastructure.md`](./testermint-infrastructure.md)

### Other

- `.gitignore` for build artifacts; `devshard/host`, `user/httpsession`, mlnode client import path updates.

---

## `735ff3150` — Prune devshard database only on epoch change

**Subject:** Prune devshard database only when epoch change. Do not useless tries every 30 seconds on timer

**Scope:** 12 files — storage pruner, dapi epoch hook, docs/tests

### Major changes

- **`devshard/storage/managed.go`:** Remove 30s background ticker and `pruneInterval`; `Start()` runs one catch-up `PruneOnce` after recovery; pruning only when callers invoke `PruneOnce`.
- **Embedded dapi (`main.go`):** `ConfigManager.SetEpochChangeHandler` → `go devshardStore.PruneOnce` on epoch publish.
- **`apiconfig`:** `epoch_change_listeners.go` — fire handler when `ApplyRuntimeConfigBlockIfChanged` reason is `epoch_change` (not initial/param-only); unit test.
- **devshardd:** Unchanged hook via `runtimeconfig.OnEpochChange` (from `26e9ddbc6`); `NewManagedStorage` signature drops interval arg.
- **Docs:** `params-dataflow.md`, `storage-design.md`; Testermint comment in `DevshardPostgresStorageTests.kt`.

### Why

Epoch transitions are already observed on the runtime-config path (long-poll / block listener). Periodic prune every 30s duplicated work and logged noise without advancing retention until epoch moved.

---

## Suggested review order

1. `38c6805bf` — small, isolated lock change  
2. `26e9ddbc6` — core RPC + runtimeconfig + Testermint (largest)  
3. `735ff3150` — storage prune policy + embedded epoch handler


0xgonka (commented):
Stops devshardd from timer-polling chain governance params; dapi now owns the chain subscription and serves a `GetRuntimeConfig` long-poll over NodeManager. Long-poll plumbing, lock semantics, and epoch-prune wiring are well-built. **One must-fix proto issue** in the same v0.2.13 numbering slots that just landed in #1208 (already on the base branch) — it'll break wire compatibility with any existing v0.2.13-RC state.

**Must fix:**

- **[Critical] `DevshardEscrowParams` field renumbering breaks wire compat with the already-merged v0.2.13 base.** `inference-chain/proto/inference/inference/params.proto:374-388` inserts `bool devshard_requests_enabled = 9` and shifts the existing `default_seal_grace_nonces` (was 9 in the base from PR #1208) to 10, and `default_inference_clear_grace_seconds` (was 10) to 11. Any chain or testnet that has persisted `DevshardEscrowParams` from the prior v0.2.13 state will mis-decode after the upgrade: field 9 (was a uint32 grace = ~20) gets read as a bool (varint, decodes to `true`), field 10's uint32 inherits whatever was at field 10 in the old state (`default_inference_clear_grace_seconds` = 120, decoded as `default_seal_grace_nonces`), and the new field 11 is absent so `default_inference_clear_grace_seconds` reverts to 0. The upgrade handler at `v0_2_13/upgrades.go:115-118` backfills SealGrace only if it's currently 0, so a wrong-non-zero value from misdecoded data **will silently survive the upgrade**. Fix by appending the new field rather than inserting: `bool devshard_requests_enabled = 11` (or higher), keep 9 and 10 as they were in the base. Mainnet impact today is zero because v0.2.13 didn't pass governance, but anyone on a 0.2.13-RC build (testnet, devnet, internal clusters) hits this on upgrade.

**Should fix:**

- **[Medium] Epoch-prune goroutine has no panic recovery.** `decentralized-api/cmd/devshardd/params_provider.go:55-60` wires `store.PruneOnce(ctx)` into the epoch-change callback via `go store.PruneOnce(ctx)`. The callback dispatcher at `apiconfig/epoch_change_listeners.go:25-27` has a `defer recover()` for the listener itself, but the inner goroutine spawned by `go store.PruneOnce(ctx)` is outside that recover frame. `PruneOnce` calls `rp.pruneBefore(cutoff)` on Postgres / SQLite drivers — a panic there (e.g., driver-level nil-deref on a disconnected pool) kills the goroutine silently and pruning stops forever (no metric, no log, the 30s timer that used to catch this is gone). Wrap the inner `go` in `defer recover()` with an error log; the cost is one line.

- **[Medium] `GuardianPermission` is genesis-immutable and gates a runtime kill switch.** `inference-chain/x/inference/keeper/msg_server_set_devshard_requests_enabled.go:11` auths `MsgSetDevshardRequestsEnabled` with `GuardianPermission`, which `permissions.go:checkGuardianPermission` resolves against `GetGenesisGuardianAddresses(ctx)` — a list set at genesis with no on-chain mutation path. If any single guardian key is compromised, that key can flip `devshard_requests_enabled` to `false` and freeze the entire devshard substrate, with no way to revoke that key without a chain upgrade. Either (a) add a parallel `GovernancePermission` so a gov proposal can disable a rogue guardian's effect, or (b) require ≥2-of-N guardians by accepting the message only when N guardian signatures appear (rough; less invasive option (a) is cleaner). At minimum, document in the PR that the kill-switch trust model is "any one guardian."

- **[Low] No per-peer / per-connection cap on `GetRuntimeConfig` long-poll.** `decentralized-api/nodemanager/server.go:79-155` caps each call at 60s (`clampMaxWait` defaults to 60s, override via `DAPI_RUNTIME_CONFIG_MAX_WAIT_SECONDS`) but doesn't limit concurrent in-flight calls per peer. A misbehaving devshardd peer (or an attacker who can hit the NodeManager port) can park ~2 KB × N goroutines indefinitely at the maximum-wait ceiling. Mitigate via `grpc.WithMaxConcurrentStreams()` on server construction, or per-peer accounting in the handler. Not exploitable today since NodeManager is internal, but worth a defense-in-depth note.

- **[Low] `RuntimeConfigSnapshot` reads content under RLock but `ParamsBlockHeight` from a separate atomic — non-atomic snapshot.** `decentralized-api/apiconfig/config_manager.go RuntimeConfigSnapshot`: `cm.runtimeParamsBlockHeight.Load()` happens inside the RLock-protected return statement but the value is updated in `runtime_config_publish.go:92-99` AFTER the content has been updated under `cm.mutex.Lock()` and AFTER `runtimePublishMu` is acquired — there's a window where a snapshot reader can observe new content (cache already updated by `SetDevshardVersionsCache`) paired with the old height. Self-correcting because `Notify()` always follows the height bump, so a long-poll client wakes up and sees the right pair on the next read, but a single-shot reader at `maxWait=0` can return `Unchanged` with a snapshot whose content has actually drifted from the height. Tighten by either (a) bumping `runtimeParamsBlockHeight` under `cm.mutex.Lock()` together with the cache write, or (b) accepting the race and documenting that `params_block_height` is the canonical revision marker — `Unchanged` is "your height matches our last published height" not "no content change exists."

**Nits:**

- `decentralized-api/apiconfig/runtime_config_publish.go:92-99` uses a CAS loop on `runtimeParamsBlockHeight` while already holding `runtimePublishMu` — only one writer can be here at a time, so a plain `if h := atomic.Load; h < blockHeight { atomic.Store(blockHeight) }` is equivalent and clearer. Not wrong, just over-engineered.
- `decentralized-api/nodemanager/server.go:142-152` does the standard `if !timer.Stop() { <-timer.C }` drain dance in two `select` branches. Worth a small helper `stopAndDrain(t *time.Timer)` to remove the duplication and prevent future drift between the two copies.
- PR description mentions a known-failing e2e (DevshardAvailibilityGate) addressed in #1211 — link that PR in the failing test's `t.Skip` reason so future operators can find it.

**Looks correct:**

- **Long-poll subscribe-before-read is right.** `server.go:88-93` calls `notifier.NotifyChan()` *before* `RuntimeConfigSnapshot(epochID)`, so a `Notify()` arriving between the two doesn't get lost — the wake channel is already captured.
- **Timer cleanup is right on all `select` exits.** Both the wake branch and the ctx-done branch at `server.go:142, 152` follow the `if !timer.Stop() { <-timer.C }` pattern; no goroutine or timer leaks.
- **`params_block_height` monotonicity is enforced.** `runtime_config_publish.go:92-99` breaks the loop when `blockHeight <= cur`, so the published height never goes backwards even if block events arrive out of order.
- **RWMutex downgrade in `ManagedStorage.CreateSession` is safe.** `devshard/storage/managed.go:143-153` only reads `m.prunedUpTo` under the RLock. `m.inner.CreateSession` operates on the wrapped store's own state (not on the pruner's `prunedUpTo`), and `m.observe()` is atomic CAS. `PruneOnce` is the only writer of `prunedUpTo` and uses `m.mu.Lock()` (full write) at line 109, so the lock contract holds. Standard RWMutex pitfall (write-under-RLock) is avoided.
- **`MsgSetDevshardRequestsEnabled` wiring is consistent end-to-end.** Message permission declared at `permissions.go MessagePermissions` map (`MsgSetDevshardRequestsEnabled → GuardianPermission`), enforced at handler entry via `k.CheckPermission(...)`, GuardianPermission resolves correctly to `checkGuardianPermission`. No bypass.
- **`AvailabilityTracker` gating blocks the right subset of host operations.** `devshard/host/host.go requestBlockedWhenUnavailable` blocks only inference-bearing txs (StartInference / TimeoutInference / Validation / ValidationVote) and Payload submissions; it deliberately lets gossip-only `req.Diffs` through so the host can still apply state and reach consensus while disabled. Right call.
- **Upgrade handler properly initializes the new field.** `v0_2_13/upgrades.go setDevshardEscrowParams` unconditionally sets `DevshardRequestsEnabled = types.DefaultDevshardRequestsEnabled` rather than the "only if 0" backfill pattern used for the grace fields, which is the right choice for a bool whose zero value (`false`) would be a chain-wide kill switch on every upgrade.
- **Epoch-driven prune is idempotent.** `managed.go PruneOnce:103-105` bails early when `prunedUpTo >= cutoff`, so duplicate epoch-change events from the notifier are safe to fire. Removing the 30s timer doesn't lose safety as long as the epoch-change callback fires at least once per epoch — which `ApplyRuntimeConfigBlockIfChanged` guarantees at the dispatcher level.
- **Notifier broadcast via channel-close + swap.** `runtime_config_notifier.go` is unconventional but correct: closing the old channel wakes all waiters simultaneously, and the swap to a fresh channel happens under mutex. No lost-wake possible if callers re-subscribe after each wake, which `server.go` does.

**Bottom line:** Hold for the proto fix (must-fix #1). Everything else is shippable as-is or follow-up. The long-poll implementation is genuinely well-built — subscribe-before-read, timer cleanup, monotonicity, idempotent prune are all handled correctly, which is the failure mode I was most worried about going in.

0xgonka:
also some merge conflicts now assuming that is the right base branch

a-kuprin:
@0xgonka I have fixed them. Except this concerns:

1. [Medium] GuardianPermission is genesis-immutable and gates a runtime kill switch.
Is applicable to 0.2.13 that is already in prod.
Maybe we can think about it separately. 
But it definitely more broader problem than this PR.

2. [Low] No per-peer / per-connection cap on GetRuntimeConfig long-poll.
As this is the internal call

0xgonka:
@a-kuprin agreed on (1) and (2): Guardian model is a broader chain-wide discussion (track separately, not a blocker here) and NodeManager-internal callers don't need a per-peer cap today. Both noted as won't-fix-here.

One outstanding: the Critical proto-field renumbering. params.proto:380-388 still has bool devshard_requests_enabled = 9 rather than appended at the end. If your read is "no chain state with the old numbering exists in any environment we care about so this is a moot concern," that's a reasonable position... please confirm explicitly so it's on the record. Otherwise, the one-line fix is to renumber devshard_requests_enabled = 11 and let SealGrace + InferenceClearGrace keep their 9/10 slots from the base; this is wire-compatible with any pre-existing v0.2.13-RC state.

a-kuprin:
@0xgonka 
Currently released is `gm/microrelease` branch where `devshard_requests_enabled` index is set to 9

```go
// DevshardEscrowParams defines governance-controlled parameters for devshard escrow creation and settlement.
message DevshardEscrowParams {
  uint64 min_amount = 1;
  uint64 max_amount = 2;
  uint32 max_escrows_per_epoch = 3;
  uint32 group_size = 4;
  repeated string allowed_creator_addresses = 5;
  uint64 token_price = 6;
  repeated DevshardApprovedVersion approved_versions = 7;
  uint32 max_nonce = 8;
  bool devshard_requests_enabled = 9;
}
```

So we are appending to existing release proto:

```go
  uint32 default_seal_grace_nonces = 10;
  uint32 default_inference_clear_grace_seconds = 11;
```

untouching existed fields

0xgonka (approved):
Stops devshardd from timer-polling chain governance params; dapi owns chain subscription and serves a `GetRuntimeConfig` long-poll over NodeManager. Six commits since my prior review address the snapshot race and the epoch-prune panic recovery directly. **One blocker remains: merge conflicts with base.**

**PR state:**
- CI: pass (6/6: Build and Test API Wrapper, Devshard, Versioned, chain, build, lint)
- **Merge: DIRTY** — conflicts with `devshard-0.2.13-v2`. PR #1235 (`devshard_meta` drain on client cancel) and PR #1238 (Finish recovery gossip) landed on base after this PR's head and touch the same files. 11 conflicting files: `devshard/cmd/devshardctl/proxy.go`, `devshard/cmd/devshardctl/proxy_test.go`, `devshard/docs/proxy.md`, `devshard/host/host.go`, `devshard/host/host_test.go`, `devshard/host/mempool.go`, `devshard/host/mempool_test.go`, `devshard/storage/shared_test.go`, `decentralized-api/internal/devshard/bridge_test.go`, `decentralized-api/internal/devshard/manager_test.go`, `decentralized-api/internal/server/public/post_chat_handler_interruption_test.go`.
- Prior reviews: 1 COMMENTED (mine, 2026-05-25) with a 4-comment follow-up thread.
- New activity: author landed 6 commits between 2026-05-26 12:05 and 13:36 addressing the prior review. Status of each prior thread:
  - **Critical (proto field 9 renumber):** author was right — I was wrong on this one. Confirmed: PR makes no changes to `params.proto`, and `bool devshard_requests_enabled = 9` is the baseline state on both `devshard-0.2.13-v2` and `main`. Retracting (mirrors the retraction I posted on #1258).
  - **Medium (epoch-prune panic recovery):** addressed by commit `1d8724698` — `PruneOnceAsync` in [`devshard/storage/managed.go:93-103`](devshard/storage/managed.go#L93-L103) wraps the goroutine in `defer recover() { slog.Error }`, and [`decentralized-api/cmd/devshardd/params_provider.go:77`](decentralized-api/cmd/devshardd/params_provider.go#L77) switched to it. Closed.
  - **Medium (Guardian permission immutable):** author declined as broader chain-wide discussion; I agreed in 2026-05-27 12:37 comment. Tracked separately, not a blocker here.
  - **Low (per-peer long-poll cap):** author declined as NodeManager is internal; I agreed. Closed.
  - **Low (non-atomic snapshot — content+height read separately):** addressed by `1d8724698` — both fields now read under one `runtimePublishMu.RLock()` window in [`decentralized-api/apiconfig/config_manager.go RuntimeConfigSnapshot`](decentralized-api/apiconfig/config_manager.go). Regression test `TestConfigManager_RuntimeConfigSnapshot_StaleUntilPublish` locks the new "stale-until-publish" semantic. Closed.

**Must fix:**

- **[Blocker] Rebase against current `devshard-0.2.13-v2`.** mergeable=CONFLICTING. The list above is what needs resolution; PR #1235's proxy-cancel work and PR #1238's recovery-gossip work both refactor the same hot-path files this PR touches. Until rebased, CI is running on a stale base and the protocol-level interactions between long-poll runtime config and the new mempool gossip can't be evaluated.

**Should fix:**

- **[Low] Out-of-order epoch-handler invocation when `runtimePublishMu` is released before `notifyEpochChange`** ([`runtime_config_publish.go:106-115`](decentralized-api/apiconfig/runtime_config_publish.go#L106-L115), introduced by `1bd69811a`). Two concurrent `ApplyRuntimeConfigBlockIfChanged` writers can have their epoch-handler calls scheduled in arbitrary order: writer A captures `(oldEpoch=1, newEpoch=2)` and unlocks; writer B captures `(2, 3)` and unlocks; the scheduler can run B's `notifyEpochChange(2, 3)` before A's `notifyEpochChange(1, 2)`. The current sole subscriber (`PruneOnceAsync`) is idempotent and bounded by `prunedUpTo`, so benign today. But the new `TestApplyRuntimeConfigBlockIfChanged_EpochHandlerMayReadSnapshot` documents that handlers may now read `RuntimeConfigSnapshot()`, and a handler that reads will see the LATEST published epoch — not necessarily the epoch matching its `(old, new)` args. Either add a comment on `SetEpochChangeHandler` ("handlers must not assume in-order delivery") or serialize-callbacks queue so handler invocation order matches publish order. Comment is the cheaper option.

**Looks correct (newly fixed since prior review):**

- **Atomic content + height snapshot.** Commit `1d8724698` collapses the prior content-RLock + atomic-height-load into a single `runtimePublishMu.RLock()` window that reads both. `RuntimeConfigSnapshot` now reflects the LAST PUBLISHED revision with live-cache fallback only before the first publish. `TestConfigManager_RuntimeConfigSnapshot_StaleUntilPublish` locks the semantic: cache writes alone don't move the snapshot until `ApplyRuntimeConfigBlockIfChanged` runs.
- **Epoch-prune panic recovery.** `PruneOnceAsync` in [`managed.go:93-103`](devshard/storage/managed.go#L93-L103) wraps the goroutine in `defer recover() { slog.Error }`, and [`params_provider.go:77`](decentralized-api/cmd/devshardd/params_provider.go#L77) switches the epoch-change hook to use it. Storage-driver fault no longer silently kills pruning forever.
- **Self-deadlock guard.** `1bd69811a` releases `runtimePublishMu` before calling notifier + epoch handler; `TestApplyRuntimeConfigBlockIfChanged_EpochHandlerMayReadSnapshot` is the right regression test — a handler that calls `RuntimeConfigSnapshot()` would previously RLock-after-Lock (RWMutex isn't reentrant) and hang. Fix is correct, just with the out-of-order caveat above.
- **`stopAndDrain` helper** in [`server.go`](decentralized-api/nodemanager/server.go) extracts the duplicated `if !t.Stop() { <-t.C }` from two select branches into a helper. Less drift risk going forward.
- **`SetRuntimeParamsBlockHeight` no longer needs a CAS loop.** Old `atomic.Int64.CompareAndSwap` loop is replaced with a plain `if height > cm.runtimeParamsBlockHeight` under `runtimePublishMu.Lock()` — equivalent under the new lock discipline, simpler to read.

**Bottom line:** Rebase first; once the conflicts with #1235/#1238 resolve cleanly, this is mergeable. The Low-Should-Fix is a one-line comment.


---

Looks good once merge conflicts are resolved

PR #1235, opened 2026-05-23
`devshard` fix(devshardctl): drain upstream SSE for devshard_meta after client cancel

a-kuprin:
When a user aborts a streaming chat (Ctrl+C / client disconnect), **devshardctl** must still read the host’s SSE stream until **`devshard_meta`** arrives. That event carries mempool updates (e.g. `MsgFinishInference`) needed for correct session sequencing. Previously, canceling the client could stop the upstream read too early and drop protocol completion.

**Behavior:** On client disconnect, the proxy stops writing to the client but keeps the **upstream** request alive (bounded drain, default **5s**). It continues parsing SSE until `devshard_meta` or timeout. The OpenAI-facing request context is not canceled solely because the client went away.

## Commits

| Commit | Summary |
|--------|---------|
| `85f2cb79d` | devshardctl: upstream SSE drain through `devshard_meta` on client cancel |
| `651bfab53` | Follow-up cleanup in `proxy.go` |
|`6fe5e29c2`|Added inference_clear_grace_seconds missed|

Other commits are for aligning with 0.2.13

## Files

- `devshard/cmd/devshardctl/proxy.go` — cancel flag, bounded meta drain, streaming/non-streaming paths
- `devshard/cmd/devshardctl/proxy_test.go` — client-cancel + meta drain coverage

### DIFF
--- a/decentralized-api/internal/devshard/bridge_test.go
+++ b/decentralized-api/internal/devshard/bridge_test.go
@@ -72,7 +72,8 @@ func TestChainBridge_GetEscrow_SealGraceFromParams(t *testing.T) {
 	qc.On("Params", mock.Anything, mock.Anything).Return(&inferenceTypes.QueryParamsResponse{
 		Params: inferenceTypes.Params{
 			DevshardEscrowParams: &inferenceTypes.DevshardEscrowParams{
-				DefaultSealGraceNonces: 88,
+				DefaultSealGraceNonces:            88,
+				DefaultInferenceClearGraceSeconds: 99,
 			},
 		},
 	}, nil)
@@ -81,6 +82,7 @@ func TestChainBridge_GetEscrow_SealGraceFromParams(t *testing.T) {
 	info, err := cb.GetEscrow("42")
 	require.NoError(t, err)
 	assert.Equal(t, uint32(88), info.SealGraceNonces)
+	assert.Equal(t, uint32(99), info.InferenceClearGraceSeconds)
 }
 
 func TestChainBridge_GetEscrow_ParamsError_LeavesSealGraceZero(t *testing.T) {
@@ -103,6 +105,7 @@ func TestChainBridge_GetEscrow_ParamsError_LeavesSealGraceZero(t *testing.T) {
 	info, err := cb.GetEscrow("1")
 	require.NoError(t, err)
 	assert.Equal(t, uint32(0), info.SealGraceNonces)
+	assert.Equal(t, uint32(0), info.InferenceClearGraceSeconds)
 }
 
 func TestChainBridge_GetEscrow_DefaultSealGraceZero_UsesGroupFallback(t *testing.T) {
@@ -131,4 +134,5 @@ func TestChainBridge_GetEscrow_DefaultSealGraceZero_UsesGroupFallback(t *testing
 	info, err := cb.GetEscrow("1")
 	require.NoError(t, err)
 	assert.Equal(t, uint32(30), info.SealGraceNonces)
+	assert.Equal(t, inferenceTypes.DefaultDevshardInferenceClearGraceSeconds, info.InferenceClearGraceSeconds)
 }
--- a/decentralized-api/internal/devshard/manager_test.go
+++ b/decentralized-api/internal/devshard/manager_test.go
@@ -756,12 +756,13 @@ func TestCreateSession_FreezesSealGraceFromBridge(t *testing.T) {
 
 	br := &mockBridge{
 		escrow: &bridge.EscrowInfo{
-			EscrowID:        "escrow-1",
-			Amount:          100000,
-			CreatorAddress:  user.Address(),
-			Slots:           addresses,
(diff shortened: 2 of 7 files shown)


0xgonka (approved):
Focused, well-built fix. When a streaming client aborts, the proxy now keeps the upstream host SSE alive until `devshard_meta` arrives (bounded by `DEVSHARD_META_DRAIN_TIMEOUT_SECONDS`, default 10s) so `MsgFinishInference` makes it into the session's pending txs rather than being dropped mid-protocol. Tests cover both the cancel-during-attempt-1 path (attempt 2 skipped, timeout recorded) and the happy-path-while-cancelled (finish merged anyway).

**Must fix:**

None blocking.

**Should fix:**

- **PR body's "default 5s" disagrees with the code's `defaultMetaDrainTimeout = 10 * time.Second`.** `devshard/cmd/devshardctl/proxy.go:23`. Operators will read the body and tune around the wrong baseline. Commit `175f3e0c6` explicitly bumped it to 10s — update the body before merge.
- **`metaDrainTimeout` is a package-init singleton** (`devshard/cmd/devshardctl/proxy.go:27`: `var metaDrainTimeout = metaDrainTimeoutFromEnv()`). Setting `DEVSHARD_META_DRAIN_TIMEOUT_SECONDS` after process start is silently ignored, and the unit tests intentionally call the parser function directly to work around this. Document the requires-restart semantic in `devshard/docs/proxy.md`, or read the env var inside `sendAndProcess` instead — runtime-config cost is negligible per-inference. Not blocking, but it's the kind of thing operators discover only by adjusting tuning during an outage.
- **`46fd15a05 "uncommitted"`** — sloppy commit message; squash or amend before merge.

**Nits:**

- **Skip-attempt-2 trades some recoverable inferences for fewer no-op host requests.** `devshard/cmd/devshardctl/proxy.go:258-272`. When a client disconnects after attempt 1 fails but before attempt 2, the proxy now goes straight to `handleTimeout` instead of retrying. This means the executor doesn't get its second chance even though it might have succeeded. The comment captures the rationale ("no point streaming a reset to nobody") and on balance it's the right call, but the network sees a slightly higher `MsgTimeoutInference` rate for streaming clients that abort. Worth a one-liner in `devshard/docs/proxy.md` so the metrics-watching team knows where the new timeout rate comes from.
- **`time.After(metaDrainTimeout)` inside the goroutine at `proxy.go:299-307`** creates a Timer that isn't `Stop()`'d when `sendCtx.Done()` wins the outer select. Not a real leak (the runtime cleans up after the duration elapses), but at 10s timeouts under high concurrency it's still a small steady-state churn. `time.NewTimer` + `defer t.Stop()` is the idiomatic fix.

**Looks correct:**

- **Decoupling upstream from `r.Context()` is exactly right for protocol completion.** `handleStreaming` and `handleNonStreaming` pass `context.Background()` into `runInference`. The host SSE drain and `ProcessResponse` run to completion regardless of whether the client is still listening. The session's mempool merge happens; the next user diff sees the right state.
- **`proxyWriter` pretend-success (`return len(p), nil` at `proxy.go:402`) is necessary for `ProcessResponse` to keep parsing.** Without this, the SSE parser would see a write error and bail before `devshard_meta`. The trade-off (callers think bytes landed but they didn't) is contained because `proxyWriter`'s only consumer is the response pipeline, not delivery accounting.
- **`cancelFlag` is nil-safe (`Gone()`, `Done()`, `Trigger()` all guard on `cf == nil`).** Lets the test helpers and the non-streaming path pass `nil` without ceremony. `sync.Once` makes `Trigger()` idempotent; the tests at `proxy_test.go:TestCancelFlag_NilSafeAndOneShot` lock in both properties.
- **`watchClientCancel` goroutine exits cleanly.** Blocks on `<-r.Context().Done()` which Go's `net/http` cancels when `ServeHTTP` returns — so even if the client never disconnects, the goroutine ends when the handler finishes. No leak per request.
- **`sendAndProcess`'s context-derivation pattern (`sendCtx, cancel := context.WithCancel(ctx)`; `defer cancel()`) ensures the timeout-watcher goroutine always exits** via either the `flag.Done() + metaDrainTimeout` branch or the `sendCtx.Done()` (from defer) branch.
- **Skip-attempt-2 invariant verified by test.** `TestRunInference_ClientGoneSkipsRetry` checks that `devshard_stream_reset` was NOT written (which would have come from attempt 2) AND that `MsgTimeoutInference` was still recorded. That's the exact contract.
- **`inference_clear_grace_seconds` added to `statusSessionConfig`** is a purely additive JSON field (`proxy.go:543`, `proxy.go:662`); no client breaks. The corresponding test `TestHandleStatus_ExposesSealGraceConfig` covers both the seal-grace and the new field together, which is the right granularity.

**Bottom line:** Land it. The two should-fixes (PR-body number, init-time singleton) are documentation-tier, and the design (decouple upstream context + pretend-success writer + bounded drain) is the right one.

PR #1238, opened 2026-05-23
`devshard` Finish gossip on missed diff

a-kuprin:
## Summary

When the executor host produces `MsgFinishInference` but the user sequencer does not include it in a signed diff for several nonces (e.g. devshardctl client disconnect before `devshard_meta`, proxy crash, or other stall), other hosts never learn the Finish and the session can wedge.

This change adds **recovery gossip**: after a grace period, the executor re-broadcasts its locally proposed Finish via tx gossip so peers can absorb it and return it to the user through normal mempool paths.

## Behavior

- **Grace:** `finishGossipGraceRotations × len(group)` nonces (default **2 rotations** → `2 × groupSize`). With round-robin executor selection (`nonce % len(group)`), that gives the user two natural passes to pick up the Finish from the executor’s `devshard_meta` tail before P2P recovery.
- **Trigger:** On each `HandleRequest` that applied diffs, and after `ApplyCatchUpDiffs`, the host checks the mempool for stale local Finishes and calls `gossip.BroadcastTxs` (async, hash-deduped).
- **Scope:** Only **locally proposed** `MsgFinishInference` (`ProposedAt > 0`). Peer-imported mempool entries (`ProposedAt == 0`) are not re-broadcast.

## Code touchpoints

| Area | Change |
|------|--------|
| `devshard/host/mempool.go` | `StaleFinishes(currentNonce, grace)` |
| `devshard/host/host.go` | `collectStaleFinishesLocked`, recovery broadcast in `HandleRequest` / `ApplyCatchUpDiffs` |
| `devshard/host/*_test.go` | Unit + integration tests for trigger, no-trigger when included, no amplification of peer imports |

## Test plan

- [ ] `cd devshard/host && go test -run FinishGossip -count=1 .`
- [ ] `cd devshard/host && go test -count=1 .`
- [ ] `cd devshard && go test ./...` (CI **Build and Test Devshard**)

### DIFF
--- a/decentralized-api/internal/server/public/post_chat_handler_interruption_test.go
+++ b/decentralized-api/internal/server/public/post_chat_handler_interruption_test.go
@@ -113,6 +113,7 @@ type interruptionTestSuite struct {
 	mockRecorder    *cosmosclient.MockCosmosMessageClient
 	mockQueryClient *mockInterruptionQueryClient
 	mockMLServer    *httptest.Server
+	mockClientFactory *mlnodeclient.MockClientFactory
 	server          *Server
 	configManager   *apiconfig.ConfigManager
 	nodeBroker      *broker.Broker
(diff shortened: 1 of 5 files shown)


0xgonka (approved):
Adds peer-to-peer recovery gossip when the user sequencer drops a locally-proposed `MsgFinishInference` — after `finishGossipGraceRotations × len(group)` nonces (default 2 rotations) without inclusion, the executor re-broadcasts via tx gossip so peers can absorb and re-surface the Finish on their own mempool returns. **Clean, well-tested, no blockers.** Anti-amplification is the most important correctness property and it's both implemented and tested.

**PR state:**
- CI: pass (5/5: Build and Test API Wrapper, Devshard, Versioned, chain, build)
- Merge: clean (`mergeable=true`, `mergeStateStatus=CLEAN`)
- Prior reviews: none on GitHub. Latest commit `d870c6fc6` is from 2026-05-28; no GitHub-side review activity since.
- New activity: N/A — head SHA unchanged since prior local analysis.

**Must fix:**

None.

**Should fix:**

- **Each `HandleRequest` cycle that finds *any* stale Finish spawns a fresh `broadcastTxsBestEffort` goroutine, even when the gossip layer's permanent dedup (`devshard/gossip/gossip.go BroadcastTxs:broadcastedTxs` map) will immediately no-op the call.** `devshard/host/host.go:404-409` (HandleRequest) and `devshard/host/host.go:706-708` (ApplyCatchUpDiffs). For a wedged session (user permanently disconnected), every subsequent diff cycle indefinitely spawns one goroutine that takes the gossip mutex, checks the dedup map, and exits. Microseconds per call, but it's steady-state churn for the lifetime of the session. Either (a) track "already broadcast" on the host side and skip the spawn entirely, or (b) expose `gossip.HasBroadcast(txHash)` and gate `collectStaleFinishesLocked` on it. Not a blocker — the steady-state cost is negligible — but worth a one-liner before merge.

**Nits:**

- **`ApplyCatchUpDiffs` lost its `defer h.mu.Unlock()` in favor of a manual unlock at `devshard/host/host.go:701`.** Correct today because the only path between `Lock()` and `Unlock()` is the diff-apply loop + `collectStaleFinishesLocked`, neither of which `return`s early. But if anyone later adds an error-handling early-return in the diff-apply loop, the lock is held forever. A wrapped helper (`func collectStaleAfterCatchupLocked(...) []*types.DevshardTx { defer h.mu.Unlock(); ... }`) preserves the manual-control intent without leaving the bare-unlock footgun.
- **The mempool dedup keyed by `ProposedAt > 0` is a single sentinel value (`0` = peer-imported, anything else = local proposal nonce).** Works, but the field's dual purpose isn't obvious from its name. A typed wrapper (`Origin: OriginLocal | OriginPeer` plus `ProposedAt uint64`) would document intent at the type level. Pre-existing pattern in `mempool.go`, not introduced here.

**Looks correct:**

- **Anti-amplification via `ProposedAt > 0` filter is the most important property and it's both implemented (`devshard/host/mempool.go StaleFinishes:88-90`) and explicitly tested.** `TestHost_FinishGossipRecovery_PeerImportedFinishNotAmplified` injects a peer-imported Finish (sentinel `ProposedAt=0`), advances well past `grace+2`, and asserts zero broadcasts. Without this filter the network would amplify exponentially as each receiver re-broadcasts.
- **Grace-boundary test is rigorous, not just "advance many nonces and check it fired."** `TestHost_FinishGossipRecovery_TriggersOnMissedDiff` asserts (a) no broadcast at `currentNonce == ProposedAt + grace` (strict-less-than), and (b) broadcast triggers at the next nonce. Pins both sides of the threshold.
- **Lock discipline is right: collect under `h.mu`, broadcast in a separate goroutine off-lock.** The slice of `*types.DevshardTx` pointers is built inside `Mempool.StaleFinishes`'s own `m.mu`, and the goroutine that receives the slice does not need either mutex. Tx values are effectively immutable after insertion, so the cross-goroutine read is safe.
- **`grace = finishGossipGraceRotations * len(group)` is the right binding** — under round-robin executor selection (`nonce % len(group)`), every `len(group)` nonces brings the user back into direct contact with the original executor's `devshard_meta` tail. Two full rotations = two natural pickup chances before falling back to P2P recovery. The constant is documented at `devshard/host/host.go:25-33` including the tuning trade-off.
- **`broadcastTxsBestEffort(context.Background())` is the right context choice** for fire-and-forget recovery. Binding to the request context would mean a client disconnect mid-`HandleRequest` cancels the recovery broadcast for an unrelated user's pending Finish.
- **`diffsApplied` flag at `devshard/host/host.go:367,372` gates the staleness check correctly** — no point scanning the mempool when no new state was applied; existing entries' staleness hasn't changed.
- **Test scaffolding (`recordingPeer`, `awaitTxsCall`, `assertNoTxsCallFor`) does poll-with-deadline instead of fixed sleeps.** `assertNoTxsCallFor` fails fast on the first observed call rather than waiting out a full sleep — the right shape for "assert no broadcast" tests and keeps the suite quick.

**Bottom line:** Ship it. The should-fix is a steady-state-cleanliness improvement (skip the no-op goroutine when gossip's dedup will swallow the call) that I'd ask for before merge but isn't blocking.

PR #1240, opened 2026-05-24
fix(devshard): reassemble fragmented SSE events for raceWriter content classification

qdanik:
## Summary
- `raceWriter.Write` classified each call in isolation, so when transport delivered an SSE event split across `Write` calls (TLS fragmentation, proxy flush mid-event, slow connections), `sseChunkContentSource` / `sseChunkErrorDetails` saw only a fragment and the response was misclassified as empty — triggering false `empty_stream` quarantine.
- Added `inflight.classifyPartial` + `raceWriter.takeParseable(p)` so classification always parses a buffer ending on `\n` and stashes the trailing partial for the next `Write`. Forwarding to the client is unchanged — only the bytes fed to the classifier are reassembled.
- Memory is bounded by two caps: a generous **per-attempt soft cap of 1 MiB** (orders of magnitude above any plausible single SSE event) and a **global hard cap of 100 MiB** across all in-flight attempts, tracked via `atomic.Int64`. When either cap would be exceeded, the per-attempt buffer is dropped and classification falls back to the current `Write`. Buffer memory is released via `defer` in `monitorInflight` — the single per-attempt cleanup point — so the global gauge stays balanced.
- Added embedded SSE fixtures + a regression suite that exercises the full `raceWriter` pipeline across a wide chunk-size range, plus a cap-drop test that verifies the global counter returns to baseline.

## Why
On the route we kept seeing `empty_stream` events for responses that actually carried full content. Root cause: classifier ran per `Write`, but SSE event boundaries don't align with `Write` boundaries. Validated against real captured vLLM responses (not checked in — too noisy for the diff): before the fix 100% misclassified empty at sub-KiB chunks; after the fix 0% empty at every chunk size from 1 byte through 8 KiB.

The two-tier cap design replaces an earlier single per-attempt cap of 256 KiB. The new shape addresses two opposite concerns raised in review: per-attempt 1 MiB is generous enough that no legitimate SSE event from vLLM-class servers will hit it (real events are sub-KiB to single-digit KiB), while the global 100 MiB ceiling is the hard guarantee that no number of misbehaving upstreams can exhaust process memory.

## Tests
Embedded fixtures in [sse_fixtures_test.go](devshard/cmd/devshardctl/sse_fixtures_test.go) mirror the vLLM `chat.completion.chunk` shape for both `delta.content` and `delta.tool_calls` paths:

- `TestSseBlobClassifierNotEmpty` — sanity-checks each fixture as a single blob (2/2 PASS)
- `TestSseRaceWriterAllChunkSizes` — feeds each fixture through the writer at chunk sizes `{1, 64, 256, 1024, 4096, 8192}`, asserts `isEmptyStreamAttempt == false` and `contentSource` matches expected (12/12 PASS)
- `TestSseRaceWriterRandomChunking` — deterministic xorshift random chunk sizes 1–64 bytes (2/2 PASS)
- `TestSseRaceWriterClassifyCapDrops` — writes `maxClassifyPartial + 1` newline-less bytes, asserts buffer is dropped and global counter returns to baseline (1/1 PASS)

`go test ./...` under `devshard/` — all packages green.

## Test plan
- [x] All embedded-fixture tests pass at every chunk size 1 B → 8 KiB
- [x] Per-attempt cap drops the buffer; global gauge stays balanced
- [x] `go test ./...` in `devshard/` green

### DIFF
--- a/devshard/cmd/devshardctl/redundancy.go
+++ b/devshard/cmd/devshardctl/redundancy.go
@@ -873,6 +873,7 @@ type inflight struct {
 	// classifyPartial keeps the tail after the last '\n' from prior Writes;
 	// used to reassemble fragmented SSE events for the classifier.
 	classifyPartial []byte
+	classifyCapLog  sync.Once
 
 	// contentSource labels the field that produced the first content event
 	// ("delta.content", "delta.reasoning_content", "delta.tool_calls", or the
@@ -1280,13 +1281,33 @@ func (rw *raceWriter) takeParseable(p []byte) []byte {
 	return parseable
 }
 
+// logClassifyDrop reports the first reassembly-buffer drop for this attempt.
+// Capped at once per attempt so a persistently newline-less or oversized
+// transport can't spam the log.
+func (rw *raceWriter) logClassifyDrop(reason string) {
+	rw.inf.classifyCapLog.Do(func() {
+		ctx := context.Background()
+		if rw.group != nil {
+			ctx = rw.group.logCtx
+		}
+		logInferenceStage(ctx, rw.inf.escrowID, rw.nonce, "classify_buffer_dropped",
+			"host", rw.inf.hostID,
+			"reason", reason,
+			"buffered", len(rw.inf.classifyPartial),
+			"global_bytes", classifyPartialBytes.Load(),
+		)
+	})
+}
+
 // growClassify appends tail to classifyPartial if both caps still fit.
 func (rw *raceWriter) growClassify(tail []byte) bool {
 	if len(rw.inf.classifyPartial)+len(tail) > maxClassifyPartial {
+		rw.logClassifyDrop("per_attempt_cap")
 		return false
 	}
 	if classifyPartialBytes.Add(int64(len(tail))) > maxClassifyPartialGlobal {
 		classifyPartialBytes.Add(-int64(len(tail)))
+		rw.logClassifyDrop("global_cap")
 		return false
 	}
 	rw.inf.classifyPartial = append(rw.inf.classifyPartial, tail...)
@@ -1296,11 +1317,13 @@ func (rw *raceWriter) growClassify(tail []byte) bool {
 // replaceClassify swaps classifyPartial for buf, adjusting global accounting.
 func (rw *raceWriter) replaceClassify(buf []byte) bool {
 	if len(buf) > maxClassifyPartial {
+		rw.logClassifyDrop("per_attempt_cap")
 		return false
 	}
 	delta := int64(len(buf) - len(rw.inf.classifyPartial))
 	if delta > 0 && classifyPartialBytes.Add(delta) > maxClassifyPartialGlobal {
 		classifyPartialBytes.Add(-delta)
+		rw.logClassifyDrop("global_cap")
 		return false
 	}
 	if delta < 0 {
--- a/devshard/cmd/devshardctl/redundancy.go
+++ b/devshard/cmd/devshardctl/redundancy.go
@@ -1387,7 +1387,7 @@ func (rw *raceWriter) Write(p []byte) (int, error) {
 					rw.inf.errorCode = details.Code
 					rw.inf.errorType = details.Type
 					rw.inf.errorMessage = details.Message
-					rw.inf.errorBodySample = append(rw.inf.errorBodySample, p...)
+					rw.inf.errorBodySample = append(rw.inf.errorBodySample, parseable...)
 				}
 			}
 		}
--- a/devshard/cmd/devshardctl/redundancy.go
+++ b/devshard/cmd/devshardctl/redundancy.go
@@ -1281,6 +1281,47 @@ func (rw *raceWriter) takeParseable(p []byte) []byte {
 	return parseable
 }
 
+// classifyParseable records the first content/non-retriable-error signal for
+// this attempt, returning whether the buffer carried either.
+func (rw *raceWriter) classifyParseable(parseable []byte) (hasContent, hasError bool) {
+	if len(parseable) == 0 {
+		return false, false
+	}
+	if src, ok := sseChunkContentSource(parseable); ok {
+		hasContent = true
+		if rw.inf.contentSource == "" {
+			rw.inf.contentSource = src
+		}
+	} else if details, ok := sseChunkErrorDetails(parseable); ok {
+		src := "error"
+		if details.Type != "" {
+			src = "error." + details.Type
+		}
+		hasError = !isRetriableCapabilityErrorMessage(details.Message)
+		if rw.inf.errorSource == "" {
+			rw.inf.errorSource = src
+			rw.inf.errorCode = details.Code
+			rw.inf.errorType = details.Type
+			rw.inf.errorMessage = details.Message
+			rw.inf.errorBodySample = append(rw.inf.errorBodySample, parseable...)
+		}
+	}
+	return hasContent, hasError
+}
+
+// flushClassify classifies a newline-less final SSE event left in classifyPartial
+// once the stream ends, so a truncated tail isn't misread as empty_stream.
+func (rw *raceWriter) flushClassify() {
+	if rw.inf.probe || len(rw.inf.classifyPartial) == 0 {
+		return
+	}
+	hasContent, hasError := rw.classifyParseable(rw.inf.classifyPartial)
+	rw.dropClassify()
+	if hasContent || hasError {
+		rw.inf.contentChunks.Add(1)
+	}
+}
+
 // logClassifyDrop reports the first reassembly-buffer drop for this attempt.
 // Capped at once per attempt so a persistently newline-less or oversized
 // transport can't spam the log.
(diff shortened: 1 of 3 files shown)


devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +999,70 @@ type raceWriter struct {
...
+	}
+
+	// Defensive: malformed stream growing without newlines — drop partial.
+	if len(rw.inf.classifyPartial)+len(p) > maxClassifyPartial {
+		rw.inf.classifyPartial = rw.inf.classifyPartial[:0]
+		lastNL := bytes.LastIndexByte(p, '\n')
+		if lastNL == -1 {
+			return nil
+		}
+		if lastNL+1 < len(p) {
```
akup: We've just got `lastNL` from `p`
Why it can ever be `>= len(p)`
Do we need this check?

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +999,70 @@ type raceWriter struct {
 	inf   *inflight
 }
 
+// maxClassifyPartial caps the SSE-reassembly buffer to defend against
+// malformed transports that stream indefinitely without newlines. If we
+// hit this cap the accumulated partial is dropped and classification falls
+// back to processing only the current Write.
+const maxClassifyPartial = 256 * 1024
```
akup: I'm not sure.
Is 256Kb enough for response without new lines?

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +999,70 @@ type raceWriter struct {
...
+		}
+		if lastNL+1 < len(p) {
+			rw.inf.classifyPartial = append(rw.inf.classifyPartial[:0], p[lastNL+1:]...)
+		}
+		return p[: lastNL+1]
+	}
+
+	// Combine prior partial with new bytes; classify whole view.
+	combined := append(rw.inf.classifyPartial, p...)
+	lastNL := bytes.LastIndexByte(combined, '\n')
```
akup: I think it should be better to first try to find '\n' at `p`

```go
lastNLatP := bytes.LastIndexByte(p, '\n')
if lastNLatP == -1 {
    rw.inf.classifyPartial = combined
    return nil
} else {
    lastNL = lastNLatP + len(rw.inf.classifyPartial)
}
```

Currently if there is no '\n' symbol we iterate over all `combined` that grows but we need to check for '\n' only at appendable fresh part

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +999,70 @@ type raceWriter struct {
...
+		rw.inf.classifyPartial = combined
+		return nil
+	}
+	// Copy out the trailing partial — combined shares backing with the prior
+	// classifyPartial, which we're about to replace.
+	if lastNL+1 < len(combined) {
+		tail := make([]byte, len(combined)-lastNL-1)
+		copy(tail, combined[lastNL+1:])
+		rw.inf.classifyPartial = tail
+	} else {
```
akup, line 1313: I can't understand how we can get into this else.
`lastNL` is looked at `combined`
How it can be > len(combined), is this the real case?

Or what do I miss?

qdanik: The else fires when lastNL+1 == len(combined) — i.e. the buffer ends exactly on \n

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +994,81 @@ type raceWriter struct {
...
+		return false
+	}
+	rw.inf.classifyPartial = append(rw.inf.classifyPartial, tail...)
+	return true
+}
+
+// replaceClassify swaps classifyPartial for buf, adjusting global accounting.
+func (rw *raceWriter) replaceClassify(buf []byte) bool {
+	if len(buf) > maxClassifyPartial {
+		return false
```
a-kuprin, line 1408: Let's add error logging here

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +994,81 @@ type raceWriter struct {
...
+	}
+	delta := int64(len(buf) - len(rw.inf.classifyPartial))
+	if delta > 0 && classifyPartialBytes.Add(delta) > maxClassifyPartialGlobal {
+		classifyPartialBytes.Add(-delta)
+		return false
```
a-kuprin: Let's add error logging here

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +994,81 @@ type raceWriter struct {
...
+}
+
+// growClassify appends tail to classifyPartial if both caps still fit.
+func (rw *raceWriter) growClassify(tail []byte) bool {
+	if len(rw.inf.classifyPartial)+len(tail) > maxClassifyPartial {
+		return false
+	}
+	if classifyPartialBytes.Add(int64(len(tail))) > maxClassifyPartialGlobal {
+		classifyPartialBytes.Add(-int64(len(tail)))
+		return false
```
a-kuprin, line 1398: Let's add error logging here

devshard/cmd/devshardctl/redundancy.go
```diff
@@ -990,6 +994,81 @@ type raceWriter struct {
...
+	if !rw.replaceClassify(p[lastNL+1:]) {
+		rw.dropClassify()
+	}
+	return parseable
+}
+
+// growClassify appends tail to classifyPartial if both caps still fit.
+func (rw *raceWriter) growClassify(tail []byte) bool {
+	if len(rw.inf.classifyPartial)+len(tail) > maxClassifyPartial {
+		return false
```
a-kuprin, line 1394: Let's add error logging here

devshard/cmd/devshardctl/redundancy.go
qdanik: done

devshard/cmd/devshardctl/redundancy.go
qdanik: done

devshard/cmd/devshardctl/redundancy.go
qdanik: done

devshard/cmd/devshardctl/redundancy.go
qdanik: done

a-kuprin:
There are also some AI review neats. I've checked all concerns are valid, and could be added for cleaner codebase:

## Must fix
- **Inaccurate errorBodySample in inflight struct** (sources: devshard-transport-ops, ./devshard/cmd/devshardctl/redundancy.go:1139-1140)
  When an SSE event is classified as an error, the `errorBodySample` field currently appends the current chunk of bytes (`p`) instead of the reassembled parseable buffer. If the error event was fragmented across multiple `Write` calls, `p` would only contain a fragment, leading to an incomplete or misleading `errorBodySample`, which degrades diagnostic quality.
  **Fix:** Change `rw.inf.errorBodySample = append(rw.inf.errorBodySample, p...)` to `rw.inf.errorBodySample = append(rw.inf.errorBodySample, parseable...)`.
  
## Should fix
- **Add t.Parallel() to subtests for robustness** (sources: test-mechanics, ./devshard/cmd/devshardctl/sse_fixtures_test.go:52)
  The tests for `raceWriter` deal with a component designed for handling potentially concurrent network writes and interact with a global `atomic.Int64` counter. While `t.Run` is used, the subtests themselves do not call `t.Parallel()`. Explicitly marking relevant subtests in `TestSseRaceWriterAllChunkSizes` and `TestSseRaceWriterRandomChunking` with `t.Parallel()` would validate their behavior under parallel execution and help catch any subtle race conditions, ensuring better test robustness.

## Consider
- **Redundant custom itoa implementation; use `strconv.Itoa`** (sources: test-mechanics, style, ./devshard/cmd/devshardctl/sse_fixtures_test.go:124-144)
  The file includes a custom implementation of `itoa` (lines 124-144). The Go standard library provides `strconv.Itoa` for this exact purpose, which is more idiomatic, improves readability, and avoids maintaining duplicate functionality. Using the standard library function is generally more optimized and less prone to errors.
  **Fix:** Remove the custom `itoa` function and use `strconv.Itoa(sz)` where integer-to-string conversion is needed.
- **Custom PRNG implementation** (sources: style, ./devshard/cmd/devshardctl/sse_fixtures_test.go:111)
  The `pseudoRandSeed` function implements a custom xorshift32 PRNG. For deterministic testing, it is more idiomatic to use the standard library's `math/rand/v2` (or `math/rand`) package with a fixed seed. This approach makes the code more readable for other Go developers and leverages a vetted, well-understood implementation.
  **Fix:** Use `rand.New(rand.NewPCG(seed, 0))` from `math/rand/v2`.

a-kuprin (approved):

a-kuprin:
@qdanik 

## The following attack is possible

A hostile participant can keep many streams open and send bytes **without newlines**, holding up to ~1 MiB per attempt. ~100 such concurrent inflights fill the shared pool. This will lead **new buffering fails and partial data is dropped**. Under normal gateway concurrency this is realistic for a participant that receives a steady share of traffic.

## Consider

**Per-participant buffer cap.** The right isolation unit is withheld classify buffer per participant, not concurrent request count. A request cap is coarse and overlaps existing gateway limits; the attack is specifically about holding unassembled bytes. A per-participant ceiling (e.g. 10 MiB) lets a bad actor only harm itself while a global cap remains the process-wide backstop.

**Environment variables.** These limits are machine-scale knobs (memory, concurrency), not governance policy — configure them via env at startup, not `gateway.db`. Defaults can stay conservative (100 MiB global); larger hosts can raise global and per-participant caps together.

**Graceful degradation.** On cap failure, falling back to classifying the raw write chunk (best-effort) may be better than dropping classification entirely.

a-kuprin (changes requested):
Possible attack was found, should mitigate it before merging

a-kuprin (approved):

PR #1248, opened 2026-05-26
fix: validate Ethereum destination on bridge withdrawal

ouicate:
## Summary

This fixes a validation gap where `MsgRequestBridgeWithdrawal` accepted malformed `DestinationAddress` values at `ValidateBasic` and only failed later during BLS signature preparation. Users paid gas for transactions that could never succeed, while `MsgRequestBridgeMint` already rejected the same malformed addresses up front.

## Root Cause

`MsgRequestBridgeWithdrawal.ValidateBasic` checked `DestinationAddress` for non-empty length only. `DestinationBridgeAddress` was validated with `isValidEthereumAddress`, but the user's ultimate Ethereum recipient was not. The handler called `ethereumAddressToBytes(recipient)` in `prepareBridgeWithdrawalSignatureData`, which failed after the message was already included in a block.

## Fix

- Apply `isValidEthereumAddress` to `DestinationAddress` in `MsgRequestBridgeWithdrawal.ValidateBasic`, matching `MsgRequestBridgeMint`.
- Add unit tests rejecting short/invalid destination addresses.

Malformed withdrawal destinations are rejected before broadcast, eliminating wasted gas and inconsistent UX between symmetric bridge mint and withdrawal flows. Execution-time failures from invalid recipient encoding on the withdrawal path are no longer reachable through `ValidateBasic`-accepted messages.

## Test plan

- `go test ./x/inference/types/... -run 'Withdrawal|BridgeWithdrawal'`
- Submit `MsgRequestBridgeWithdrawal` with `DestinationAddress = "0x12345"` and confirm `ValidateBasic` error before execution

---

Supersedes #1197 (closed because the head fork was accidentally deleted; this branch is a clean rebase onto current `main`).


a-kuprin:
This is valid fix, targeted for user experience preventing users from mistakes in address.
Should be classified as Neat.

a-kuprin (approved):

ouicate:
Please take a look @GLiberman @DimaOrekhovPS

PR #1250, opened 2026-05-26
fix: claw back validator shares on invalidation

ouicate:
## Summary

This fixes a consensus-accounting bug where validator work shares credited by `MsgValidation{passed=true}` could remain claimable even if the same inference was later invalidated. In that case, only the executor was charged for the refund, while validator `CoinBalance` stayed intact and could still be claimed as `WorkCoins`, creating an unbacked, repeatable drain on the inference module account.

## Root Cause

`MsgValidation` allowed normal, non-revalidation validations to run after the inference had already left `FINISHED`, including `VALIDATED` and `VOTING`. Separately, `refundInvalidatedInference` debited only the executor by `ActualCost` and never reversed the `ShareWork` distribution already recorded for validators in `inference.ValidatedBy`.

## Fix

- Restrict normal `MsgValidation` processing to `FINISHED` inferences only. `INVALIDATED` remains idempotent, `STARTED` remains an error, and duplicate validations still return `ErrDuplicateValidation`.
- Prevent already `INVALIDATED` inferences from recording validation credit.
- On same-epoch invalidation, recompute the exact executor plus validator work distribution using the existing `ShareWork` algorithm and claw back every credited worker's share before settlement can make stale positive balances claimable.
- Preflight all validated-worker participant lookups before issuing the refund, then apply clawbacks in deterministic worker order.
- Make validation share accounting fail closed if `ShareWork` produces invalid or imbalanced adjustments.

The exploit required two conditions: a normal validation could move a `VALIDATED` inference into invalidation voting, and invalidation left prior validator shares untouched. This PR removes the invalid state transition and also fixes the accounting invariant by reversing all previously assigned shares when an invalidation refund is issued.

## Test plan

- `go test ./x/inference/keeper/... -run 'Invalidate|Validation|Clawback'`

---

Supersedes #1131 (closed because the head fork was accidentally deleted; this branch is a clean rebase onto current `main`, including a small conflict resolution against an upstream refactor that now passes `inference.RequestedBy` directly to `IssueRefund`).


ouicate:
Superseded by #1373, which keeps this validator-share clawback **and** adds a fix for the root cause it doesn't cover: an uncapped `ActualCost` on the start-then-finish path can exceed the escrow collected, so refunds/work-shares/clawbacks could move more out of the module pool than was ever escrowed. #1373 caps `ActualCost <= EscrowAmount` at every consumption site (plus a crisis invariant). Closing this in favor of the stronger combined PR.

PR #1252, opened 2026-05-26
fix: exclude INVALIDATED inferences from ClaimRewards missed-validation check

ouicate:
`MsgValidation` grants validation credit on `INVALIDATED` inferences, allowing validators to reduce their missed-validation count without performing actual work. This bypasses the `ClaimRewards` missed-validation penalty.

Fix: filter `INVALIDATED` inferences from both sides of the missed-validation computation in `ClaimRewards`:

1. `getValidatedInferences`: exclude `INVALIDATED` inference IDs from the `wasValidated` map, directly blocking the exploit vector.
2. `getMustBeValidatedInferences`: exclude `INVALIDATED` inferences from the `mustBeValidated` list, so honest validators are not penalized for skipping invalidated inferences.

`MsgValidation` still records credit in `EpochGroupValidations` (design decision: validations can be submitted during the whole epoch), but that credit no longer affects reward calculations for invalidated inferences.

---

Supersedes #829 (closed because the head fork was accidentally deleted).


a-kuprin:
I think we shouldn't work on legacy validation workflow.
devshard currently is the only way to pass inferences to Gonka

PR #1253, opened 2026-05-26
fix: stop stale PoC validation promptly

ouicate:
## Summary

- Cancels off-chain PoC validation when the validation phase ends or the PoC stage changes, and propagates cancellation into proof fetch / ML-node validation work.
- Keeps retry requeue handling cancellable and avoids treating exhausted transient failures as invalid participants.

This continues the work in the original `fix: add global deadline and mitigate PoC validation timeout attack` #827, with a single narrow commit and no unrelated changes.

## Why this matters

Without cancellation, validation workers can keep running past the acceptance window when the PoC stage advances. The off-chain side then submits stale results (or exhausts retries on participants that simply moved on), feeding the missed-validation accounting noise that the on-chain side cannot distinguish from real misses.

## Test plan

- `go test ./decentralized-api/poc/...`
- Manual: trigger a stage change mid-validation, confirm the in-flight ML-node fetch is cancelled and the worker exits cleanly rather than marking the participant invalid.

---

Supersedes #1124 (closed because the head fork was accidentally deleted).


ouicate:
@patimen, earlier you stated:

> "This still looks like the right idea to me. The underlying issue appears to still be live on current main, and this looks like a good narrow vehicle for fixing it. We should review the details carefully and get this into v0.2.14."

That was in reference to PR #1124.

Given that assessment, I think it makes sense to include this fix in v0.2.14

patimen (commented):
Nicely scoped, useful, correct.

patimen (approved):
Nicely scoped, useful, correct.

PR #1255, opened 2026-05-26
fix(inference): settle accounts before releasing matured unbonding collateral

ouicate:
## Summary
This fixes an ordering bug in the end-of-PoC-validation epoch transition where the collateral module's `AdvanceEpoch` (which releases matured unbonding entries) ran before `SettleAccounts` (which evaluates participant performance and can trigger collateral slashing). Any unbonding entry whose `completion_epoch` equaled the just-completed epoch was paid out and removed before slashing could reach it, allowing a participant to time their unbonding to mature exactly at a slash epoch and shield that collateral from the slash.

## Root Cause
`onEndOfPoCValidationStage` in `inference-chain/x/inference/module/module.go` invoked `am.keeper.GetCollateralKeeper().AdvanceEpoch(ctx, effectiveEpoch.Index)` before `am.keeper.SettleAccounts(ctx, effectiveEpoch.Index, previousEpochIndex)`. `AdvanceEpoch` calls `ProcessUnbondingQueue(completedEpoch)`, which iterates every entry with `completionEpoch == completedEpoch`, sends the coins back to the participant via `SendCoinsFromModuleToAccount`, and then removes the entry from the indexed map. `SettleAccounts` then walks participants and calls `SetParticipant`, which goes through `UpdateParticipantStatus` and, on a transition to `INACTIVE`/`INVALID`, invokes `SlashForDowntime`/`SlashForInvalidStatus`. The collateral `Slash` function slashes both active collateral and the participant's unbonding entries by iterating `GetUnbondingByParticipant`, but because the matured entries for the completed epoch had already been released and removed, the slash on those amounts was unreachable. A rational participant expecting downtime or planning misbehavior could therefore unbond `UnbondingPeriodEpochs` epochs in advance so that the unbonding matured at the slash epoch, removing up to 100% of the would-be-slashed collateral from the slash base.

## Fix
- Reorder `onEndOfPoCValidationStage` so `SettleAccounts` runs before the collateral module's `AdvanceEpoch`. Settlement-driven slashing now happens while unbonding entries for the completed epoch are still in the queue and reachable by `Slash`. `AdvanceEpoch` then processes whatever remains and releases only the post-slash residual to the participant. The streamvesting `AdvanceEpoch` continues to run alongside collateral `AdvanceEpoch` after settlement.
- Add `TestEpochProcessing_SlashBeforeUnbondingRelease` in the collateral keeper tests, which exercises the post-fix ordering directly: it sets up an active + unbonding balance with `completionEpoch == completedEpoch`, calls `Slash` (50%), asserts the unbonding entry is reduced (not removed) and that the gov transfer covers both active and unbonding portions, then calls `AdvanceEpoch(completedEpoch)` and asserts the participant only receives the post-slash remainder.

With settlement now running first, any slashing performed via `SetParticipant -> UpdateParticipantStatus` for the just-completed epoch reaches the unbonding entries before the unbonding queue processor releases them. A participant can no longer use the predictable epoch timing to convert about-to-be-slashed collateral into a release that happens in the same epoch transition: by the time `ProcessUnbondingQueue` runs, the entry has already been debited proportionally, so only the slashed-down balance is paid out. The economic deterrent of `SlashFractionDowntime`/`SlashFractionInvalid` once again applies to the full active + unbonding base it was designed to cover.

### DIFF
--- a/inference-chain/x/collateral/keeper/epoch_processing_test.go
+++ b/inference-chain/x/collateral/keeper/epoch_processing_test.go
@@ -63,41 +63,21 @@ func (s *KeeperTestSuite) TestEpochProcessing_ProcessUnbondingQueue() {
 	s.Require().True(found, "future-dated unbonding entry should not be processed")
 }
 
-// TestEpochProcessing_SlashBeforeUnbondingRelease pins the ordering contract used
-// by the inference module's onEndOfPoCValidationStage: slashing for the just-
-// completed epoch must be applied before the collateral unbonding queue is
-// processed for that same epoch. If the unbonding release ran first, any entry
-// whose completion_epoch equals the completed epoch would be paid out and
-// removed before Slash could reach it, allowing a participant to time their
-// unbonding to mature exactly at the slash epoch and shield that collateral
-// from the slash.
-//
-// This test simulates the correct ordering (slash, then AdvanceEpoch) and
-// verifies that:
-//  1. The unbonding entry maturing at the completed epoch is slashed
-//     proportionally alongside the active collateral.
-//  2. The remaining (post-slash) unbonding amount is what gets released to
-//     the participant when AdvanceEpoch subsequently processes the queue.
 func (s *KeeperTestSuite) TestEpochProcessing_SlashBeforeUnbondingRelease() {
 	participantStr := sample.AccAddress()
 	participant, err := sdk.AccAddressFromBech32(participantStr)
 	s.Require().NoError(err)
 
-	// Active = 20, Unbonding = 80 (maturing at completedEpoch). Mirrors the
-	// attack scenario where the bulk of collateral is parked in the unbonding
-	// queue and timed to mature at the slash epoch.
 	activeAmount := int64(20)
 	unbondingAmount := int64(80)
 	activeCoin := sdk.NewInt64Coin(inftypes.BaseCoin, activeAmount)
 	unbondingCoin := sdk.NewInt64Coin(inftypes.BaseCoin, unbondingAmount)
 	completedEpoch := uint64(7)
+	slashFraction := math.LegacyNewDecWithPrec(50, 2)
 
 	s.Require().NoError(s.k.SetCollateral(s.ctx, participant, activeCoin))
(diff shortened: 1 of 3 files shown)


inference-chain/x/inference/module/module.go
```diff
@@ -588,8 +588,38 @@ func (am AppModule) onEndOfPoCValidationStage(ctx context.Context, blockHeight i
 		return
 	}
 
+	previousEpoch, found := am.keeper.GetPreviousEpoch(ctx)
+	previousEpochIndex := uint64(0)
+	if found {
+		previousEpochIndex = previousEpoch.Index
+	}
+
+	// SettleAccounts must run before the collateral module's AdvanceEpoch.
```
patimen: Dang... that's a heck of a comment here, can we tighten this or rely on commit? Big long comments aren't great in code.

inference-chain/x/collateral/keeper/epoch_processing_test.go
```diff
@@ -61,3 +62,86 @@ func (s *KeeperTestSuite) TestEpochProcessing_ProcessUnbondingQueue() {
 	_, found = s.k.GetUnbondingCollateral(s.ctx, futureParticipant, futureEpoch)
 	s.Require().True(found, "future-dated unbonding entry should not be processed")
 }
+
+// TestEpochProcessing_SlashBeforeUnbondingRelease pins the ordering contract used
```
patimen: Again, comments, not documentation.

patimen (changes requested):
First priority, we need that unit test fixed. It seems persistent and only in this branch.
Second, if you could be a bit more terse in the in-line comments... nobody is going to read a gigantic comment like that.

ouicate:
> First priority, we need that unit test fixed. It seems persistent and only in this branch.
> Second, if you could be a bit more terse in the in-line comments... nobody is going to read a gigantic comment like that.

Thanks - pushed `74b10ab`.

Fixed the unit test and shortened the inline comments.

patimen (approved):

PR #1258, opened 2026-05-26
`devshard` 0.2.13 optimizations: strict version naming, limit nonces, db optimizations and bug fix

### LINKED ISSUE #977: [P0] `devshards`: Limit amount of inferences
As described in https://github.com/gonka-ai/gonka/issues/914#issuecomment-4090483233, we want to limit the amount of inferences that can be done in a `devshard`, as a way of limiting the amount of damage a hijacked `devshard` can cause on the network.

For now, let's say each `devshard` can run up to 2k inferences.

Upon settlement, the protocol should verify "Missed inferences + Invalidated inferences" does not exceed 2k, for the whole group.

a-kuprin:
# Devshard 0.2.13 optimizations — PR brief

Related docs:

- [protocol-version.md](./protocol-version.md) — runtime vs state-root version
- [params-dataflow.md](./params-dataflow.md) — runtime config bind lanes
- [storage-design.md](./storage-design.md) — storage layout, migrations, and boot-time backend selection

---

## `37f1177c4` — Enforce `max_nonce` on hosts

**Title:** `feat(devshard): enforce max_nonce on hosts before settlement fails`

### What changed

Hosts read **`DevshardEscrowParams.max_nonce`** from the dapi config cache or
devshardd runtime long-poll and reject completion work in the active phase once
the session would exceed **`MaxActiveNonce`** (`max_nonce` minus a reserve for
finalization and on-chain settlement).

- `MaxNonceProvider` wired through **HostManager** on dapi and devshardd.
- Host tests and `devshard/types` helpers for the limit math.
- **Hybrid storage:** `PruneEpoch`, `pruneBefore`, and `Close` now return
  **`errors.Join`** for SQLite and Postgres failures so a SQLite error is not
  hidden when Postgres also fails.

### Why it matters

Settlement failures from exceeding chain nonce limits are caught **before**
hosts accept more diffs, instead of failing only at settlement time.

### Review focus

- `devshard/host/host.go`, `devshard/types/max_nonce.go`
- `decentralized-api/internal/devshard/max_nonce_provider.go`
- Hybrid prune/close error propagation (pre–thin-hybrid behavior)

---

## `c2c470940` — Strict version naming

**Title:** `Making versions naming strict`

### What changed

Separates concepts that were previously conflated under `DEVSHARD_LOG_PREFIX`:

| Concept | Where it lives | Purpose |
| --- | --- | --- |
| **Runtime / binary version** | `approved_versions` name, `DEVSHARD_BINARY_VERSION`, storage `CreateSessionParams.Version` | Which devshard binary may run; versiond sets oracle + build env vars |
| **State root & protocol version** | `types.DevshardStateRootAndProtocolVersion` (`"v2"`) | Hashed into state roots; on `MsgSettleDevshardEscrow.state_root_and_protocol_version` |
| **Legacy route tag** | `types.LegacyRouteSessionVersion` (`"v1"`) | `/v1/devshard` and embedded dapi hosts only |

- devshardd checks **oracle version** vs **build version** at startup.
- Settlement and hash tests use the protocol constant explicitly.
- Storage enforces **session version** at create/recover boundaries
  (`session_version.go`, conflicts on mismatch).
- New [protocol-version.md](./protocol-version.md); updates to [upgrade.md](./upgrade.md).

### Why it matters

Parallel `devshardd` versions can share Postgres; pinning **binary** vs **protocol**
tags prevents the wrong state-machine rules from attaching to live escrows and
keeps settlement verification aligned with hashed roots.

### Review focus

- `devshard/types/domain.go`, `devshard/paths.go` (`VersionForRoutePrefix`)
- `devshard/state/settlement.go`, `devshard/state/hash.go`
- `decentralized-api/cmd/devshardd/main.go` (`resolveRuntimeVersion`)

---

## `1cc9d2c6d` — Runtime config snapshot for session create

**Title:** `feat(devshard): route session config through dapi runtime snapshot`

### What changed

Session creation no longer relies on bridge defaults alone. **HostManager** binds
escrow config from chain fields plus a **frozen runtime snapshot** at create time:

- Fees frozen on create; **GetRuntimeConfig** extended with refusal/execution
  timeouts, `validation_rate`, `vote_threshold_factor`.
- **Grace, validation, vote threshold** fixed at session bind; **timeouts and
  availability** stay live via `RuntimeParamsProvider`.
- Removed nested `QueryParams` / bridge grace defaults; documented bind lanes
  A/B/C in [params-dataflow.md](./params-dataflow.md).
- Proto/nodemanager and Testermint coverage for new snapshot fields.
- HTTP transport classifies errors for retry.

### Why it matters

Operators can tune devshard behavior from governance/runtime config without
each host inferring different defaults from the bridge. Per-session parameters
are stable for settlement while still allowing live timeout/availability updates.
Removed hardcoded params

### Review focus

- `decentralized-api/internal/devshard/manager.go` (`create`, providers)
- `decentralized-api/internal/devshard/runtime_params_provider.go`
- `devshard/docs/params-dataflow.md`
- `devshard/state/machine_phase7_test.go`

---

## `567cc2fad` — Forward-only storage schema migrations

**Title:** `feat(devshard): add forward-only session storage schema migrations`

### What changed

**Devshard session storage** moves parent/meta/epoch DDL into ordered
**`*_migrate.go`** steps with shared `devshard/storage/migrate/` (`schema_migrations`,
`ApplyPG` / `ApplySQLite`). Schema runs **once at startup** (or first epoch DB
open); per-epoch **`PARTITION OF`** stays in **`ensurePartition` only**.

**Bug fix:** `SaveSnapshot` no longer runs partition DDL on every call — it uses
cached **`ensurePartition`** (regression:
`TestSaveSnapshot_SameEpoch_PartitionCreateOnce`).

**CI:** `scripts/check-storage-ddl.sh` + `verify.yml` job `check-storage-ddl`.

**Docs:** [storage-design.md](./storage-design.md) schema-evolution section;
removed standalone `storage-migrations-plan.md`. Minor dapi payload/stats storage
tweaks (comments / `ensureSchema` errors only — still inline schema, not migrate
framework).

### Why it matters

Multiple `devshardd` versions can share Postgres; forward-only migrations allow
safe additive schema while older binaries still run. Hot paths stay free of parent
DDL.

### Review focus

- `devshard/storage/postgres_migrate.go`, `sqlite_meta_migrate.go`,
  `sqlite_epoch_migrate.go`, `devshard/storage/migrate/`
- `devshard/storage/postgres.go` (`SaveSnapshot` / `ensurePartition`)
- `scripts/check-storage-ddl.sh`

---

## `6cbe3a7a7` — Boot-time storage backend selection (fork fix)

**Title:** `fix(devshard): select storage backend at boot to prevent reboot fork`

### What changed

Replaces **dual-backend hybrid** (SQLite + Postgres live together, sticky
in-memory routes, PG reconnect, silent fallback) with **one backend per process**:

| Condition | Backend |
| --- | --- |
| `escrow_epoch` has rows in `_meta.db` | **SQLite** (drain; WARN if `PGHOST` set) |
| `PGHOST` set and meta empty | **Postgres** (boot **fails** if PG unreachable) |
| `PGHOST` unset | **SQLite** |
| `.pg-bound` without `_meta.db`, `PGHOST` unset | **Boot fails** |

- `storage_mode.go`: `HasSQLiteSessions`, `.pg-bound` marker.
- `HybridStorage`: thin wrapper forwarding to a single `backend`.
- Removed: sticky routing, duplicate-session WARN merge, `PG_RETRY_INTERVAL`,
  mid-run PG reconnect.
- Tests: `factory_test.go` (full matrix + **fork regression**), `hybrid_test.go`,
  `storage_mode_test.go`; deleted obsolete dual-hybrid tests from
  `postgres_test.go`.
- [storage-design.md](./storage-design.md): storage mode selection; removed
  sticky hybrid / SQLite fallback sections.
- `migrate/pg_test.go`: always use testcontainers (ignore shell `PGHOST`).

### Why it matters (fork bug)

Previously: Postgres session → reboot while PG down → in-memory routes lost →
`CreateSession` for same escrow could **land in SQLite** while append log stayed
in Postgres → **permanent fork**, often only a log WARN. Now: Postgres mode either
connects at boot or **fails**; no silent SQLite path for PG-only deployments.

### Review focus

- `devshard/storage/factory.go`, `decideStorageBackend`, `logSQLiteTransitionWarn`
- `devshard/storage/hybrid.go` (size reduction)
- `TestNewStorage_postgresModeNoForkWhenPGDownAfterSessionInPG`
- Operator transition: drain SQLite until `escrow_epoch` empty, then restart for
  Postgres-only — [Storage Mode Selection](./storage-design.md#storage-mode-selection)

---

## `fa119b2c7` — CI, proto, and v0.2.13 upgrade backfill

**Title:** `fix(ci,devshard,chain): CI ripgrep, single nodemanager proto, migration and upgrade backfill`

### What changed

- **CI:** Pin `check-storage-ddl` to `ubuntu-22.04` and `ripgrep=13.0.0-2` for reproducible DDL guard runs.
- **Proto:** Single `devshard/nodemanager` source (dapi re-exports generated stubs); removes duplicate nodemanager definitions.
- **Migrations:** Forward-only `Statements []string` per step (no naive `;` SQL split); v0.2.13 upgrade handler backfills new `DevshardEscrowParams` and per-escrow fee fields when zero.

### Why it matters

Unblocks reliable CI and safe chain upgrades before devshardd can depend on a
stable `GetRuntimeConfig` RPC on dapi.

### Review focus

- `.github/workflows/verify.yml`, `devshard/storage/migrate/`
- `inference-chain/app/upgrades/v0_2_13/upgrades.go`

---

## `f4e2c6704` — devshardd chain-poll runtime config fallback

**Title:** `feat(devshardd): chain-poll runtime config fallback when dapi lacks GetRuntimeConfig`

**Commit:** `f4e2c6704e83543b514057c5dbf7c30f8c66f0b4` (branch tip at time of writing)

### Problem

Standalone **devshardd** (versioned hosts via versiond) loads governance params only
through dapi **NodeManager `GetRuntimeConfig` long-poll** (`devshard/runtimeconfig`).
**`gm/microrelease` dapi** exposes only `AcquireMLNode` / `ReleaseMLNode` on
NodeManager — no `GetRuntimeConfig`. In that mixed deploy:

| Symptom | Cause |
| --- | --- |
| `devshard_requests_enabled` ignored | `AvailabilityTracker` stuck at constructor seed `Enabled=true`; gRPC `apply()` never runs |
| `max_nonce` not enforced | `Snapshot.MaxNonce` stays `0` → host skips ceiling |
| Epoch prune hook dead | `OnEpochChange` gated on `prev.ParamsBlockHeight > 0` |
| Log spam | gRPC loop retries forever on `codes.Unimplemented` |

Embedded dapi **v1** hosts are unaffected — they use `ConfigManagerAvailability` fed
by the block dispatcher, not devshardd’s provider.

### What changed

**Two providers, one `Snapshot` + shared `baseProvider`:**

| Path | When | How params update |
| --- | --- | --- |
| **gRPC** (`runtimeconfig.New`) | Current dapi; `DEVSHARDD_PARAMS_SOURCE=grpc` or auto probe OK | Long-poll `GetRuntimeConfig` → `apply()` |
| **Chain** (`runtimeconfig.NewChain`) | Old dapi; auto probe `Unimplemented`; `DEVSHARDD_PARAMS_SOURCE=chain` | `QueryParams` + `QueryEpochInfo` every **60s** (same cadence as pre–long-poll devshardd) |

**Selection** (`decentralized-api/cmd/devshardd/params_provider.go`):

- `DEVSHARDD_PARAMS_SOURCE=auto` (default): one-shot probe with `MaxWaitSeconds=0`;
  **`codes.Unimplemented` → chain**, transient errors → stay on gRPC.
- Env overrides: `grpc` | `chain`; chain refresh via `DEVSHARDD_PARAMS_CHAIN_REFRESH_SECONDS` (default 60).

**Other behavior:**

- **`seedAvailabilityFromChain`** in `cmd/devshardd/main.go` — one 3s `QueryParams`
  at boot so the kill-switch matches chain before the first HTTP request (closes
  the optimistic `Enabled=true` window).
- **gRPC provider** exits on `Unimplemented` (no infinite retry loop).
- **`ChainParamsFetcher`** in `decentralized-api/internal/devshard` — maps cosmos
  queries to `Snapshot` without pulling `inference-chain` into `devshard/runtimeconfig`.
- v0.2.12 chain: new proto fields decode as zero → existing **“if > 0 override”**
  bind semantics keep compiled defaults (state-root safe).

### Why it matters

Operators can roll out **new devshardd** before **new dapi** without losing
`devshard_requests_enabled`, `max_nonce`, logprobs mode, epoch-driven prune, or
(v0.2.13 chain) session-bind params — at the cost of **~60s** propagation latency
on the chain path vs near-instant long-poll.

`ApprovedVersions` remain dapi-only (not on chain); devshardd session lifecycle does
not consume them today.

### Review focus

- `devshard/runtimeconfig/provider_base.go`, `chain_provider.go`, `grpc_provider.go`
- `decentralized-api/internal/devshard/chain_params_fetcher.go`
- `decentralized-api/cmd/devshardd/params_provider.go`, `main.go` (`seedAvailabilityFromChain`)
- Unit tests: `devshard/runtimeconfig/*_test.go`, `cmd/devshardd/params_provider_test.go`

---

## Test plan (high level)

```bash
# Devshard module
cd devshard && go test ./...

# Runtime config providers (gRPC + chain fallback)
cd devshard && go test ./runtimeconfig/... -count=1

# devshardd params selector + chain fetcher
cd decentralized-api && go test ./cmd/devshardd/... ./internal/devshard/... -run 'ChainParams|ParamsProvider|RuntimeConfig' -count=1

# DDL guard
bash scripts/check-storage-ddl.sh

# Storage + migrate (Docker)
cd devshard && go test ./storage/... ./storage/migrate/...
```

Postgres integration tests require Docker (testcontainers). Use `-short` to skip
them locally. **Testermint** (`DevsharddRuntimeConfigTests`) still exercises the
**gRPC** path against current dapi; chain fallback is covered by unit tests
(optional future e2e: `DEVSHARDD_PARAMS_SOURCE=chain` or stub `Unimplemented`).


a-kuprin:
at `MsgSettleDevshardEscrow` field `version` was renamed to `state_root_and_protocol_version`
It keeps the wiring and meaning, just naming, so it's legal proto change

0xgonka (commented):
**Must fix:**

- **[Critical] `DevshardEscrowParams` field 9 still inserted as `bool devshard_requests_enabled`, displacing the grace fields to 10/11.** [`inference-chain/proto/inference/inference/params.proto:386-388`](inference-chain/proto/inference/inference/params.proto#L386-L388). The base branch `devshard-0.2.13-v2` has field 9 = `default_seal_grace_nonces` (uint32) and field 10 = `default_inference_clear_grace_seconds` (uint32). This PR inserts a bool at slot 9, pushing them to 10/11, then appends new fields 12-17 (those are fine). Same Critical I flagged on PR #1234 — it carries forward unfixed. Anything that persisted `DevshardEscrowParams` under the prior numbering (every v0.2.13-RC build, devnets, internal clusters) will mis-decode after this upgrade: field 9's stored uint32 (~20) reads as a bool, field 10's old `default_inference_clear_grace_seconds` uint32 (120) gets read as `default_seal_grace_nonces`, field 11 is absent → `default_inference_clear_grace_seconds` reverts to 0. The upgrade handler at [`v0_2_13/upgrades.go:115-118`](inference-chain/app/upgrades/v0_2_13/upgrades.go#L115-L118) only backfills SealGrace when it's currently 0, so the wrong non-zero value silently survives. **One-line fix: move `bool devshard_requests_enabled` to a number ≥ 18 (after the new appended block), restore 9/10 to the grace fields.** Mainnet impact is zero today (v0.2.13 never went live), but every RC environment fails on upgrade.

- **[High] `EscrowState.Version` → `EscrowState.StateRootAndProtocolVersion` rename silently zeroes out the field when loading pre-PR snapshots.** [`devshard/types/domain.go EscrowState`](devshard/types/domain.go) renames the field with no `json` tag and no custom `UnmarshalJSON`. Snapshots from the base branch were marshaled by [`devshard/host/snapshot.go MarshalStateSnapshot`](devshard/host/snapshot.go) with key `"Version"` (Go's default = struct field name). After this PR, the new binary's `UnmarshalStateSnapshot` ignores the unknown `"Version"` key and leaves `StateRootAndProtocolVersion = ""`. Subsequent state-root computation hashes `sha256("")` instead of `sha256("v1")` — different preimage, different state-root. Intra-session consistency is preserved *if* every host in the session upgrades simultaneously (they all see `""`), but: (a) the chain has no idea the field's empty-string is a "lost v1," so debugging future settlement failures will lie about what protocol version actually ran; (b) any mixed-binary moment between hosts in the same session diverges silently. The fix is either a `json:"Version"` tag on the new field, or a custom `UnmarshalJSON` that falls back to the old key. There's also no cross-version snapshot test in [`devshard/host/snapshot_test.go`](devshard/host/snapshot_test.go) — add one that round-trips an old-format snapshot through the new unmarshal and asserts the version field is preserved.

**Should fix:**

- **[Medium] `.pg-bound` marker has a hole when `_meta.db` exists but is empty.** [`devshard/storage/factory.go decideStorageBackend`](devshard/storage/factory.go) (lines 76-100): if `.pg-bound` exists AND PGHOST is unset, the code only refuses to boot when `_meta.db` does NOT exist. If the file exists but has zero rows (e.g., after a clean SQLite drain that didn't delete the file), the code falls through to `return storageBackendSQLite, nil` — silently orphaning any Postgres sessions. Either delete the marker as part of the drain procedure (and document that), or refuse-boot whenever `.pg-bound` is present and PGHOST is unset, regardless of `_meta.db` state. The current asymmetry is a footgun under operator error. Also worth a `TestNewStorage_PGBoundWithEmptyMetaDB` regression.

- **[Medium] `max_nonce` is read live on every diff, not snapshotted at session create.** [`devshard/host/host.go checkDiffNonceLimitLocked`](devshard/host/host.go) (~line 459) reads `h.maxNonce.MaxNonce()` per diff. If governance lowers `max_nonce` mid-session (rare but possible — admin error or contested gov), a diff at nonce N can be accepted against the old limit and then leave the session over the new chain-side limit. Settlement is what fails at that point, which is exactly what this PR is trying to prevent. Either snapshot `max_nonce` at session create alongside the other lane-A params in [`decentralized-api/internal/devshard/manager.go create`](decentralized-api/internal/devshard/manager.go), or change semantics to "live but monotonically clamped — never decrease intra-session." The former is consistent with how `RefusalTimeout` / `ExecutionTimeout` / `ValidationRate` / `VoteThresholdFactor` are handled by the lane-A freeze.

**Nits:**

- **[Low] Auto-probe Unimplemented latch is permanent across the devshardd process lifetime** — [`decentralized-api/cmd/devshardd/params_provider.go:142-160 probeUnimplemented`](decentralized-api/cmd/devshardd/params_provider.go#L142-L160). This is acceptable design (once you know dapi is the old `gm/microrelease` binary, polling chain is correct forever), but the failure-recovery story is "operator restarts devshardd after dapi upgrade." Worth documenting that explicitly in [`devshard/docs/params-dataflow.md`](devshard/docs/params-dataflow.md) so operators don't waste cycles wondering why GetRuntimeConfig still isn't being used after they upgraded dapi.
- **[Low] Schema migration framework assumes single-process per data dir** — [`devshard/storage/migrate/sqlite.go ApplySQLite`](devshard/storage/migrate/sqlite.go) uses `BEGIN IMMEDIATE` which is correct against concurrent migrators in the same process, but SQLite's default DELETE journal mode allows two devshardd processes on the same `_meta.db` to race the schema_migrations table writes. Devshard is single-process per data dir by design, but worth a one-line comment or a `PRAGMA journal_mode=WAL` to make the assumption explicit.

**Looks correct:**

- **`SaveSnapshot` partition-DDL fix is real and clean.** [`devshard/storage/postgres.go SaveSnapshot`](devshard/storage/postgres.go) (~line 285) moves `ensurePartition(ctx, epochID)` out of the upsert hot path and caches per-epoch hits in `knownEpochs` under `mu sync.RWMutex`. At 500+ snapshots per epoch this was real overhead; the cache is concurrent-safe and the regression test (`TestSaveSnapshot_SameEpoch_PartitionCreateOnce`) locks in the no-DDL-on-second-call behavior.
- **Forward-only migration framework is sound.** [`devshard/storage/migrate/sqlite.go ApplySQLite`](devshard/storage/migrate/sqlite.go) creates `schema_migrations` atomically, reads `MAX(id)` before each step, runs each step in its own `BEGIN IMMEDIATE ... COMMIT` with the migration-record insert inside the transaction — crash mid-step leaves the step unrecorded, next boot retries it. Idempotent and correct.
- **Boot-time backend selection eliminates the dual-write fork mode.** Old `HybridStorage` could route the same escrow to SQLite OR Postgres across restarts depending on in-memory routing state, leaving append-log fragments in both backends. New design picks one backend at `NewStorage()` and stays there, with the `.pg-bound` marker preventing accidental "I'm now SQLite, your PG sessions are orphans" downgrades. Modulo the hole flagged above, this is the right architecture.
- **Auto-probe correctly distinguishes Unimplemented from transient errors.** [`params_provider.go probeUnimplemented`](decentralized-api/cmd/devshardd/params_provider.go) latches to chain-poll ONLY on `codes.Unimplemented`; any other error (Unavailable, DeadlineExceeded) returns `false` and keeps the long-poll path so devshardd self-heals when dapi comes back. Comment at the function explicitly documents this intent. Correct design — *not* a permanent transient-error latch.
- **Lane-A session-create params freeze the right things.** [`decentralized-api/internal/devshard/manager.go create`](decentralized-api/internal/devshard/manager.go) snapshots `RefusalTimeout`, `ExecutionTimeout`, `ValidationRate`, `VoteThresholdFactor`, `SealGraceNonces` at bind time and writes them into `SessionConfig` — these are hashed into the state-root and need to be identical across hosts, so freezing them avoids inter-host divergence from a gov change landing during session create. The keep-live set (timeouts for availability, max_nonce for enforcement) is the right complement.
- **Upgrade handler ordering is correct.** [`v0_2_13/upgrades.go`](inference-chain/app/upgrades/v0_2_13/upgrades.go) calls `setDevshardEscrowParams` (which initializes the new fee fields with defaults) **before** `backfillDevshardEscrowFees` (which reads from now-populated params and writes per-escrow rows). The earlier review concern about reading-zero is unfounded — the params are populated first.
- **`MsgSettleDevshardEscrow` field 9 rename `version` → `state_root_and_protocol_version` is wire-compatible.** Proto identifies fields by number; only the name changed at slot 9. Cross-version chain/host pairs continue to decode the same string in/out, and the chain verifier hashes whatever string the field carries (no allowlist check), so settlement succeeds regardless of which name the Go binary uses internally.
- **Single nodemanager.proto source.** Consolidating the two duplicated proto files into one under `devshard/nodemanager/nodemanager.proto` and re-exporting the generated stubs from `decentralized-api/nodemanager` removes a real drift risk. New runtime-config fields (refusal/execution timeouts, validation_rate, vote_threshold_factor) are appended to `GetRuntimeConfigResponse`, no renumbering.

**Bottom line:** Hold for the proto fix (must-fix #1, one-line). The snapshot-field-rename (must-fix #2) is also one line if you accept the json-tag fix; if a custom UnmarshalJSON is preferred, slightly more. Everything else is shippable as-is or a follow-up. The storage-mode rework is a genuine architectural improvement and the lane-A/B/C param freeze story is well thought-through.

a-kuprin:
> * **[Critical] `DevshardEscrowParams` field 9 still inserted as `bool devshard_requests_enabled`, displacing the grace fields to 10/11.** [`inference-chain/proto/inference/inference/params.proto:386-388`](inference-chain/proto/inference/inference/params.proto#L386-L388). The base branch `devshard-0.2.13-v2` has field 9 = `default_seal_grace_nonces` (uint32) and field 10 = `default_inference_clear_grace_seconds` (uint32). This PR inserts a bool at slot 9, pushing them to 10/11, then appends new fields 12-17 (those are fine). Same Critical I flagged on PR #1234 — it carries forward unfixed. Anything that persisted `DevshardEscrowParams` under the prior numbering (every v0.2.13-RC build, devnets, internal clusters) will mis-decode after this upgrade: field 9's stored uint32 (~20) reads as a bool, field 10's old `default_inference_clear_grace_seconds` uint32 (120) gets read as `default_seal_grace_nonces`, field 11 is absent → `default_inference_clear_grace_seconds` reverts to 0. The upgrade handler at [`v0_2_13/upgrades.go:115-118`](inference-chain/app/upgrades/v0_2_13/upgrades.go#L115-L118) only backfills SealGrace when it's currently 0, so the wrong non-zero value silently survives. **One-line fix: move `bool devshard_requests_enabled` to a number ≥ 18 (after the new appended block), restore 9/10 to the grace fields.** Mainnet impact is zero today (v0.2.13 never went live), but every RC environment fails on upgrade.

This is the reference to `devshard-0.2.13-v2` base branch

https://github.com/gonka-ai/gonka/blob/611186b2b52110cf0ec15dfb951d8f7c5c5ae99a/inference-chain/proto/inference/inference/params.proto#L384

It is
```go
bool devshard_requests_enabled = 9;
```

And this is exactly what is currently onchain (link to `main` branch):
https://github.com/gonka-ai/gonka/blob/17808620293b57112896bcbb7f99c4c2f554d6c8/inference-chain/proto/inference/inference/params.proto#L384

a-kuprin:
> * [Medium] `max_nonce` is read live on every diff, not snapshotted at session create.

According to this one. I don't think we should use snapshot on session creation, because if `max_nonce` parameter is changed, it anyway will break the settlement if any devshard has more nonces.
So if we error on devshard it is not worst case. Handling this edge case by snapshotting anyway doesn't solve the problem that `max_nonce` could be set to lower value during existing devshard sessions run and it could be lower then existing in session `nonce`

Actually I don't see a good way to handle it

a-kuprin:
> * **[High] `EscrowState.Version` → `EscrowState.StateRootAndProtocolVersion` rename silently zeroes out the field when loading pre-PR snapshots.**

I don't think it is valid concern. The main point that session is binded to devshard version. And it is just impossible that old devshard binary version is participating in session that has newer version snapshot. So there will no be conflict with this renaming

a-kuprin:
@0xgonka 
These are fixed:
[Medium] .pg-bound marker has a hole when _meta.db exists but is empty.
[Low] Schema migration framework assumes single-process per data dir

This is also fixed:
[Low] Auto-probe Unimplemented latch is permanent across the devshardd process lifetime
But I just added auto-fallback. It always prefers gRPC and if falled back to chain requests it keeps trying gRPC connection each 300 seconds. So if gRPC is alife (dapi was updated to support endpoint, or mayb dapi was temporary unavailable) it will continue to use gRPC. If dapi is down, it will use chain directly

I've also replaced snapshot json to protobuf, as it is slightly more effcient to marshall/unmarshall

0xgonka (approved):

0xgonka (changes requested):
Delta review since commit `b15644882` — one new squashed commit `84fd11b6a` addresses three of my prior Should-Fix items cleanly (`.pg-bound` boot-fail, WAL on migrate, adaptive grpc/chain failover with failback) and resolves the prior High about the EscrowState JSON snapshot field rename by switching the snapshot format to protobuf. **The Critical proto-field renumbering and the Medium max_nonce TOCTOU are still outstanding.** One newly-introduced concern around the protobuf snapshot cutover.

**Must fix (carried from prior reviews, still unaddressed):**

- **[Critical] `DevshardEscrowParams` field 9 still inserted as `bool devshard_requests_enabled`, displacing `default_seal_grace_nonces` to 10 and `default_inference_clear_grace_seconds` to 11.** `inference-chain/proto/inference/inference/params.proto:386-388`. Same issue I flagged on #1234 and the prior pass of this PR. Mainline state on `main` doesn't have fields 10/11, but the `devshard-0.2.13-v2` base branch (which several open PRs target) does — anything that persisted under that prior numbering will mis-decode on upgrade. One-line fix: move `bool devshard_requests_enabled` to a number ≥ 18, restore 9/10 to the grace fields, append the rest.

**Should fix:**

- **[Medium, carried] `max_nonce` is read live per-diff in `devshard/host/host.go checkDiffNonceLimitLocked` (~line 459)** instead of being snapshotted at session create alongside the other lane-A params in `decentralized-api/internal/devshard/manager.go create`. A downward governance change to `max_nonce` mid-session can let the host accept a diff at nonce N that puts the session over the new chain-side limit, and settlement then fails — exactly what this PR's max_nonce gate is supposed to prevent. Either snapshot it like `RefusalTimeout` / `ExecutionTimeout` / `VoteThresholdFactor` are, or clamp monotonically ("max_nonce may grow intra-session, never shrink").

- **[Medium, new] Protobuf snapshot cutover is a hard break for existing JSON snapshots; no migration path, no startup error message guiding the operator.** `devshard/host/snapshot.go` now calls `types.UnmarshalStateSnapshotProto(data)` directly with no JSON fallback, and `devshard/host/snapshot_test.go TestUnmarshalStateSnapshot_RejectsJSON` explicitly locks in the rejection. Any operator with an existing JSON snapshot from a prior v0.2.13-RC binary boots into a proto unmarshal error and devshardd presumably crashes-and-restarts in a loop. Acceptable IF the release notes prominently say "wipe devshard state on upgrade," but the current PR has no such notice. Either (a) add a JSON-fallback decoder that converts old snapshots in-place on first load and logs a one-time warning, or (b) emit a clear error on the proto-parse failure path explaining "this looks like a legacy JSON snapshot, please wipe `--storage-dir`," and document the upgrade requirement in `devshard/docs/upgrade.md`. The two-line wrapper version of (b) is the minimum:

      if state, _, _, err := UnmarshalStateSnapshotWithCommitted(data); err != nil {
          if json.Valid(data) {
              return nil, fmt.Errorf("legacy JSON snapshot detected; v0.2.13 uses protobuf snapshots — wipe storage dir and re-bootstrap: %w", err)
          }
          return nil, err
      }

- **[Low, new] Schema drift risk between `EscrowState` Go struct and `EscrowStateProto`.** `devshard/types/snapshot_codec.go` is the manual mapping layer. Future additions to `EscrowState` need to be added to BOTH the proto schema AND the codec, with no compiler help to enforce the parallelism. One missed field becomes a silent data-loss bug on restart. Mitigate with a `TestEscrowStateProtoRoundTrip_AllFields` test that constructs a fully-populated `EscrowState` (reflection-walked or property-tested), round-trips through proto, and asserts equality. The existing `TestMarshalStateSnapshotWithCommitted_RoundTrip` is the right pattern, just needs every field exercised explicitly.

**Looks correct (newly fixed since prior review):**

- **`.pg-bound` boot-fail is now unconditional.** `devshard/storage/factory.go decideStorageBackend:91-95` — removed the `os.Stat(MetaDBPath(...))` gate that previously allowed the SQLite fallback when an empty `_meta.db` existed alongside `.pg-bound`. Now any `.pg-bound`-without-PGHOST boot fails with `ErrStoragePGBoundWithoutPostgres` regardless of MetaDB state. Exactly the fix I suggested; closes the orphan-PG-sessions hole.
- **WAL + busy_timeout on the migrate path.** `devshard/storage/migrate/sqlite.go ensureSQLiteMigratePragmas` sets `PRAGMA journal_mode=WAL` and `PRAGMA busy_timeout=5000` before opening the migration transaction. The added comment also explicitly documents the single-process-per-data-dir invariant. Addresses the prior nit.
- **Adaptive grpc/chain failover and failback is well-built.** `devshard/runtimeconfig/adaptive_provider.go` + `adaptive_config.go`. Boot probe → grpc or chain based on `codes.Unimplemented`. Supervisor watches `GRPCStale` (90s default) and fails over to chain when grpc accumulates errors without successful applies. On chain, reprobes grpc every `GRPCReprobe` (300s default) and fails back after `FailbackProbes` (2) consecutive healthy probes. Single supervisor goroutine, atomic counters for stats, mutex-protected runner cancel, `runnerWG.Wait()` in `stopRunner` for clean shutdown. `startGRPC` and `startChain` both call `stopRunner` first so only one runner is active at a time. Resolves my prior "permanent latch" concern; devshardd now self-heals when dapi comes back. Test coverage in `adaptive_provider_test.go` (+471 lines) looks comprehensive.
- **Protobuf snapshot format is the right long-term call.** `devshard/proto/devshard/v1/snapshot.proto` defines `EscrowStateProto` with the correct field-by-field mapping, and the codec at `devshard/types/snapshot_codec.go` maps both directions cleanly. Fixes my prior High by making the snapshot format wire-stable independent of Go struct field names; future renames of `EscrowState` fields no longer silently drop data. The migration story (Should-Fix #2) is the only loose end.
- **Adaptive probe correctly distinguishes Unimplemented from transient errors.** Boot probe latches to chain only on `codes.Unimplemented`; failback-side `grpcProbeHealthy` returns true on success, false on Unimplemented OR any other error (and consecutive-streak resets on either failure mode — conservative). This is consistent with the prior probe semantics.
- **Lock + shutdown discipline in `adaptiveProvider` is clean.** `runnerCancel` is mutex-guarded but `lastGRPCApply` / `grpcErrorsSinceApply` / `failbackStreak` are atomics / single-goroutine-owned, no over-locking. `cancelRunner` clears the cancel pointer before invoking it so a concurrent `stopRunner` doesn't double-cancel.

**Bottom line:** Three of four prior Should-Fixes are resolved cleanly. The protobuf snapshot rework is a genuine improvement over the JSON field-rename land mine. Still hold for the Critical proto-field renumbering. The protobuf-cutover migration gap is small but operator-facing — the two-line "legacy JSON detected" error message would meaningfully reduce on-call pain. Max-nonce TOCTOU is the same lingering item from the prior pass.

0xgonka:
@a-kuprin — verified your pushback in-repo, retracting both findings.

**Critical (proto field 9):** confirmed `origin/devshard-0.2.13-v2:inference-chain/proto/inference/inference/params.proto` already has `bool devshard_requests_enabled = 9` followed by `default_seal_grace_nonces = 10` and `default_inference_clear_grace_seconds = 11`, and `origin/main` has the same field 9. PR #1258's only proto diff is appending fields 12-17 — no renumber, no displacement. I was reasoning from a stale snapshot of the base state in my head. Withdraw.

**Medium (max_nonce TOCTOU):** verified `inference-chain/x/inference/keeper/devshard_settlement.go:95-96` checks `params.MaxNonce` live at settlement, not a per-escrow snapshot. So your point lands exactly: if gov lowers MaxNonce mid-session, snapshotting at session create just lets the host produce more diffs past the new limit before settlement deterministically rejects on `nonce > MaxNonce`. The current live-read gate is strictly safer than what I was suggesting — host can refuse new diffs at the new limit and the session might still settle if existing nonces stay below it. Withdraw.

Net: the only outstanding item from my delta review is the protobuf-snapshot migration UX (Should-Fix Medium — emit a clearer "legacy JSON detected, wipe storage dir" error message on parse failure, or just document the wipe-on-upgrade requirement in `devshard/docs/upgrade.md`). That one's documentation-tier, not a blocker. Approving.

0xgonka (approved):
Approved. See retraction comment above for the Critical/Medium I'm withdrawing. The protobuf snapshot UX gap is the only outstanding item and is documentation-tier — fine as follow-up.

a-kuprin:
> Approved. See retraction comment above for the Critical/Medium I'm withdrawing. The protobuf snapshot UX gap is the only outstanding item and is documentation-tier — fine as follow-up.

It's also impossible because snapshot is used only inside devshard session. And session is binded to version.
It is impossible that different devshard version instances are in one session. It means, that if some snapshot already exist, it will be read only by instance of version, that can read it. Newer version will be running in parallel and work with new sessions

PR #1259, opened 2026-05-27
Gateway: 0.2.13 conflict resolution

DimaOrekhovPS:


qdanik (approved):

0xgonka (approved):
Merges the v0.2.13 microrelease (already in main) into the gateway feature branch (`dl/devshards-gateway-to-main`). 16k+ additions but the manual conflict-resolution surface is small — the merge commit auto-resolved most files, with two follow-up commits fixing real issues introduced by the auto-merge. **Clean merge; no consensus-critical regressions.** The proto/upgrade/storage state matches main, the duplicate-symbol fix is real, and the gateway storage path normalization is sound.

**Must fix:**

None blocking. Pre-fix state had a real compile error (duplicate `var ErrSnapshotNotFound`), but commit `7a34aaab4` already addresses it.

**Should fix:**

- **[Low] Empty PR description.** PR body is blank. For a merge-resolution PR this size (16k+ additions, three resolution commits, cross-branch semantics), the body should at minimum link the two parents being merged (`dl/devshards-gateway-to-main` ← main post-#1143), name the two conflict-resolution commits, and call out the storage-path semantic change (legacy `state.db` file paths → directory paths) so reviewers and operators know what to test. The two follow-up commits already have good messages; the PR body just needs to summarize them.

- **[Low] Gateway storage-path normalization isn't enforced on the cold-recover path.** [`devshard/cmd/devshardctl/gateway.go`](devshard/cmd/devshardctl/gateway.go) — `normalizeStorageDir` is called from `buildRuntime`, `handleAdminAddDevshard`, `handleAdminImportDevshard`, `addCreatedEscrowRuntime`, `removeDevshardStorage`, and `finalizeRuntimeConfigs`, which covers the live-creation paths. But cold-recovery from `GatewayStore` (loading persisted `RuntimeConfig` rows that were written before this PR landed) reads `StoragePath` as-is. Operators with existing devshard gateways on disk that have legacy `state.db` paths persisted will hit `os.MkdirAll(cfg.StoragePath, ...)` against a file path → directory-creation error at boot. Either run `normalizeStorageDir` when loading from `GatewayStore` too, or add a one-time DB migration to rewrite persisted paths. Worth a regression test that round-trips a pre-PR persisted config.

**Nits:**

- The PR pulls in the full v0.2.13 microrelease as a single commit (`178086202`), so almost all the substantive changes are pre-reviewed via #1143's merge — there's no new code path to vet here beyond the conflict-resolution surface. Reviewers can scope their attention to `decentralized-api/completionapi/` + `devshard/cmd/devshardctl/` + the resolution commits, not the full 100-file diff.
- The two new helpers `normalizeStorageDir` and `legacyPerfSourcePath` are mutually inverse-ish utilities — worth a comment on `normalizeStorageDir` documenting the round-trip contract with `legacyPerfSourcePath` so a future refactor doesn't accidentally break it.

**Looks correct:**

- **Duplicate `ErrSnapshotNotFound` fix is necessary and clean.** [`devshard/storage/interface.go`](devshard/storage/interface.go) — the merge commit `517ab6b90` left two `var ErrSnapshotNotFound = errors.New("snapshot not found")` at lines 24 and 55, both top-level. That's a Go compile error (duplicate package-level identifier), not just a code smell — the branch literally wouldn't build without the fix in `7a34aaab4`. The surviving declaration at line 24 is the right one (it sits with the other `ErrSession*` declarations).
- **Storage-path semantic change is correctly threaded through every code path.** v0.2.13 moves from `escrow-{id}/state.db` file paths to `escrow-{id}/` directory paths. `defaultStoragePath` was updated, and `normalizeStorageDir` is called at every live entry point that takes a user-supplied path (buildRuntime, both admin handlers, finalize, addCreatedEscrow). The legacy file-path form is still accepted and silently converted. `legacyPerfSourcePath` is the right escape hatch — `BackfillLegacyEscrowSamples` is a file-level operation against the historical `state.db` and needs the file path, not the directory.
- **`removeDevshardStorage` simplification is safe.** Before: hand-maintained list of WAL/SHM siblings to clean up. After: `normalizeStorageDir` → safety checks → `os.RemoveAll(storagePath)`. Now that all paths are directories, the recursive remove handles WAL/SHM/journal naturally.
- **Test coverage matches the new semantics.** `TestFinalizeRuntimeConfigsNormalizesLegacyStateDBPath` and `TestResolveBaseStorageDirNormalizesLegacyStateDBPath` (added in `1f091ab92`) lock in the legacy-path-acceptance contract.
- **v0.2.13 surface preserved end-to-end.** Upgrade handler at [`inference-chain/app/upgrades/v0_2_13/upgrades.go`](inference-chain/app/upgrades/v0_2_13/upgrades.go) still runs the full sequence (setDevshardEscrowParams, backfillConfirmationWeightScales, grantRespondDealerComplaintsAuthz, disableConfirmationPocForUpgradeEpoch, executeBridgeSetup). `UpgradeName = "v0.2.13"` unchanged. Proto state matches main exactly (`DevshardEscrowParams` fields 1-9, no renumber). No storage-backend logic from main's `factory.go` / `hybrid.go` / `migrate.go` was lost in the merge.
- **Gateway functions count went UP (114 → 116), not down.** The two new functions are `normalizeStorageDir` and `legacyPerfSourcePath`. No gateway capability was silently dropped during merge resolution.

**Bottom line:** Land as-is, modulo the cold-recovery normalization concern (Should-Fix above) — that's a real operator-facing gap for anyone with a pre-PR gateway on disk. Everything else in the conflict-resolution work is correct.

a-kuprin (approved):

PR #1263, opened 2026-05-27
fix(validation): force return_token_ids as required param

qdanik:
## Summary

Fixes false-positive `inflated_tokens` invalidations on Kimi-K2.6 caused by vLLM's reasoning parser stripping `<think>` / `</think>` / `<|tool_calls_section_begin|>` special tokens from delta-streamed SSE events while still counting them in `usage.completion_tokens`. Result: validator's stream-derived `enforced_tokens` list is shorter than executor's claim, so the strict `claim > validator.usage.completion_tokens` check fires on otherwise-honest inferences.

Production sweep across 25 diverse multi-turn / reasoning / CJK / emoji prompts: **9 false-positive inflations → 0**.

## Root cause

`extract_reasoning_streaming` in [`vllm/reasoning/kimi_k2_reasoning_parser.py`](https://github.com/vllm-project/vllm/blob/main/vllm/reasoning/kimi_k2_reasoning_parser.py) returns `None` when a single-token delta carries `_start_token_id` or `_end_token_id`, suppressing the SSE event entirely. The token is still produced and counted in `usage.completion_tokens`, but the stream consumer cannot observe it.

[vllm-project/vllm#29074](https://github.com/vllm-project/vllm/pull/29074) changed that path to emit an empty `DeltaMessage()` (with the token id attached) instead of `None` when `return_token_ids=true` is set. The control event lets stream consumers see every token vLLM accounts for, so:
- executor stream's logprob entries match `usage.completion_tokens` 1-to-1
- validator's stream-derived `enforced_tokens` is complete
- validator's `usage.completion_tokens` matches executor's

## The fix

1. **Force `return_token_ids=true`** on every upstream chat-completions request (executor and validator paths). Gateway registers it with `ForceLiteralParameterHandler{Value: true}` so client overrides cannot disable it. Comment links to vllm-project#29074.

2. **Bump validator `max_tokens`** in `BuildValidationBody` to `max(original_max_tokens, 2 × len(enforced_tokens))`. Empirically observed on multi-turn Kimi-K2.6 sequences: when `max_tokens == len(enforced_tokens)`, vLLM finishes the validator run early with `finish_reason=length` at roughly half the budget. Doubling the headroom unblocks vLLM; the output is still strictly bounded by the enforced list plus the natural EOS continuation.

3. **Strip vLLM-internal fields from client-facing responses** so the gateway-forced `return_token_ids` doesn't leak `prompt_token_ids` / `token_ids` / `prompt_logprobs` into OpenAI-compatible client payloads. Devshard internals (`payloadStorage`, `BuildValidationBody`, `EvaluateValidationResponse`) read the raw host response before this proxy boundary, so the strip has no impact on validation.

4. **De-duplicate `numericJSONValueAsUint64`** that was byte-for-byte identical between `decentralized-api/internal/devshard/shared_runtime.go` and `devshard/cmd/devshardctl/request_filters.go` — unified as `devshard.JSONNumericUint64` with the canonical implementation living in `devshard/json.go` alongside `CanonicalizeJSON`.

## Verification

25-prompt sweep on staging covering: CJK / Japanese / Korean / Emoji / Mixed-language / Multi-turn (up to 5 messages) / Reasoning-heavy / Structured-output / Tool-call-like patterns. `max_tokens` ranges 512–1024, `temperature=0.5`, deterministic seeds.

| Metric | Before fix | After fix |
|---|---|---|
| Inflated false-positives | 9 / 24 | **0 / 25** |
| `len(enforced_tokens) == usage.completion_tokens` | partial (drift 1–10) | **25 / 25** |
| `executor.completion ≤ validator.completion` | mismatched on 9 prompts | **25 / 25** |

Unit coverage:
- `JSONNumericUint64` — 21 sub-cases (float64 / int / json.Number / nil / string / fractional / negative)
- `rewriteRequest` — 7 cases (bump / preserve / threshold edge / missing fields / non-numeric / empty enforced / JSON-loaded shape)
- `ModifyRequestBody` `return_token_ids` injection — both default and client-override paths
- `filterClientInternalFields` — token_ids / prompt_token_ids strip

## References

- [vllm-project/vllm#29074](https://github.com/vllm-project/vllm/pull/29074) — vLLM parser change this PR depends on
- [vllm.ai/blog/agent-lightning](https://vllm.ai/blog/agent-lightning) — `return_token_ids` rationale
- [kimi_k2_reasoning_parser.py](https://github.com/vllm-project/vllm/blob/main/vllm/reasoning/kimi_k2_reasoning_parser.py) — source of the strip logic


a-kuprin:
**Recommendation:** Adopt conditional stripping for `logprobs` and `top_logprobs` on the client-facing proxy path. Keep unconditional stripping for `token_ids`, `prompt_token_ids`, and `prompt_logprobs`.

The gateway forces logprobs upstream for validation, but clients who never asked for logprobs should not see them in the response (OpenAI-compatible default). Clients who explicitly set `logprobs: true` or `top_logprobs` should get them back. `return_token_ids` is forced for protocol correctness ([vLLM #29074](https://github.com/vllm-project/vllm/pull/29074)), we shouldn't expose them, as it is not supported API.

PR #1267, opened 2026-05-28
Upgrade v0.2.14

### LINKED ISSUE #891: [P0] Proxy server for `devshards`: timeout handling
At the moment, the proxy server in `/subnet/cmd/subnetctl` handles 3 types of actions: chat completions (with diff propagation), finalizing `devshards`, and querying the `devshards` status.

It should also handle timeout-related mechanisms.

patimen:
# Upgrade Proposal: v0.2.14

This PR prepares the v0.2.14 release.

The mainnet chain/API work focuses on PoC duplicate-artifact protection, early share detection, classic inference API deprecation (disabling `/v1/chat/completions` billing on mainnet and removing embedded `/v1/devshard` from the API binary), reward recipient routing, and upgrade-time safety fixes.

The devshard part prepares the v3 runtime so brokers can serve inference during the chain upgrade without depending on the deprecated classic API path, improving RAM utilization and enabling safe switching between SQLite and Postgres storage.

## Upgrade Plan

The node binary is upgraded through an on-chain software upgrade proposal. Existing hosts are not required to manually update their `api` or `node` containers as part of the upgrade.

A separate devshard v3 release from this branch will be proposed and rolled out before the mainnet chain upgrade. Brokers who switch inference traffic to `/devshard/v3` ahead of time can keep serving inference while the chain upgrade runs.

## Proposed Process

2. The devshard v3 release is proposed and rolled out before the mainnet chain upgrade.
3. Brokers switch inference traffic to `/devshard/v3`.
4. If the on-chain proposal is approved, this PR will be merged immediately after the upgrade is executed on-chain.

### inference-chain / decentralized-api

- Replace the PoC artifact store with SMST-based nonce indexing to prevent duplicate artifacts, with benchmarks, params, and snapshot rebuild fixes [`9a63468`](https://github.com/gonka-ai/gonka/commit/9a6346862e9b365db7994cfa2f07aa81a2a5f958), [`22a21b0`](https://github.com/gonka-ai/gonka/commit/22a21b0b6d893200af0a88f5bb551d238c64f556) by @gmorgachev, @DimaOrekhovPS, @0xMayoor.
- Add DAPI-side early-share detection as an additional security layer, later reworked to check inclusion by nonce instead of dense index so it stays correct against the SMST store [#1382](https://github.com/gonka-ai/gonka/pull/1382), [#1416](https://github.com/gonka-ai/gonka/pull/1416) by @libermans, @DimaOrekhovPS, @gmorgachev.
- Remove deprecated classic decentralized-api inference paths and bind devshard routes to runtime versions [#1386](https://github.com/gonka-ai/gonka/pull/1386) by @gmorgachev.
- Track the last software upgrade height and fix disabling cPoC in the window around chain upgrades [#1268](https://github.com/gonka-ai/gonka/pull/1268) by @patimen.
- Seed maintenance-window params in the v0.2.14 handler with maintenance disabled by default, so governance can enable it later [#998](https://github.com/gonka-ai/gonka/pull/998), [#1428](https://github.com/gonka-ai/gonka/pull/1428) by @Ryanchen911, @DimaOrekhovPS.
- Zero mint inflation during upgrade, burn the fee collector base-denom balance, and add Dahl (@paranjko) to allowed devshard escrow creators [#1418](https://github.com/gonka-ai/gonka/pull/1418) by @gmorgachev.
- Add on-chain configurable reward claim recipients and apply them to stale devshard escrow payouts [#889](https://github.com/gonka-ai/gonka/pull/889), [#1377](https://github.com/gonka-ai/gonka/pull/1377) by @alancapex, @DimaOrekhovPS.
- Prevent reward payments for incoming delegations from excluded hosts [#1378](https://github.com/gonka-ai/gonka/pull/1378) by @gmorgachev.
- Add a total supply endpoint for integrations that need plain-text GONKA supply [#1346](https://github.com/gonka-ai/gonka/pull/1346) by @DimaOrekhovPS.
- Sync bridge block progress so Ethereum transactions are not lost if nodes restart during upgrade [#1376](https://github.com/gonka-ai/gonka/pull/1376) by @GLiberman.
- Smaller settlement, validation, fee, and setup-report fixes: [#1100](https://github.com/gonka-ai/gonka/pull/1100), [#1101](https://github.com/gonka-ai/gonka/pull/1101), [#1129](https://github.com/gonka-ai/gonka/pull/1129), [#1160](https://github.com/gonka-ai/gonka/pull/1160), [#1253](https://github.com/gonka-ai/gonka/pull/1253), [#1255](https://github.com/gonka-ai/gonka/pull/1255), [#1278](https://github.com/gonka-ai/gonka/pull/1278), [#1307](https://github.com/gonka-ai/gonka/pull/1307), [#1383](https://github.com/gonka-ai/gonka/pull/1383), [#1413](https://github.com/gonka-ai/gonka/pull/1413) by @0xMayoor, @0xgonka, @ouicate, @redstartechno, @GLiberman, @DimaOrekhovPS.

### devshard

- Serve inference during the PoC validation phase on validation-inference-capable nodes [#1348](https://github.com/gonka-ai/gonka/pull/1348) by @qdanik, @gmorgachev.
- Enable standalone `devshardd` to continue serving during DAPI outages using an ML-node cache, add per-escrow SQLite/Postgres storage routing, and snapshot `validation_rate` at escrow creation [#1417](https://github.com/gonka-ai/gonka/pull/1417) by @akup.
- Fix a production host leak where failed session resolution left validation workers alive [#1417](https://github.com/gonka-ai/gonka/pull/1417) by @gmorgachev.
- Collect gateway v3 fixes: runtime versioning, observability, escrow rotation, request parameter validation, SSE handling, drain barriers, and gateway v2 cleanup [#1427](https://github.com/gonka-ai/gonka/pull/1427) by @gmorgachev, @qdanik, @libermans, @a-kuprin.
- Distribute unsettled devshard escrow by slot instead of unique address [#1347](https://github.com/gonka-ai/gonka/pull/1347) by @0xMayoor.

## Testing

The devshard v3 runtime was deployed and verified first. During testing, `node` and `api` containers were stopped while devshard traffic was running; active requests were allowed to finish, and new requests were successfully created and executed through `/devshard/v3`.

After devshard v3 validation, the mainnet-style upgrade from `v0.2.13` to `v0.2.14` was tested.

## Contributors (sorted alphabetically)

- @0xgonka
- @0xMayoor
- @a-kuprin
- @akup
- @alancapex
- @d-bogdan-engenious (testing)
- @DimaOrekhovPS
- @GLiberman
- @gmorgachev
- @libermans
- @maria-mitina (testing)
- @ouicate
- @patimen
- @qdanik
- @redstartechno
- @Ryanchen911

## Proposed Bounties
Bounty ID | Sum USDT | Bounty Explanation | GitHub ID
-- | -- | -- | --
RM | 2000 | devshards v3 release management, review of 0.2.14 upgrade, HackerOne reviews | @akup
RM | 6000 | release management, HackerOne reviews | @x0152
RM | 11500 | release management, HackerOne reviews, R&D for MiniMax M2.7, Inference During PoC (including cost of GPUs for testing)  and SMST optimization (PR #1432)| @qdanik
PR #1253 | 1000 | fix: stop stale PoC validation promptly | @ouicate
PR #1255 | 1000 | fix(inference): settle accounts before releasing matured unbonding collateral | @ouicate
PR #1278 | 1000 | Bound event-listener tx queue to stop a single zero-fee attacker from OOMing validator nodes | @ouicate
PR #1100 | 500 | fix: prevent uint64 wrap in settle amount sums | @0xMayoor
PR #1101 | 750 | fix: widen ShouldValidate to uint64 to prevent silent validation skip | @0xMayoor
PR #1347 | 500 | fix(devshard): distribute unsettled escrow per slot, not per unique address | @0xMayoor
PR #1376 | 2000 | vulnerability finding (Bridge block sync) | @0xMayoor
PR #889 | 3000 | On-chain configurable reward claim recipients | @alancapex
PR #998 | 7500 | Implementing maintenance windows | @Ryanchen911 
PR #1307 | 500 | fix(setup-report): avoid query-gas-limit on grant check | @redstartechno
PR #1438 | 5000 | vulnerability report | @Lelouch33 
-- | 1000 | vulnerability report for reward power-cap | @Lelouch33 
PR #1415 | 3000 | vulnerability report and fix | @maksimenkoff 
v0.2.13 | 1000 | Review of the upgrade | @blizko


inference-chain/x/inference/module/confirmation_poc.go
```diff
@@ -441,12 +441,22 @@ func (am AppModule) evaluateConfirmation(
...
+	// maintenance-covered absence from CPoC duties.
+	maintenanceAddrs := am.keeper.CollectActiveMaintenanceAddresses(ctx)
+
 	for _, vw := range epochGroupData.ValidationWeights {
 		addr := vw.MemberAddress
 		ratio, ok := ratios[addr]
 		if !ok {
 			continue
 		}
+		if _, inMaint := maintenanceAddrs[addr]; inMaint {
```
qdanik: Maintenance exemption here is asymmetric: it skips the INACTIVE ratio but not the ConfirmationWeight fold, so a maintenance-covered participant still loses rewards.

This block exempts maintenance-covered participants from the ConfirmationPoCRatio assignment (so they aren't marked INACTIVE), but foldEventReadings above has already lowered their ConfirmationWeight to preserved + measured, and that fold is maintenance-blind. For a participant offline during the window, measured ≈ 0 and preserved only counts whichever of their nodes happened to land in the (also maintenance-blind) preserved sample — so their ConfirmationWeight is reduced, and drops to 0 if none of their nodes are preserved for this event.

That reduction is not cosmetic. Reward effectiveWeight is directly proportional to ConfirmationWeight in bitcoin_rewards.go (effectiveWeight = ConfirmationWeight * Weight / rawTotal), so a host whose maintenance window overlaps a cPoC event still loses rewards for that epoch — up to all of them — despite the exemption. That contradicts the stated intent of this block ("expected to be offline … must not be marked INACTIVE due to maintenance-covered absence"): losing rewards is itself a penalty for the expected absence.

Suggest making the exemption symmetric — also neutralize the ConfirmationWeight lowering for maintenance-covered participants (e.g. preserve their prior ConfirmationWeight for covered epochs), not just the ratio. Maintenance is disabled by default, so this isn't a ship blocker, but it should be fixed before governance enables the feature.

decentralized-api/poc/validator.go
```diff
@@ -417,6 +541,54 @@ func (v *OffChainValidator) ValidateAll(pocStageStartBlockHeight int64, pocStart
...
+		"pocStageStartBlockHeight", pocStageStartBlockHeight)
+	cancel()
+	return true
+}
+
+func shouldStopValidationForStage(state *chainphase.EpochState, pocStageStartBlockHeight int64) bool {
+	if state == nil {
+		return false
+	}
+	return !ShouldAcceptValidatedArtifacts(state) || GetCurrentPocStageHeight(state) != pocStageStartBlockHeight
```
qdanik, line 614: Question on how this reacts to a transient un-sync rather than an actual phase end. shouldStopValidationForStage returns true when !ShouldAcceptValidatedArtifacts(state), and ShouldAcceptValidatedArtifacts returns false whenever state.IsNilOrNotSynced() (phase.go:33-36) — i.e. not only when the validation phase has ended, but also if the phase tracker is momentarily nil / not yet re-synced. Since cancelWhenValidationPhaseEnds cancels the shared context on the first true and never resumes for that stage, would a brief sync hiccup mid-window (e.g. a short RPC/catch-up lag) permanently cancel all in-flight validation for the whole stage on this node, so it abstains on the remaining participants until the next stage?

If that can happen, is it intended — or should a not-synced reading be treated as "wait/retry" and the cancel reserved for an actual phase-end or a stage-height change (GetCurrentPocStageHeight != pocStageStartBlockHeight)? I might be over-estimating how often the tracker goes not-synced mid-phase, so wanted to check.

decentralized-api/poc/artifacts/smst.go
x0152, line 1: Suggestion: what if insertAt didn't mutate nodes in place, but
copied the ~depth nodes along the path instead (path copying)?

```go
// copy the path, share everything else
newNode := &smstNode{left: node.left, right: node.right}
// ... update the one child, count, hash on newNode ...
return newNode
```

Then old roots stay alive, and the early-guard case becomes just:

```go
// at the early checkpoint
s.earlyRoot = s.root   // O(1), shares all unchanged subtrees with the live tree

// serving proofs against the early root
proof := buildProof(s.earlyRoot, nonce)   // O(log n), no disk replay
```

With no rebuilds left, the snapshot cache, the rebuild-from-disk path, the
warm-up and the distinct-count limiter could all be dropped. Hashes and the
committed root format don't change, so no consensus impact, and the change to
Insert itself is tiny

And a thought for later: with a versioned tree the early guard doesn't have to
check at a fixed 1/3 - it could pick an arbitrary version (or even a few) and
verify inclusion against it, which is much harder to game than a known
checkpoint. Needs more thought...

Also worth a look for the future: https://github.com/pokt-network/smt

decentralized-api/poc/validator.go
```diff
@@ -154,6 +180,67 @@ func NewOffChainValidator(
...
+	if v.guard.Enabled() {
+		stage, target, ok := EarlyShareCaptureTarget(&epochState, v.guard.FirstFraction())
+		if ok && epochState.CurrentBlock.Height >= target {
+			go v.guard.MaybeCapture(context.Background(), v.recorder.NewInferenceQueryClient(), stage, target, epochState.CurrentBlock.Height)
+		}
+	}
+}
+
+func (v *OffChainValidator) maybeWarmEarlySnapshot(epochState chainphase.EpochState) {
+	stage, target, ok := EarlyShareCaptureTarget(&epochState, earlyshare.DefaultFirstFraction)
```
x0152: I think it should be v.guard.FirstFraction() instead of earlyshare.DefaultFirstFraction?

decentralized-api/poc/earlyshare/decision.go
```diff
@@ -0,0 +1,92 @@
...
+		if isConfirmation {
+			// Only a confirmation PoC pass is trusted to clear the streak.
+			next.ConsecutiveMisses = 0
+		}
+		// Regular PoC pass: leave the streak untouched.
+		return MissOutcome{VoteNo: false, NewState: next}
+	}
+
+	// Failure in either phase always accrues a miss.
+	next.ConsecutiveMisses = prev.ConsecutiveMisses + 1
```
x0152, line 94: I would add an early return when prev.UpdatedStageHeight == stageHeight to keep miss-streak updates idempotent in all scenarios as a proactive safety fix

decentralized-api/poc/early_guard.go
```diff
@@ -0,0 +1,505 @@
...
+			passed := !d.shareFail && d.share >= threshold
+
+			prev, _, err := g.store.GetGuardState(ctx, addr, modelID)
+			if err != nil {
+				logging.Warn("EarlyShareGuard: guard-state load failed; skipping participant", types.PoC,
+					"stage", stageHeight, "participant", addr, "model", modelID, "error", err)
+				continue
+			}
+			outcome := earlyshare.ApplyMissStreak(prev, passed, isConfirmation, stageHeight)
+			if err := g.store.UpsertGuardState(ctx, outcome.NewState); err != nil {
```
x0152, line 358: Would it make sense to reset (or ignore) miss streaks accumulated in observe when switching to enforce?

decentralized-api/poc/early_guard.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	if len(checkpoints) > 0 {
+		if err := g.store.UpsertCheckpoints(ctx, checkpoints); err != nil {
+			logging.Error("EarlyShareGuard: failed to persist early checkpoints", types.PoC, "stage", stageHeight, "error", err)
+			return
+		}
+	}
+	if err := g.store.MarkStageCaptured(ctx, stageHeight, target, capturedAt, perModel); err != nil {
+		logging.Error("EarlyShareGuard: failed to mark capture run", types.PoC, "stage", stageHeight, "error", err)
```
x0152, line 157: Should an empty capture be treated as failed instead of completed, so it can retry on the next block?

devshard/mlnode/manager.go
```diff
@@ -0,0 +1,255 @@
...
+
+// DefaultCacheTTL is the live-flow window. A model counts as actively served
+// when it has a node observed within this span; pruning runs only for such
+// models (see pruneAll).
+const DefaultCacheTTL = 10 * time.Minute
+
+// DefaultStaleTTL is the node-retirement age: an unobserved node is dropped
+// after this long, and only while its model still has a live flow. Must be
+// >= the fresh window.
+const DefaultStaleTTL = time.Hour
```
a-kuprin, line 18: The devshardd ML-node cache (staleTTL 1h + flow-gated pruning) keeps serving indefinitely while DAPI is down, but the fallback path never counts in-flight requests, so the broker's MaxConcurrent ceiling is bypassed.
When devshardd lives without dapi it can hammer its ML nodes with unbounded concurrency.

The cache was meant to survive DAPI upgrades/short reboots — a long-lived cache is over-engineered for that goal and just widens this window

**It's informational note**, and it is not proposed to change anything, as I think, this situation will never happen when both intersects: dapi is down and a lot of concurrent inferences/validations

x0152:
Three pre-existing gaps (not from this PR), none looks like a real attack on its
own, but the new proof paths touch all three:

1. The client reads proofs but never checks the response signature - so a
broken response can't be pinned on the participant vs network tampering.
The new by-nonce path uses the same unverified responses

2. A participant could serve valid proofs to some validators and 503 the ones
whose sample would catch bad leaves - turning their "no" into a silent abstain

3. A validator isn't excluded from validating itself, so its own weight is a
guaranteed "yes"

None of these looks like a real attack vector on its own but they're not written down anywhere,
might be worth documenting

gmorgachev:
> A participant could serve valid proofs to some validators and 503 the ones
whose sample would catch bad leaves - turning their "no" into a silent abstain

it should vote invalid if can't get 


> A validator isn't excluded from validating itself, so its own weight is a
guaranteed "yes"

i think it's correct, as we are checking against 2/3 of weight, not against 2/3 - validatorWeight

decentralized-api/poc/artifacts/smst.go
gmorgachev: Could you please experiment with such approach? Let's say 500_000 nonces, 1_000_000 nonces, 10_000_000
Looking at RAM and time to access proof first time (right now it's re-build time)

decentralized-api/poc/validator.go
gmorgachev: I'd fix that

inference-chain/x/inference/module/confirmation_poc.go
gmorgachev: Overall, maintainance window is disabled for now. So not fthe first priority but seems like we need to fix

devshard/mlnode/manager.go
gmorgachev: > but the fallback path never counts in-flight requests

could you add this one?

i made it work indefinitely for the long term strategy to be close to 99.9% SLA

inference-chain/x/inference/module/confirmation_poc.go
Ryanchen911: Thanks for catching. It will be fixed before governance enables the feature. I'll make the exemption symmetric — preserving the prior ConfirmationWeight for maintenance-covered participants over covered epochs, not just the ratio — and add a regression test, in a separate follow-up PR.

Quick question @gmorgachev : should I target that fix at upgrade-v0.2.14, or against main / the next release branch once v0.2.14 merges? Happy to align with whatever's cleanest for the release.

devshard/mlnode/manager.go
a-kuprin: I think it could be added to next release, and is not a blocker

I will add it

a-kuprin:
> 3\. A validator isn't excluded from validating itself, so its own weight is a
>     guaranteed "yes"

It was discussed multiple times at H1.
The main point, as for me - participant always can farm a lot of addresses and distribute it's weight, and self-validate. So hardening here gives nothing to real security and it's ok to keep current behavior when participant can self-validate. Anyway he will need 2/3 of total weight to make it be the attack

a-kuprin:
There was a host leak fix (400GB RSS production incident) Commit: ae2b525 — merge cache manager (#1417)

For every escrow it is creating 20 go routings for validation, if there are N escrows there are 20*N validation go routines. But anyway concurrent execution is limited by maxConcurrent at ML node manager level, and mostly all routines are at stale state as they are awaiting results from ML nodes.

I think it's better to use shared validation routines pool. Share across all escrows, for one devshardd process.

It's also not a blocker and could be added to next release

a-kuprin:
Routes and chain msgs are disabled for old inference path, but related files and chain queries are kept. Maybe we should cleanup them, or this is deferred and scheduled to next releases?

a-kuprin:
decentralized-api isn't listening `maintenance_scheduled` / `maintenance_canceled` events, and doesn't provide api endpoint to expose them.

I think it's easy to add it to current release, so gateway can poll it from dapi to support maintenance at gateway release and don't wait for next 0.2.15 release

inference-chain/x/inference/module/confirmation_poc.go
DimaOrekhovPS: @Ryanchen911 I suggest the following fix. Can you take a look and review: https://github.com/gonka-ai/gonka/pull/1438/changes/a6902931e55cda3ae2c3a40a5e77370a08fca02f

decentralized-api/poc/validator.go
DimaOrekhovPS: Fixed: https://github.com/gonka-ai/gonka/pull/1438/changes/1ccbcff7e88ab1c01ec43321c0f6b5b1e4bd5e26

decentralized-api/poc/validator.go
DimaOrekhovPS: Good catch
Fixed: https://github.com/gonka-ai/gonka/pull/1438/changes/0a2ce1a36b49a4663f4fa6b9b9d4af5a068d5779

DimaOrekhovPS:
@a-kuprin 

> Routes and chain msgs are disabled for old inference path, but related files and chain queries are kept. Maybe we should cleanup them, or this is deferred and scheduled to next releases?

Yes, this upgrade is the first step

decentralized-api/poc/early_guard.go
DimaOrekhovPS: the capture query is pinned to the target block height. The chain's state at a given height is frozen forever. So retrying won't yield any different results

inference-chain/x/inference/module/confirmation_poc.go
Ryanchen911: Reviewed — LGTM. This is exactly the symmetric exemption qdanik and I discussed: hoisting maintenanceAddrs into foldEventReadings as skipAddrs protects both the ConfirmationWeight fold and the ratio in one place, so a maintenance-covered host no longer loses effectiveWeight (proportional to ConfirmationWeight in bitcoin_rewards.go) for a covered epoch. Nice cleanup collapsing the now-redundant downstream if inMaint block too. The new TestFoldEventReadings_MaintenanceExemptSkipsWeightAndRatio covers the key invariant (weight preserved + no ratio for the maintenance addr, online host still folded). Thanks for picking this up. Confirming base = upgrade-v0.2.14 works for me.

decentralized-api/poc/early_guard.go
DimaOrekhovPS: I think it's ok to not reset. Reset mechanic probably requires a new postrgres relation, to persist the prev value. In my opinion it's too much state handling for this case

the default setting in enforced anyway + I don't think it's obvious that we need to reset the counts from the observe mode, it's still valuable history

decentralized-api/poc/earlyshare/decision.go
DimaOrekhovPS: Fixed: https://github.com/gonka-ai/gonka/pull/1438/changes/be08b4abd1a9a3db28d3f6366203cc145076e8e2

decentralized-api/poc/early_guard.go
DimaOrekhovPS: Made sure the default setting is enforce: https://github.com/gonka-ai/gonka/pull/1438/changes/5621f454cdd7f9e39c7f8d0621857bc47ef00cc2

Ryanchen911:
The chain already exposes this via query endpoints, not just events — `MaintenanceScheduled` and `MaintenanceActive` (plus `MaintenanceStatus`/`MaintenanceSchedulability`) are all in `query_maintenance.go` and available over gRPC/REST in this release. So the gateway can poll maintenance state today without waiting for 0.2.15.

You're right that decentralized-api itself doesn't subscribe to the `maintenance_scheduled` / `maintenance_canceled` events or re-expose them. If the gateway can query the chain node directly, it can use the query endpoints above as-is. If the gateway is constrained to talk only to dapi, I can add a thin dapi passthrough endpoint that proxies `MaintenanceScheduled`/`MaintenanceActive` — small change, happy to include it in this release.

@a-kuprin which is it on the gateway side — can it hit the chain query API directly, or does it need a dapi-side endpoint?

decentralized-api/poc/artifacts/smst.go
a-kuprin: I have another suggestion.

We currently insert nodes one by one and on each insert are recalculating affected path in a Merkle tree.
But nonces comes in a batch, and commit is sent for the batch.

When we are inserting batch there are always for each insert shared nodes in affected path (especially near root). 

So it should be faster to **Insert all, then rehash affected once**
1. Place all new leaves (structure / counts) in a batch, mark nodes on their paths dirty
2. Rehash each dirty node once, bottom-up

I think this optimization could be easily added

a-kuprin: > And a thought for later: with a versioned tree the early guard doesn't have to
> check at a fixed 1/3 - it could pick an arbitrary version (or even a few) and
> verify inclusion against it, which is much harder to game than a known
> checkpoint. Needs more thought...

Don't we already have this ability? We can ask for proof on any count and tree will be build for the requested count by stored ordered artifacts

x0152: > I have another suggestion.
> 
> We currently insert nodes one by one and on each insert are recalculating affected path in a Merkle tree. But nonces comes in a batch, and commit is sent for the batch.
> 
> When we are inserting batch there are always for each insert shared nodes in affected path (especially near root).
> 
> So it should be faster to **Insert all, then rehash affected once**
> 
> 1. Place all new leaves (structure / counts) in a batch, mark nodes on their paths dirty
> 2. Rehash each dirty node once, bottom-up
> 
> I think this optimization could be easily added

Yes, agree, the same mechanic (along with all the suggestions above) is already implemented #1432

x0152: > > And a thought for later: with a versioned tree the early guard doesn't have to
> > check at a fixed 1/3 - it could pick an arbitrary version (or even a few) and
> > verify inclusion against it, which is much harder to game than a known
> > checkpoint. Needs more thought...
> 
> Don't we already have this ability? We can ask for proof on any count and tree will be build for the requested count by stored ordered artifacts

Technically any count is possible, but in practice it is capped because cache misses cause O(N) rebuilds and the limiter protects against dos (in the current implementation)

a-kuprin: @gmorgachev Are we ready to merge https://github.com/gonka-ai/gonka/pull/1432 into this upgrade?

x0152:
> > 3. A validator isn't excluded from validating itself, so its own weight is a
> > guaranteed "yes"
> 
> It was discussed multiple times at H1. The main point, as for me - participant always can farm a lot of addresses and distribute it's weight, and self-validate. So hardening here gives nothing to real security and it's ok to keep current behavior when participant can self-validate. Anyway he will need 2/3 of total weight to make it be the attack

My suggestion in the comment above was that maybe we should document these three assumptions explicitly, since they’re easy to miss during review - not to question why it works this way or to change it

a-kuprin:
> The chain already exposes this via query endpoints, not just events — `MaintenanceScheduled` and `MaintenanceActive` (plus `MaintenanceStatus`/`MaintenanceSchedulability`) are all in `query_maintenance.go` and available over gRPC/REST in this release. So the gateway can poll maintenance state today without waiting for 0.2.15.
> 
> You're right that decentralized-api itself doesn't subscribe to the `maintenance_scheduled` / `maintenance_canceled` events or re-expose them. If the gateway can query the chain node directly, it can use the query endpoints above as-is. If the gateway is constrained to talk only to dapi, I can add a thin dapi passthrough endpoint that proxies `MaintenanceScheduled`/`MaintenanceActive` — small change, happy to include it in this release.
> 
> @a-kuprin which is it on the gateway side — can it hit the chain query API directly, or does it need a dapi-side endpoint?

Yes I've seen the exposed query endpoints and it works.
My point is that gateway can fetch them, but it is every 5 second fetch, while dapi has already event listener and listens for block updates, and it's just nicer architecturally that dapi already fetched the chain and can poll it for gateway.

Target design is proposed here: https://github.com/gonka-ai/gonka/discussions/1367

It's not an issue, but some polishing

decentralized-api/poc/artifacts/smst.go
x0152: from my mind, merging it into this upgrade could be risky, so it needs additional testing. I'd rather ship it in the next update

devshard/mlnode/manager.go
a-kuprin: @gmorgachev 
I've finished the review and if we do not start voting on this release in a few hours, I can add this today. Anyway I will prepare PR soon

decentralized-api/poc/artifacts/smst.go
DimaOrekhovPS: I agree with @x0152, seems risky. @qdanik, the author of the PR is also aligned that the approach needs more testing

a-kuprin:
I have created follow-up PR related to PR comments:
https://github.com/gonka-ai/gonka/pull/1443

I think it is safe to merge without testing, as it only adds functionality for decentralized-api and doesn't touch life code paths. Devshard part can go in separate devshard-only release.

Added:
- Listens for chain escrow (created/settled) and maintenance (scheduled/canceled) events into a bounded HostEventRing
- Exposes GetHostEvents long-poll (cursor + subscribe set + max wait); clients get only the kinds they subscribe to
Subscribe-filtered wakes — unsubscribed kinds don’t wake waiters / reset deadlines

Before devshardd is initing lazily on first served inference, and it asks chain node for escrow data. When escrow is created but no inferences was sent, and node is down (or during upgrade) if first inference to devshard will be send it will fail.
Now even if node is down devshard already will have the data long-polled from dapi on escrow creation. It's making serving inferences stability during upgrade even better

Also added discussed:
- When dapi is down fallback path counts in-flight requests and takes into account average escrows loads
- decentralized-api listens for maintenance events from chain and can serve maintenance state via long-polling

devshard/mlnode/manager.go
a-kuprin: Done
https://github.com/gonka-ai/gonka/pull/1443

I think it is safe to merge without testing, as it only adds functionality for decentralized-api and doesn't touch life code paths. Devshard part can go in separate devshard-only release.

decentralized-api/internal/server/admin/server.go
```diff
@@ -59,6 +61,7 @@ func NewServer(
 	}
 
 	e.Use(middleware.LoggingMiddleware)
+	e.Any("/debug/pprof/*", echo.WrapHandler(http.DefaultServeMux))
```
blizko, line 64: Is it safe to leave this as a public route?

gmorgachev: It's on admin api, private network

DimaOrekhovPS (approved):

GLiberman (approved):

PR #1270, opened 2026-05-28
fix(validation): iterate all logprobs.content per chunk in GetEnforcedTokens

qdanik:
## Summary

Single validation-path fix for the devshard inference validator, targeting production `similarity_below` / `inflated_tokens` / `different_length` rejections on responses that were actually valid.

**Capture every token in a multi-content SSE chunk.** `StreamedCompletionResponse.GetEnforcedTokens` only read `Logprobs.Content[0]` per chunk. A single `chat.completion.chunk` can carry **multiple** tokens in `logprobs.content`

## Why

On the validation route, valid responses were rejected with `similarity_below`, `inflated_tokens`, and `different_length`. Root cause: whenever an executor emitted more than one token in a single SSE chunk, `GetEnforcedTokens` kept only `Content[0]` and dropped the rest, so the enforced replay was 1–N tokens shorter than the original. A single dropped token mid-stream shifts every position after it, which simultaneously tanks similarity and inflates the token-count delta — both symptoms trace to the same dropped tokens.

Reproduced end-to-end on the live fleet: a multi-content chunk carrying two tokens (`" parameter"`, `" name"`) in one `logprobs.content` produced `enforced=N-1` and an `inflated_tokens` gap of 1 against the executor's count.

## Changes

- `completionapi/completionresponse.go` — `StreamedCompletionResponse.GetEnforcedTokens` iterates all `logprobs.content` entries per chunk instead of only the first.
- `completionapi/enforced_multitoken_test.go` (new) — regression tests: a multi-token-per-chunk stream contributes every token to the enforced sequence, and the enforced-token count equals the `ExtractLogits` position count.
- `internal/devshard/shared_runtime.go` — removed the unused `rewriteRequest` (`2×` `max_tokens` bump); `BuildValidationBody` no longer rewrites `max_tokens`. Added a comment noting that enforced replay appends a terminator (response one token longer) unless the sequence already ends on a stop token.

> The `max_tokens` context-window clamp explored earlier in this branch was reverted; this PR is scoped to the multi-content fix only.

## Tests

- [x] Multi-content chunk contributes all tokens to enforced sequence
- [x] Enforced-token count matches `ExtractLogits` position count
- [x] `go build ./...` / `go vet ./...` green under `decentralized-api/`


decentralized-api/cmd/devshardd/validation.go
```diff
@@ -57,9 +64,36 @@ func (v *devshardValidator) Validate(ctx context.Context, req devshardpkg.Valida
...
+// Known models are cached for the process lifetime; unknown models cause a
+// fresh ModelsAll query so newly-added governance models get picked up
+// without a restart.
+func (v *devshardValidator) modelContextWindow(ctx context.Context, model string) uint64 {
+	v.contextMu.Lock()
+	if cw, ok := v.contextWindows[model]; ok {
+		v.contextMu.Unlock()
+		return cw
+	}
+	v.contextMu.Unlock()
```
a-kuprin: It is hot-path and mostly read that can be done without locking when nothing updates the map (usual in this case)
It is better to use `sync.RWMutex` for `contextMu`

```go
v.contextMu.RLock()
cw, ok := v.contextWindows[model]
v.contextMu.RUnlock()
if ok {
    return cw
}
```

qdanik: done

decentralized-api/internal/devshard/shared_runtime.go
```diff
@@ -178,21 +179,41 @@ func ProcessExecutionHTTPResponse(
...
 }
 
-func rewriteRequest(requestMap map[string]interface{}, enforcedTokens completionapi.EnforcedTokens) {
-	minMaxTokens := uint64(len(enforcedTokens.Tokens) * 2)
-	if maxTokens, ok := devshardpkg.JSONNumericUint64(requestMap["max_tokens"]); ok && minMaxTokens > maxTokens {
-		requestMap["max_tokens"] = minMaxTokens
+// contextWindowSafetyMargin reserves a few tokens of headroom so the
+// recomputed max_tokens never lands exactly at the model context limit
+// (vLLM rejects when prompt + max_tokens > max_model_len).
+const contextWindowSafetyMargin uint64 = 8
```
a-kuprin: subject to remove.

If it is > 0, There is possible attack when inputTokens + outputTokens == contextWindow

PR #1272, opened 2026-05-28
perf: gRPC caching

x0152:
Supersedes #542

## Summary

Implemented a gRPC query cache in DAPI:
- Cache keying by (block_height, method, request_hash)
- Block-pinned reads via x-cosmos-block-height for consistent per-block queries
- Height hint fallback for unpinned requests
- Explicit cache bypass for EpochInfo to avoid stale network state
- Query cache stats support in the client/admin layer

## Endpoints

- `GET /admin/v1/cache/stats`
- `POST /admin/v1/cache/stats/reset`

## Load test

Tested under load (localtestnet , ~5s blocks). The cache stores results per block, so many identical requests in the same block are served from one backend call (extra ones are deduplicated).

1000 concurrent requests per endpoint (bypassing the proxy, ~5s blocks, cache memory limit = 1 GiB):

| Endpoint | gRPC calls | cache hits | backend calls | cache size | wall (ON) | wall (OFF) |
|---|---|---|---|---|---|---|
| `/v1/governance/models` | 1000 | 999 | 1 | ~3.5 KB | 45 ms | 223 ms |
| `/v1/models` | 1000 | 999 | 1 | ~3.6 KB | 37 ms | 196 ms |
| `/v1/governance/pricing` | 3000 | 2999 | 1 | ~3.6 KB | 37 ms | 536 ms |
| `/v1/pricing` | 1000 | 1000 | 0 | ~3.6 KB | 35 ms | 223 ms |
| `/v1/participants` | 1000 | 999 | 1 | ~3.8 KB | 34 ms | 175 ms |

The latency gain is small on the local testnet (tiny state, backend ~150 ms). On prod (bigger state, real load) it will be more noticeable.

## Out of scope (follow-up PR)

`/v1/status` and ABCI queries are not cached - they use CometBFT RPC, not the gRPC layer. Caching them needs a separate cache, so it's left for a follow-up pr

### DIFF
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -79,14 +79,15 @@ type ApiConfig struct {
 }
 
 type ChainNodeConfig struct {
-	Url              string `koanf:"url" json:"url"`
-	IsGenesis        bool   `koanf:"is_genesis" json:"is_genesis"`
-	SeedApiUrl       string `koanf:"seed_api_url" json:"seed_api_url"`
-	AccountPublicKey string `koanf:"account_public_key" json:"account_public_key"`
-	SignerKeyName    string `koanf:"signer_key_name" json:"signer_key_name"`
-	KeyringBackend   string `koanf:"keyring_backend" json:"keyring_backend"`
-	KeyringDir       string `koanf:"keyring_dir" json:"keyring_dir"`
-	KeyringPassword  string `json:"-"`
+	Url               string `koanf:"url" json:"url"`
+	IsGenesis         bool   `koanf:"is_genesis" json:"is_genesis"`
+	SeedApiUrl        string `koanf:"seed_api_url" json:"seed_api_url"`
+	AccountPublicKey  string `koanf:"account_public_key" json:"account_public_key"`
+	SignerKeyName     string `koanf:"signer_key_name" json:"signer_key_name"`
+	KeyringBackend    string `koanf:"keyring_backend" json:"keyring_backend"`
+	KeyringDir        string `koanf:"keyring_dir" json:"keyring_dir"`
+	KeyringPassword   string `json:"-"`
+	QueryCacheEnabled bool   `koanf:"query_cache_enabled" json:"query_cache_enabled"`
 	// MinGasPriceNgonka is the gas price in ngonka used for transaction fee calculation.
 	//
 	// When set to a non-zero value, the DAPI uses it as-is (this is an override
--- a/decentralized-api/apiconfig/config.go
+++ b/decentralized-api/apiconfig/config.go
@@ -79,17 +79,17 @@ type ApiConfig struct {
 }
 
 type ChainNodeConfig struct {
-	Url               string `koanf:"url" json:"url"`
-	IsGenesis         bool   `koanf:"is_genesis" json:"is_genesis"`
-	SeedApiUrl        string `koanf:"seed_api_url" json:"seed_api_url"`
-	AccountPublicKey  string `koanf:"account_public_key" json:"account_public_key"`
-	SignerKeyName     string `koanf:"signer_key_name" json:"signer_key_name"`
-	KeyringBackend    string `koanf:"keyring_backend" json:"keyring_backend"`
-	KeyringDir        string `koanf:"keyring_dir" json:"keyring_dir"`
-	KeyringPassword   string `json:"-"`
-	QueryCacheEnabled bool   `koanf:"query_cache_enabled" json:"query_cache_enabled"`
-	QueryCacheMaxEntries int `koanf:"query_cache_max_entries" json:"query_cache_max_entries"`
-	QueryCacheMaxBytes int64 `koanf:"query_cache_max_bytes" json:"query_cache_max_bytes"`
+	Url                  string `koanf:"url" json:"url"`
+	IsGenesis            bool   `koanf:"is_genesis" json:"is_genesis"`
+	SeedApiUrl           string `koanf:"seed_api_url" json:"seed_api_url"`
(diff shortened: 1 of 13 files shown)


a-kuprin:
# Performance

Single global mutex on the hot read path.
Every cache hit takes the full sync.Mutex because lookup mutates the LRU list:

```go
func (c *QueryCache) lookup(height int64, key string) ([]byte, bool) {
	c.mu.Lock()
	defer c.mu.Unlock()
	...
	c.lru.MoveToFront(el)   // write under lock on every read
	return v, true
}
```

The whole point of this cache is to absorb bursts (the PR tests 1000 concurrent requests/endpoint), but every hit serializes on one lock — including the MoveToFront pointer surgery and a map lookup on `lruIndex`. An `RWMutex` won't help because reads write the list.

**LRU recency is mostly wasted work**

With `keepLast = 3` heights and 200k-entry / 1 GiB limits, real-world eviction is almost always height pruning, **not LRU**. So the per-read MoveToFront cost buys you almost nothing in the common case — it's pure contention with no payoff. Consider dropping read-side reordering at all.

x0152:
> # Performance
> Single global mutex on the hot read path. Every cache hit takes the full sync.Mutex because lookup mutates the LRU list:
> 
> ```go
> func (c *QueryCache) lookup(height int64, key string) ([]byte, bool) {
> 	c.mu.Lock()
> 	defer c.mu.Unlock()
> 	...
> 	c.lru.MoveToFront(el)   // write under lock on every read
> 	return v, true
> }
> ```
> 
> The whole point of this cache is to absorb bursts (the PR tests 1000 concurrent requests/endpoint), but every hit serializes on one lock — including the MoveToFront pointer surgery and a map lookup on `lruIndex`. An `RWMutex` won't help because reads write the list.
> 
> **LRU recency is mostly wasted work**
> 
> With `keepLast = 3` heights and 200k-entry / 1 GiB limits, real-world eviction is almost always height pruning, **not LRU**. So the per-read MoveToFront cost buys you almost nothing in the common case — it's pure contention with no payoff. Consider dropping read-side reordering at all.

Good catch, missed it. Updated the table

Also fixed stale reads in the cache

a-kuprin (approved):

patimen:
@x0152 Did you consider using golang-lru or some other cache library here instead of implementing the cache logic by hand? The block-height logic is specific to Gonka, but the LRU/eviction/concurrency parts look like generic cache machinery.

x0152:
> @x0152 Did you consider using golang-lru or some other cache library here instead of implementing the cache logic by hand? The block-height logic is specific to Gonka, but the LRU/eviction/concurrency parts look like generic cache machinery.

Yes, agree - rewrote the generic cache machinery to use otter/v2 (better fit here then golang-lru). 

ran the benchmark and got results a bit faster than the hand-written version

snevolin:
Looks like the cache never fills up when running.

The DAPI with this configuration does not use a gRPC connection to the chain. Ignite's cosmosclient builds the client.Context with WithClient(c.RPC), not WithGRPCClient. On the ABCI path the read path keys entries by the height we already know (the pin, or the hint), but the write path keys them by the height the node reports back in the response header.

Before 77206b721 the cache put the block height into the outgoing request itself:

    pinnedCtx = setPinnedHeight(ctx, height)
    ...
    c.invokeBackend(pinnedCtx, method, args, reply, callOpts...)

The node echoed that height back, so `responseHeight` matched the hint the read path uses, entries were stored, and the load test in the description was accurate.

After 77206b721 you addded `hintTrusted := c.cache.hintFresh()`, but the write path was left unchanged. It still takes the key from the response header (query_cache.go:316), and now nothing puts a height in the request, so there is nothing for the node to echo back. On CometBFT RPC the response height is just whatever the request contained (baseapp/grpcrouter.go:125), so the header comes back as 0 and store() is skipped — no entry is ever written, and the read path at query_cache.go:281 never finds anything.

To confirm: run the load test again and check /admin/v1/cache/stats. If cache_write_total is 0 and cache_write_skipped_height_total is high, this is it.

x0152:
> Looks like the cache never fills up when running.
> 
> The DAPI with this configuration does not use a gRPC connection to the chain. Ignite's cosmosclient builds the client.Context with WithClient(c.RPC), not WithGRPCClient. On the ABCI path the read path keys entries by the height we already know (the pin, or the hint), but the write path keys them by the height the node reports back in the response header.
> 
> Before [77206b7](https://github.com/gonka-ai/gonka/commit/77206b721142b361825c32d583ab9db4c98a130f) the cache put the block height into the outgoing request itself:
> 
> ```
> pinnedCtx = setPinnedHeight(ctx, height)
> ...
> c.invokeBackend(pinnedCtx, method, args, reply, callOpts...)
> ```
> 
> The node echoed that height back, so `responseHeight` matched the hint the read path uses, entries were stored, and the load test in the description was accurate.
> 
> After [77206b7](https://github.com/gonka-ai/gonka/commit/77206b721142b361825c32d583ab9db4c98a130f) you addded `hintTrusted := c.cache.hintFresh()`, but the write path was left unchanged. It still takes the key from the response header (query_cache.go:316), and now nothing puts a height in the request, so there is nothing for the node to echo back. On CometBFT RPC the response height is just whatever the request contained (baseapp/grpcrouter.go:125), so the header comes back as 0 and store() is skipped — no entry is ever written, and the read path at query_cache.go:281 never finds anything.
> 
> To confirm: run the load test again and check /admin/v1/cache/stats. If cache_write_total is 0 and cache_write_skipped_height_total is high, this is it.

Thanks, valid point!

I rechecked this in code and agree it's a real regression after https://github.com/gonka-ai/gonka/commit/77206b721142b361825c32d583ab9db4c98a130f

PR #1275, opened 2026-05-29
fix(inference): refund client escrow for VOTING inferences on timeout (quorum miss)

### LINKED ISSUE #1265: Stuck VOTING inferences orphan client escrow when x/group proposals miss quorum
## Summary

`expireInferences` (`inference-chain/x/inference/module/module.go:226-234`) filters by `Status == STARTED` only. When a failing `MsgValidation` transitions an inference to `VOTING` and the resulting x/group proposals don't reach quorum, the inference is silently skipped by the timeout cleanup, the timeout entry is removed unconditionally at line 391, and the client's escrow is permanently held in the inference module account.

Surfaced 2026-05-27 by a multi-seed sim sweep (related to #982 Phase 3 work) — seed 99 triggered a custom `NoStuckVoting` invariant: an inference from epoch 0 was still in `Status=VOTING` at epoch 10.

## Reproducer

A black-box test driving the real `expireInferences` is in #1275:
`TestExpireInferences_VotingInferenceRefundedOnTimeout`
(`inference-chain/x/inference/module/expire_voting_stuck_test.go`).

It seeds a `VOTING` inference plus a matching `InferenceTimeout`, then calls the
real (unexported) `expireInferences` through an `export_test.go` wrapper with a
mocked bank keeper. On current `main` it fails — the refund is never issued and
the inference stays `VOTING`; with the fix in #1275 it passes (inference marked
`EXPIRED`, client escrow refunded).

## Failure mode

1. Client calls `MsgStartInference` — escrow ngonka to inference module account, `Status=STARTED`, `InferenceTimeout` queued at `start_block + ExpirationBlocks`.
2. Executor calls `MsgFinishInference` — `Status=FINISHED`.
3. Validator submits failing `MsgValidation` (`msg_server_validation.go:178`) — `Status=VOTING`, two x/group proposals (invalidate, revalidate) created via `submitValidationProposalsWithPolicy` at line 280.
4. x/group voting window closes without quorum. Neither proposal reaches majority. x/group `EndBlock` tallies and prunes but does not auto-execute — execution requires `MsgExec` or a vote with `Exec_EXEC_TRY` (`cosmos-sdk/x/group/keeper/msg_server.go:771-777`).
5. Block reaches `InferenceTimeout.ExpirationHeight`. EndBlocker calls `expireInferences` (`module.go:208`).
6. Filter at `module.go:231` only handles `Status == STARTED`:

   ```go
   if inference.Status == types.InferenceStatus_STARTED {
       am.handleExpiredInferenceWithContext(...)
   }
   ```

   VOTING falls through. No refund.
7. `RemoveInferenceTimeout` (`module.go:391`) unconditionally removes the timeout entry.
8. Inference now permanently `VOTING`, client escrow stuck in inference module account.

## Severity

Liveness for client funds. Not a chain halt — block processing continues. But on every failing `MsgValidation` that misses x/group quorum within the voting window, the client's escrow is lost permanently. The trigger (validators not reaching majority within the voting window) is a routine production condition — network lag, validator restart, split votes — not an exotic edge case.

## Minimal recommended fix

Extend the filter at `module.go:231` to also handle `VOTING`:

```go
switch inference.Status {
case types.InferenceStatus_STARTED:
    am.handleExpiredInferenceWithContext(ctx, inference, expiryCtx)
case types.InferenceStatus_VOTING:
    // Voting window expired without consensus. Refund client, mark
    // EXPIRED, leave x/group proposals to be pruned by x/group EndBlock.
    am.expireInferenceAndIssueRefund(ctx, inference)
}
```

Smallest change that restores liveness. Implements default-to-refund semantics on quorum miss: when validators don't reach consensus, client gets escrow back, executor receives no payment, no slashing.

## Alternative semantics (for separate discussion)

The minimal fix picks one of several reasonable semantics. The following alternatives are worth considering as a separate design discussion (not bundled with this fix):

- **Re-vote on quorum miss.** Open a new proposal pair with current epoch members; retry up to N times; fall back to refund if still no quorum. Maximizes consensus chance but introduces new state (retry counter), a new governance param, proposer-identity handling on retry (the invalidate proposer is the invalidator; the revalidate proposer is the executor, who may have been dropped from the active set), and would require a proposal-close API not currently in the inference module's group interface.
- **Default-to-invalidate.** Quorum miss = treat as invalidated. Protects client but penalizes executor for validators' technical failures (offline, network lag).

Not recommended: default-to-validate (passive non-voting becomes implicit approval — attack vector); slash non-voting validators (DoS vector via spurious failing validations).

## Open questions for maintainers

- Is default-to-refund acceptable on quorum miss, or do you prefer a different fallback (re-vote, default-to-invalidate)?
- If re-vote is preferred, that probably belongs in a separate design issue — happy to draft it once the immediate liveness gap is closed.

---

@patimen — touching code you originally wrote (`module.go:231`, commit `2f33567dd7`). Flagging directly since you'd have the most context on the intended semantics here.

vitaly-andr:
## Problem

`expireInferences` (`x/inference/module/module.go`) filtered timed-out inferences by `Status == STARTED` only. A failing `MsgValidation` moves an inference to `VOTING`; if its x/group proposals miss quorum within the voting window, the timeout cleanup silently skips it — the `InferenceTimeout` entry is removed, the inference stays `VOTING` forever, and the client's escrow is stranded in the inference module account. The quorum-miss trigger (network lag, validator restart, split vote) is a routine production condition, not an exotic edge case.

## Fix

Add a `VOTING` branch: on timeout, mark the inference `EXPIRED` and refund the client via `expireInferenceAndIssueRefund` (default-to-refund on quorum miss — client made whole, executor unpaid, no slashing). The x/group proposals are left to be pruned by x/group `EndBlock`.

## Test

`TestExpireInferences_VotingInferenceRefundedOnTimeout` drives the real `expireInferences` (via an `export_test.go` wrapper, since it is unexported) with one `VOTING` timeout — it fails on current `main` (no refund, inference stays `VOTING`) and passes with the fix.

## Context

Discovered while working on #998 (maintenance windows); the reproducer was built with the inference-chain simulation/fuzz harness (#982), where a `NoStuckVoting` invariant flagged an inference still `VOTING` at epoch 10.

Closes #1265.

cc @patimen (#982 author)

### DIFF
--- a/inference-chain/x/inference/module/expire_voting_stuck_test.go
+++ b/inference-chain/x/inference/module/expire_voting_stuck_test.go
@@ -1,0 +1,87 @@
+package inference_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+
+	keepertest "github.com/productscience/inference/testutil/keeper"
+	"github.com/productscience/inference/testutil/sample"
+	inference "github.com/productscience/inference/x/inference/module"
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// TestExpireInferences_VotingInferenceRefundedOnTimeout exercises the real
+// expireInferences filter (module.go) and asserts that a VOTING inference
+// whose timeout fires is refunded and marked EXPIRED.
+//
+// Bug (pre-fix): the filter handled Status == STARTED only. When a failing
+// MsgValidation moves an inference to VOTING and the resulting x/group
+// proposals miss quorum, the timeout cleanup silently skips it — the timeout
+// entry is removed but the inference stays VOTING forever and the client's
+// escrow is stranded in the inference module account.
+//
+// Without the fix this test FAILS twice over: the BankKeeper refund call is
+// never made (unmet gomock expectation) and the inference stays VOTING. With
+// the fix the VOTING branch calls expireInferenceAndIssueRefund → refund +
+// EXPIRED.
+//
+// expireInferences is unexported, so the test drives it through the
+// ExpireInferencesForTest wrapper (export_test.go). currentEpoch is passed as
+// nil: the VOTING path never consults the epoch, and
+// GetLatestPoCOrCPoCRangeWithData short-circuits to an inactive PoC range when
+// currentEpoch == nil — so the expiry context builds without any epoch/PoC
+// state and the real STARTED/VOTING filter runs.
+func TestExpireInferences_VotingInferenceRefundedOnTimeout(t *testing.T) {
+	k, sdkCtx, mocks := keepertest.InferenceKeeperReturningMocks(t)
+	am := inference.NewAppModule(nil, k, nil, nil, nil, nil)
+
+	// Refund path: IssueRefund -> PayParticipantFromEscrow ->
+	// SendCoinsFromModuleToAccount (+ best-effort LogSubAccountTransaction).
+	// Times(1) is the fail-without-fix guard: pre-fix the VOTING inference is
+	// skipped and this expectation goes unmet.
+	mocks.BankKeeper.EXPECT().
+		SendCoinsFromModuleToAccount(gomock.Any(), types.ModuleName, gomock.Any(), gomock.Any(), gomock.Any()).
+		Return(nil).Times(1)
+	mocks.BankKeeper.EXPECT().LogSubAccountTransaction(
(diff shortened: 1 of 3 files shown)


vitaly-andr:
Closing: the code path this fixes was removed from the chain.

PR #1276, opened 2026-05-29
fix(inference): skip revalidation votes when voter is not an epoch-group member

### LINKED ISSUE #1269: x/inference: revalidation vote fails when voter absent from epoch x/group
# x/inference: revalidation vote fails when voter absent from epoch x/group

## Summary

`MsgValidation{Revalidation:true}` for an inference whose VOTING window
is still open fails with `voter address: <bech32>: not found` when the
submitter is in `ActiveParticipantsSet[inference.EpochId]` but is NOT a
member of the x/group that owns
`inference.ProposalDetails.ReValidatePolicyId` **at the moment the
second of two sequential votes is dispatched**.

The handler issues two `k.group.Vote` calls back-to-back (Invalidate
then Revalidate). Either vote can hit a state where the voter is
missing from the group. Two independent mechanisms can produce that
state — one demonstrated by sim, one plausible from code inspection.

Related to #1265 (manifestation #1: VOTING-stuck escrow on timeout,
narrow fix already discussed there).

## Reproduction (simulation)

Two prerequisites:

1. **Sim infrastructure** — check out PR #1228 (Phase 3 of #982; adds
   `make sim-full-test`, simsx wiring, first-wave `x/inference`
   operations):

   ```bash
   gh pr checkout 1228
   ```

2. **Manifestation #1 production fix** — apply this local patch. Without
   it sim-full fails earlier on the timeout-stuck path (#1265) and
   never reaches the vote race below:

   ```diff
   --- a/inference-chain/x/inference/module/module.go
   +++ b/inference-chain/x/inference/module/module.go
   @@ -228,6 +228,11 @@ func (am AppModule) expireInferences(
    		if inference.Status == types.InferenceStatus_STARTED {
    			am.handleExpiredInferenceWithContext(ctx, inference, expiryCtx)
   +		} else if inference.Status == types.InferenceStatus_VOTING {
   +			// VOTING inferences whose x/group proposals missed quorum
   +			// would otherwise be silently dropped here, stranding client
   +			// escrow in the inference module account.
   +			am.expireInferenceAndIssueRefund(ctx, inference)
    		}
    	}
    	return nil
   ```

Then from `inference-chain/`, run sim-full in canonical Docker per
`docs/simulation.md` **with `-Verbose=true`** so the intra-tx trace is
captured:

```bash
go test -mod=readonly -tags "sims muslc" ./app/ \
  -run "TestFullAppSimulation" \
  -Enabled=true -Commit=true -Verbose=true \
  -NumBlocks=500 -BlockSize=200 -Seed=99 -GenesisTime=1700000000 \
  -v -timeout 180m
```

On seed=99 the run terminates at block 39/500, op 222. The verbose
trace within the failing `MsgValidation` tx shows the full sequence:

```
INF Voting               vote="proposal_id:170 voter:gonka1acqhn... Invalidate ..."
INF Voted on validation  vote="proposal_id:170 ..."         ← first vote succeeds
INF Inference invalidated
INF Participant status updated  new=INVALID original=ACTIVE
                                reason=statistical_invalidations
WRN Participant invalidated
INF Slashing participant        slash_fraction=0.2          ← voter removed from group here
INF Voting               vote="proposal_id:171 voter:gonka1acqhn... Revalidate ..."
ERR Error voting   error="voter address: gonka1acqhn...: not found"
                   vote="proposal_id:171 ..."               ← second vote fails
simulate.go:346: error on block 39/500, operation (222/236) ...
```

## Root cause

The handler `revalidateInferenceVote`
(`x/inference/keeper/msg_server_validation.go:226`) calls
`k.voteValidationProposal` twice in sequence — first on
`InvalidatePolicyId`, then on `ReValidatePolicyId`
(`msg_server_validation.go:246`, `:253`). Both calls bottom out at
`k.group.Vote` (`msg_server_validation.go:304`), which suppresses one
exact error string only — `"proposal not open for voting: invalid
value"` (`msg_server_validation.go:306`); every other error
propagates as `DeliverTx` failure.

Two independent mechanisms can leave the voter absent from the group:

**Mechanism A — self-invalidation cascade between the two votes
(DEMONSTRATED by sim verbose trace above).** The first vote on
`InvalidatePolicyId` can succeed and immediately trigger the
inference-invalidation handler. That handler updates participant
statistics; if `statistical_invalidations` thresholds are breached
(in the seed=99 trace, after `invalidated_inferences=24`), the voter
themselves is marked `INVALID`, slashed, and removed from the epoch
sub-group. The handler then proceeds to the second `k.group.Vote` on
`ReValidatePolicyId` — voter is no longer a member, vote
hard-errors. The entire tx then reverts (atomicity preserves funds
but discards both votes, the invalidation, and the slash — see
Consequences).

**Mechanism B — permissive `addEpochMembers` leaves
`ActiveParticipantsSet ⊆ group members` invariant unmaintained
(plausible from code inspection; not separately demonstrated here).**
`SetActiveParticipants` writes `ActiveParticipantsSet` first
(`x/inference/keeper/module.go:757`, `:785`), then `addEpochMembers`
adds those participants to the epoch x/group
(`module.go:1119`, `x/inference/keeper/epoch_group.go:391`).
`addEpochMembers` is permissive:

- skips participants whose seed is nil (`module.go:1134`, logged as
  `ILLEGAL STATE`);
- continues after an `AddMember` error rather than rolling back
  (`module.go:1143`, `:1145`).

Both paths leave `ActiveParticipantsSet` populated while the x/group
for that epoch is missing those members. A later `MsgValidation`
picking such a participant as voter would then fail on the **first**
vote, before the cascade above is even reached.

A correct fix has to cover both: the voter may be absent at handler
entry (mechanism B) or become absent between the two sequential votes
(mechanism A).

## Production consequences

- **Atomicity preserves funds**: a handler error reverts the entire
  tx. The first vote, the invalidation, the slash, and any partial
  refund all roll back together with the failed second vote. Net
  effect: nothing happens, voter remains ACTIVE, inference status
  unchanged.
- **Fee impact**: `MsgValidation` is fee-exempt via
  `NetworkDutyFeeBypassDecorator` (`app/ante_fee.go::isNetworkDuty`).
  No fee is deducted on the failed delivery.
- **Silent retry trap**: the cascade is deterministic from the same
  state, so a validator that retries the same `MsgValidation` will
  hit the same failure. Stuck only escapes via state change from
  other validators' votes or eventual timeout.
- **Quorum impact**: this validator's votes do not count on either
  proposal. If multiple validators reach the cascade independently,
  the `ReValidatePolicyId` proposal can miss quorum and expire.
  Liveness for the inference is bounded by the manifestation #1
  timeout fix (#1265) — client refunded, inference marked EXPIRED.
- **Frequency**:
  - Mechanism A triggers whenever a validator voting YES on
    `InvalidatePolicyId` breaches the statistical-invalidation
    threshold via that same tx. Reproducible: any validator near the
    threshold who lands on a will-invalidate inference.
  - Mechanism B depends on the rate of nil-seed participants and of
    `AddMember` failures during epoch construction. Lower frequency,
    but each occurrence persists for the rest of the epoch.

## Suggested fix (narrow)

Structural membership precheck in `revalidateInferenceVote`, applied
**before each `k.group.Vote` call** — not once at the top. Mechanism A
specifically requires a recheck between the two votes; checking only
at handler entry mirrors the state in which the voter is still a
member and is not sufficient.

Sketch:

```go
isMember := func() bool {
    sg, err := k.GetEpochGroup(ctx, inference.EpochId, inference.Model)
    if err != nil || sg == nil || sg.GroupData == nil || sg.GroupData.EpochGroupId == 0 {
        return false
    }
    members, mErr := sg.GetGroupMembers(ctx)
    if mErr != nil {
        return false
    }
    for _, m := range members {
        if m.Member != nil && m.Member.Address == voter {
            return true
        }
    }
    return false
}

if !isMember() {
    // log + no-op return; do not reach k.group.Vote
}
// ... first vote on InvalidatePolicyId ...

if !isMember() {
    // voter removed mid-tx (mechanism A); skip second vote
}
// ... second vote on ReValidatePolicyId ...
```

`GroupMessageKeeper.GroupMembers` is already exposed
(`x/inference/types/expected_keepers.go:46`) and
`epochgroup.EpochGroup.GetGroupMembers` paginates correctly
(`x/inference/epochgroup/epoch_group.go:464`). No interface
extension required.

Out of scope for this issue (deferred):

- Whether the second vote *should* still be cast in some form when
  mechanism A fires (e.g. the inference is already invalidated by
  the first vote, so the revalidate proposal is moot anyway). That
  is a design question for `revalidateInferenceVote` semantics.
- Tightening `addEpochMembers` itself (rollback on `AddMember`
  failure vs the current continue-and-log) — separate behavioural
  change that touches epoch transition semantics.
- Re-vote semantics for inferences whose voting window closes without
  quorum — covered by the broader hybrid-revote discussion mentioned
  on #1265.

## How discovered

Surfaced by the x/inference simulation work in #982 (Phase 3-B
invariants + custom decoders, follow-up to PR #1228). Initial
hypothesis from code inspection (mechanism B above) was filed first;
running sim with `-Verbose=true` produced the intra-tx trace that
demonstrates mechanism A is the actual failure mode on seed=99
b39/op 222.

## References

- Manifestation #1 (related): #1265 (VOTING-stuck escrow on timeout)
- Sim infrastructure prerequisite: PR #1228 (Phase 3 of #982)
- Handler: `x/inference/keeper/msg_server_validation.go:226`, `:246`,
  `:253`, `:304`, `:306`
- Permissive epoch-build path (mechanism B):
  `x/inference/keeper/module.go:1119`, `:1134`, `:1143`, `:1145`
- Group member API: `x/inference/types/expected_keepers.go:46`,
  `x/inference/epochgroup/epoch_group.go:464`

vitaly-andr:
## Problem

`revalidateInferenceVote` cast the Invalidate/Revalidate x/group votes unconditionally. The voter can be absent from the inference's epoch sub-group — either at handler entry (group membership is permissive), or after the Invalidate vote statistically invalidates the voter themselves and removes them from the group mid-transaction. The real x/group keeper rejects a non-member vote with an error (`voteValidationProposal` only tolerates the "proposal already decided" case), which fails the whole `MsgValidation`.

## Fix

Add a membership precheck before voting: confirm the voter is in the epoch sub-group, otherwise skip the vote as a no-op. Checked both at handler entry and again after the Invalidate vote (the self-invalidation cascade).

## Tests

- `TestMsgServer_Validation_Revalidation_SkipsVoteWhenVoterNotInGroup` — new reproducer; fails pre-fix (the unconditional vote is attempted on a non-member), passes with the precheck.
- `TestMsgServer_Validation_Invalidate` — updated to establish real x/group membership (non-zero `EpochGroupId` + stubbed `GroupMembers`) so its votes still fire under the precheck.

## Context

Surfaced by simulation/fuzz testing for inference-chain (#982), where the revalidation-vote factory hit a non-member vote that hard-errored mid-run.

Closes #1269.

cc @patimen (#982 author)

### DIFF
--- a/inference-chain/x/inference/keeper/msg_server_validation.go
+++ b/inference-chain/x/inference/keeper/msg_server_validation.go
@@ -236,6 +236,37 @@ func (k msgServer) revalidateInferenceVote(
 		revalidationOption = group.VOTE_OPTION_YES
 	}
 
+	// Structural membership precheck: the voter may be absent from the epoch
+	// sub-group either at handler entry, or after the first vote triggers a
+	// self-invalidation cascade that removes the voter mid-transaction. Voting
+	// as a non-member makes x/group return an error (voteValidationProposal only
+	// tolerates the "proposal already decided" case), which would fail the whole
+	// MsgValidation. Skip each vote as a no-op unless membership is confirmed at
+	// that point.
+	isMember := func() bool {
+		sg, err := k.GetEpochGroup(ctx, inference.EpochId, inference.Model)
+		if err != nil || sg == nil || sg.GroupData == nil || sg.GroupData.EpochGroupId == 0 {
+			return false
+		}
+		members, mErr := sg.GetGroupMembers(ctx)
+		if mErr != nil {
+			return false
+		}
+		for _, m := range members {
+			if m.Member != nil && m.Member.Address == voter {
+				return true
+			}
+		}
+		return false
+	}
+
+	if !isMember() {
+		k.LogError("Voter not in epoch sub-group at handler entry; skipping both votes",
+			types.Validation, "voter", voter, "inferenceId", inference.InferenceId,
+			"epochId", inference.EpochId, "model", inference.Model)
+		return &types.MsgValidationResponse{}, nil
+	}
+
 	voteMsg := &group.MsgVote{
 		ProposalId: inference.ProposalDetails.InvalidatePolicyId,
 		Voter:      voter,
@@ -245,6 +276,15 @@ func (k msgServer) revalidateInferenceVote(
 	}
 	if err := k.voteValidationProposal(ctx, voteMsg); err != nil {
 		return nil, err
+	}
+
+	// Re-check: the Invalidate vote may have statistically invalidated the
+	// voter themselves, removing them from the group; the Revalidate vote would
+	// then hard-error.
+	if !isMember() {
+		k.LogError("Voter removed from epoch sub-group after Invalidate vote; skipping Revalidate vote",
+			types.Validation, "voter", voter, "inferenceId", inference.InferenceId)
+		return &types.MsgValidationResponse{}, nil
 	}
 
 	voteMsg.ProposalId = inference.ProposalDetails.ReValidatePolicyId
--- a/inference-chain/x/inference/keeper/msg_server_validation_test.go
+++ b/inference-chain/x/inference/keeper/msg_server_validation_test.go
@@ -78,6 +78,22 @@ func TestMsgServer_Validation_Invalidate(t *testing.T) {
 	mocks.GroupKeeper.EXPECT().SubmitProposal(gomock.Any(), gomock.Any()).Return(&group.MsgSubmitProposalResponse{
 		ProposalId: 2,
 	}, nil)
+
+	// The membership precheck in revalidateInferenceVote performs a real x/group
+	// membership query. Give the sub-group a non-zero EpochGroupId and stub the
+	// membership query so the revalidation voter is recognized as a member;
+	// otherwise both votes are correctly skipped as no-ops.
+	groupData, _ := k.GetEpochGroupData(ctx, 0, MODEL_ID)
+	groupData.EpochGroupId = 1
+	k.SetEpochGroupData(ctx, groupData)
+	mocks.GroupKeeper.EXPECT().GroupMembers(gomock.Any(), gomock.Any()).Return(
+		&group.QueryGroupMembersResponse{
+			Members: []*group.GroupMember{
+				{GroupId: 1, Member: &group.Member{Address: testutil.Requester, Weight: "100"}},
+			},
+		}, nil,
+	).AnyTimes()
+
 	ms := inferenceHelper.MessageServer
 	_, err = ms.Validation(ctx, &types.MsgValidation{
 		InferenceId:  expected.InferenceId,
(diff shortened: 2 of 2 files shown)


inference-chain/x/inference/keeper/msg_server_validation.go
```diff
@@ -236,6 +236,37 @@ func (k msgServer) revalidateInferenceVote(
...
 	}
 
+	// Structural membership precheck: the voter may be absent from the epoch
+	// sub-group either at handler entry, or after the first vote triggers a
+	// self-invalidation cascade that removes the voter mid-transaction. Voting
+	// as a non-member makes x/group return an error (voteValidationProposal only
+	// tolerates the "proposal already decided" case), which would fail the whole
+	// MsgValidation. Skip each vote as a no-op unless membership is confirmed at
+	// that point.
+	isMember := func() bool {
```
qdanik, line 246: It would be better to move it into independent function

vitaly-andr: Thanks — agreed, this reads better as a dedicated helper (e.g. `func (k msgServer) isVoterInEpochGroup(ctx, inference, voter) bool`), and it makes the membership check unit-testable in isolation. It's only called within this handler today, so I've kept it inline for now to keep the fix minimal and focused on the bug; I'll fold the extraction in once a maintainer takes a first pass on the PR, so the structural change rides along with any other requested adjustments rather than churning the diff twice. The fail-closed semantics (any lookup error → treat as non-member → skip the vote as a no-op) would be preserved verbatim.

gmorgachev:
Hi, thank you for PR. I think the change doesn't want to support any more inference with per inference billing on mainnet, so this change is probably also not prioritized for now
I would focus attention on the devshard's

vitaly-andr:
Thanks @gmorgachev — that's clear. If per-inference-billing on mainnet is being retired, then both this fix (#1276) and the escrow-refund one (#1275) sit on a flow that's going away, so it makes sense they're not a priority. I'll leave them open with their reproducers for the record — feel free to close them if you'd rather keep the queue clean, your call.

One thing still open (I asked patimen on #1265): the #982 simulation infrastructure (#1228) itself is flow-agnostic and would be the natural base for devshard sim once its validation logic settles. Is it worth keeping as merged infra, or should the simulation work be treated as parked until devshard is ready? Either way works — just keeping things tidy.

vitaly-andr:
Closing: the path this guards was removed from the chain.

PR #1278, opened 2026-05-30
fix: Bound event-listener tx queue to stop a single zero-fee attacker from OOMing validator nodes

ouicate:
## Summary

This fixes an unbounded-memory bug in the validator API's (`decentralized-api`) chain event-ingestion path, where committed tx events are flattened into an in-memory queue that has **no size limit, no producer backpressure, and no pre-enqueue filtering**. The `BlockObserver` (single producer) walks up to a 500-block catch-up window and synthesizes one queue item for **every tx result in every observed block**, pushing them into a backing slice that grows without bound; only **10 fixed workers** drain it, and relevance filtering happens **after** dequeue. Whenever the consumer is slower than the producer, the backing slice grows until the DAPI process exhausts memory and crashes — and because progress is persisted only when a per-block barrier is consumed, a crash mid-backlog replays the same window on restart, turning a memory spike into a restart loop.

The realistic trigger is not generic traffic (irrelevant events drain in nanoseconds, faster than the RPC-bound producer) but the `MsgRequestThresholdSignature` path: it is open to anyone, has no per-participant rate limit (`ante_fee.go` explicitly documents this), is **free** under the current rollout (`MinGasPriceNgonka = 0`, set by the v0.2.12 upgrade), and faces **no block-gas ceiling** (`max_gas = -1` in genesis, ~21 MB/block). Each such request makes **every** active participant's DAPI do real BLS work (a scalar multiplication per owned slot, up to `i_total_slots = 1000`) plus a **blocking** `BroadcastTxSync`, on only 10 workers — a 1→N amplification with the slow side decoupled from the producer by an unbounded queue. A single funded attacker can therefore drive a target validator's DAPI to OOM while the chain itself stays healthy. This is the same event-ingestion code path validators run in production today.

## Root Cause

`UnboundedQueue.manage()` stores items in a slice with no upper bound and no mechanism to slow producers when consumers fall behind. `BlockObserver.processBlock` enqueues a synthetic event for every tx result before any handler check, and always advances `lastQueriedBlockHeight` the instant a block is enqueued — fully decoupling the producer from the 10-worker consumer. Relevance filtering (`hasHandler`) runs only after dequeue, so irrelevant and malicious-but-relevant traffic alike consume queue memory ahead of any gate. There is no per-process cap on buffered events and no backpressure: the producer never blocks, so the backing slice is bounded only by available RAM. Persistence stores `lastProcessedBlockHeight` (advanced on barrier consumption), not `lastQueriedBlockHeight`, so an OOM crash rebuilds the in-flight backlog from the persisted height on restart.

## Fix

- Add `NewBoundedQueue(capacity)` alongside `NewUnboundedQueue` (same `*UnboundedQueue` type, same `In`/`Out`/`Close` API, same FIFO guarantee). When bounded and full, `manage()` stops reading `input` (so producer sends block) while continuing to serve `output`, giving clean backpressure with no busy-loop and no deadlock as long as any consumer drains. `NewUnboundedQueue` is byte-for-byte unchanged (capacity 0).
- Construct the `BlockObserver` tx queue with `NewBoundedQueue(txEventQueueCapacity)` (`= 20000`). In-memory backlog and worst-case memory are now O(1) in spam volume instead of O(N). The separate NewBlock event queue stays unbounded on purpose, so the height-coordination path `waitForEventHeight` depends on can never backpressure (no deadlock cycle).
- Add a producer-side relevance filter on `BlockObserver`, wired to the exact same predicate the consumer already uses (`el.hasHandler`). Events with no handler are dropped before they take a queue slot — behavior-preserving for every event that matters, since the consumer gates on the identical function. **Barrier events bypass the filter and are always enqueued**, so `lastProcessedBlockHeight` still advances for every observed block.
- Route enqueues through a `ctx`-aware `enqueue()` helper that selects on `ctx.Done()`, so a backpressured producer shuts down cleanly instead of hanging on a full queue; a non-enqueued (cancelled) block leaves `lastQueriedBlockHeight` unadvanced and is simply retried, matching existing fetch-failure semantics.

The failure required three conditions: the queue could grow without bound, the producer ran ahead of the consumer with no backpressure, and a crash mid-backlog replayed the same window. This PR removes all three. Memory is now capped at ~`txEventQueueCapacity` items regardless of how much an attacker spams; when the queue is full the single producer blocks instead of allocating, so the process survives and resumes once load subsides; and because memory is bounded, the persisted-`lastProcessedBlockHeight` replay-on-restart can no longer escalate into an OOM restart loop (and it now replays a tiny set, because irrelevant traffic never enters the queue). Ordering, barrier semantics, and replay/progress guarantees are all preserved by construction. The chain-side validation surface is unchanged; this is a narrowly scoped resource-control fix to the consumer process.

## Test plan

- [ ] `go test ./internal/event_listener/...` in `decentralized-api` (run on Linux/CI — this tree's CGO deps `wasmvm`/`blst` do not build on a Windows host; the edits add no new imports and are type-checked clean by the language server).
- [ ] New `TestBoundedQueueBackpressureNoDropOrReorder`: a fast producer with a slow consumer over a small-capacity bounded queue delivers every item exactly once, in FIFO order, with no deadlock.
- [ ] Existing `TestBlockObserver_StressBackpressure` (200 blocks × 10 txs, slow consumer) still reaches `lastQueriedBlockHeight == 200` and drains all `blocks × (txs+1)` events through the now-bounded queue.
- [ ] Confirm a block with zero handler-matched tx events still emits its barrier and advances `lastProcessedBlockHeight` (relevance filter must not suppress barriers).
- [ ] Soak: synthetic backlog ≫ `txEventQueueCapacity` keeps DAPI RSS flat (bounded) instead of climbing, and the producer goroutine parks on `In` send rather than allocating.

### DIFF
--- a/decentralized-api/internal/event_listener/unbounded_queue.go
+++ b/decentralized-api/internal/event_listener/unbounded_queue.go
@@ -4,8 +4,15 @@ import (
 	"sync"
 )
 
-// UnboundedQueue[T] represents an unbounded thread-safe FIFO queue
-// that exposes channels for enqueuing and dequeuing elements of type T
+// UnboundedQueue[T] represents a thread-safe FIFO queue that exposes channels
+// for enqueuing and dequeuing elements of type T.
+//
+// By default the queue is unbounded (see NewUnboundedQueue). When constructed
+// with NewBoundedQueue it instead enforces a maximum number of buffered items
+// and applies backpressure to producers: a send on In blocks once the queue is
+// full, rather than letting the internal backing slice grow without bound. The
+// In/Out channel API, FIFO ordering, and Close semantics are identical in both
+// modes, so callers do not need to change how they use the queue.
 type UnboundedQueue[T any] struct {
 	// Public channels for interacting with the queue
 	In  chan<- T // Send-only channel for producers
@@ -17,20 +24,43 @@ type UnboundedQueue[T any] struct {
 	done      chan struct{}
 	wg        sync.WaitGroup
 	closeOnce sync.Once // Ensures Close is only executed once
+	capacity  int       // 0 = unbounded; >0 = max buffered items before producers block
 }
 
 // NewUnboundedQueue creates a new unbounded queue that exposes channels
 func NewUnboundedQueue[T any]() *UnboundedQueue[T] {
+	return newQueue[T](0)
+}
+
+// NewBoundedQueue creates a FIFO queue that applies backpressure to producers
+// once `capacity` items are buffered. It preserves the same In/Out channel API
+// and ordering guarantees as NewUnboundedQueue; the only behavioral difference
+// is that a send on In blocks while the queue is full instead of the backing
+// store growing without bound. A capacity <= 0 is treated as unbounded.
+//
+// Note: because the internal input channel is buffered for performance, the
+// effective worst-case number of in-flight items is approximately
+// capacity + len(input buffer) + len(output buffer); the bound is intended to
+// cap memory, not to be an exact admission-control limit.
+func NewBoundedQueue[T any](capacity int) *UnboundedQueue[T] {
+	if capacity < 0 {
+		capacity = 0
+	}
+	return newQueue[T](capacity)
+}
+
+func newQueue[T any](capacity int) *UnboundedQueue[T] {
 	input := make(chan T, 100)  // Buffer size is just for performance
 	output := make(chan T, 100) // Buffer size is just for performance
 	done := make(chan struct{})
 
 	q := &UnboundedQueue[T]{
-		In:     input,  // Public producer channel (send-only)
-		Out:    output, // Public consumer channel (receive-only)
-		input:  input,  // Private full access
-		output: output, // Private full access
-		done:   done,
+		In:       input,  // Public producer channel (send-only)
+		Out:      output, // Public consumer channel (receive-only)
+		input:    input,  // Private full access
+		output:   output, // Private full access
+		done:     done,
+		capacity: capacity,
 	}
 
 	q.wg.Add(1)
@@ -58,8 +88,18 @@ func (q *UnboundedQueue[T]) manage() {
 			first = items[0]
 		}
 
+		// Backpressure: when the queue is bounded and full, stop reading from
+		// the input channel. Producers then block on their send to In until a
+		// consumer drains an item. We keep `out` active so consumers can always
+		// make progress and relieve the pressure (so this never deadlocks as
+		// long as some consumer keeps draining).
+		in := q.input
+		if q.capacity > 0 && len(items) >= q.capacity {
+			in = nil
+		}
+
 		select {
-		case item := <-q.input:
+		case item := <-in:
 			// Store new item from producer
 			items = append(items, item)
 
--- a/decentralized-api/internal/event_listener/block_observer.go
+++ b/decentralized-api/internal/event_listener/block_observer.go
@@ -15,6 +15,14 @@ import (
 	"github.com/productscience/inference/x/inference/types"
 )
 
+// txEventQueueCapacity bounds the in-memory backlog of synthetic tx events
+// produced by the BlockObserver. When the queue is full, processBlock blocks
+// (applying backpressure to block fetching) instead of letting the backing
+// slice grow without bound and eventually OOM the process. It is sized to
+// comfortably hold several blocks' worth of *relevant* events (irrelevant ones
+// are filtered out before enqueue) while capping worst-case memory.
+const txEventQueueCapacity = 20000
+
 type BlockObserver struct {
 	lastProcessedBlockHeight atomic.Int64
 	lastQueriedBlockHeight   atomic.Int64
@@ -24,6 +32,20 @@ type BlockObserver struct {
 	caughtUp                 atomic.Bool
 	tmClient                 TmHTTPClient
 	notify                   chan struct{}
+	// relevanceFilter, when set, is consulted before enqueuing a synthetic tx
+	// event. Only events for which it returns true are enqueued; this keeps
+	// chain traffic the DAPI has no handler for from consuming queue capacity
+	// and worker time. Barrier events are always enqueued regardless, so block
+	// progress (lastProcessedBlockHeight) still advances for every block.
+	// A nil filter preserves the legacy behavior of enqueuing every event.
+	relevanceFilter func(*chainevents.JSONRPCResponse) bool
+}
+
+// SetRelevanceFilter installs a predicate used to drop irrelevant tx events at
+// the producer before they enter the queue. It must be safe for concurrent use
+// and is expected to be set once during wiring, before Process starts.
+func (bo *BlockObserver) SetRelevanceFilter(filter func(*chainevents.JSONRPCResponse) bool) {
+	bo.relevanceFilter = filter
 }
 
 // TmHTTPClient abstracts the subset of RPC methods we need
@@ -33,7 +55,7 @@ type TmHTTPClient interface {
 }
 
 func NewBlockObserver(manager *apiconfig.ConfigManager) *BlockObserver {
-	queue := NewUnboundedQueue[*chainevents.JSONRPCResponse]()
+	queue := NewBoundedQueue[*chainevents.JSONRPCResponse](txEventQueueCapacity)
 	// Initialize Tendermint RPC client
 	httpClient, err := cosmosclient.NewRpcClient(manager.GetChainNodeConfig().Url)
 	if err != nil {
@@ -64,7 +86,7 @@ func NewBlockObserver(manager *apiconfig.ConfigManager) *BlockObserver {
 
 // NewBlockObserverWithClient allows injecting a custom Tendermint RPC client (used in tests)
 func NewBlockObserverWithClient(manager *apiconfig.ConfigManager, client TmHTTPClient) *BlockObserver {
-	queue := NewUnboundedQueue[*chainevents.JSONRPCResponse]()
+	queue := NewBoundedQueue[*chainevents.JSONRPCResponse](txEventQueueCapacity)
 
 	bo := &BlockObserver{
 		ConfigManager: manager,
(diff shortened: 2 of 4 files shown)


decentralized-api/internal/event_listener/unbounded_queue.go
```diff
@@ -58,8 +88,18 @@ func (q *UnboundedQueue[T]) manage() {
...
 		}
 
+		// Backpressure: when the queue is bounded and full, stop reading from
+		// the input channel. Producers then block on their send to In until a
+		// consumer drains an item. We keep `out` active so consumers can always
+		// make progress and relieve the pressure (so this never deadlocks as
+		// long as some consumer keeps draining).
+		in := q.input
+		if q.capacity > 0 && len(items) >= q.capacity {
+			in = nil
```
a-kuprin, line 98: Let's add warn logging here

a-kuprin:
I think that actually we are more bounded by chain throughput as we are getting messages from chain. So even sending `MsgRequestThresholdSignature` will also be limited by chain, and dapi will drain the queue faster.

So while it is correct improvement of code base, it is more like low-priority neat of a path that will never be reached.

But I think someone else should look into why are we making `MsgRequestThresholdSignature` zero-gas fee?
It looks like it should be limited to prevent DoS attachs via `MsgRequestThresholdSignature`.

Maybe @patimen or @0xgonka could advice on this?

ouicate:
> I think that actually we are more bounded by chain throughput as we are getting messages from chain. So even sending `MsgRequestThresholdSignature` will also be limited by chain, and dapi will drain the queue faster.
> 
> So while it is correct improvement of code base, it is more like low-priority neat of a path that will never be reached.
> 
> But I think someone else should look into why are we making `MsgRequestThresholdSignature` zero-gas fee? It looks like it should be limited to prevent DoS attachs via `MsgRequestThresholdSignature`.
> 
> Maybe @patimen or @0xgonka could advice on this?

The suggestion that this path is "low priority" because chain throughput limits `MsgRequestThresholdSignature` and the DAPI "will drain the queue faster" does not match how the code is wired. Chain block time limits how fast **new blocks arrive**; it does **not** limit how fast the DAPI **producer** pushes already-committed tx events into the in-memory queue, nor how long the **consumer** holds each worker while doing BLS work and a synchronous broadcast.

---

## Why "chain throughput bounds the producer" is wrong

`BlockObserver` is a single goroutine. Once the node is caught up, it does **not** enqueue at one-block-per-chain-block pace. On each wakeup it runs a tight loop: fetch `BlockResults` for the next height, flatten **every tx** in that block into synthetic events, enqueue them, advance `lastQueriedBlockHeight`, and immediately continue to the next height until it reaches the current tip.

```190:202:decentralized-api/internal/event_listener/block_observer.go
			// Process as many contiguous blocks as available (based on lastQueried)
			for {
				nextHeight := bo.lastQueriedBlockHeight.Load() + 1
				if nextHeight > bo.currentBlockHeight.Load() || nextHeight <= 0 {
					break
				}
				if !bo.processBlock(ctx, nextHeight) {
					// stop on fetch error; next status change will retry
					break
				}
				// Successfully enqueued events for nextHeight; advance lastQueried
				bo.lastQueriedBlockHeight.Store(nextHeight)
			}
```

Per tx, `processBlock` builds a map and enqueues (pre-fix: **all** txs; no relevance filter). Producer cost is local CPU plus one RPC **per block**, not per tx. The producer can therefore inject a multi-block backlog in seconds while the chain only produced those blocks over many block times.

The consumer is a fixed pool of **10 workers**:

```196:201:decentralized-api/internal/event_listener/event_listener.go
func (el *EventListener) processEvents(ctx context.Context, mainQueue *UnboundedQueue[*chainevents.JSONRPCResponse]) {
	const numWorkers = 10
	for i := 0; i < numWorkers; i++ {
		worker(ctx, mainQueue, el.processEvent, "process_events_"+strconv.Itoa(i))
	}
}
```

For `EventThresholdSigningRequested`, a worker runs `ProcessThresholdSigningRequested`, which performs a **BLST scalar multiplication per owned slot** and then submits a partial signature via `SendTransactionAsyncWithRetry`. That path performs a synchronous `BroadcastTxSync` before returning — the worker is blocked for the full RPC round-trip:

```167:182:decentralized-api/internal/bls/threshold_signing.go
	// For each relative slot offset, compute per-slot signature
	var concatenated []byte
	for rel := 0; rel < len(result.AggregatedShares); rel++ {
		sk := result.AggregatedShares[rel]
		skBytes := sk.Bytes()
		// Convert to little-endian for blst
		for i := 0; i < 16; i++ {
			skBytes[i], skBytes[31-i] = skBytes[31-i], skBytes[i]
		}

		sig := new(blst.P1)
		sig.FromAffine(messageG1Blst)
		sig.MultAssign(skBytes[:], 255)
		concatenated = append(concatenated, sig.ToAffine().Compress()...)
	}
```

```909:909:decentralized-api/cosmosclient/tx_manager/tx_manager.go
	resp, err := m.client.Context().BroadcastTxSync(txBytes)
```

So the relevant comparison is not "events per block ÷ block time" vs "drain rate". It is **burst enqueue rate** (many blocks × many txs, at local RPC speed) vs **10 workers × (per-slot BLS + BroadcastTxSync)** per handler-matched event. When the consumer is slower, the backing slice in `UnboundedQueue.manage()` grows without bound until the process OOMs.

---

## The attack surface is real and documented on-chain

`MsgRequestThresholdSignature` is intentionally **not** fee-exempt; the ante code documents that it has **no per-participant rate limit**:

```145:148:inference-chain/app/ante_fee.go
	// NOTE: MsgRequestThresholdSignature is intentionally NOT exempt.
	// It has no per-participant rate limit — anyone can request signatures
	// with arbitrary RequestIds.
```

Under the v0.2.12 rollout, consensus `MinGasPriceNgonka` is set to **0** (temporary, per the upgrade handler — `setFeeParams` in `app/upgrades/v0_2_12/upgrades.go`), so non-exempt txs still pass with zero fee. Genesis also sets `max_gas: "-1"` and `max_bytes: "22020096"` (~21 MiB), so inclusion is bounded by block bytes, not gas — many small requests can pack into one block.

Each committed request emits `EventThresholdSigningRequested`. **Every** active participant's DAPI that owns slots runs the expensive handler above — a 1→N cost asymmetry (one cheap chain tx, N validators doing BLS + broadcast).

---

## Three mechanisms that make queue growth reachable

### 1. Multi-block catch-up burst

If `lastQueriedBlockHeight` is more than ~500 blocks behind the tip (or invalid), the observer resets and replays from `max(currentHeight - 500, earliestAvailable)`:

```179:188:decentralized-api/internal/event_listener/block_observer.go
			if lastQueried < (currentHeight-500) || lastQueried <= 0 {
				startBlock := bo.getStartProcessingBlock(ctx, currentHeight)
				logging.Info("Resetting lastQueriedBlockHeight for block availability", types.EventProcessing,
					"oldLastQueried", lastQueried,
					"currentHeight", currentHeight,
					"newStartBlock", startBlock)
				bo.lastQueriedBlockHeight.Store(startBlock - 1)
			}
```

That can deliver up to **500 blocks** of tx events into the queue in one producer run — far faster than those blocks were originally committed. This is hit after long downtime, snapshot-node bring-up, or any gap where persisted progress falls behind the tip. It does **not** require an attacker; it only requires enough handler-relevant txs in that window to outpace 10 workers.

### 2. OOM → restart loop

Progress is persisted only when a **barrier** event is consumed (`signalAllEventsRead` → `SetLastProcessedHeight`). `lastQueriedBlockHeight` advances when events are **enqueued**, not when they are **processed**:

```292:306:decentralized-api/internal/event_listener/block_observer.go
func (bo *BlockObserver) signalAllEventsRead(height int64) {
	if height < bo.lastProcessedBlockHeight.Load() {
		logging.Warn("BlockObserver: signalAllEventsRead called for out-of-order block", types.EventProcessing, "height", height)
	} else if height == bo.lastProcessedBlockHeight.Load() {
		logging.Warn("BlockObserver: signalAllEventsRead called for already processed block", types.EventProcessing, "height", height)
	} else {
		bo.lastProcessedBlockHeight.Store(height)
		if err := bo.ConfigManager.SetLastProcessedHeight(height); err != nil {
			logging.Warn("BlockObserver: Failed to persist last processed height", types.Config, "error", err)
		}
	}
}
```

A crash mid-backlog leaves a large gap between persisted `lastProcessedBlockHeight` and what was already queued. On restart, the same txs are replayed into a still-unbounded queue; if drain rate is unchanged, the process OOMs again — a self-reinforcing outage of the validator DAPI while the chain stays healthy.

### 3. Chain-side retry re-emits events

When a threshold signing request expires or fails, `maybeAutoRetryThresholdSigningRequest` emits **another** `EventThresholdSigningRequested`:

```379:387:inference-chain/x/bls/keeper/threshold_signing.go
	if err := cacheCtx.EventManager().EmitTypedEvent(&types.EventThresholdSigningRequested{
		RequestId:           retryReq.RequestId,
		CurrentEpochId:      retryReq.CurrentEpochId,
		EncodedData:         retryReq.EncodedData,
		MessageHash:         retryReq.MessageHash,
		DeadlineBlockHeight: retryReq.DeadlineBlockHeight,
	}); err != nil {
```

`DefaultParams()` in `x/bls/types/params.go` sets `max_signing_attempts: 3` (initial attempt + 2 retries), so a single logical request can produce up to 3× handler-matched events. Expiry sweeps are also chain-driven (`ProcessThresholdSigningDeadlines`, up to `defaultMaxExpiredRequestsPerBlock = 200` per block), so retry bursts arrive at the DAPI as bunched events — exactly the burst pattern the unbounded queue can't shed.

---

## A fee on `MsgRequestThresholdSignature` is complementary, not a substitute

Agree that limiting or pricing `MsgRequestThresholdSignature` on-chain is worth a follow-up — the codebase already flags the missing rate limit. But it does not fix this bug:

- **Catch-up and crash-replay** (1 and 2 above) can OOM the DAPI without any attacker.
- A fee raises the cost of spam; it does not cap **in-process** queue memory or remove the producer/consumer decoupling.
- Even with a fee, each chain tx still amplifies to **every** participant's DAPI doing BLS + broadcast; validator-side cost still dominates a single submitter's cost.

Chain-side admission control and DAPI-side load shedding address different layers. Both are useful; neither replaces the other.

---

## What this PR does (minimal, behavior-preserving fix)

On `main` today, `UnboundedQueue.manage()` appends to a slice with no cap and no producer backpressure — that is the root defect.

This PR:

1. Adds `NewBoundedQueue(capacity)` with the same `In`/`Out`/`Close` API; `NewUnboundedQueue` is unchanged.
2. Bounds the BlockObserver tx queue (`txEventQueueCapacity = 20000`); the NewBlock queue stays unbounded to avoid deadlocking height coordination.
3. Filters at the producer with the **same** `hasHandler` predicate the consumer already uses; barrier events always enqueue so per-block progress still advances.
4. Uses a context-aware `enqueue()` so shutdown under backpressure does not hang.

Net effect: worst-case memory is O(capacity) instead of O(spam volume); the producer blocks instead of allocating without limit; the OOM-restart loop cannot escalate because the backlog is bounded. Ordering, barrier semantics, and handled-event behavior are preserved by construction.

patimen (approved):
I see no problems with the specific implementation.
We initially created UnboundedQueue to solve temporary problems, with the knowledge that an unbounded queue had risks. In principle, it still does, and this PR addresses a big chunk of them.
However, I still don't fully buy that this is a current exploit. I think it is a good change, but was unlikely to ever reproduce in production. I'm not going to reject it though, let's get this possible hole fixed.

decentralized-api/internal/event_listener/unbounded_queue.go
patimen: Wouldn't that potentially cause massive log spam at the worst time?

PR #1046, opened 2026-04-13
Implement dapi & devshard observability 

qdanik:
## Summary

Adds observability for the API, devshard runtime, and local join deployment.

The branch wires traces, metrics, logs, dashboards, and docs so we can debug inference execution, validation, devshard sessions, chain queries, and node health from one place.

## What Changed

This PR adds a new observability layer for `decentralized-api` and `devshardd`.

On the API side, the main inference and validation paths now create OpenTelemetry spans and Prometheus metrics. This covers request handling, executor forwarding, ML node calls, finish submissions, validation sampling, payload retrieval, logit comparison, chain queries, and tx broadcasts.

On the devshard side, the runtime now reports request lifecycle metrics, terminal outcomes, interruptions, payload-serving results, validation state, ML node attempts, token usage, queue depth, mempool size, and HTTP connection state. devshardd also exports traces and metrics when running as a standalone binary behind versiond.

The local join stack now includes a full observability overlay: Jaeger for traces, Prometheus for metrics, Loki/Promtail for logs, Grafana for dashboards, and cAdvisor for container metrics. Grafana and Jaeger can be exposed through the proxy at `/grafana/` and `/jaeger/`.

Grafana dashboards were added for devshard behavior, chain/node health, query latency, storage/container state, and detailed devshard debugging. Logs in Grafana can link back to Jaeger traces when a `trace_id` is present.

The branch also adds documentation in `docs/observability/observability-overview.md` with setup instructions, metric/tracing details, operational checks, and troubleshooting notes.

## Why

Before this, debugging a failed or slow inference meant jumping between logs, chain state, API behavior, devshard state, and ML node output manually.

With this change we get a connected picture of what happened: request traces, runtime metrics, structured logs, dashboards, and container/node health.

This should make it easier to investigate:

- slow or failed inference requests;
- missing finish submissions;
- validation failures;
- payload retrieval issues;
- devshard terminal states;
- ML node errors or latency;
- chain query / tx broadcast problems;
- node or container resource pressure.

## Notes

- Tracing is controlled by `DAPI_OTEL_ENABLED` and `DEVSHARD_OTEL_ENABLED`.
- Metrics are exposed through Prometheus `/metrics` endpoints.
- Logs are shipped to Loki through Promtail.
- Jaeger stores traces.
- Grafana is the main UI for dashboards and log exploration.
- `X-Request-Id` is still separate from W3C `traceparent`.

PR #1288, opened 2026-06-01
Merge dapi & devshard observability (#1046) to devshard-0.2.13-v2

a-kuprin:
This PR is based on https://github.com/gonka-ai/gonka/pull/1046 by @qdanik 
It only merges observability to devshard release.

**What should be added to observability in next releases:**

1. `trace_id` should be automatically injected into log (as a label for loki) based on current otel context
2. Currently scraping is used only for logs (promtail), otel data is sent to jaeger directly. We should also collect it locally and have delivery guarantees.

And we should log all failed inferences and validation prompts (according to comment https://github.com/gonka-ai/gonka/pull/1046#discussion_r3256765126) that will be binded to traces by `trace_id`


a-kuprin:
As `feat/observability` branch was already tested, reviewed and approved, I merge this branch into devshard-0.2.13-v2

PR #1293, opened 2026-06-03
[P2] Bridge: cap completed post-process retries + abandonment monitoring (#1286)

### LINKED ISSUE #1286: Bridge: add retry cap or monitoring for stale refund cleanup retries
## Summary

When a threshold signing request reaches COMPLETED, the inference BLS hook calls `CleanupBridgePendingRefundByBlsRequestID` to remove pending bridge refund state.

If this cleanup persistently fails, the request is re-queued under `CompletedPostProcessRetryPrefix`, but `ProcessCompletedPostProcessRetries` currently has no maximum retry cap.

## Impact

This does not appear to create a direct safety issue:

- `cancelThresholdSigningRequest` rejects cancellation once the BLS request is COMPLETED.
- Therefore a stale pending refund entry should not allow double-spend.

However, persistent cleanup failures can cause:

- stale refund entries remaining in state indefinitely
- retry queue growth
- noisy monitoring / operational confusion

## Suggested fix

Either:

1. add a maximum retry count for completed post-process retries, or
2. add monitoring / alerting for retry queue depth and old retry entries.

## Relevant code

- `inference-chain/x/inference/module/bls_hooks.go`
  - `AfterThresholdSigningCompleted`
- `inference-chain/x/inference/keeper/bridge_pending_refund.go`
  - `CleanupBridgePendingRefundByBlsRequestID`
- `inference-chain/x/bls/keeper/threshold_signing.go`
  - `ProcessCompletedPostProcessRetries`

## Acceptance criteria

- Persistent cleanup failures cannot retry forever without visibility.
- Either a max retry cap exists, or an alert/metric exists for stale completed post-process retries.
- Existing double-spend protection remains unchanged.

bonujel:
## Summary

Closes #1286. When a threshold-signing request reaches `COMPLETED`, the BLS completion post-process runs the inference hook that cleans up the pending bridge refund (`CleanupBridgePendingRefundByBlsRequestID`). On failure the request is re-queued under `CompletedPostProcessRetryPrefix` and retried every block in `EndBlock` — previously with **no cap**, so a persistently failing cleanup could retry forever with no per-entry visibility.

This adds a hard cap **and** an observable abandonment signal (the issue asked for either; both are cheap together).

- **Attempt counter** — the retry queue entry value goes from empty to a 4-byte big-endian attempt count. Empty decodes to `0`, so in-flight/legacy entries are backward compatible (no state migration, no upgrade-handler change).
- **Cap** — `maxCompletedPostProcessRetryAttempts` (default `50`), a `var`-with-default constant mirroring the existing `maxCompletedPostProcessRetriesPerBlock`. Retries run once per block, so ~50 blocks before giving up — long enough to ride out a transient failure, bounded enough to never retry indefinitely.
- **Abandonment** — at the cap the entry is deleted, logged at `Error`, and a raw `completed_post_process_retry_abandoned` event (`request_id` / `current_epoch_id` / `attempts`) is emitted for monitoring. The per-block summary log gains `abandoned_count` and `max_attempts`.

### Safety

- **Double-spend protection unchanged.** Abandoning leaves a stale pending refund, which is harmless: `cancelThresholdSigningRequest` rejects cancellation once a BLS request is `COMPLETED`. That code path is untouched.
- **Deterministic.** Slice iteration (not map) + fixed big-endian encoding → identical across validators, no app-hash divergence. The event is not part of the app hash.
- **Self-contained.** The retry value is read/written only inside `threshold_signing.go`; no other module depends on its format.

## Test plan

- [x] `go build ./...` (whole inference-chain) — green
- [x] `go test ./x/bls/...` — green, incl. the new `TestProcessCompletedPostProcessRetries_AbandonsAfterMaxAttempts` (cap lowered to 3: entry survives + counter increments twice, then is removed with a correctly-attributed event)
- [x] `go test ./x/inference/...` (regression — the post-process hook drives the inference bridge-refund cleanup) — green
- [x] `gofmt` clean
- [ ] Testnet — the abandonment branch only fires on a *persistent* post-process failure, which doesn't occur naturally; covered by unit test via fault injection rather than a live run


bonujel:
Hi @tcharchian , I've opened this PR for #1286 (retry cap + monitoring). Whenever you have a moment, I'd really appreciate it if you could take a look. Thank you!

x0152 (approved):
valid hardening

bonujel:
hi @x0152, I updated the base branch, could you please approve the workflows once more?

bonujel:
Rebased onto upgrade-v0.2.16 to match the milestone (the old upgrade-v0.2.14 base is no longer shipping), could you please take a look when you have a moment? @tcharchian

bonujel:
Hi @tcharchian, @DimaOrekhovPS — a small request, and not a review one this time.

This PR and #1294 are both waiting on CI rather than on feedback. The workflow runs are sitting in `action_required`: GitHub holds runs for contributors who don't yet have a merged PR in the repo, which is the `contributor:first-time` tier from your #1440, @tcharchian. Because of that the two required checks haven't run, so the PRs can't merge, and the tier stays as it is. There are ten runs queued across the two branches.

If either of you has a moment, clicking "Approve and run workflows" would clear it. It should only need doing once — the gate lifts by itself once a first PR merges.

**This one (#1293)** — 2 files, +147/-6. @x0152 approved it back in June ("valid hardening") and the code hasn't changed since; I compared the approved commit against the current head and both files are identical. Rebased onto current `upgrade-v0.2.16` today, so it's ready to merge once the checks pass.

**#1294** — 3 files, docs only. @x0152's feedback is addressed, and @Ryanchen911 checked the field order and BLS sizes against `BridgeContract.sol` before leaving an LGTM. That one needs a formal approval as well, so passing it to a reviewer would work just as well.

I've run both required suites locally against the current heads and they pass, so hopefully the CI time won't go to waste.

No rush on any of this — and if approving runs isn't something you'd normally handle, I'd be grateful for a pointer to whoever does. Thanks!

bonujel:
Thanks very much for approving those runs — both went green on Aug 25, all
checks including the two required ones.

I'm afraid I've undone it since. The base moved by one commit, and with the
strict up-to-date policy a `BEHIND` branch can't merge, so I rebased; that
invalidated the passing runs and the new ones are sitting at `action_required`
again. Apologies for the extra round trip.

If you have a moment for one more approval, #1293 is ready to go in straight
afterwards: it already has @x0152's approval from June, and the content hasn't
changed since — I compared the approved commit against the current head and both
files are identical.

One thing worth mentioning in case it saves you being asked again: once any one
of my PRs actually merges, GitHub stops gating the runs, so this shouldn't need
repeating.

Thanks again.

bonujel:
@tcharchian thanks for approving the runs on the 29th — all nine checks came back green, and GitHub now reports this as mergeable with no conflicts.

Would you or @DimaOrekhovPS be able to merge it when you have a moment? x0152's approval from 24 June still stands, and the two files this touches are byte-identical to what was reviewed then — `threshold_signing.go` at blob `ec17e060` and the retry test at `70f76b99`, unchanged across the rebases — so there is nothing new to look at.

The reason I'm asking rather than just waiting: the same thing happened on 25 August. Everything went green, then the base moved and the runs were invalidated before anyone merged, which cost a second approval cycle on an unchanged diff. Happy to rebase again if it comes to that — I just did not want a green run to quietly go stale a second time without flagging it.

bonujel:
Following up on the above. The nine checks are still green on `4dd31f31`, but `upgrade-v0.2.16` has moved six times since 2 September, so this is now BEHIND and the strict policy blocks the merge.

Rather than ask again, it would help more if one of you could tell me which of these you'd prefer:

1. I rebase now, someone approves the runs, and it gets merged before the base moves again. I can do the rebase whenever suits you — happy to work to your timing so the green run does not go stale in between.
2. Someone with admin rights merges it as it stands.
3. It is not wanted after all, in which case I will close it. That is a fine outcome and I would rather know than keep it open.

Any of the three works for me. It is mainly the uncertainty that is hard to plan around, so a one-line answer would settle it.

@tcharchian @patimen — you have both touched this one, so tagging you rather than the whole list.

PR #1303, opened 2026-06-03
Add Testermint upgrade rehearsal workflow

patimen:
## Summary

Adds a dedicated Testermint upgrade rehearsal workflow that runs an actual local software upgrade before spending Testnet time on it.

The workflow:
- discovers the target upgrade and previous canonical release dynamically;
- checks out the previous release and pulls the production images declared by that release;
- applies a small prep-test patch to the old checkout so the old Testermint harness creates meaningful state;
- keeps the old Docker cluster alive;
- builds candidate production-style upgrade archives from the PR branch;
- serves those archives on the Testermint Docker network;
- attaches the current Testermint harness to the existing old cluster, submits the upgrade proposal, waits for cosmovisor, and verifies post-upgrade inference plus devshard settlement.

## Why

Testing upgrades only in Testnet is too risky and expensive. This gives us a repeatable CI rehearsal that catches version-discovery, archive-layout, cosmovisor, migration, and post-upgrade behavior issues before Testnet.

## Validation

Local checks:
- `git diff --check`
- `zsh -lc 'cd inference-chain && GOTOOLCHAIN=go1.24.2 go test ./app/upgrades/v0_2_14 ./x/inference/module -run TestDoesNotExist'`
- `python3 scripts/upgrade-rehearsal/resolve_versions.py --repo . --target-upgrade v0.2.14 --previous-release release/v0.2.13`
- `python3 scripts/upgrade-rehearsal/resolve_previous_images.py --compose deploy/join/docker-compose.yml`

Workflow validation before squashing history:
- Intentional crash proof failed correctly: https://github.com/gonka-ai/gonka/actions/runs/26902286237
- Clean full rehearsal succeeded: https://github.com/gonka-ai/gonka/actions/runs/26905450054
- Clean full rehearsal succeeded again: https://github.com/gonka-ai/gonka/actions/runs/26908298954

The squashed PR commit has the same tree as the twice-green validated commit. I am also dispatching the new upgrade rehearsal workflow against this PR head so the PR shows fresh evidence on the final commit.

## Notes

The workflow is currently `workflow_dispatch` only. We can decide in review whether to keep it manual for release branches or wire it into a release gate later.

### DIFF
--- a/docs/testermint-upgrade-rehearsal.md
+++ b/docs/testermint-upgrade-rehearsal.md
@@ -45,6 +45,20 @@ The workflow applies that patch to the old checkout before running the prep test
 
 The workflow must not tear the cluster down after this phase. The live Docker containers and volumes are the upgrade subject.
 
+### Maintaining The Prep Patch
+
+`previous-release-prep.patch` is intentionally version-coupled to the previous release's Testermint harness. When the previous canonical release changes, refresh the patch against that release instead of assuming the old patch still applies.
+
+Recommended refresh loop:
+
+1. Check out the new previous release tag in a scratch worktree, for example `git worktree add /tmp/gonka-prev release/v0.2.14`.
+2. Add or update `testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt` in that scratch checkout.
+3. Run the prep test locally or in a temporary workflow run until it creates the expected manifest and leaves the cluster running.
+4. From the scratch checkout, regenerate the patch with `git diff -- testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt > /path/to/current/testermint/upgrade-rehearsal/previous-release-prep.patch`.
+5. Re-run `testermint-upgrade-rehearsal.yml` with explicit `target_upgrade` and `previous_release` overrides before relying on the default discovery path.
+
+Keep the patch narrow. It should add only the old-checkout prep test and should not alter cluster teardown, Docker image naming, or production chain code in the old release.
+
 ## Phase 2: Build Candidate Upgrade Archives
 
 The candidate checkout builds the upgrade archives with:
--- a/docs/testermint-upgrade-rehearsal.md
+++ b/docs/testermint-upgrade-rehearsal.md
@@ -42,7 +42,7 @@ The workflow applies that patch to the old checkout before running the prep test
 - runs a normal inference;
 - creates and settles a devshard escrow;
 - waits into another epoch;
-- writes an upgrade rehearsal manifest with heights, epoch indexes, participant ids, inference id, and devshard escrow id.
+- writes an upgrade rehearsal manifest with heights, epoch indexes, participant ids, baseline participant weights, inference id, and devshard escrow id.
 
 The workflow must not tear the cluster down after this phase. The live Docker containers and volumes are the upgrade subject.
 
@@ -96,7 +96,7 @@ The attach test:
 
 By default, the power-stability check allows at most a 50% per-miner and total-power change across that post-upgrade PoC cycle. Use `UPGRADE_REHEARSAL_MAX_POWER_CHANGE_PERCENT` only for debugging a known intentional power-model change; do not loosen it to hide unexpected miner removal or cPoC fallout.
 
-The completion manifest records the post-upgrade PoC start block, `SET_NEW_VALIDATORS` block, before/after epoch ids, and miner weights so a failed or suspicious run can be compared without replaying the full logs first.
+The completion manifest records the post-upgrade PoC start block, `SET_NEW_VALIDATORS` block, pre-upgrade baseline epoch/weights, and post-upgrade epoch/weights so a failed or suspicious run can be compared without replaying the full logs first.
 
 ## Running In GitHub Actions
 
--- a/docs/testermint-upgrade-rehearsal.md
+++ b/docs/testermint-upgrade-rehearsal.md
@@ -1,0 +1,128 @@
+# Testermint Upgrade Rehearsal
+
+This workflow tests an actual local software upgrade before we spend Testnet time on it. It starts a Testermint cluster from the previous production release, creates meaningful pre-upgrade state, then upgrades the still-running cluster with binaries built from the candidate branch.
+
+The rehearsal is intentionally separate from the normal integration matrix. It needs two repository checkouts, two versions of the Testermint harness, and live Docker state that must survive between phases.
+
+## What It Proves
+
+- The previous release can still boot in the current CI environment using the production images declared by that release.
+- The previous release's own Testermint harness can create realistic pre-upgrade state.
+- The candidate branch can submit a real Cosmos SDK software-upgrade proposal against that existing cluster.
+- Cosmovisor downloads and installs the candidate `inferenced` and `decentralized-api` upgrade archives.
+- The chain resumes block production after the upgrade and records the expected `last-upgrade-height`.
+- Normal inference and devshard settlement still work after the upgrade.
+
+## Version And Image Discovery
+
+The workflow treats deploy scripts as the canonical source for production image references.
+
+1. The target upgrade defaults to the newest semantic `UpgradeName` found under `inference-chain/app/upgrades/*/constants.go` in the candidate checkout.
+2. The previous release defaults to the highest canonical release tag below that target, where canonical means `release/vX.Y.Z`.
+3. Suffix tags such as `release/v0.2.13-testnet-3`, `release/v0.2.13-testnet-rehearsal-1`, or ad hoc tags are ignored.
+4. The old checkout is made at that previous release tag.
+5. Previous production image refs are read from the old checkout's `deploy/join/docker-compose.yml`.
+6. Those production images are pulled and retagged to the local Testermint image names, such as `ghcr.io/product-science/inferenced`, `ghcr.io/product-science/api`, and `ghcr.io/product-science/proxy:latest`.
+
+Manual workflow inputs can override the target upgrade or previous release when debugging a non-standard branch.
+
+## Phase 1: Prepare Old State
+
+The old release checkout runs the old release's Testermint harness against the previous production images.
+
+Because old release tags are immutable, the candidate branch stores a small patch at:
+
+`testermint/upgrade-rehearsal/previous-release-prep.patch`
+
+The workflow applies that patch to the old checkout before running the prep test. The patched test:
+
+- boots the old cluster from scratch;
+- waits through an epoch;
+- runs a normal inference;
+- creates and settles a devshard escrow;
+- waits into another epoch;
+- writes an upgrade rehearsal manifest with heights, epoch indexes, participant ids, inference id, and devshard escrow id.
+
+The workflow must not tear the cluster down after this phase. The live Docker containers and volumes are the upgrade subject.
+
+## Phase 2: Build Candidate Upgrade Archives
+
+The candidate checkout builds the upgrade archives with:
+
+`make build-for-upgrade`
+
+The generated archives live under:
+
+- `public-html/v2/inferenced/inferenced-amd64.zip`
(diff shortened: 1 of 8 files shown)


patimen:
Started the upgrade rehearsal workflow against this PR head commit.

Run: https://github.com/gonka-ai/gonka/actions/runs/26912719322
Head SHA: bb36e5c6a0ba27458c11a11d77c46e9c76c2a659

This is the same workflow that passed twice on the identical pre-squash tree; this run is fresh evidence on the final PR commit.

patimen:
Upgrade rehearsal workflow passed on the final PR head.\n\nRun: https://github.com/gonka-ai/gonka/actions/runs/26912719322\nHead SHA: bb36e5c6a0ba27458c11a11d77c46e9c76c2a659\nResult: success\n\nKey phases all passed: old-checkout prep, candidate archive build, archive serving, current-checkout upgrade completion, artifact upload, and test result publication.

patimen:
Addressed a few AI-review comments in `4e07424b`:

- Confirmed the reported proto mismatch is a false positive: `inference-chain/proto/inference/inference/params.proto` already defines `bool devshard_requests_enabled = 9`, and generated Go includes it.
- Pinned the upgrade rehearsal runner from `ubuntu-latest` to `ubuntu-24.04`.
- Added maintenance guidance for regenerating `testermint/upgrade-rehearsal/previous-release-prep.patch` when the previous release moves forward.
- Replaced manual completion-manifest JSON construction with `cosmosJson.toJson(...)`.

Intentionally not changed:

- Kept `ghcr.io/product-science/...` in docs/scripts because the current deploy compose resolves those exact image refs.
- Kept `devshardRequestsEnabled` nullable in the Testermint export model so older exports can distinguish an absent field from an explicit `false`.
- Left the longer sleep/polling structure as-is for now because the full workflow has repeatedly passed and the sleep is guarding cosmovisor restart timing; we can refine separately if it becomes a source of cost/flakiness.

Validation after the changes:

- `git diff --check` passed.
- `./gradlew compileTestKotlin --stacktrace` passed locally after Gradle fell back from a local Kotlin daemon connection hiccup.

Started a fresh upgrade rehearsal on the new PR head:
https://github.com/gonka-ai/gonka/actions/runs/26916035036

Head SHA: `4e07424b16f2d78d03b2bbc574b96669b4d37d32`

patimen:
Upgrade rehearsal rerun is green on the current PR head.

- Latest run: https://github.com/gonka-ai/gonka/actions/runs/26921900932
- Head SHA: `68e254c8b63883b03e009a6a0bb4b7da7523443e`
- Result: success

Follow-up from reviewer/CI feedback:
- Pinned the workflow runner to `ubuntu-24.04`.
- Added patch-maintenance documentation.
- Switched the completion manifest writer to Gson-backed JSON serialization.
- Stabilized post-upgrade API recovery so it waits on node block production while polling API health, instead of using an API-dependent wait helper immediately after API restart.
- Added bounded post-upgrade inference-routing retries for transient 503/empty-active-participant readiness at the start of a new inference window.

The latest green run covers previous-release prep, candidate archive build/serving, upgrade execution, post-upgrade normal inference, and post-upgrade dev-shard settlement.

0xgonka (commented):
My clanker has this feedback

Adds a `workflow_dispatch`-only GitHub Actions workflow that runs a full local upgrade rehearsal (boot old release, apply candidate upgrade, verify post-upgrade inference + devshard settlement) before spending Testnet time on it. Sound scope, generally clean implementation, but **one Copilot-flagged concern is unaddressed** and will fire on the next standard integration run.

**PR state:**
- CI: 2 fail / 7 pass. Both failures (`Build and Test API Wrapper`, `Decentralized-API Tests`) come from the same single Go test — `TestInterruption_ClientDisconnect_StreamingComplete_VerifyFinishInference` in `decentralized-api/internal/server/public/` (timing-sensitive client-disconnect-during-streaming test). This PR doesn't touch that area, and the base branch's latest run shows the same test passing (992 ✅ / 1 skipped). **Looks like a flake, not a regression** — recommend a rerun.
- Merge: clean tree-wise, but `mergeStateStatus=BLOCKED` due to `REVIEW_REQUIRED`.
- Prior reviews: 1 COMMENTED (`copilot-pull-request-reviewer[bot]` at 2026-06-03 21:05) with 1 inline comment at [`testermint/src/test/kotlin/UpgradeRehearsalTests.kt:19`](testermint/src/test/kotlin/UpgradeRehearsalTests.kt#L19) about `upgrade-rehearsal` tag not being excluded from the default suite.
- New activity: author landed three commits after the Copilot review (`4e07424b "Address upgrade rehearsal review feedback"`, `308aeaac9`, `68e254c8b`). The first claimed to address feedback but only refactored JSON-writing inside the test class (replaced a string heredoc + custom `escapeJson` helper with `cosmosJson.toJson(map)`). **The Copilot tag-exclusion concern was NOT addressed by any of these three commits** — see Must-Fix below.

**Must fix:**

- **[High] `UpgradeRehearsalTests` will get pulled into the next standard integration matrix and fail immediately.** [`testermint/src/test/kotlin/UpgradeRehearsalTests.kt:20`](testermint/src/test/kotlin/UpgradeRehearsalTests.kt#L20) has `@Tag("upgrade-rehearsal")` at the **method** level only — no class-level annotation. The `listAllTestClasses` task at [`testermint/build.gradle.kts:301`](testermint/build.gradle.kts#L301) (used by `run-tests` to build the matrix) calls `hasExcludeOrUnstableTag` which only recognizes class-level `@Tag("unstable")` or `@Tag("exclude")`. So `UpgradeRehearsalTests` will be enumerated into the matrix, `./gradlew test --tests UpgradeRehearsalTests` will run, and `requiredEnv("UPGRADE_REHEARSAL_TARGET")` at [`UpgradeRehearsalTests.kt:355`](testermint/src/test/kotlin/UpgradeRehearsalTests.kt#L355) will throw `error("Required environment variable UPGRADE_REHEARSAL_TARGET is not set")` → matrix turns red on every subsequent integration run. Fix: add `@Tag("exclude")` at the class level (just below the imports, on the line before `class UpgradeRehearsalTests`). The dedicated rehearsal workflow uses `--tests "UpgradeRehearsalTests.complete upgrade rehearsal"` (specific method name) which bypasses the `excludeTags` filter, so the rehearsal itself still runs. Same fix should go on the `UpgradeRehearsalPrepTests` class introduced via the `previous-release-prep.patch`, applied to the old checkout — if that patch ever gets cherry-picked or the patch fails clean and someone hand-applies, the same gap exists there.

**Should fix:**

None.

**Nits:**

- The `AppExport.kt` change at [`testermint/src/main/kotlin/data/AppExport.kt:305-306`](testermint/src/main/kotlin/data/AppExport.kt#L305-L306) (new `devshardRequestsEnabled: Boolean? = null` field) also drops a "missing newline at end of file" — fine, just noting in case the repo enforces it.
- The dedicated workflow's `concurrency: cancel-in-progress: false` is the right call for a 4-hour rehearsal, but worth mentioning in `docs/testermint-upgrade-rehearsal.md`: if you re-dispatch on the same ref before the first run completes, the second sits in queue rather than canceling. That can surprise operators who hit Re-run by reflex.

**Looks correct:**

- **No shell-injection surface in the workflow inputs.** All three `inputs.*` values flow into bash via `'${{ inputs.X }}'` (single-quoted), and the python scripts ([`scripts/upgrade-rehearsal/resolve_versions.py`](scripts/upgrade-rehearsal/resolve_versions.py), [`resolve_previous_images.py`](scripts/upgrade-rehearsal/resolve_previous_images.py)) parse via `argparse` and shell out via `subprocess.check_output([list-form])` — no `shell=True`, no string-formatted commands.
- **`workflow_dispatch`-only trigger is the right default for v1.** Per the PR body, this avoids burning runner time on every PR while the rehearsal is still maturing. Wiring it into a release gate later is a one-line change.
- **`AppExport.kt` field is purely additive and nullable** (`val devshardRequestsEnabled: Boolean? = null`). Existing Gson-decoded payloads without the field unmarshal cleanly to `null`; no Kotlin-side reflection breakage. Matches the chain proto's `bool devshard_requests_enabled = 9` already on `upgrade-v0.2.14`.
- **The "boot old, apply prep patch, run prep test, build candidate, serve via httpd container, run upgrade test" sequence is logically sound** and the docker network (`chain-public`) reuses Testermint's existing convention, so the candidate archives are reachable from cosmovisor inside the cluster without a host-network port.
- **The python scripts do produce the right outputs for the matrix step that consumes them** (`steps.images.outputs.*` for `prepare_previous_images.sh`, `steps.versions.outputs.*` for the previous-release ref). The local validation list in the PR body covers them.
- **`git apply ../new/testermint/upgrade-rehearsal/previous-release-prep.patch`** at a known previous-release tag is deterministic — the patch is checked in, the old ref is pinned by `resolve_versions.py`, no fuzz risk in practice.

**Bottom line:** Land the one-line class-level `@Tag("exclude")` fix on `UpgradeRehearsalTests` (and the same on whatever `UpgradeRehearsalPrepTests` the prep patch introduces) and the PR is ready. Recommend rerunning CI to confirm the Go test failure is a flake.

docs/testermint-upgrade-rehearsal.md
```diff
@@ -0,0 +1,142 @@
...
+
+The rehearsal is intentionally separate from the normal integration matrix. It needs two repository checkouts, two versions of the Testermint harness, and live Docker state that must survive between phases.
+
+## What It Proves
+
+- The previous release can still boot in the current CI environment using the production images declared by that release.
+- The previous release's own Testermint harness can create realistic pre-upgrade state.
+- The candidate branch can submit a real Cosmos SDK software-upgrade proposal against that existing cluster.
+- Cosmovisor downloads and installs the candidate `inferenced` and `decentralized-api` upgrade archives.
+- The chain resumes block production after the upgrade and records the expected `last-upgrade-height`.
```
DimaOrekhovPS, line 13: Maybe one more condition should be "a PoC is successful after the upgrade"

patimen:
Addressed Dima’s post-upgrade PoC concern.

What changed:
- Phase 1 prep manifest now records baseline participant PoC weights from the previous release.
- Phase 2 completion waits for a candidate-binary `START_OF_POC` -> `SET_NEW_VALIDATORS` cycle after upgrade.
- Completion asserts every prepared miner remains active, not excluded, has non-zero PoC power, and stays within the configured power-change threshold.
- Docs now call out the epoch/PoC guarantee and troubleshooting path.

Validation:
- Testermint `testClasses` passes locally.
- Patched `release/v0.2.13` prep test compiles in a scratch worktree.
- Fresh upgrade rehearsal passed on head `1cb58d46`: https://github.com/gonka-ai/gonka/actions/runs/27030917144

Successful manifest evidence: pre-upgrade baseline weights were 10/10/10; post-upgrade PoC advanced from epoch 5 to 13 and post-upgrade weights remained 10/10/10.

patimen:
Follow-up for the epoch-boundary question: I extended the post-upgrade PoC verification so the completion test now waits through `START_OF_POC`, `SET_NEW_VALIDATORS`, and `CLAIM_REWARDS`, then captures a second active-participant power snapshot in the new epoch before running post-upgrade inference/devshard checks.

Validation run: https://github.com/gonka-ai/gonka/actions/runs/27034193789 passed on `bf335b32c`.

Artifact evidence from the completion manifest:
- prep final epoch: `5`
- post-upgrade PoC epoch after `SET_NEW_VALIDATORS`: `13`
- post-upgrade new epoch after `CLAIM_REWARDS`: `14`
- blocks: `START_OF_POC=510`, `SET_NEW_VALIDATORS=516`, `CLAIM_REWARDS=557`
- miner power stayed `10/10/10` before PoC, after PoC, and in the new epoch

The completion log also shows `CLAIM_REWARDS+2` reached, `after post-upgrade new epoch PoC power snapshot: epoch=14`, followed by successful post-upgrade normal inference and devshard settlement.

PR #1305, opened 2026-06-04
Devshard 0.2.13 v2 - inference observability after pruning and e2e fixes

a-kuprin:
## Summary

Lean inference lifecycle (prune + seal), validation observability persisted on diff apply, runtime-config / Testermint harness fixes from the v2 integration branch, and targeted Testermint stabilizers for collateral, confirmation PoC, and parallel devshard runs.

### Devshard — sealed inference stats after prune

When an inference is pruned from RAM, the host now **seals** pricing, token counts, model id, hashes, costs, and timestamps into session storage so dAPI and settlement paths keep working without the live `Mutable.Inferences` entry.

- `devshard/state/seal.go` — seal on prune triggers; late `MsgValidation` updates the sealed row
- `devshard/storage/*` — `inferences` table (SQLite epoch migrate, Postgres); interface cleanup; removed dead single-row `RecordValidationAppliedOnce` in favor of batch API
- `devshard/docs/inference-lifecycle.md` — documents storage layering at prune time

### Devshard — validation observability on diff apply

Validation and validation-vote txs in an applied diff asynchronously batch-write observability rows (`RecordValidationsAppliedOnce`), with in-flight cap, dedupe, and best-effort error handling.

- `devshard/host/validation_obs.go` (+ tests)
- `devshard/host/host.go` — hook from diff apply path
- Postgres: batch insert via `unnest`; SQLite: equivalent batch path
- `decentralized-api/internal/devshard/manager_test.go` — manager coverage

### Testermint — e2e and harness (`327e2f971`)

- **DevshardTests**: parallel sessions with isolated settlement; streaming inference e2e adjustments; shared helpers in `DevshardTestSupport.kt`
- **RuntimeConfigTests** / node admin / transaction fee tests aligned with v2 runtime config
- **LocalInferencePair**: join-validator readiness helpers (used by collateral and multi-join flows)
- `devshardctl` proxy tweak for test proxy path

### Testermint — collateral timing (`8bd360ba9`)

- `waitThroughEpochRewardClaim(targetEpoch)` — waits through `CLAIM_REWARDS` until stream-vesting credits land before balance assertions
- Fixes unbonding epoch math to use **current** epoch index when computing expected completion

### Testermint — confirmation PoC stability (`89186ce69`)

- `ConfirmationPoCPassTests` / `ConfirmationPoCFailTests`: `resetMlNodes = false` on `initCluster`
- Keeps docker-default `wiremock` on `*-mock-server` instead of mid-run swap to `wiremock2` @ `ml-0000.*.test` (which stayed **STOPPED** and yielded zero off-chain validations during automatic confirmation PoC in epoch 2)

`ConfirmationPoCMultiNodeTests` already used `resetMlNodes = false`.

## Test plan

- [ ] `go test ./devshard/...` (storage, host, state, bind)
- [ ] `go test ./decentralized-api/internal/devshard/...`
- [ ] Testermint: `DevshardTests` (parallel sessions, streaming e2e)
- [ ] Testermint: `CollateralTests` — deposit/withdraw with vesting wait
- [ ] Testermint: `ConfirmationPoCFailTests`, `ConfirmationPoCPassTests`
- [ ] Testermint: `PruningTests` (sealed row behavior after prune)
- [ ] CI `test-workflow` on gonka-ai/gonka (no fork-specific registry-cache write flags required)

## Notes for reviewers

- **Prune + seal** is the main behavioral change: confirm sealed fields match what dAPI needs post-prune and that late validations update the sealed row only, not RAM.
- **Validation obs** is best-effort by design; failures are logged, not fatal to diff apply.
- **Confirmation PoC tests** still use high `expectedConfirmationsPerEpoch` in spec; they now avoid `resetMlNodes` so setup does not replace operational nodes during an automatic confirmation event while `initialize()` is still running.


0xgonka (approved):
Backports three v0.2.13-v2 changes to main: sealed-inference snapshot on prune (so dAPI keeps serving observability after the live record evicts), async batch-write of validation observability rows on diff apply, and Testermint stabilizers. Consensus-safe by construction (late MsgValidation for sealed inferences rejects before touching RAM), no proto changes, no state-root composition shift. **No blockers.**

**PR state:**
- CI: 6/6 pass (Build and Test API Wrapper / Devshard / Versioned / chain, Storage schema DDL guard, build).
- Merge: clean (`mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`).
- Prior reviews: none.
- New activity: N/A — fresh PR.

**Must fix:**

None.

**Should fix:**

- **Async observability sync-fallback can stall `applyAndPersist` under storage hiccups.** [`devshard/host/validation_obs.go:60-89`](devshard/host/validation_obs.go#L60-L89) — when `validationObsInFlight` exceeds 64, the next call falls back to `writeValidationObsBatch` **synchronously, under `h.mu`**. The comment even acknowledges this: "If this warning shows up in practice, replace this fallback with a bounded worker queue or a drop-with-warn policy." There's no timeout on the synchronous write either, so a single slow Postgres write under contention stalls all `HandleRequest` calls. Two cheap mitigations: (a) put a `context.WithTimeout` on the sync path (e.g., 500ms), and (b) on timeout, log + drop rather than block. Observability is best-effort by design and recovery replays the diff anyway — locked-up host availability is the worse failure mode. Not blocking for this PR but worth folding in before this lands on main.

- **No graceful drain of in-flight obs goroutines at host shutdown.** [`devshard/host/validation_obs.go:84-88`](devshard/host/validation_obs.go#L84-L88) — `go func() { defer h.validationObsInFlight.Add(-1); writeValidationObsBatch(...) }()` fires and forgets; if the process exits while goroutines are mid-write, those rows are lost. Mitigated in practice because diff replay re-records them idempotently on next boot (the `ON CONFLICT DO NOTHING` against the `(epoch_id, escrow_id, inference_id, slot_id)` PK), so eventual consistency holds. A `sync.WaitGroup` drained from `Host.Close()` would make the property explicit instead of relying on the replay invariant; tiny diff, removes a class of "I shut the host down and rows were missing for 30s" surprises.

**Nits:**

- **`waitThroughEpochRewardClaim(targetEpoch)` has no overall timeout** ([`testermint/src/test/kotlin/CollateralTests.kt`](testermint/src/test/kotlin/CollateralTests.kt) helper). The inner `waitForStage` presumably has its own timeout, but if the chain genuinely halts the outer `while (... < targetEpoch)` loop spins on a non-advancing index. CI hangs are easier to debug than CI hangs with no failing assertion — a `withTimeout(15.minutes)` wrap would help.
- **`validationObservabilityFromStore` sums `uint32` slot totals into a `uint32` Totals field** ([`decentralized-api/internal/devshard/manager.go:713-741`](decentralized-api/internal/devshard/manager.go#L713-L741)). Wrap would require ~4B validations in one escrow — practically impossible, but a `uint64` Totals is one keystroke and removes a theoretical landmine.
- **`ConfirmationPoCPassTests`/`FailTests` `resetMlNodes = false`** is the right test-harness fix — production ML nodes are real, never reset — but worth a one-line comment on the test class so the next person doesn't "tidy up" by re-enabling it and re-hitting the `wiremock2 @ ml-0000.*.test STOPPED` bug the author chased down.

**Looks correct:**

- **Late MsgValidation for sealed inferences rejects deterministically across all hosts.** [`devshard/state/machine.go:781-785`](devshard/state/machine.go#L781-L785) (`applyValidation`) and [`:860-864`](devshard/state/machine.go#L860-L864) (`applyValidationVote`) both check `sm.sealedNonces[msg.InferenceId]` BEFORE any `state.Inferences` read or mutation. Returns `ErrInferenceSealed` cleanly. No SealedAcc fold, no live-map write — so no state-root divergence between hosts that have sealed at slightly different points. This is the critical correctness property for the whole feature and it's right.
- **`drainLiveIntoSealedAccLocked` sorts IDs ascending before folding** ([`devshard/state/seal.go:189-193`](devshard/state/seal.go#L189-L193)), so the SealedAcc preimage is identical across hosts even though Go map iteration order is randomized.
- **Per-host seal nonce is `sm.state.LatestNonce`** ([`devshard/state/seal.go:235`](devshard/state/seal.go#L235)) — same value across all hosts at the moment of the deterministic drain. No wall-clock or RNG dependency in the seal path.
- **In-host dedup before the batch write** ([`validation_obs.go:13-34`](devshard/host/validation_obs.go#L13-L34)) — `map[ValidationObsEntry]struct{}` collapses duplicate `(inference_id, slot_id)` within one diff so the storage round-trip is minimal and the storage `ON CONFLICT DO NOTHING` is only a defense-in-depth layer, not the primary dedup mechanism.
- **Storage path correctly fires AFTER `ApplyDiff` + `AppendDiff`.** The hook is gated on those two having succeeded, so any sealed-id rejection has already happened — the obs writer never receives validation entries for inferences the state machine rejected. The doc comment on `recordValidationObsFromAppliedDiff` explicitly calls this ordering out as load-bearing.
- **`sealed_inferences` table schema is forward-only with idempotent upsert** ([`devshard/storage/sqlite_epoch_migrate.go:41-74`](devshard/storage/sqlite_epoch_migrate.go#L41-L74), Postgres equivalent in `postgres_migrate.go`): `CREATE TABLE IF NOT EXISTS`, `ON CONFLICT(escrow_id, inference_id) DO UPDATE`. A double-seal of the same inference (replay) is a no-op; a partial migration on crash gets retried because step IDs survive in `schema_migrations`.
- **All three new methods (`InsertSealedInference`, `GetSealedInference`, `DeleteSealedInferences`) are implemented in every backend** — Memory, SQLite, Postgres, Hybrid all wire through. No surprise panic from a Hybrid wrapper hitting an unimplemented method.
- **`liveTimeouts` override in the proxy** ([`devshard/cmd/devshardctl/proxy.go:153-176`](devshard/cmd/devshardctl/proxy.go#L153-L176)) is the right shape: snapshot timeouts in `SessionConfig` are immutable per session, but governance can rotate them mid-session via the runtime params provider. Falling back to the snapshot when the live source isn't wired preserves the previous behavior.
- **`Storage schema DDL guard` CI check is green**, which is the right place to catch any hidden raw-SQL drift across the SQLite/Postgres backends for the new `inferences` table — the guard exists precisely so an asymmetric DDL between backends turns CI red instead of producing two divergent shapes in production.

**Bottom line:** Ship as-is, with the sync-fallback timeout and `WaitGroup` drain as the two follow-up improvements. The core property — "late-validation can't divergently mutate state" — holds cleanly.

a-kuprin:
@0xgonka 

I have

**Concern:** When `validationObsInFlight` exceeds 64, the code wrote obs rows
synchronously under `h.mu`, so a slow Postgres call could block every
`HandleRequest` on that host.

The stall is not obs-specific — `AppendDiff` (saving state) and seal writes are already synchronous under the same lock with no timeout.

So I have bounded all Postgres ops (`statement_timeout`, `lock_timeout`, `opCtx`).

Removed the sync fallback: at saturation the batch is dropped with a warn; It is the only observability path and if we are extremely slow on db we just don't try to write everything. But I don't think it is realistic because state is also written to same backend, and backpressure is something that we need to handle if we really observe the cases.
Moreover, I think that digging deeper not the right place at this PR and relase.
Digging deeper is something like apply high-availability patterns with external message queues, that is definitely out of scope now

**Concern:** Fire-and-forget goroutines may lose rows if the process exits mid-write.

Not implemented. Obs is idempotent (`ON CONFLICT DO NOTHING`) and `RecoverSessions` re-records on replay, so shutdown loss is self-healing.

## Nits

validationObservabilityFromStore sums uint32 slot totals into a uint32 Totals field ([decentralized-api/internal/devshard/manager.go:713-741](https://github.com/gonka-ai/gonka/pull/decentralized-api/internal/devshard/manager.go#L713-L741)).

and

ConfirmationPoCPassTests/FailTests resetMlNodes = false comment

are both fixed

0xgonka (approved):

0xgonka:
The blocker bar is "could this break consensus or settlement?" and the answer is no:

Late-validation rejection (the load-bearing consensus property) is implemented correctly in `machine.go` and verified: late `MsgValidation` returns `ErrInferenceSealed` before any state mutation, so no inter-host divergence.
Seal order is deterministic (sorted ascending IDs, shared `LatestNonce`), so `SealedAcc` fold matches across hosts.
CI is clean, schema-DDL guard is green, no proto changes, no migration that doesn't survive a partial crash.
The two Should-Fix items are real but not blocking:

Async-obs sync fallback can stall under storage hiccups (the comment in the code itself admits this is a known follow-up)
No `WaitGroup` drain at shutdown: eventually consistent via diff replay, so observable lost-row windows are bounded

PR #1284, opened 2026-05-31
Devshard Gateway

gmorgachev:
## Summary

This PR brings the devshard gateway work into `main`. It adds the `devshard gateway` service, request filtering/normalization, capacity-aware routing, redundancy handling, escrow lifecycle automation.

It also adds deployment support for running the devshard gateway alongside a join node via `docker-compose.devshard-gateway.yml` and `config.devshard.env.template`.

## Major Changes

- Adds the devshard gateway runtime in `devshard/cmd/devshardctl`, including gateway routing, host/session picking, redundancy, request accounting, request capture, response caching, OpenAPI/status/admin surfaces, and metrics.
- Adds capacity-aware request admission and routing, including live capacity state, participant limits, PoC-aware behavior, ghost probes, and per-host performance tracking.
- Adds escrow lifecycle and rotation support, including escrow checking, temp escrow workflows, settlement controls, phase gates, and REST-based chain transaction helpers.
- Adds request filters for chat completions compatibility, including message normalization, unsupported parameter stripping/rejection, structured output bounds, tool schema caps, reasoning/thinking handling, Kimi-specific behavior, metadata/user/stream options support, and vLLM crash mitigations.
- Improves devshard protocol/state/session handling, including connection tracking, session recovery, settlement validation, hash/bitmap/config updates, and transport client behavior.
- Adds deployment assets for the devshard gateway image and join-node gateway compose setup.
- Adds and updates docs for chat API behavior, supported parameters, host health, gateway architecture, known issues, protobuf transport, speculative proxy behavior, and multimodal planning.
- Adds broad unit/integration coverage across gateway routing, request filters, capacity state, escrow automation, redundancy, sessions, transport, state validation, and chat API parameter handling.

## Test Plan

Before merge, run the relevant Go test suites:

- `go test ./devshard/...`
- `go test ./decentralized-api/...`

For deployment validation:

- Copy `deploy/join/config.devshard.env.template` to `config.devshard.env`.
- Start the gateway compose file with `docker compose -f docker-compose.devshard-gateway.yml up -d`.
- Verify `/v1/status` and a representative `/v1/chat/completions` request.

## Contributors

Thanks to everyone who contributed commits to this branch:

- David Liberman (@libermans) — 45 commits
- Daniil Yankouski (@qdanik) — 23 commits
- Mykola Baichak (@baychak) — 9 commits
- clanster (@clanster) — 8 commits
- DimaOrekhovPS (@DimaOrekhovPS) — 2 commits
[...]

PR #1266, opened 2026-05-27
Create Gonka Network Development Roadmap

paranjko:
This PR adds the Gonka Network Development Roadmap as a strategic direction document for Gonka’s future development tracks.

If approved through governance and merged into the repository, the roadmap should become a shared reference point for future proposals, funding discussions, and network development.

> [!NOTE]
> This roadmap does not approve budgets, contractors, implementation details, Foundation structure, or immediate execution of any listed project.
>
> It sets the general strategic direction first.
>
> Future changes to roadmap tracks and projects should go through the Gonka Improvement Protocol process: public discussion, community feedback, a GitHub pull request showing the proposed changes to the roadmap document, and then a governance proposal to approve the update.

## Execution Framework

If this roadmap is approved, it should open two follow-up directions.

### 1. External Teams and Contributors

External teams, contributors, and ecosystem participants can start preparing concrete proposals for specific roadmap tracks and projects.

These proposals should define the actual scope, delivery plan, team, budget, timeline, milestones, KPIs, maintenance expectations, and handoff requirements.

### 2. Foundation Framework Discussion

The roadmap also opens a separate discussion around the legal and operational framework Gonka may need to support the network’s future development.

The Foundation’s legal structure, responsibilities, accountability model, treasury scope, and relationship with governance should be discussed and approved through a separate governance process.

PR #1307, opened 2026-06-04
fix(setup-report): avoid query-gas-limit by querying grants per cold->warm pair

redstartechno:

The admin endpoint `GET /admin/v1/setup/report` reports `permissions_granted` as `UNAVAILABLE` on a populated chain. Its `checkPermissions` step queries the node with `authz.GranteeGrants(grantee = warm key)`, which the Cosmos SDK implements as a scan of the **entire** authz grant store (grants are keyed granter-first), filtering by grantee in the pagination callback. As the total number of grants on the network grows, that query exceeds the node's `query-gas-limit` and fails with out-of-gas, so the warm-key permission check can no longer complete. Only the setup report is affected; the node and chain are unaffected (this is a read-only query, not a transaction).

- `decentralized-api/internal/server/admin/setup_report.go` calls `GranteeGrants(Grantee: warmKeyAddr)` and then filters the result in memory with `if grant.Granter == coldKeyAddr`.
- In cosmos-sdk v0.53.3 (the pinned version), the grant key layout is granter-first (`0x01|granter|grantee|msgType`). `x/authz/keeper/grpc_query.go` implements `GranteeGrants` as `prefix.NewStore(store, GrantKey)` — i.e. it iterates and unmarshals **every** grant in the store and keeps only those whose grantee matches. Cost scales with the total number of grants on chain, not with the grants relevant to this account.
- The repo's own upgrade handlers note the scale (~2,000 grants at ~100 hosts in `app/upgrades/v0_2_12`), which is enough to exceed the default `query-gas-limit` (10,000,000, set in `inference-chain/scripts/init-docker.sh`).

It replaces the grantee-wide scan with a pair-scoped query of the exact cold→warm grant:

```go
authzQueryClient.Grants(ctx, &authztypes.QueryGrantsRequest{
    Granter: coldKeyAddr,
    Grantee: warmKeyAddr,
})
```

`Grants(granter, grantee)` prefix-scans only the `granter|grantee` sub-tree (the ~20 grants in `InferenceOperationKeyPerms`), so query gas no longer grows with the size of the chain. Because the query already constrains both sides, the in-memory `grant.Granter == coldKeyAddr` filter is removed. This mirrors the keeper's existing `HasWarmKeyGrant` (`inference-chain/x/inference/keeper/devshard_settlement.go`), and the chain's own `AuthzKeeper` interface already exposes only `GranterGrants` and `Grants`, never `GranteeGrants`.

Low. The check's result is unchanged — the same `permissions_granted` PASS/FAIL/expiring-soon logic runs over the same set of cold→warm grants; only the underlying query is narrowed. No pagination is needed: a single cold→warm pair holds at most ~20 grants, below the SDK's default page size of 100. No other code path is touched (`checkFeegrant` already uses a precise point query and is unchanged).

The query changes from an O(all grants on chain) full-store scan to an O(grants for one pair) prefix scan, so it stays well under `query-gas-limit` regardless of network size, and the out-of-gas condition that produced `UNAVAILABLE` no longer occurs. The expected check result is identical to what `GranteeGrants` produced before it began hitting the gas limit.

`decentralized-api` only — `internal/server/admin/setup_report.go`, function `checkPermissions`. No chain/state-machine code, no other endpoints.

## Testing & evidence

- `gofmt` clean.
- Local toolchain note: this repo's full build requires cgo (the `blst` BLS dependency) and a C compiler, which I do not have set up locally, so I could not run the full `go build`/`go test` here. The change is a single, type-level substitution (`GranteeGrants`→`Grants`, with `Granter` added and the now-redundant granter filter dropped), verified against the cosmos-sdk v0.53.3 `authz` types and matching the in-repo `HasWarmKeyGrant` usage. CI/maintainer build will exercise the full compile and tests.
- Maintainer-verifiable end-to-end: on a chain with many grants, call `GET /admin/v1/setup/report` and confirm `permissions_granted` returns PASS/FAIL rather than `UNAVAILABLE` with an out-of-gas message.

Diff: 1 file, +13 / -10.


redstartechno:
Thanks for running integration. Looked into the 3 failing suites — they appear unrelated to this change, which only touches the `/admin/v1/setup/report` authz query:

- `StreamVestingTests` also failed on `main` itself the same day (push run [27322767077](https://github.com/gonka-ai/gonka/actions/runs/27322767077) after #1333 merged) and on another PR's run ([27323513130](https://github.com/gonka-ai/gonka/actions/runs/27323513130)).
- Those runs each failed different additional suites (`UpgradeTests`, `DevshardStandaloneTests`), while this PR's run failed `BLSDKGSuccessTest`/`InferenceRetryTests` — varying suites per run rather than a consistent failure.

Happy to dig further if you see a plausible path from this change to any of those suites.

patimen:
Thanks for working on this. I can't really review this in its current shape because the branch/commit history still appears to include unrelated roadmap and devshard gateway scope alongside the setup-report gas-limit fix.

Could you please clean this up or recut it so the PR contains only the setup-report gas-limit change? Once the diff/history is focused, I'll take another look.

redstartechno:
Done — recut the branch onto `upgrade-v0.2.14` with only the setup-report gas-limit commit. The unrelated roadmap/devshard scope appeared because the branch was originally cut from `main` (after #1266 and #1284 merged there) and the PR base later moved to `upgrade-v0.2.14`, which doesn't include those yet. The diff is now a single commit touching only `decentralized-api/internal/server/admin/setup_report.go`.

x0152 (approved):

tcharchian:
@redstartechno please send your wallet address so the bounty payment for your work can be processed.

redstartechno:
> @redstartechno please send your wallet address so the bounty payment for your work can be processed.

gonka105ce4495mj0mwkxqeasgdzqfq5jjrfq32eza5l

PR #1308, opened 2026-06-04
fix(devshard): return logprobs to clients who explicitly requested them (#1264)

### LINKED ISSUE #1264: logprobs, top_logprobs conditional stripping
The gateway forces logprobs upstream for validation, but clients who never asked for logprobs should not see them in the response (OpenAI-compatible default).
Clients who explicitly set `logprobs: true` or `top_logprobs` should get them back.

Now we always strip them even if client asked this fields in request

Recommendation: Adopt conditional stripping for logprobs and top_logprobs on the client-facing proxy path. Keep unconditional stripping for token_ids, prompt_token_ids, and prompt_logprobs.

redstartechno:

The devshard gateway returns an OpenAI-compatible API, but it never returns `logprobs`/`top_logprobs` to clients — even when a client explicitly asks for them. Fixes #1264.

The gateway force-enables logprobs on every upstream request for inference validation (`request_filters_parameters.go`: `newParameter("logprobs").withRule(RequestFilterStagePostLimits, ForceLiteralParameterHandler{Value: true})` and the matching `top_logprobs` rule). It then unconditionally strips `logprob`, `logprobs` and `top_logprobs` from the client-facing response on both response paths:

- non-streaming / chunked streaming: `filterClientInternalFields` → `stripClientInternalFields` deletes those keys recursively;
- the SSE `chat.completion` → `chat.completion.chunk` rewrite (`rewriteStreamingDataEvent`) drops the logprobs entirely when synthesizing chunks.

So a client sending `{"logprobs": true}` or `{"top_logprobs": 5}` receives a response with the field removed — a deviation from OpenAI behavior, where those fields should be returned when requested. (`token_ids`, `prompt_token_ids` and `prompt_logprobs` are genuine internals and must stay stripped.)

Conditional stripping driven by the client's **original** request intent:

1. Capture the client's original `logprobs`/`top_logprobs` into `chatRequest` in `readChatRequestFields`, **before** the gateway force-overwrites them. The pre-force values are preserved through `SyncRequestView` (the same pattern already used for `max_tokens`), so the forced upstream values don't clobber the captured intent.
2. Plumb the intent (`logprobClientIntent`) to the two response-strip call sites via request context, mirroring the existing `requestLogFromContext` pattern.
3. Strip `logprob`/`logprobs` only when the client did **not** request logprobs; the internal fields (`token_ids`, `prompt_token_ids`, `prompt_logprobs`) are still stripped unconditionally.
4. Both response paths are covered: the recursive field strip, and the streaming rewrite which now reconstructs the per-chunk `logprobs` object (`reconstructChunkLogprobs`). When logprobs are requested without `top_logprobs`, `top_logprobs` is emitted as a present-but-empty array (OpenAI's shape) on both paths.

The zero value of `logprobClientIntent` is "strip everything" — i.e. the exact prior behavior — so clients that don't ask for logprobs see no change.

- **Leaking logprobs to clients who didn't ask.** Mitigated by the safe-default design: intent is captured only from an explicit boolean `logprobs: true` (non-bool shapes are ignored, never rejected), `keep_top_logprobs` is gated on `keep_logprobs`, and the context zero value strips everything. Unit tests assert the default path still strips.
- **Leaking true internal fields.** `token_ids`/`prompt_token_ids`/`prompt_logprobs` are always in the strip set regardless of intent; covered by an existing test.
- **Validation impact.** None — the strip happens at the client-facing proxy boundary; devshard internals consume the pre-boundary host response (see the comment on `rewriteStreamingPayload`). The forced upstream logprobs used for validation are unchanged.
- **Note (not a regression):** the gateway clamps `top_logprobs` to a fixed value upstream for validation determinism, so a client requesting more alternatives than the clamp receives the clamped count. This matches the prior forcing behavior and is independent of this change.

New unit tests (`stream_rewrite_logprobs_test.go`) assert, for both response paths:

- client requested logprobs (+top_logprobs) → logprobs and top_logprobs returned;
- client requested logprobs without top_logprobs → logprobs returned, top_logprobs emptied;
- no request → logprobs/top_logprobs stripped (prior behavior);
- internals (`token_ids`/`prompt_*`) always stripped;
- the request pipeline captures the original intent even though the upstream body is still force-enabled (`logprobs:true`).

`decentralized-api`'s devshard gateway only (`devshard/cmd/devshardctl/`):
- `stream_rewrite.go` — conditional strip + streaming logprobs reconstruction + intent type/context helpers
- `proxy.go` — capture intent, plumb to both strip sites
- `request_filters.go`, `request_filters_parameters.go` — capture original logprobs intent
- `stream_rewrite_logprobs_test.go` (new), `proxy_test.go` — tests

No chain, consensus, proto, or genesis changes.

## Testing & evidence

`devshard` is a standalone Go module with no cgo dependencies, so the package builds and tests locally:

```
cd devshard
GOTOOLCHAIN=go1.24.2 go build ./cmd/devshardctl/
GOTOOLCHAIN=go1.24.2 go vet ./cmd/devshardctl/
GOTOOLCHAIN=go1.24.2 go test ./cmd/devshardctl/ -run 'TestFilterClientInternalFields|TestRewriteStreamingPayload|TestNormalizeChatRequest_CapturesOriginalLogprobIntent|TestDeferredWriter' -count=1
```

All new and existing related tests pass; `go build` and `go vet` are clean.

### Before
Client request: `{"messages":[...],"logprobs":true,"top_logprobs":3}`
Client response (logprobs requested but stripped):
```
{"choices":[{"index":0,"message":{"role":"assistant","content":"Hi"},"finish_reason":"stop"}]}
```

### After
Same request, logprobs returned (internals still stripped):
```
{"choices":[{"index":0,"message":{"role":"assistant","content":"Hi"},"logprobs":{"content":[{"token":"Hi","logprob":-0.5,"bytes":[72,105],"top_logprobs":[{"token":"Hi","logprob":-0.5},{"token":"Hey","logprob":-1.5}]}]},"finish_reason":"stop"}]}
```

### DIFF
--- a/devshard/cmd/devshardctl/stream_rewrite.go
+++ b/devshard/cmd/devshardctl/stream_rewrite.go
@@ -2,9 +2,38 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 )
+
+// logprobClientIntent records whether the client's ORIGINAL request asked for
+// logprobs / top_logprobs. The gateway force-enables logprobs upstream for
+// validation, so without this the client-facing strip cannot tell a client who
+// asked for logprobs from one who did not. The zero value (keep nothing) is the
+// historical behavior: strip everything.
+type logprobClientIntent struct {
+	keepLogprobs    bool
+	keepTopLogprobs bool
+}
+
+func logprobClientIntentFromRequest(req chatRequest) logprobClientIntent {
+	return logprobClientIntent{
+		keepLogprobs:    req.Logprobs,
+		keepTopLogprobs: req.Logprobs && req.TopLogprobs > 0,
+	}
+}
+
+type logprobIntentContextKey struct{}
+
+func withLogprobClientIntent(ctx context.Context, intent logprobClientIntent) context.Context {
+	return context.WithValue(ctx, logprobIntentContextKey{}, intent)
+}
+
+func logprobClientIntentFromContext(ctx context.Context) logprobClientIntent {
+	intent, _ := ctx.Value(logprobIntentContextKey{}).(logprobClientIntent)
+	return intent
+}
 
 type streamingRewritePayload struct {
 	ID                string          `json:"id"`
@@ -48,7 +77,7 @@ type rewriteLogprob struct {
 // Only if a host sent SSE-wrapped chat.completion JSON do we synthesize
 // chat.completion.chunk events for the client. The synthetic role chunk exists
 // only in that streaming rewrite.
-func rewriteStreamingPayload(p []byte) []byte {
+func rewriteStreamingPayload(p []byte, intent logprobClientIntent) []byte {
 	needsCompletionRewrite := bytes.Contains(p, []byte(`data: {`)) && bytes.Contains(p, []byte(`"message"`))
 	needsInternalFieldsFilter := containsAnyInternalField(p)
 	if !needsCompletionRewrite && !needsInternalFieldsFilter {
@@ -71,9 +100,9 @@ func rewriteStreamingPayload(p []byte) []byte {
 			continue
 		}
 		payload := bytes.TrimSpace(event[len("data: "):])
-		rewritten, ok := rewriteStreamingDataEvent(payload)
+		rewritten, ok := rewriteStreamingDataEvent(payload, intent)
 		if !ok {
-			filtered := filterClientInternalFields(payload)
+			filtered := filterClientInternalFields(payload, intent)
 			if !bytes.Equal(filtered, payload) {
 				fmt.Fprintf(&out, "data: %s\n\n", filtered)
 				changed = true
@@ -91,7 +120,7 @@ func rewriteStreamingPayload(p []byte) []byte {
 	return out.Bytes()
 }
 
-func rewriteStreamingDataEvent(payload []byte) ([]byte, bool) {
+func rewriteStreamingDataEvent(payload []byte, intent logprobClientIntent) ([]byte, bool) {
 	var resp streamingRewritePayload
 	if err := json.Unmarshal(payload, &resp); err != nil {
 		return nil, false
@@ -116,7 +145,7 @@ func rewriteStreamingDataEvent(payload []byte) ([]byte, bool) {
 			continue
 		}
 		if role := choice.Message.Role; role != "" {
-			writeStreamingChunkEvent(&out, resp, choice.Index, map[string]any{"role": role}, nil, nil)
+			writeStreamingChunkEvent(&out, resp, choice.Index, map[string]any{"role": role}, nil, nil, nil)
 		}
 
 		tokens := []rewriteLogprob(nil)
@@ -130,16 +159,16 @@ func rewriteStreamingDataEvent(payload []byte) ([]byte, bool) {
 				if i == len(tokens)-1 {
 					finish = choice.FinishReason
 				}
-				writeStreamingChunkEvent(&out, resp, choice.Index, delta, finish, choice.StopReason)
+				writeStreamingChunkEvent(&out, resp, choice.Index, delta, finish, choice.StopReason, reconstructChunkLogprobs(token, intent))
 			}
 			continue
 		}
 
 		if choice.Message.Content != "" {
-			writeStreamingChunkEvent(&out, resp, choice.Index, map[string]any{"content": choice.Message.Content}, nil, nil)
+			writeStreamingChunkEvent(&out, resp, choice.Index, map[string]any{"content": choice.Message.Content}, nil, nil, nil)
 		}
 		if choice.FinishReason != nil || len(bytes.TrimSpace(choice.StopReason)) > 0 {
-			writeStreamingChunkEvent(&out, resp, choice.Index, map[string]any{}, choice.FinishReason, choice.StopReason)
+			writeStreamingChunkEvent(&out, resp, choice.Index, map[string]any{}, choice.FinishReason, choice.StopReason, nil)
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


qdanik (approved):

redstartechno:
Rebased onto `main` after the devshard v2 merge (#1289), which rewrote `devshard/cmd/devshardctl/` and reintroduced the unconditional logprobs strip this PR fixes (`stream_rewrite.go` `clientStrippedFields` again lists `logprobs`/`top_logprobs` alongside the always-strip internals). The fix is re-applied on top of v2's new request/response flow rather than mechanically merged:

- Client logprobs intent is captured in `DecodeRequest` (before the gateway force-enables logprobs upstream for validation) and preserved across `SyncRequestView`, the same way v2 already preserves `MaxTokens`.
- Response strip is now intent-aware: internals (`token_ids`/`prompt_token_ids`/`prompt_logprobs`) are always stripped; `logprobs`/`top_logprobs` are returned only when the client requested them, on both the non-streaming and the SSE chunk-rewrite paths. `top_logprobs` requested-without is emptied to OpenAI's present-but-empty shape.

Verification: `go build ./...` and `go vet ./cmd/devshardctl/...` clean; `go test ./cmd/devshardctl/ -run Logprob -v` — all logprobs tests pass (the package's one unrelated failure is a pre-existing Windows-only path-separator test, fails identically on clean `main`).

@qdanik heads-up: this was force-pushed (history rewritten vs the commit you approved), so it likely needs a fresh look.

redstartechno:
Re-verified locally on Linux (matching CI's environment) after the devshard v2 rebase, at head `66b5fa03c`:

- `go build ./...` and `go vet ./...` — clean.
- Full logprobs suite under the race detector — `go test ./cmd/devshardctl/ -race -run Logprob` — all pass (20 tests, including the SSE chunk-rewrite reconstruction and the original-intent capture cases).

For full transparency: a whole-module `go test ./... -race` run does surface failures in `devshard/cmd/devshardctl`, but they are **not** from this PR. They are the pre-existing `receiptTime` data race in `redundancy.go` (write at `:1491` via `receiptOnce.Do`, read at `:2251` in `escalationForInflight`) that I reported separately in #1341 — it reproduces identically on a clean `main` (`8a35022be`) with this PR's commit removed, and this PR doesn't touch `redundancy.go`. The only other failure was a local environment gap (the storage DDL-lint test needs `ripgrep` installed); it passes once `rg` is present, and CI's "Storage schema DDL guard" already runs green.

Net: the change is green on its own tests under `-race`, and the diff is limited to `proxy.go`, `request_filters*.go`, and `stream_rewrite.go` (plus tests). Ready for a fresh review whenever convenient, @qdanik.

devshard/cmd/devshardctl/stream_rewrite.go
```diff
@@ -2,10 +2,39 @@ package main
...
+type logprobClientIntent struct {
+	keepLogprobs    bool
+	keepTopLogprobs bool
+}
+
+func logprobClientIntentFromRequest(req chatRequest) logprobClientIntent {
+	return logprobClientIntent{
+		keepLogprobs:    req.Logprobs,
+		keepTopLogprobs: req.Logprobs && req.TopLogprobs > 0,
+	}
```
x0152, line 24: can we clarify expected behavior for top_logprobs without logprobs:true? right now it is treated as no logprobs and stripped

redstartechno:
@x0152 Good question — yes, that's intentional. `top_logprobs` without `logprobs: true` is treated as "no logprobs" and stripped, which mirrors OpenAI's own semantics: in their API `top_logprobs` requires `logprobs: true`, and a request that sets `top_logprobs` alone is rejected. Since there's no top-level `logprobs` object on the response in that case, there's nowhere for `top_logprobs` data to live, so both come back stripped — the client gets the exact same shape as a plain no-logprobs completion.

The one deliberate difference from OpenAI: rather than 400-ing a `top_logprobs`-only request, we strip gracefully and return a valid response. The gateway already force-enables logprobs internally for validation, so erroring on the client's behalf felt more surprising than just giving back the normal (logprob-free) shape. If you'd prefer to match OpenAI strictly and reject it, that's a small addition in request validation — happy to add it, though it's a behavior change slightly beyond this PR's scope, so I left it as graceful-strip for now.

This case is covered by tests: `TestFilterClientInternalFields_EmptiesTopLogprobsWhenNotRequested` and `TestNormalizeChatRequest_CapturesOriginalLogprobIntent/client_asked_logprobs_with_explicit_top_logprobs_0`.

Separately — I re-applied the fix onto the new base `devshard-0.2.13-v2-r2` (the retarget left it showing conflicts against the old merge-base). The change itself is unchanged; build, vet, and the full logprobs suite pass locally.

a-kuprin:
I have updated target branch to take it to devshard-v4 release.

@redstartechno please rebase and resolve conflicts

redstartechno:
Rebased onto `ak/pixelplex-refactoring-into-r2` and force-pushed. The fix cherry-picked cleanly onto the v4 base — no content conflicts; every hunk landed in still-present context, so the PR remains the single original commit (6 files, +370/−27, all under `devshard/cmd/devshardctl/`).

I re-verified the fix is coherent on the new base, not just a clean textual merge. The client-intent plumbing still connects end-to-end: the original `logprobs`/`top_logprobs` intent is captured in `readChatRequestFields` before the PostLimits force-enable, preserved through `SyncRequestView` alongside `MaxTokens`, and consumed at both strip sites — `rewriteStreamingPayload` for the SSE chunk path and `filterClientInternalFields` for the non-streaming path. Internal fields (`token_ids` / `prompt_*`) stay unconditionally stripped regardless of client intent.

Tests: the full logprobs / stream / filter suite passes (`ok devshard/cmd/devshardctl`), covering both response paths, the Route A/B `top_logprobs` shape parity, and the always-strip-internal invariant.

x0152 (approved):

a-kuprin (approved):

PR #1311, opened 2026-06-06
fix: verify host signature in fetchSignature before storing

0xMayoor:
`processResponse` (`session.go:484`) marshals `StateSignatureContent`, recovers the address from `resp.StateSig`, and rejects unless it matches `expectedAddr`. the sibling `fetchSignature` (`session.go:1179`) doesn't — bytes returned by a host's GET `/sessions/:id/signatures` are stored after only a slot-ownership check.

`processResponse` only stores when `resp.StateSig != nil` (line 473), so a host that withholds its sig in the streaming response always lands in the next missing set — `CollectSignatures` routes it through `fetchSignature` every time. once the bytes are stored, `sigWeight` counts by slot not by validity, `hasQuorum` trips, collection cancels, `Finalize` short-circuits at line 974, settlement ships poisoned and the chain rejects. `s.signatures` is never reset, retries hit the same cache.

the existing TODO at `server.go:114` (restrict GET endpoints to group members) doesn't close this — a byzantine group member passes membership auth and still serves whatever its `HandleGetSignatures` returns raw from local storage. orthogonal to the missing content verification.

Fix: mirror `processResponse:472-493` in `fetchSignature` — `postStateRootForNonce`, marshal, RecoverAddress, match against expectedAddr (warm-key fallback). regression in `fetch_signature_verify_test.go`.

#1284

### DIFF
--- a/devshard/user/session.go
+++ b/devshard/user/session.go
@@ -472,24 +472,8 @@ func (s *Session) processResponse(hostIdx int, resp *host.HostResponse, inferenc
 	// Verify and store state signature.
 	if resp.StateSig != nil {
 		expectedAddr := s.group[hostIdx].ValidatorAddress
-		sigContent := &types.StateSignatureContent{
-			StateRoot: resp.StateHash,
-			EscrowId:  s.escrowID,
-			Nonce:     resp.Nonce,
-		}
-		sigData, err := proto.Marshal(sigContent)
-		if err != nil {
-			return fmt.Errorf("marshal state sig content: %w", err)
-		}
-		addr, err := s.verifier.RecoverAddress(sigData, resp.StateSig)
-		if err != nil {
-			return fmt.Errorf("%w: host %d: %v", types.ErrInvalidStateSig, hostIdx, err)
-		}
-		if addr != expectedAddr {
-			if !s.sm.CheckWarmKey(addr, expectedAddr) {
-				return fmt.Errorf("%w: host %d: expected %s, got %s",
-					types.ErrInvalidStateSig, hostIdx, expectedAddr, addr)
-			}
+		if err := s.verifyStateSignature(resp.Nonce, resp.StateHash, resp.StateSig, expectedAddr); err != nil {
+			return fmt.Errorf("host %d: %w", hostIdx, err)
 		}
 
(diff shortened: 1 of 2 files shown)


devshard/user/session.go
```diff
@@ -1279,11 +1196,32 @@ func (s *Session) fetchSignature(ctx context.Context, hostIdx int, nonce uint64,
...
+	}
+	sigContent := &types.StateSignatureContent{
+		StateRoot: postRoot,
+		EscrowId:  s.escrowID,
+		Nonce:     nonce,
+	}
+	sigData, err := proto.Marshal(sigContent)
+	if err != nil {
+		return false
+	}
```
a-kuprin, line 1212: The logic for constructing `StateSignatureContent`, marshaling it, and verifying the signature is duplicated from `processResponse`.
It is better to extract the signature verification logic into a helper method, like: `verifyStateSignature(nonce uint64, postRoot []byte, signature []byte, expectedAddr types.Address)`.

devshard/user/session.go
```diff
@@ -1279,11 +1196,32 @@ func (s *Session) fetchSignature(ctx context.Context, hostIdx int, nonce uint64,
...
 		if addr != expectedAddr {
 			continue
 		}
+		recovered, err := s.verifier.RecoverAddress(sigData, sigs[slotID])
+		if err != nil {
+			return false
+		}
+		if recovered != expectedAddr && !s.sm.CheckWarmKey(recovered, expectedAddr) {
+			return false
+		}
```
a-kuprin, line 1223: We should add logging the failures on `return false`

a-kuprin:
For information
The finalization step is subject for change https://github.com/gonka-ai/gonka/discussions/1369 and we should revisit it.

But anyway this missed signature fix is very useful, and added primitive will be reused

0xMayoor:
both done:
- verify logic pulled into verifyStateSignature, reused in processResponse and fetchSignature
- logging on the failure returns in fetchSignature

x0152 (approved):
looks good

non-blocking: I'd also add a valid wrong-key signature test

a-kuprin (approved):

PR #1312, opened 2026-06-06
fix: verify gossiped tx proposer signature before mempool entry

0xMayoor:
`HandleGossipNonce` (`server.go:560`) does `RecoverAddress` + slot-owner match before storing. the sibling `HandleGossipTxs` (`server.go:584`) doesn't — txs land via `OnTxsReceived` → `mempool.AddTx` with only a sender-membership check.

a byzantine group member POSTs a forged `MsgValidationVote{InferenceId: target, VoterSlot: <victim's own slot>, ProposerSig: garbage}`. it passes membership auth, enters the victim's mempool, and the victim's `collectValidationJobs` (`host.go:775`) calls `hasMempoolValidationOrVote` which trips purely on slot match — no proposer-sig check. the host skips its honest validation for that inference. same shape for `MsgRevealSeed` via `maybeRevealSeed`'s early-return at `host.go:675-677`.

never evicted: bad-sig tx never applies (`applyValidationVote` `machine.go:862` rejects via `verifyProposerSig`), so `mempool.RemoveIncluded` never sees it; persists for the whole session.

restricting gossip to group members doesn't close this — the route is already members-only. a byzantine member passes auth and still serves a forged inner tx. orthogonal to content verification.

Fix: add `StateMachine.VerifyGossipedTx` mirroring the existing per-tx `verifyProposerSig` calls in `applyValidation`, `applyValidationVote`, `applyRevealSeed`. expose a `Host.VerifyGossipedTx` wrapper. filter at ingestion in `HandleGossipTxs` (`server.go:584`) — drop bad-sig txs before passing to `OnTxsReceived`, same shape as `HandleGossipNonce`. regression in `gossip_tx_verify_test.go` with reject-forged + accept-valid pairs for each of the three affected tx kinds, plus unknown-slot and pass-through-other-kinds guards.

related: #1311 (same class on the `GET /sessions/:id/signatures` path), both introduced by #1284.

### DIFF
--- a/devshard/host/gossip_tx_verify_test.go
+++ b/devshard/host/gossip_tx_verify_test.go
@@ -1,0 +1,199 @@
+package host
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+
+	"devshard/internal/testutil"
+	"devshard/signing"
+	"devshard/state"
+	"devshard/stub"
+	"devshard/types"
+)
+
+// signMsg signs the canonical proposer preimage for a message using signer.
+// Mirrors host.signProposer: marshal the message (with ProposerSig already zero
+// or unset) and Sign(data).
+func signMsg(t *testing.T, signer *signing.Secp256k1Signer, msg proto.Message) []byte {
+	t.Helper()
+	data, err := proto.Marshal(msg)
+	require.NoError(t, err)
+	sig, err := signer.Sign(data)
+	require.NoError(t, err)
+	return sig
+}
+
+// newVerifyTestHost builds a Host with a writable state-machine the test can
+// observe. Reuses the same shape as newTestHost in host_test.go but renamed so
+// this file can be read in isolation.
+func newVerifyTestHost(t *testing.T, ownIdx int) (*Host, []*signing.Secp256k1Signer) {
+	t.Helper()
+	hosts := []*signing.Secp256k1Signer{
+		testutil.MustGenerateKey(t),
+		testutil.MustGenerateKey(t),
+		testutil.MustGenerateKey(t),
+	}
+	user := testutil.MustGenerateKey(t)
+	group := testutil.MakeGroup(hosts)
+	config := testutil.DefaultConfig(len(hosts))
+	verifier := signing.NewSecp256k1Verifier()
+	store := testutil.MustMemoryStore(t, "escrow-1", user.Address(), config, group, 100000)
+	sm, err := state.NewStateMachine("escrow-1", config, group, 100000, user.Address(), verifier, store)
+	require.NoError(t, err)
+	engine := stub.NewInferenceEngine()
+	h, err := NewHost(sm, hosts[ownIdx], engine, "escrow-1", group, nil, WithGrace(10))
+	require.NoError(t, err)
+	return h, hosts
+}
+
+// VerifyGossipedTx must reject a MsgValidationVote whose ProposerSig does not
+// recover to the claimed VoterSlot's address. Forged bytes are exactly the
+// payload a byzantine group member can serve via POST /sessions/:id/gossip/txs.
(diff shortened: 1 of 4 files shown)


a-kuprin:
`MsgRevealSeed` is deprecated as we have removed seed revealing stage at devshard `v2`. So you can remove it from signature validation

Reason is that we are going to change validation logic.

Also please rebase

a-kuprin:
Must fix:
- The system is vulnerable to a remote attacker crashing the host by sending malformed protobufs over gossip. Specifically, when checking the type of a protobuf wrapper (`tx.Tx.(type)`), the inner message field (e.g., `inner.Validation`) can be `nil`. If a peer gossips a `DevshardTx` containing a `DevshardTx_Validation` wrapper with a `nil Validation` field, accessing `msg.ValidatorSlot` will result in a panic, causing a liveness failure.
**Implement explicit `nil` checks for the inner message in each relevant case (Validation, ValidationVote, and RevealSeed) before attempting to dereference it.**
- The `VerifyGossipedTx` method acquires and holds the main state machine lock (`sm.mu.Lock()`) for its entire execution. A malicious peer could exploit this by flooding the node.
**Refactor `VerifyGossipedTx` to release the state machine lock before performing expensive cryptographic signature verification and other computationally intensive operations.**


Test nits:
- The failure messages in `TestVerifyGossipedTx_RejectsForgedValidation` (line 104), `TestVerifyGossipedTx_AcceptsValidValidation` (line 120), `TestVerifyGossipedTx_RejectsForgedRevealSeed` (line 134), and `TestVerifyGossipedTx_AcceptsValidRevealSeed` (line 151), are not as verbose or explicit as those in the ValidationVote tests. **Enhance these failure messages to provide more detailed and explicit information about why a test failed, matching the verbosity and clarity of the ValidationVote tests.**
- Add `t.Parallel()` to test functions, such as `TestVerifyGossipedTx_RejectsForgedValidationVote` and others, to enable parallel execution. This will speed up the overall test suite execution time.

0xMayoor:
rebased onto devshard-0.2.13-v2-r2 @a-kuprin, all points addressed:

- dropped MsgRevealSeed from the validation (deprecated in v2)
- nil-check on the inner message in each case before deref
- VerifyGossipedTx releases the state lock before the sig recovery now
- made the Validation test failure messages match the vote ones
- added t.Parallel() across the tests

couldn't get the nil case through proto.Unmarshal over gossip, added the guard anyway.

a-kuprin (approved):

x0152:
Nice fix

Two questions:
1. On gossip, VerifyGossipedTx writes to WarmKeys (part of state root). Could this drift nodes' state roots?
2. The warm-key fallback does a network call while holding sm.mu. Move it outside the lock?

a-kuprin (changes requested):
> 1. On gossip, VerifyGossipedTx writes to WarmKeys (part of state root). Could this drift nodes' state roots?

I think it could theoretically

The main problem is how Warm Key is rotated.
For example host can send some host signed message like `ConfirmInferenceStart` or `FinishInference` back to sequensor and it will be carried with further nonces to other hosts. But gossip message could be accidentally signed by a fresh key.

I think it is better to withhold current PR (while it useful) and think deeper, also taking into account this discussion:
https://github.com/gonka-ai/gonka/pull/986#discussion_r3523465880

@0xMayoor 
Currently gossip is disabled. And it should be disabled by design for state sharing before any dispute needed.

Idea is that we can synchronize using nonces and on regular devshard load it perfectly does so. Adding gossiping just adds extra traffic (each nonce cause cascade of gossip messages), and as one host can participate in different devshards, there will be a lot of unneeded traffic.
So we disable gossiping, and we should gossip only on finalization or disputes:
https://github.com/gonka-ai/gonka/discussions/1369

So let's don't push adding features to gossiping path before finalization and disputes aren't fully discussed and approved

0xMayoor:
both real @x0152, checked both against the code. nice catch

@a-kuprin understood, withholding.

PR #1318, opened 2026-06-06
fix(devshard): validate & clamp chat-completion params at the gateway instead of vLLM

qdanik:
## Summary
- The gateway forwarded out-of-range, non-positive, and wrong-typed chat-completion parameters straight to vLLM, which returned opaque upstream 400s — and `max_tokens: 0` (non-Kimi) was worse: vLLM emitted no content, the redundancy layer waited for a winner that can never arrive, and the request returned **0 bytes and timed out**, tying up a request slot/escrow for the full deadline.
- Added gateway-side validation so the backend only ever sees legal values. Out-of-range values with a sane target are clamped; values whose only out-of-bound is an exclusive lower bound or an invalid discrete value are rejected with a clear, field-named 400:

  | Param | Gateway rule |
  |-------|--------------|
  | `temperature` | clamp to `[0, 2]` |
  | `min_p` | clamp to `[0, 1]` |
  | `top_p` | clamp `>1` → `1`; reject `≤0` (must be in `(0, 1]`) |
  | `repetition_penalty` | clamp `>2` → `2`; reject `≤0` (must be `>0`) |
  | `top_k` | reject unless `-1` (disabled) or `≥1` |
  | `max_tokens` / `max_completion_tokens` | reject `0` on non-Kimi routes; Kimi-K2.6 floors small budgets to 16 |
  | `seed`, `min_tokens` | reject non-integer |
  | `stream`, `skip_special_tokens`, `detokenize`, `parallel_tool_calls` | reject non-boolean |
  | `stop`, `bad_words` | reject non-string array elements |
  | `stop_token_ids` | reject non-integer array elements |

- Collapsed the reject/validate logic into three predicate-driven handlers in the `paramvalidators` package — `RejectNumberParameter{Allow, Message}`, `ValidateScalarParameter`, `ValidateListElementsParameter` — driven from the catalog via reusable stateless `ParameterHandlerAdapter` instances (`rejectNonPositiveNumber`, `rejectInvalidTopK`, `mustBeBool`, `mustBeUint`, `elementsMustBeString`, `elementsMustBeUint`). Per-model behavior reuses the existing `ModelScopedParameterHandler` (Kimi clamps `max_tokens`, others reject `0`).

## Why
These parameters are all valid OpenAI/vLLM fields, but the gateway passed any value through and let the engine reject it — producing opaque upstream 400s with no field name, and in the `max_tokens: 0` case a silent hang instead of an error. Catching the bad value at the boundary gives the client an immediate, specific error and keeps invalid requests from consuming a slot.

Reject vs clamp is deliberate: `temperature`/`min_p`/`top_p`/`repetition_penalty` have a sane in-range target so they are clamped silently (in-range client behavior is unchanged), but `top_p`/`repetition_penalty` `≤ 0` have an exclusive lower bound where clamping to `0` would itself be illegal, and `top_k`/`max_tokens: 0` are invalid discrete values — those are rejected. `max_tokens` is per-model: Kimi-K2.6 already floors small budgets to 16 as a think-burn mitigation, so on that route `0` becomes `16` rather than a 400.

The handler unification and file split carry no behavior change — they remove near-duplicate reject/validate types and break up a ~960-line file that mixed handler primitives, the catalog, the JSON document, and parsing.

## Tests
New unit tests, each exercising `normalizeChatRequest` end-to-end (RED → GREEN):

- `TestNormalizeChatRequestClampsTemperatureBelowMin` — `temperature: -1` → `0`
- `TestNormalizeChatRequestClampsMinPIntoRange` — `min_p` clamped into `[0, 1]`
- `TestNormalizeChatRequestTopPClampsHighRejectsNonPositive` — `>1` → `1`; `0` / `-0.2` rejected
- `TestNormalizeChatRequestRejectsNonPositiveRepetitionPenalty` — `0` rejected; `>2` clamped to `2`
- `TestNormalizeChatRequestRejectsInvalidTopK` — `0` / `-2` rejected; `-1` / `1` / `50` accepted
- `TestNormalizeChatRequestRejectsZeroMaxTokens` — `max_tokens` / `max_completion_tokens` = `0` rejected (non-Kimi)
- `TestNormalizeChatRequestKimiClampsZeroMaxTokensInsteadOfRejecting` — Kimi route: `0` → `16`
- `TestNormalizeChatRequestRejectsNonBoolFlags` — non-bool `skip_special_tokens` / `detokenize` / `parallel_tool_calls` rejected
- `TestNormalizeChatRequestRejectsNonIntStopTokenIds` — non-int `stop_token_ids` element rejected
- `TestNormalizeChatRequestRejectsNonStringStopAndBadWords` — non-string `stop` / `bad_words` element rejected
- `TestNormalizeChatRequestRejectsNonIntMinTokens` — non-int `min_tokens` rejected

Existing parameter tests (clamp-above-max, NaN/string sanitize, greedy `n` coercion, Kimi `max_tokens` floor) unchanged. `go build ./...`, `go vet ./cmd/devshardctl/`, and `go test ./cmd/devshardctl/...` all green.

## Test plan
- [x] Out-of-range sampling values clamped or rejected per the table above
- [x] `max_tokens: 0` rejected on non-Kimi routes, floored to 16 on Kimi-K2.6
- [x] Wrong-typed params (bool / int / array element) rejected with a field-named 400
- [x] Handler unification + file split carry no behavior change (full suite green)
- [x] `go test ./...` in `devshard/` green

### DIFF
--- a/devshard/cmd/devshardctl/paramvalidators/handlers.go
+++ b/devshard/cmd/devshardctl/paramvalidators/handlers.go
@@ -206,6 +206,74 @@ func (h ValidateUintParameter) HandleParameter(ctx ParameterContext) error {
 	return nil
 }
 
+// RejectNumberParameter rejects the request when the parameter is present and parses as a
+// number but Allow returns false for it. Non-numeric or absent values pass through so the
+// dedicated type/shape validators own those cases. Used for range gates whose lower bound is
+// exclusive (e.g. top_p > 0), where clamping to the bound would itself be an illegal value.
+type RejectNumberParameter struct {
+	Allow   func(float64) bool
+	Message string
+}
+
+func (h RejectNumberParameter) HandleParameter(ctx ParameterContext) error {
+	value, exists := ctx.Document[ctx.Parameter]
+	if !exists {
+		return nil
+	}
+	number, ok := devshard.JSONNumericFloat64(value)
+	if !ok {
+		return nil
+	}
+	if h.Allow == nil || h.Allow(number) {
+		return nil
+	}
+	return fmt.Errorf("%s: %s", ctx.Parameter, h.Message)
+}
+
+// ValidateScalarParameter rejects the request when the parameter is present (non-null) but
+// Valid returns false for its raw JSON value. Catches wrong-typed scalars at the boundary
+// instead of forwarding them for an opaque upstream 400.
+type ValidateScalarParameter struct {
+	Valid   func(any) bool
+	Message string
+}
+
+func (h ValidateScalarParameter) HandleParameter(ctx ParameterContext) error {
+	value, exists := ctx.Document[ctx.Parameter]
+	if !exists || value == nil {
+		return nil
+	}
+	if h.Valid == nil || h.Valid(value) {
+		return nil
+	}
+	return fmt.Errorf("%s: %s", ctx.Parameter, h.Message)
+}
+
+// ValidateListElementsParameter rejects the request when any array element fails Valid.
+// Pass-through when the field is absent or not an array.
+type ValidateListElementsParameter struct {
+	Valid   func(any) bool
+	Message string
+}
+
+func (h ValidateListElementsParameter) HandleParameter(ctx ParameterContext) error {
+	raw, ok := ctx.Document[ctx.Parameter].([]any)
+	if !ok {
+		return nil
+	}
+	for index, item := range raw {
+		if h.Valid != nil && !h.Valid(item) {
+			return fmt.Errorf("%s[%d]: %s", ctx.Parameter, index, h.Message)
+		}
+	}
+	return nil
+}
+
+// JSON type predicates for the Validate*Parameter handlers above.
+func IsJSONBool(value any) bool   { _, ok := value.(bool); return ok }
+func IsJSONString(value any) bool { _, ok := value.(string); return ok }
+func IsJSONUint(value any) bool   { _, ok := devshard.JSONNumericUint64(value); return ok }
+
 // LengthCapListParameter bounds the number of entries in an array field, and
 // optionally the byte length of each string entry. MaxEntries=0 disables the
 // array cap; MaxEntryLen=0 disables the per-string cap.
--- a/devshard/cmd/devshardctl/paramvalidators/validate_test.go
+++ b/devshard/cmd/devshardctl/paramvalidators/validate_test.go
@@ -1,0 +1,102 @@
+package paramvalidators
+
+import (
+	"strings"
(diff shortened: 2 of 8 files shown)


qdanik:
Successfully tested and verified on the gateway.
Results: https://github.com/qdanik/gonka-inference-tests/blob/main/artifacts/2026-06-14/gateway-chat-params/README.md

devshard/cmd/devshardctl/request_filters_document.go
```diff
@@ -0,0 +1,207 @@
...
+func (d *ChatRequestDocument) String(name string) (string, bool) {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	value, ok := d.raw[name].(string)
+	return value, ok
+}
+
+// Object and Array return references into the document; do not retain them past
+// the immediate use or mutate them concurrently with other writers.
+func (d *ChatRequestDocument) Object(name string) (map[string]any, bool) {
```
a-kuprin: return raw references (`map[string]any` and `[]any`) to the document's internal data.

As `ChatRequestDocument` uses a `sync.RWMutex` for concurrent access, returning these references allows callers to mutate the internal state or read from it after the lock is released, potentially leading to race conditions or unexpected side effects.

Consider using deep copy of the data to prevent external modification of the internal state.

devshard/cmd/devshardctl/request_filters_document.go
```diff
@@ -0,0 +1,207 @@
...
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+	value, ok := d.raw[name].(map[string]any)
+	return value, ok
+}
+
+func (d *ChatRequestDocument) Array(name string) ([]any, bool) {
```
a-kuprin: Same here. Consider using deep copy of the data to prevent external modification of the internal state.

a-kuprin:
Free to skip nits

Tests:
1. Unmarshal the JSON body into a `map[string]any` or a specific struct before assertion. This makes tests more robust to formatting variations.
2. Tests currently rely on `require.Contains(t, err.Error(), field)` for identifying problematic fields in error messages. This approach is sensitive to changes in error message wording or structure. If possible, assert on a specific error type or a more structured error return. This provides a more robust and less brittle way to validate error conditions.

Styling:
A comment for `DocumentValidator` uses a colon style (`// DocumentValidator: validators in...`). Reformat the comment to follow Go conventions, e.g., `// DocumentValidator is the interface implemented by...`.

a-kuprin (approved):
Few nits possible to fix, but LGTM

PR #1325, opened 2026-06-08
[TEST ONLY] Validate PR #998 + local ARM64 follow-ups (v2)

patimen:
Test-only PR to validate CI for the full maintenance-window stack plus the local ARM64 / upgrade-test follow-up commits.

This branch is not intended to merge anywhere. It exists only to run GitHub checks in a repo where we have permission to execute workflows.

Contents:
- full maintenance-window base from Ryan Chen's `rc/maintenance_windows_v2`
- local upgrade/ARM64 follow-ups stacked on top
- CI-only arch scoping change for upgrade artifact builds

If checks pass here, we can use the result to inform the real stacked PR in Ryan's repo.


patimen:
Validation complete: this test-only PR is closing now that all checks have passed. Reference result: https://github.com/gonka-ai/gonka/pull/1325

PR #1326, opened 2026-06-09
fix(devshard): make sealing diff-driven to stop state-root divergence

a-kuprin:
## Problem

We are cleaning inferences from the devshard state, as they are polluting the state. But inference first should be finished/timedouted and after it is finished it could be validated.

The problem is that we don't know will be the inference selected for validation or not, so we need to keep all inferences for some period. We use grace nonces and grace timeout. **The grace timeout for sealing inferences (remove from state to db for stats) is the bad option.**

To be deterministic we should use only grace nonces, but it can introduce another problem for devshards under low-load. Number of inferences forms the number of nonces, it using only grace nonces is very uncertain on how long inference will be in state. So we added grace timeout, that breaks determenism as clock of hosts can diverge and it can lead to diverged state (sealed inferences has accumulator in state, also we are clearing inferences from state).

In target redesign we should know for certain in some N nonces, that inference will be selected for validation or not, so target design will be free of this bad option of using clocks to clear state.

At current intermidiate solution user is sequencing diffs and in diffs there is data to seal inferences. Hosts has gates for it, they check if inference is eligable to be cleaned, and also we have special parameter `sealAdmissionClockTolerance` that is set to `2 minutes` to handle clock divergence between hosts and user (inference grace timeout is `1 hour`). Using 2 minutes as a reliable clock-divergence makes it stable, but end design protocol should be free of grace timeout solution, that brings undeterminism.

## Implementation

Hosts could diverge from the user on `SealedAcc` / `post_state_root` because sealing used a wall-clock grace gate outside the signed diff. Sealing is now an explicit, signed part of each diff (`sealed_inference_ids` in `DiffContent`), folded deterministically in `applyCore` before the state root is computed.

- **User:** event-driven seal ledger; terminal inferences sealed on the next diff, `Finished` after nonce + wall-clock grace.
- **Host:** deterministic fold on apply/replay; live-only admission check (status + 10s clock tolerance) before signing; payload pruning decoupled (prune events emitted from `diff.SealedInferenceIDs` after apply).
- **Wire:** `sealed_inference_ids` carried in the `DiffContent` proto blob on HTTP transport and in SQL `txs_proto` (no schema change).

No protocol-version bump (v2 debugging branch).

## Test plan

- `cd devshard && go test ./...`
- `cd decentralized-api && go test ./...`
- `cd testermint && ./gradlew :test --tests "DevshardStandaloneTests" -DexcludeTags=unstable,exclude`

### DIFF
--- a/decentralized-api/internal/devshard/manager.go
+++ b/decentralized-api/internal/devshard/manager.go
@@ -483,7 +483,7 @@ func (m *HostManager) recoverStoredSession(escrowID string) (*transport.Server, 
 
 		for _, rec := range records {
 			sm.InjectWarmKeys(rec.WarmKeyDelta)
-			root, applyErr := sm.ApplyLocal(rec.Nonce, rec.Txs)
+			root, applyErr := sm.ApplyLocalSealed(rec.Nonce, rec.Txs, rec.SealedInferenceIDs)
 			if applyErr != nil {
 				return nil, fmt.Errorf("replay nonce %d: %w", rec.Nonce, applyErr)
 			}
--- a/devshard/host/host.go
+++ b/devshard/host/host.go
@@ -124,14 +124,12 @@ type Host struct {
 	// Payload prune tracking. All fields below are host-local off-state and
 	// must NOT participate in the state root or snapshot. They drive Tier A
 	// (terminal-status) and Tier C (stale-finished) pruning emission only.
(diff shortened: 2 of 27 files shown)


a-kuprin:
Better solution at
https://github.com/gonka-ai/gonka/pull/1328#pullrequestreview-4464427818

PR #1329, opened 2026-06-10
feat: optional mTLS between DAPI and ML nodes

x0152:
## What
Optional mTLS between the DAPI and ML nodes (alternative to #537).

## Why
Protects all PoC endpoints with transport-level auth + encryption, instead of per-request signatures. Locks down PoC traffic both ways: only the holder of the pinned cert can reach the ML node's PoC port (8080) or post callbacks to the DAPI (9100). No CA, nothing expires - just two self-signed certs that pin each other. It's fully optional and off by default.

## How to enable

**Same host:**
```bash
cd deploy/join
./gen-mlnode-certs.sh        # generate the two self-signed certs
source config.env
docker compose -f docker-compose.yml -f docker-compose.mlnode.yml -f docker-compose.mtls.yml up -d
```

**Remote ML node (separate machine):**

Put each machine's public IP in its cert SAN (the SAN must match the address the other side connects to).

```bash
DAPI_SANS="DNS:api,IP:<DAPI_PUBLIC_IP>" MLNODE_SANS="DNS:inference,IP:<MLNODE_PUBLIC_IP>" ./gen-mlnode-certs.sh
```

Then:
1. Copy mtls-certs/ to both machines.
2. On the DAPI host: `MTLS_POC_CALLBACK_URL=https://<DAPI_PUBLIC_IP>:9100` (must be https).
3. Use the ML node's public IP as `host` in `node-config.json`.

Hostnames work the same way: use `DNS:<name>` in the SANs and the matching names
in `MTLS_POC_CALLBACK_URL` / `node-config.json`.

## How to disable
Drop -f docker-compose.mtls.yml from the docker compose up command and restart:
```bash
docker compose -f docker-compose.yml -f docker-compose.mlnode.yml up -d
```
There's no silent fallback - mTLS is fully off once the override file is removed.


gmorgachev:
Hi @x0152 ! A little question, does it require MLNode to have domain name? At least locally resolved? Or IP would still work?

x0152:
> Hi @x0152 ! A little question, does it require MLNode to have domain name? At least locally resolved? Or IP would still work?

Yes, it could be used with ip as well

Updated doc and pr description

PR #1336, opened 2026-06-11
fix(payloadstorage): retry failed epoch prunes instead of skipping

### LINKED ISSUE #850: Bug: ManagedStorage silently skips failed epoch pruning — minPruned advanced before goroutines complete

redstartechno:

Fixes #850. When an automatic epoch prune fails in `ManagedStorage` (DB outage, file I/O error), that epoch is permanently skipped — its payload data is never pruned and accumulates on disk indefinitely on API nodes.

In `decentralized-api/payloadstorage/managed_storage.go`, `cleanup()` spawned one goroutine per epoch to call `storage.PruneEpoch`, only *logged* failures, and advanced `m.minPruned = threshold` immediately — before the goroutines completed. The next tick starts from the new `minPruned`, so a failed epoch is never retried. The failure modes are realistic: `HybridStorage.PruneEpoch` already anticipates and logs PostgreSQL/file prune errors, and the issue reproduces them (intermittent DB connectivity). The package is unchanged since v0.2.8, so the bug is live on `main`.

`cleanup()` now computes the prune range under the cache mutex, releases it, and prunes epochs **sequentially** in a new `pruneEpochs` method, advancing `minPruned` only past epochs that pruned successfully. On the first failure it stops and returns, so that epoch is retried on the next 30-second tick. All three `PruneEpoch` implementations are idempotent (`os.RemoveAll`, `DROP TABLE IF EXISTS`), so retries are safe.

Note one deviation from the fix sketched in the issue: pruning runs *outside* the cache mutex rather than while holding it. Holding `m.mu` during storage I/O would block all `Retrieve`/`Store` cache operations for the duration of a slow prune (e.g. a PostgreSQL timeout). Since `cleanupLoop` is the only caller of `cleanup()`, sequential pruning is still guaranteed not to overlap with itself — which also resolves the secondary concern in the issue (concurrent `PruneEpoch` calls for the same epoch from overlapping ticks).

- **Pruning is now sequential instead of parallel.** Bounded by `maxPruneLookback` to at most 10 epochs per tick, and it runs on a background 30-second ticker, so throughput is not a concern.
- **A cleanup tick now blocks until pruning finishes.** Cache eviction completes before pruning starts, and a delayed next tick only delays further pruning — cache reads/writes are unaffected because the mutex is released first.
- **Persistently failing epoch.** It is retried each tick until the existing `maxPruneLookback` guard jumps `minPruned` forward — exactly the documented design ("only last 10 epochs, older data requires manual prune"), so failures cannot stall pruning forever.

Mitigation: regression tests below cover the stop-at-failure and retry-then-catch-up paths; existing tests cover the unchanged lookback and no-prune-below-retain behaviors.

Two new regression tests using the package's existing mock seam (`pruneCb` hook follows the existing `storeCb` pattern):

- `TestManagedStorage_AutoPruneStopsAtFailedEpoch` — a failing epoch halts the advance; `minPruned` stays at the failed epoch; later epochs are not pruned out of order.
- `TestManagedStorage_AutoPruneRetriesFailedEpoch` — a transient failure is retried on the next cleanup; all epochs below threshold end up pruned exactly once; `minPruned` reaches threshold.

Both tests were verified to **fail against the previous implementation** (run with the old `managed_storage.go` restored via `git stash`) and pass with this change. The six pre-existing `ManagedStorage` tests still pass; three of them contained `time.Sleep` waits for the old async goroutines, which are removed since pruning is now synchronous.

- `decentralized-api/payloadstorage/managed_storage.go`
- `decentralized-api/payloadstorage/managed_storage_test.go`

No API or on-chain changes.

## Testing & evidence

`GOTOOLCHAIN=go1.24.2 go test ./payloadstorage/ -run TestManagedStorage -count=1 -v` (the `TestPostgresStorage_*`/testcontainers tests need Docker and are unrelated).

### Before

New regression tests run against the previous `managed_storage.go` (restored via `git checkout main -- ...`). Exact counts vary between runs because the old pruning was asynchronous — one representative run:

```
=== RUN   TestManagedStorage_AutoPruneStopsAtFailedEpoch
    managed_storage_test.go:220: expected 3 epochs pruned before failure, got 4: [0 1 2 4]
    managed_storage_test.go:224: epoch 4 should not be pruned past the failed epoch 3
    managed_storage_test.go:232: minPruned should stop at failed epoch 3, got 8
--- FAIL: TestManagedStorage_AutoPruneStopsAtFailedEpoch (0.00s)
=== RUN   TestManagedStorage_AutoPruneRetriesFailedEpoch
    managed_storage_test.go:261: expected 8 epochs pruned after retry, got 1: [1]
    managed_storage_test.go:272: epoch 0 was never pruned
    managed_storage_test.go:272: epoch 2 was never pruned
    managed_storage_test.go:272: epoch 3 was never pruned
    managed_storage_test.go:272: epoch 4 was never pruned
    managed_storage_test.go:272: epoch 5 was never pruned
    managed_storage_test.go:272: epoch 6 was never pruned
    managed_storage_test.go:272: epoch 7 was never pruned
--- FAIL: TestManagedStorage_AutoPruneRetriesFailedEpoch (0.00s)
FAIL
FAIL    decentralized-api/payloadstorage        0.099s
```

### After

```
--- PASS: TestManagedStorage_CacheHit (0.00s)
--- PASS: TestManagedStorage_CacheExpiration (0.02s)
--- PASS: TestManagedStorage_StoreTracksMaxEpoch (0.00s)
--- PASS: TestManagedStorage_AutoPruneTriggersInCleanup (0.03s)
--- PASS: TestManagedStorage_AutoPruneSkipsOldEpochs (0.00s)
--- PASS: TestManagedStorage_AutoPruneStopsAtFailedEpoch (0.00s)
--- PASS: TestManagedStorage_AutoPruneRetriesFailedEpoch (0.00s)
--- PASS: TestManagedStorage_NoPruneWhenBelowRetainCount (0.00s)
PASS
ok      decentralized-api/payloadstorage        0.138s
```


redstartechno:
Rebased onto `main` after the devshard v2 merge (#1289). v2 added a `DeleteInference` method to `ManagedStorage` but did not touch the prune loop, so the fix applies directly — the only conflict was an additive one in the test file (both sides appended new tests at the same spot), resolved by keeping both sets.

Behavior is unchanged from the original: failed epoch prunes are retried on the next cleanup tick instead of being skipped forever (#850). `pruneEpochs` prunes `[minPruned, threshold)` in order outside the lock, advances `minPruned` only past each epoch that pruned successfully, and returns on the first failure so that epoch is retried next tick; the `maxPruneLookback` catch-up cap is preserved.

Verification (WSL/Linux, the only env where the package builds here): `go build ./payloadstorage/` and `go vet ./payloadstorage/` clean; `go test ./payloadstorage/ -race` passes, including the two regression tests, which I confirmed still fail on the pre-fix code. The Postgres-backed container tests self-skip without Docker in my local env; this change only touches the in-memory prune-loop control flow, not any storage backend.

x0152 (approved):

x0152:
@redstartechno 
one nit: can you add a timeout to each PruneEpoch so a stuck prune doesn't freeze cleanupLoop?

redstartechno:
Thanks for the review @x0152 — good call.

Pushed `16a88adde`: each `PruneEpoch` is now wrapped in a per-epoch `context.WithTimeout` (default 30s, a named const that matches the 30s cleanup tick so a stalled epoch costs at most one tick). On timeout `PruneEpoch` returns a context error, which the loop already treats like any other failure: `minPruned` is not advanced and the epoch is retried on the next cleanup tick — so the timeout composes with the retry behavior from this PR rather than skipping anything.

One honesty note on scope: the deadline cancels backends that observe `ctx`. `PostgresStorage.PruneEpoch` runs `pool.Exec(ctx, "DROP TABLE …")`, so pgx cancels the in-flight query server-side — that's the realistic stall (a `DROP TABLE` blocked on a lock), and it's now bounded. `HybridStorage` forwards the same `ctx` to both backends. `FileStorage.PruneEpoch` is an `os.RemoveAll` on a local dir, which doesn't observe `ctx`; it's synchronous local I/O rather than a network/lock wait, so I left it as-is. If you'd prefer the file path bounded too, I can run it in a goroutine and select on `ctx.Done()` — happy to add that, just didn't want to leak goroutines on a path that doesn't realistically hang.

Evidence (WSL, Go 1.24.2):
- New regression test `TestManagedStorage_AutoPruneTimesOutStuckEpoch`: a mock backend whose epoch-3 prune blocks until its context is cancelled. With the timeout it returns `context deadline exceeded`, epochs 0–2 are pruned, `minPruned` stops at 3 for retry. Reverting just the timeout (non-cancelling context) reproduces the freeze the test guards against — it fails at 2.00s with "cleanup never returned".
- Full `go test ./payloadstorage/ -race` green, including all existing prune/retry tests.

The PR stays a single commit.

redstartechno:
Follow-up — I went ahead and made the bound cover every backend, not just the ctx-aware ones, so this supersedes my note above about leaving `FileStorage` as-is.

`pruneEpochWithTimeout` now (a) still passes the deadline to `PruneEpoch`, so ctx-aware stores cancel the in-flight work server-side (Postgres `pool.Exec`, and Hybrid forwards it), and (b) runs the call in a goroutine and selects on the deadline, so `cleanupLoop` is released at the timeout even for a backend that ignores ctx — i.e. `FileStorage.os.RemoveAll`. That's the case that mattered for your "a stuck prune doesn't freeze cleanupLoop" framing and a bare `ctx` couldn't guarantee it.

A timed-out prune is still treated as a normal failure: `minPruned` is not advanced and the epoch is retried next tick. The send channel is buffered, so the prune goroutine never blocks after we've moved on; and since prunes are idempotent (`DROP TABLE IF EXISTS`, `RemoveAll` on a missing dir is a no-op), a goroutine that finishes after the deadline — or the next-tick retry — is harmless.

Added a second regression test, `TestManagedStorage_AutoPruneTimesOutCtxIgnoringBackend`, with a mock that blocks **without** observing ctx. Evidence (WSL, Go 1.24.2):
- With the goroutine wrapper: `cleanup` returns at the deadline, epochs 0–2 pruned, `minPruned` stops at 3 for retry.
- Reverting to a bare `ctx` pass: that test fails at 2.00s ("a ctx-ignoring backend froze cleanupLoop"), confirming the wrapper is what provides the guarantee.
- Full `go test ./payloadstorage/ -race` green (both timeout tests + all existing prune/retry tests).

Still a single commit (force-pushed `0560bcca6`).

x0152:
I see one blocker right now: if a delete gets stuck, its background task never stops, and we start a new one every 30s - so they can keep so the number of stuck tasks can keep growing

redstartechno:
Good catch — fixed. The previous wrapper released `cleanupLoop` on timeout but left the prune goroutine running for ctx-ignoring backends (FileStorage's `os.RemoveAll`), and every 30s tick spawned another for the same stuck epoch.

I've added a per-epoch single-flight guard: before spawning, the epoch is marked in-flight under the existing lock; while that marker is set, subsequent ticks skip respawning and just retry the epoch next tick (`minPruned` still doesn't advance past it). The marker is cleared from inside the goroutine when `PruneEpoch` finally returns, so a fresh attempt starts on the next tick once it completes. Net: at most one background prune per epoch is ever alive, regardless of how long a delete stays stuck.

New regression test `TestManagedStorage_AutoPruneDoesNotStackStuckGoroutines` drives 3 cleanup ticks against a ctx-ignoring backend and asserts `PruneEpoch` is invoked exactly once for the stuck epoch (the old goroutine+select code recorded 3). Full `payloadstorage` suite passes under `-race`.

One honest caveat: if an epoch stays wedged forever, the `maxPruneLookback` cap eventually advances `minPruned` past it (existing behavior — older data then needs manual prune), and that one stale in-flight marker outlives its relevance since the leaked goroutine never returns to clear it. Bounded to a single map entry per pathologically-wedged epoch, so harmless.

redstartechno:
Rebased onto current main (`065e0fa42 → 4b2f66326`) after the 0.2.14 merge made the branch conflict. No logic changes — the diff is identical to the previously reviewed state except for upstream's import-path move (`decentralized-api/logging` → `common/logging`) and dropping two test cases for the now-removed `DeleteInference` feature. Re-verified on the new base with `go test ./payloadstorage/ -race`: all 8 `TestManagedStorage_*` tests pass, build and vet clean.

@x0152 this should now be ready for your re-review of the single-flight guard from my earlier reply.

redstartechno:
@x0152 Gentle ping on this one when you have a moment — the single-flight guard addressing your goroutine-accumulation blocker has been in since July 9, and the branch is now rebased onto current main (post-0.2.14) with no logic changes and all checks green. Happy to adjust anything if the guard isn't quite what you had in mind.

bonujel:
Not a blocker — I went and measured the caveat you flagged on July 9, since it was the one claim in this PR nobody had checked. It holds up.

Probe on this branch (`4b2f663`), `retainCount=2`, epoch 5 wedged permanently, one new epoch per cleanup tick:

```
threshold=28 minPruned=28
never pruned below threshold: [5]  (count=1)
```

So the loss really is bounded to the wedged epoch — pruning resumes immediately after the lookback cap laps past it. I had expected the lap to strand the whole run between the wedged epoch and `threshold-10`; it does not, because `minPruned` only moves to `threshold - maxPruneLookback` and the epochs from there to `threshold-1` are still pruned in the same tick. Your description was accurate.

The probe, for anyone who wants to re-run it:

```go
mock := newMockStorage()
mock.pruneCb = func(epochId uint64) error {
    if epochId == 5 {
        return errors.New("backend wedged for this epoch")
    }
    return nil
}
ms := NewManagedStorageWithSize(mock, 2, time.Minute, 100)

for e := uint64(0); e <= 30; e++ {
    _ = ms.Store(context.Background(), fmt.Sprintf("inf-%d", e), e, []byte("p"), []byte("r"))
    ms.cleanup()
}
// diff mock.getPruned() against [0, maxEpoch-retainCount)
```

### One gap left, and it is in this PR's own theme

The lookback lap in `cleanup()` is silent:

```go
if m.minPruned+maxPruneLookback < threshold {
    m.minPruned = threshold - maxPruneLookback   // no log
}
```

#850 is titled "silently skips failed epoch pruning". This PR fixes the silent skip on *failure* — a `Warn` plus a retry next tick. The skip on *falling behind* is still silent, and it is now the only one left.

It is not hypothetical under backlog. Same probe, but with the 31 epochs stored before any cleanup runs and **nothing wedged at all**:

```
threshold=28 minPruned=28
never pruned below threshold: [0 1 2 ... 17]  (count=18)
pruned: [18..27]
```

Eighteen epochs of payload data abandoned, no failure anywhere, not a line in the log. That is the documented `maxPruneLookback` contract ("only last 10 epochs, older data requires manual prune", L43) and it predates this PR, so it is not a defect here. But an operator told to do a manual prune has no signal telling them one is due — which is the same class of blind spot #850 opened over.

A single log line (or a counter) where `minPruned` is forced forward, saying how many epochs were dropped, would close the story this PR is telling. Happy to leave it out of scope if you would rather ship what is here — the branch has been waiting long enough.

Evidence: Go 1.25.9, `go test ./payloadstorage/` on `4b2f663`; the existing `TestManagedStorage_*` suite passes alongside the probes.

Issue #447, opened 2025-11-20
Node Registration Does Not Update After Migration (API stuck using old on-chain config)


Asplana92:
🐞 [BUG] Node Registration Cannot Update After Migration (stuck on old on-chain state; diff returns no changes)
Summary

After migrating the ML inference server to new infrastructure and switching from an unsupported model (Llama) to a governance-approved one (Qwen2.5-7B-Instruct), the node becomes permanently stuck using outdated on-chain registration data.

Even though node_config.json contains the correct configuration and the API reads it correctly on startup, the API never submits an update transaction, because the diff logic always reports “No differences”, even when the on-chain state is completely different.

This leaves the operator unable to recover or participate in inference, even with fully correct infrastructure.

✅ Expected Behavior

When node_config.json changes (host, port, hardware, models),
→ API should detect differences.
→ API should submit an update transaction.
→ On-chain registration should be updated.

❌ Actual Behavior

API loads correct local config

On-chain data is old and completely mismatched

But diff logic reports:
[sync nodes] Hardware diff: NewOrModified:[] Removed:[]
[sync nodes] No diff to submit
API then continues to use the old host + port, even though they are no longer valid:
ERROR: queryNodeStatus → dial tcp OLD_IP:8081 → connection refused
Node is stuck in FAILED state across epochs.

📌 Timeline & Reproduction Story

This issue appeared during a normal model migration, following official guidance.

(1) Initial setup (worked, but received 0 assignments)

Model: Meta-Llama-3.1-8B-Instruct

Framework: llama.cpp

Host: old_IP:8081

Hardware: RTX 4090

Registered successfully

Received 0 inference tasks for multiple epochs, so decision was made to switch to a governance model.

(2) Official guidance from Discord

Gonka team advised:

Llama is not governance-supported

Required: models from https://gonka.hyperfusion.io/v1/models

Recommended: use vLLM

Also required: additional disk space

So migration to new hardware was performed.

(3) Infrastructure migration
A new server with enough disk space was deployed:
Inference server:
- GPU: RTX 4090
- Disk: ~90GB
- Model: Qwen/Qwen2.5-7B-Instruct
- Framework: vLLM
- vLLM running on port 8081

Validation:
curl http://localhost:8081/v1/models
→ returns Qwen2.5-7B-Instruct   (OK)

(4) node_config.json updated
Correct configuration:
[{
  "id": "node1",
  "host": "NEW_IP",
  "inference_port": 8081,
[...]

PR #1337, opened 2026-06-11
fix(broker): sync host/port changes to chain

redstartechno:

When a hardware node's host or port changes — typically because the operator migrated the inference server to a new machine — the API node never updates the on-chain registration, which keeps pointing at the old endpoint. Addresses the diff-detection half of #447.

`convertInferenceNodeToHardwareNode` populates `Host` and `Port` and the chain validates and stores both (`message_submit_hardware_diff.go`, `msg_server_submit_hardware_diff.go`), but `areHardwareNodesEqual` — the function `calculateNodesDiff` uses to decide whether a node needs resubmission — compared only `LocalId`, `Status`, `Hardware`, `Models` and `Version`. A node whose host or port changed (with the same GPU and models, the classic migration) was reported as unchanged, so `MsgSubmitHardwareDiff` was never sent.

This also breaks the supported runtime path: updating a node via `PUT /admin/v1/nodes/:id` changes local broker state, but the 60-second sync loop (`nodeSyncWorker` → `syncNodes`) still considers it equal to the stale chain record and submits nothing.

Issue #447 reports the operator-facing symptom. The reporter's full scenario also involves the `node_config.json` merge-once behavior (a separate design question, see scope note under Risks); the host/port comparison gap is independently reproducible from code — demonstrated below.

Adds `Host` and `Port` to the field-by-field comparison in `areHardwareNodesEqual`, so a host/port change marks the node as `NewOrModified` and the sync loop submits the corrective diff. All `HardwareNode` proto fields are now compared.

- **Resubmission churn considered and ruled out.** `SubmitHardwareDiff` is the only writer of `HardwareNodes` and stores submitted nodes verbatim; `Port` is produced canonically via `strconv.Itoa(PoCPort)`. After one successful submit, local equals chain and the sync loop goes quiet. The new `TestAreHardwareNodesEqual_HostPort` includes an identical-nodes case to guard this.
- **One-time effect after upgrade:** nodes whose on-chain host/port is stale will submit a single corrective diff on the next sync tick, then converge. That is the intended healing, not churn.
- **Scope:** this does not change the `node_config.json` merge-once behavior described in #447 (config merges into the local DB once, gated by `kvKeyNodeConfigMerged`; runtime node management is the admin API). Whether that design should change is left to maintainers.
- **Known limitation (pre-existing):** `InferencePort` is not part of the `HardwareNode` proto, so inference-port-only changes remain invisible to the chain registry. Out of scope here.

The new unit test `TestAreHardwareNodesEqual_HostPort` covers the equal pair, a host-only change and a port-only change; it fails against the old implementation and passes against the new one (CI runs the package tests — see local limitation below). Additionally, an extracted-function harness reproduces the exact #447 migration scenario against both versions of the function (evidence below).

- `decentralized-api/broker/broker.go` — `areHardwareNodesEqual`
- `decentralized-api/broker/broker_test.go` — new test

No chain-side changes.

## Testing & evidence

Local limitation, stated openly: the `broker` package transitively requires cgo (`supranational/blst`) and this dev machine has no C compiler, so `go test ./broker/` cannot run here — CI covers the real package tests. To still verify behavior locally, the old and new versions of `areHardwareNodesEqual` (plus `hardwareEquals`) were extracted verbatim into a standalone module with the `HardwareNode` fields mirrored 1:1 from `hardware_node.pb.go`, and exercised with the #447 migration scenario: same `LocalId`/`Status`/`Hardware`/`Models`/`Version`, new host and port.

### Before

Old function — the harness test is written to fail loudly when the bug fires:

```
=== RUN   TestOldEquality_HostPortChangeNotDetected
    equal_test.go:32: BUG (#447): old equality reports local 203.0.113.20:8081 == chain 198.51.100.7:8080 — diff never submitted
--- FAIL: TestOldEquality_HostPortChangeNotDetected (0.00s)
```

### After

New function — detects the migration; identical nodes still compare equal (no diff churn):

```
=== RUN   TestNewEquality_HostPortChangeDetected
--- PASS: TestNewEquality_HostPortChangeDetected (0.00s)
=== RUN   TestNewEquality_IdenticalNodesStillEqual
--- PASS: TestNewEquality_IdenticalNodesStillEqual (0.00s)
```

### DIFF
--- a/decentralized-api/broker/broker.go
+++ b/decentralized-api/broker/broker.go
@@ -727,6 +727,12 @@ func areHardwareNodesEqual(a, b *types.HardwareNode) bool {
 	if a.Status != b.Status {
 		return false
 	}
+	if a.Host != b.Host {
+		return false
+	}
+	if a.Port != b.Port {
+		return false
+	}
 	if len(a.Hardware) != len(b.Hardware) {
 		return false
 	}
--- a/decentralized-api/broker/broker_test.go
+++ b/decentralized-api/broker/broker_test.go
@@ -2120,6 +2120,20 @@ func TestAreHardwareNodesEqual_Version(t *testing.T) {
 	assert.False(t, areHardwareNodesEqual(a, b), "nodes with different versions should not be equal")
 }
 
+func TestAreHardwareNodesEqual_HostPort(t *testing.T) {
+	a := &types.HardwareNode{LocalId: "node-1", Host: "10.0.0.1", Port: "9090", Models: []string{"m1"}}
+	b := &types.HardwareNode{LocalId: "node-1", Host: "10.0.0.1", Port: "9090", Models: []string{"m1"}}
+
+	assert.True(t, areHardwareNodesEqual(a, b), "identical nodes should be equal")
+
+	b.Host = "10.0.0.2"
+	assert.False(t, areHardwareNodesEqual(a, b), "nodes with different hosts should not be equal")
+
+	b.Host = a.Host
+	b.Port = "9091"
+	assert.False(t, areHardwareNodesEqual(a, b), "nodes with different ports should not be equal")
+}
+
 func TestConvertInferenceNodeToHardwareNode_Version(t *testing.T) {
 	node := createTestNode("node-1")
 	node.State.MlNodeVersion = "v2.3.4"


gmorgachev:
Do we need host and port recorded on mainnet at all? Wouldn't be better to remove these fields?

redstartechno:
Good question — I looked into it before answering.

As far as I can tell, the on-chain `HardwareNode.host`/`port` fields are currently write-only: they're submitted and stored, but nothing reads them for any real logic.

- The chain only length-validates them (`message_submit_hardware_diff.go`); no keeper logic branches on them.
- Model assignment reads only `Models` and `LocalId` from the stored nodes (`model_assignment.go`), never host/port.
- The broker routes to its own ML nodes from the operator's *local* config (`node.Host` / `InferencePort` / `PoCPort` in `broker.go`), not from the chain-stored value.
- The `HardwareNodes` query exposes them, but I couldn't find any in-repo consumer (CLI, gateway, or peer) that reads them — devshard doesn't reference on-chain hardware nodes at all, and there's no cross-participant routing off these fields.
- Minor smell pointing the same way: the stored `port` is actually the local PoC port (`strconv.Itoa(node.PoCPort)`), not the inference port.

So you're right that removing them is a reasonable — arguably cleaner — option, and if they go, this PR isn't needed. In fact, since nothing reads the fields, a fair point against this PR is that it adds an on-chain `MsgSubmitHardwareDiff` write whenever only host/port change — churn to keep an unread field accurate, which itself argues for dropping them.

The case for keeping them is mostly that they're observable state (operators/tooling inspecting `HardwareNodes` see the real values) and a possible hook for future peer discovery. If neither matters, removal wins.

Removal is a breaking proto change (regenerated `.pb.go`/`.pulsar.go`) plus existing on-chain state, so it would need a migration/upgrade handler on a live network — a call I wouldn't fold into this PR.

Happy to go either way:
- if you decide to drop the fields, I'll close this PR (and can help confirm removal is safe given the read-site audit above);
- if they stay, this keeps the stored values honest in the meantime.

Your call on the direction.

redstartechno:
Closing this from our side, since the direction question above is the real blocker and it shouldn't sit in your queue waiting on us.

To restate the conclusion of the audit in the comment above: nothing in the repo reads `HardwareNode.host`/`port`. The chain only length-validates them, model assignment reads `Models`/`LocalId`, the broker routes to ML nodes from the operator's local config, and devshard doesn't reference on-chain hardware nodes at all. So this PR spends an extra `MsgSubmitHardwareDiff` write keeping a field accurate that nothing consumes — which is a fair argument against the PR, and for removing the fields instead.

Removal is your call: it's a breaking proto change plus existing on-chain state, so it needs a migration on a live network. If you decide to go that way, the read-site list above should save you the audit, and I'm happy to answer questions on it. If the fields are staying and you'd like the values kept honest in the meantime, feel free to reopen — the branch stays up.

No action needed either way.

PR #1343, opened 2026-06-13
Recover orphaned rotation escrows on DB persist failure instead of re-minting on chain

qdanik:
## Summary

- Auto escrow rotation created the on-chain escrow and only then persisted it; when the gateway DB write failed (e.g. `database is locked`), the escrow ID was lost, the rotator re-derived the needed count from the DB (which never saw it), and minted another escrow on chain every tick — an unbounded, fund-burning loop.
- Intent-first: before the chain tx, the gateway writes a durable commitment row to the DB — `model, role, epoch, block_height`, and the `tx_hash` precomputed from the signed bytes. If that write fails (after retries), no tx is broadcast: nothing is created, so nothing can be lost.
- Recovery via chain: after the escrow is created the result is persisted and the commitment cleared, in the same flow. If the persist fails (crash/lock after broadcast), the surviving commitment lets `reconcileCommitments` resolve the `tx_hash` to its on-chain `escrow_id` and persist it — run at startup and at the start of every rotation tick. The chain is the source of truth, queried by the precomputed hash, so losing an `escrow_id` is impossible.
- Root cause of the lock fixed: `NewGatewayStore` now opens SQLite with `journal_mode=WAL`, `synchronous=NORMAL`, `busy_timeout=5000`, and `SetMaxOpenConns(1)`, matching `perfstore.go` and `storage/sqlite.go` — the gateway store was the only one opened with defaults.
- All state lives in the DB (no sidecar file). DB writes retry with backoff; a per-`(model, role)` in-memory backoff avoids hammering a down DB/chain. No hard pause is needed — every failure is now either safe (no escrow created) or recoverable (commitment + reconcile).

| Failure point | Outcome |
| --- | --- |
| commitment write (before tx) | abort, no broadcast, no escrow — safe |
| chain broadcast | no escrow created — safe |
| persist (after tx) | commitment survives → reconcile recovers from chain |
| crash / restart anytime | startup reconcile recovers from chain |

## Why

The chain assigns the `escrow_id`, so the durable record cannot carry it before the tx. Recording the *intent* (with the precomputed `tx_hash`) before the irreversible broadcast is what makes a storage failure abort *before* any escrow is created, and a post-broadcast failure recoverable by querying the tx — the transactional-outbox pattern, reconciled against the chain rather than a local file. The cosmos tx hash equals `SHA-256` of the broadcast bytes and is computable before broadcast; a defensive check fails loudly if a node's returned hash ever differs.

State is kept in the DB rather than a sidecar journal because the gateway already cannot function without its DB, and a second store shares the same failure modes (disk full, fs errors) without adding reliability. One source of truth plus chain reconciliation also closes the broader untracked-orphan case and removes the only loss window a write-after-broadcast journal has.

`SetMaxOpenConns(1)` is safe for throughput: the hot request path never touches `gateway.db` — proxying, redundancy, and session picking use in-memory state, and perf/A-B data lives in a separate `perf.db` — so it serializes only low-frequency rotation/admin writes. Both rotation paths are covered: depletion-triggered replacement (`replaceDepletedEscrow`) calls the same `createRotationEscrow`.

Note: recovery is keyed by `tx_hash` in each gateway's own DB, so it is correct under independent gateways. A single chain key shared across gateways that create concurrently is a separate concern — cosmos account-sequence contention — and should use one key per gateway.

## Tests

- `TestCreateRotationEscrowIntentFirstThenPersistAndClear` — happy path: commitment written before the chain tx, escrow persisted, commitment cleared.
- `TestCreateRotationEscrowAbortsWhenCommitmentWriteFails` — a read-only/locked DB makes the intent write fail → no broadcast, no escrow.
- `TestCreateRotationEscrowPersistFailureRecoversViaCommitment` — persist fails after create → commitment kept → reconcile resolves the tx hash to the escrow and persists it.
- `TestReconcileCommitmentsRecoversFromChain` / `...ClearsWhenTxAbsent` / `...KeepsCommitmentOnChainError` — reconcile persists+clears when the tx is found, clears when the tx never landed, and retains the commitment on a transient chain error.
- `TestRESTChainTxClient_CreateDevshardEscrow*` — the precomputed `tx_hash` matches the node's, and `onPrepared` runs before broadcast; the mock node echoes `SHA-256(txBytes)` like a real node.
- `TestGatewayStoreUsesWALAndBusyTimeout` — the store opens in WAL with `busy_timeout=5000`.
- `go test ./cmd/devshardctl/...` green; `-race` clean on the commitment/chain suite.

## Test plan

- [x] `go test ./cmd/devshardctl/...` green; `-race` clean on the commitment/chain tests
- [x] Intent-first + chain-reconciliation proven by integration tests (real SQLite for the intent/persist paths; chain resolution stubbed at the `tx_hash → escrow_id` seam)

### DIFF
--- a/devshard/cmd/devshardctl/chain_tx_rest.go
+++ b/devshard/cmd/devshardctl/chain_tx_rest.go
@@ -122,6 +122,10 @@ func txHashFromBytes(txBytes []byte) string {
 	return strings.ToUpper(hex.EncodeToString(sum[:]))
 }
 
+// errTxNotFound marks a tx that is not on chain (404) — distinct from one that
+// committed but failed. The tx may still land until its unordered TTL elapses.
+var errTxNotFound = errors.New("tx not found on chain")
+
 // GetTxEscrowID resolves a create tx hash to its escrow_id in one lookup;
 // found=false when the tx is absent or failed (no escrow), error when unreachable.
 func (c *RESTChainTxClient) GetTxEscrowID(ctx context.Context, txHash string) (uint64, bool, error) {
@@ -131,7 +135,7 @@ func (c *RESTChainTxClient) GetTxEscrowID(ctx context.Context, txHash string) (u
 		err := c.getJSONFromBaseURL(ctx, baseURL, "/cosmos/tx/v1beta1/txs/"+url.PathEscape(txHash), &payload)
 		if err != nil {
 			if isNotFoundError(err) {
-				return 0, false, nil // tx never landed → no escrow created
+				return 0, false, errTxNotFound // not on chain (yet) — caller decides by age
 			}
 			lastErr = fmt.Errorf("%s: %w", baseURL, err)
 			continue
--- a/devshard/cmd/devshardctl/chain_tx_rest.go
+++ b/devshard/cmd/devshardctl/chain_tx_rest.go
@@ -130,12 +130,14 @@ var errTxNotFound = errors.New("tx not found on chain")
 // found=false when the tx is absent or failed (no escrow), error when unreachable.
 func (c *RESTChainTxClient) GetTxEscrowID(ctx context.Context, txHash string) (uint64, bool, error) {
 	var lastErr error
+	hasNotFoundError := false
 	for _, baseURL := range c.txQueryBaseURLs() {
 		var payload txResponseEnvelope
 		err := c.getJSONFromBaseURL(ctx, baseURL, "/cosmos/tx/v1beta1/txs/"+url.PathEscape(txHash), &payload)
 		if err != nil {
 			if isNotFoundError(err) {
-				return 0, false, errTxNotFound // not on chain (yet) — caller decides by age
+				hasNotFoundError = true // this endpoint lacks it; try the fallback before deciding
+				continue
 			}
 			lastErr = fmt.Errorf("%s: %w", baseURL, err)
 			continue
@@ -147,6 +149,10 @@ func (c *RESTChainTxClient) GetTxEscrowID(ctx context.Context, txHash string) (u
 			return escrowID, true, nil
 		}
 		lastErr = fmt.Errorf("tx %s committed via %s but escrow_id event was not found", txHash, baseURL)
+	}
+	// Only conclude "not on chain" when every reachable endpoint agreed on 404.
+	if lastErr == nil && hasNotFoundError {
+		return 0, false, errTxNotFound
 	}
 	return 0, false, lastErr
 }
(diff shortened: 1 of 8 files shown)


a-kuprin:
The PR **moves the failure point from DB to a sidecar journal file**. File writes fail for the same reasons DB writes do: disk full, backup/sync tools locking files, filesystem errors, etc. A hand-rolled append-only file queue is not more reliable than the DB — the gateway already depends on a DB backend and cannot function without it.

Journal append failure after an on-chain create is especially bad: the escrow exists on chain but is untracked, and the rotator may mint duplicates (in-memory breaker resets on reboot).

## Proposed design: recoverable, fault-tolerant

All state lives in the DB. Chain is queried on reboot to close the gap.

1. **Commit intent (DB)** — Before any chain tx, write a commitment row: model, role, epoch, and current block height. If this write fails, retry exponentially, stop and log — no escrow was created; gateway is non-functional without DB anyway.

2. **Create on chain** — Send the create tx; accept OK or error.

3. **Persist result (DB, same tx)** — On OK: upsert the escrow and clear the commitment in one DB transaction. Keep `escrow_id` in memory for the retry path.

4. **Retry policy** — Chain tx: retry a bounded number of times, then log. DB persist: exponential backoff with cap, log each failure.

5. **Reboot recovery** — Read pending commitments from DB. For each, query chain for escrows created since the committed block height. If found → persist and clear commitment. If not → resume from step 2.


This will make loosing escrow ids impossible, and we have other situation with orphaned escrows, that are not tracked by gateway.

Also it is better to add exponential retryes on every db write or chain write, but this can be in separate PR

devshard/cmd/devshardctl/chain_tx_rest.go
```diff
@@ -113,7 +115,43 @@ func NewRESTChainTxClient(cfg RESTChainTxConfig) (*RESTChainTxClient, error) {
...
+// found=false when the tx is absent or failed (no escrow), error when unreachable.
+func (c *RESTChainTxClient) GetTxEscrowID(ctx context.Context, txHash string) (uint64, bool, error) {
+	var lastErr error
+	for _, baseURL := range c.txQueryBaseURLs() {
+		var payload txResponseEnvelope
+		err := c.getJSONFromBaseURL(ctx, baseURL, "/cosmos/tx/v1beta1/txs/"+url.PathEscape(txHash), &payload)
+		if err != nil {
+			if isNotFoundError(err) {
+				return 0, false, errTxNotFound // not on chain (yet) — caller decides by age
+			}
```
x0152, line 141: GetTxEscrowID returns on the first endpoint's 404 without trying the fallback query url

qdanik: @x0152 done

x0152 (approved):

a-kuprin (approved):

PR #1348, opened 2026-06-18
feat(poc): serve inference during PoC validation phase

qdanik:
### Why

PoC v2 runs proof computation inside vLLM. During the **validation** stage a node only re-runs a sampled set of nonces, so most of the GPU is free and the node can answer inference requests while it validates. Until now the gateway treated the whole PoC sprint uniformly: only preserved nodes served inference during generation and validation, and the boosted per-weight concurrency cap applied for the entire sprint.

This PR uses the spare validation-phase capacity.

### What changed

**1. Concurrency cap scoped to generation (devshard)**
`PoCMaxConcurrentPer10000Weight` now applies only during PoC generation phases (`pocGenerationActive` / `rawPoCGenerationState`). Validation and steady-state use the base `MaxConcurrentPer10000Weight`.

**2. Per-node capability reporting (MLNode + dapi)**
- MLNode `GET /api/v1/state` now includes `poc_validation_inference`; missing field decodes as `false` for older nodes.
- MLNode `GET /api/v1/versions` remains available for direct capability inspection, backed by vLLM `/api/v1/pow/versions`.
- The broker reads the capability from the regular `NodeState` poll and stores it on `NodeState.PoCValidationInference` alongside `MlNodeVersion`.
- Public dapi `GET /v1/versions` now returns `mlnodes: [{node_id, version, poc_validation_inference}]`.

**3. Validation-phase routing (devshard)**
- A `VersionsCache` polls each miner's public `/v1/versions` and answers, per `(miner, node_id)`, whether that node is validation-capable. Unknown, stale, and error cases fail closed.
- During validation phases (`PoCValidate`, `PoCValidateWindDown`, `CONFIRMATION_POC_VALIDATION`), `mergePreservedWithValidationCapable` adds each non-preserved miner with the summed chain weight of only its validation-capable nodes, per model. Generation and non-PoC behavior is unchanged.
- Node identity: `/v1/versions` `node_id` is the miner broker `Node.Id`, matching chain `MLNodeInfo.NodeId`, so the gateway can match per-node capability to per-node chain weight.

### API change

`GET /v1/versions` gains a `mlnodes` array:

```json
{
  "api_version":  { ... },
  "node_version": { ... },
  "mlnodes": [
    { "node_id": "n1", "version": "v3.0.8", "poc_validation_inference": true }
  ]
}
```

MLNode `GET /api/v1/state` gains `poc_validation_inference`:

```json
{
  "state": "INFERENCE",
  "version": "0.2.0",
  "poc_status": "IDLE",
  "inference_healthy": true,
  "loaded_model": "model-id",
  "poc_validation_inference": true
}
```

### vLLM change

PR: https://github.com/gonka-ai/vllm/pull/49


decentralized-api/broker/broker.go
```diff
@@ -285,6 +287,14 @@ func (s *NodeState) ShouldContinueInference() bool {
 	return len(s.PreservedModels) > 0
 }
 
+// ServesInferenceDuringValidation reports whether a node can take inference while
+// running PoC validation: capable nodes run PoC v2 inside vLLM, so it stays up.
+func (s *NodeState) ServesInferenceDuringValidation() bool {
+	return s.PoCValidationInference &&
+		s.IntendedStatus == types.HardwareNodeStatus_POC &&
+		s.PocIntendedStatus == PocStatusValidating
```
DimaOrekhovPS, line 295: This function will return `false` for a preserved node that is in `Inference` state during poc validation phase. Are we ok with it?

Judging from the usage it's ok, but maybe we should make this function private so only `nodeAvailable` function can use it

decentralized-api/internal/server/public/app_info_handlers.go
```diff
@@ -21,6 +21,21 @@ func (s *Server) getVersions(ctx echo.Context) error {
 	}
 	nodeVersion := resp.ApplicationVersion
 
+	mlnodes := make([]map[string]any, 0)
+	if s.nodeBroker != nil {
+		if nodes, nerr := s.nodeBroker.GetNodes(); nerr != nil {
```
DimaOrekhovPS, line 99: Each call to `GetNodes` enqueues a message into broker's command queue. Feels dangerous to expose this endpoint as public. Maybe let's add a very simple cache for this endpoint? So getNodes is called only once a minute or so?

DimaOrekhovPS (approved):

PR #1349, opened 2026-06-18
fix(devshard): retire runtime escrow

qdanik:
## Summary

Free a deactivated escrow's in-memory runtime once its in-flight requests drain. When a devshard escrow was deactivated (escrow missing on chain, balance exhausted, nonce depletion, or rotation), its `devshardRuntime` was left in the gateway registry **forever** — holding the `user.Session` and the per-runtime SQLite handle open. Over many rotations this leaks memory and file descriptors. This PR retires the runtime on every automatic deactivation path, and gates that retirement on the escrow draining to **zero** active requests, so it never closes a store out from under a live request.

- **Retire on deactivation** — `retireRuntime` drops the runtime from `g.runtimes` / `g.runtimeOrder`, removes it from capacity, and closes its session/SQLite store. Wired into every automatic deactivation path.
- **Drain-gated** — if requests are still in flight, retirement is marked `retirePending`; the last request to finish fires it from the `releaseRuntime` drain hook. Mirrors the settlement drain mechanism from #1314.
- **Settlement takes priority** — when an escrow is both settlement-pending and retire-pending, the drain edge routes to settlement (which retires the runtime itself), never a bare retire that would close the store mid-settle.
- **I/O off the lock** — the SQLite `close()` runs outside `g.mu`.

Scope: `devshard/cmd/devshardctl` (+ one `devshard/user` test). No chain, api, or ml-node changes.

---

### 1. Deactivated escrows leaked their runtime

**Before.** Deactivation flipped `active=false` and stopped traffic, but the `devshardRuntime` stayed in the registry, keeping its `user.Session` and per-runtime SQLite handle open indefinitely. Rotation churns escrows, so handles and memory accumulated.

**After.** `retireRuntime` removes the runtime from the registry and capacity and closes its session, and is called on every automatic deactivation path: escrow-missing-on-chain, balance-exhausted (rotation disabled), rotated-out (settle and no-settle), depleted-escrow replacement (no-settle), and auto-settlement (both on success and after retries are exhausted, so a permanently-unsettleable escrow still can't leak its store — on-disk state is preserved for manual recovery).

### 2. Retiring could close a store out from under a live request

**Before.** A naive retire would `close()` the SQLite store even while requests were still reading/writing it.

**After.** `retireRuntimeLocked` returns early and records `retirePending` when `activeRequests > 0`. The drain hook in `releaseRuntime` fires the deferred retire at exactly `remaining == 0`. Because `active=false` blocks new reservations, the count only drains downward, so that is the precise "last request finished" edge. The hook is mutually exclusive with settlement: a settlement-pending escrow routes to `scheduleAutoSettlement` (which retires the runtime itself once settled), so retirement never preempts an in-flight settlement.

### 3. The close held the gateway lock

**Before.** `rt.close()` ran inside `g.mu`, so disk I/O blocked every other gateway operation contending for the lock.

**After.** `retireRuntimeLocked` performs the registry mutations under `g.mu` and returns the runtime; `retireRuntime` releases the lock and only then calls `rt.close()`.

---

## Concurrency

The drain hook in `releaseRuntime` is lock-free. `retireReason` is written before `retirePending` (under `g.mu`) and read after the atomic `retirePending.Load()` in the hook — the atomic Store→Load pair supplies the happens-before, identical to the `settlementReason`/`settlementPending` pair. `retirePending` is **not** persisted (unlike `settlementPending`): retirement only frees a live in-memory handle, and a restart reopens handles fresh, so there is nothing to reconcile. Verified clean under `go test -race`.

---

## Test plan

`go test ./cmd/devshardctl/... ./user/...` (and `-race`) — all green:

- [x] `TestRetireRuntimeRemovesRuntimeFromRegistry` — retire drops the runtime from `g.runtimes` and `g.runtimeOrder`; idempotent on a missing id.
- [x] `TestRetireRuntimeDefersWhileRequestsInFlight` — retire defers (and records intent) while busy, succeeds once the count is zero.
- [x] `TestReleaseRuntimeRetiresAfterDrain` — a deferred retire fires once the last request drains through `releaseRuntime`.
- [x] `TestReleaseRuntimeRetiresWithOnlyRetirePending` — retire branch in isolation: only `retirePending` set, drain → retire.
- [x] `TestReleaseRuntimeDefersWhileRequestsRemain` — `remaining != 0` fires nothing; the drain edge only acts at zero.
- [x] `TestRetireRotatedDevshardRetiresWithoutSettlement` / `...RetiresAfterSettlement` — rotation retires in both the no-settle and settle terminal paths.
- [x] `TestSession_Close_ClosesUnderlyingStore` — proves `Session.Close` releases the SQLite store the leak was failing to free.
- [x] `go test -race` clean — confirms the lock-free drain hook's happens-before.

### DIFF
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -3278,6 +3278,36 @@ func removeRuntime(runtimes []*devshardRuntime, id string) []*devshardRuntime {
 	return out
 }
 
+// retireRuntime drops a runtime from the in-memory registry and closes it,
+// releasing its user. Session and the per-runtime SQLite handles that session owns.
+func (g *Gateway) retireRuntime(id, reason string) bool {
+	g.mu.Lock()
+	defer g.mu.Unlock()
+	return g.retireRuntimeLocked(id, reason)
+}
+
+// retireRuntimeLocked is retireRuntime's body; callers must already hold g.mu.
+func (g *Gateway) retireRuntimeLocked(id, reason string) bool {
+	rt, ok := g.runtimes[id]
+	if !ok {
+		return false
+	}
+	if inFlight := rt.activeRequests.Load(); inFlight > 0 {
+		log.Printf("runtime_retire_deferred escrow=%s reason=%q active_requests=%d", id, reason, inFlight)
+		return false
+	}
+	delete(g.runtimes, id)
+	g.runtimeOrder = removeRuntime(g.runtimeOrder, id)
+	if g.capacity != nil {
+		g.capacity.RemoveEscrow(id)
+	}
+	if err := rt.close(); err != nil {
+		log.Printf("runtime_retire_close_error escrow=%s reason=%q error=%v", id, reason, err)
+	}
+	log.Printf("runtime_retired escrow=%s reason=%q", id, reason)
+	return true
+}
+
 func (g *Gateway) sortRuntimeOrderLocked() {
 	slices.SortFunc(g.runtimeOrder, func(a, b *devshardRuntime) int {
 		return strings.Compare(a.id, b.id)
@@ -3301,13 +3331,17 @@ func (g *Gateway) attachEscrowChecker(rt *devshardRuntime) {
 	if g.escrowChecker != nil {
 		rt.proxy.redundancy.onEscrowMissing = func() {
 			go g.escrowChecker.TriggerCheck(escrowID, func() {
-				g.deactivateDevshardByID(escrowID)
+				if g.deactivateDevshardByID(escrowID) {
+					// Escrow no longer exists on chain -- nothing to settle.
+					g.retireRuntime(escrowID, "escrow confirmed missing on chain")
+				}
 			})
 		}
 	}
 	rt.proxy.redundancy.onBalanceExhausted = func() {
 		if !g.escrowRotationEnabled() {
 			g.deactivateDevshardByIDWithReason(escrowID, "escrow balance exhausted")
+			g.retireRuntime(escrowID, "escrow balance exhausted")
 			return
 		}
 		log.Printf("gateway_replacing_exhausted_escrow escrow=%s", escrowID)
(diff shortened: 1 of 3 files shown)


devshard/cmd/devshardctl/gateway.go
```diff
@@ -3258,6 +3258,36 @@ func removeRuntime(runtimes []*devshardRuntime, id string) []*devshardRuntime {
...
+}
+
+// retireRuntimeLocked is retireRuntime's body; callers must already hold g.mu.
+func (g *Gateway) retireRuntimeLocked(id, reason string) bool {
+	rt, ok := g.runtimes[id]
+	if !ok {
+		return false
+	}
+	if inFlight := rt.activeRequests.Load(); inFlight > 0 {
+		log.Printf("runtime_retire_deferred escrow=%s reason=%q active_requests=%d", id, reason, inFlight)
```
DimaOrekhovPS, line 3327: "deferred" implies that it will be tried later, but there's no guarantees for it, right?

qdanik: escrow cannot be settled while we have active request, if this request will be finished during settlement retires it will retire. 

But we already have a PR to settle escrow when the active requests are released https://github.com/gonka-ai/gonka/pull/1314, it might be part of this PR

devshard/cmd/devshardctl/gateway.go
```diff
@@ -3258,6 +3258,36 @@ func removeRuntime(runtimes []*devshardRuntime, id string) []*devshardRuntime {
...
+		return false
+	}
+	delete(g.runtimes, id)
+	g.runtimeOrder = removeRuntime(g.runtimeOrder, id)
+	if g.capacity != nil {
+		g.capacity.RemoveEscrow(id)
+	}
+	if err := rt.close(); err != nil {
+		log.Printf("runtime_retire_close_error escrow=%s reason=%q error=%v", id, reason, err)
+	}
```
a-kuprin: It is better to call I/O operations (`rt.close`) outside the lock

devshard/cmd/devshardctl/gateway.go
a-kuprin: I'm not sure reture and settle are the same. It can be retired but not settled.
And when it is retired and we have unfinished `activeRequests` it will never be deleted from memory.

You right to link https://github.com/gonka-ai/gonka/pull/1314 , but it seams we should first merge them together and also add runtime retiring at `releaseRuntime`

a-kuprin (approved):

PR #1350, opened 2026-06-18
feat: Trainshards v0 (stage 1)

### LINKED ISSUE #1219: [P0] Basic primitives for training
The framing is simple: you can prepare any container, it will run across different GPUs in the network, a protocol layer will coordinate interaction between nodes, they will perform training, and then use a protocol-level voting mechanism to determine which participants behaved correctly.

We intentionally ignore the automatic redeployment problem for now. 
This should be a small-scoped task on mainnet.
Put together a lightweight training flow without the heavy logic we have in devshards

x0152:
### Stage 1: the on-chain reserve and release cycle
This stage adds the on-chain part that lets a researcher borrow GPUs for training and give them back. The flow: a host opts a node in, the researcher opens a run with **assemble_trainshard**, the chain reserves the matching nodes and freezes their weight, those nodes step out of inference and PoC, and **settle_trainshard** or a timeout puts them back to normal work.

1. Running an experiment needs a governance proposal first (A passed vote is permission, not a guarantee)
2. Add four transactions: **assemble_trainshard**, **settle_trainshard**, **autokick_trainshard_node**, **refresh_training_node_opt_in**
3. The chain picks the right nodes for the request, only from nodes that are allowed for training and match the requested gpu profile
4. The picked nodes get reserved, their weight is frozen, with a timeout
5. While a node is reserved, it does no inference and no PoC
6. Being reserved never causes penalties: no downtime hits, no claim issues, no cPoC ratio drop, no status changes.
7. Routing, admission and rewards count only the free (non-reserved) nodes
8. There are caps on how much training can hold (overall, per model, per gpu type), checked at assemble with a safety buffer
9. On **settle_trainshard** or timeout, remaining nodes are released after a short return buffer; failed nodes can be autokicked earlier

**Result**: the whole reserve -> run -> release cycle works on-chain, and the rest of the network is not affected

**Details**: https://docs.google.com/document/d/1LLZngQ7VoIL3DVT8St40XLE8HcRcxyNXZueyzoQfWuE/edit?usp=sharing


Ryanchen911:
Reviewed at `6ef021c8a` against base `6e92e9089`, Five things below. Only the first is something I think needs a decision before training is enabled; the rest are smaller.

### 1. Lending a machine zeroes its PoC income, and there is no on-chain compensation in v0

I want to make sure I read this correctly, because it is the part that matters most to operators deciding whether to opt in.

`applyTrainingReservationAdjustment` strips the reserved node's weight from the host's numerator, and the denominator stays `totalFullWeight`. That denominator choice is pre-existing (`bitcoin_rewards.go:990`, "prevent redistribution of unclaimed shares"), so the other miners' shares are unchanged — the reserved node's share simply goes to governance rather than being redistributed. Your test `pre-formation reserved node adds frozen weight to denominator, numerator untouched` pins exactly this, so I read it as deliberate, and the reasoning makes sense: the researcher borrowed the compute, the lender should not also collect PoC rewards on it, and the rest of the network should not get a windfall either.

The gap is that `MsgCreateTrainshardProposal` has no payment field, and neither does `Trainshard`. So for an operator, opting in currently means: this machine earns nothing for the epoch it is lent, and any compensation has to be arranged entirely off-chain.

That may well be intended for v0, and if so no code change is needed.

### 2. The new zero-weight accept path in `pocValidated` is guarded from a distance

`chainvalidation.go:284` now returns true unconditionally when `TotalNetworkWeight <= 0`:

```go
if wc.TotalNetworkWeight <= 0 {
    // no enforceable weight left in scope: accept rather than slash or divide by zero
    return true
}
```

Reservations reduce that total via `FreeShareOfWeight` in `getEffectiveValidationBaseState`, so this branch is newly reachable in principle, and if it ever fires PoC validation becomes a no-op for that model.

I traced what actually keeps it unreachable, and it is not the bps caps — it is the per-model guard in `selectTrainshardNodes`:

```go
if view.modelCapacity[e.ModelId]-(reserved.perModel[e.ModelId]+takenModel[e.ModelId]+1) < 1 {
    fits = false
}
```

Worth noting the per-profile condition right next to it is `< 0`, and does allow a profile to be filled completely. I confirmed the model guard holds with a local probe: both a 1-of-1 and a 2-of-2 shard are refused with `insufficient free training capacity` even at `max_reserved_share_per_model_bps = 10000`. So there is always at least one free node per model, and the total cannot reach zero. The current behaviour is correct.

My concern is the coupling rather than the code. These two sites are far apart, and someone loosening that model guard later — for the small-network usability reason in point 5 — would silently open the PoC accept path with no test failing. A cross-reference comment in both places would be enough to prevent that.

### 3. `autokick` accepts a host kicking its own node, with cooldown defaulting to 0

`msg.Creator != msg.Participant` in the authorization check means a host can `autokick` its own node. That path is clearly needed for a host to report its own hardware failure, so I am not suggesting removing it.

The side effect is a sequence like: opt in → get reserved → immediately self-kick → sit in the release buffer, where routing, downtime detection, and cPoC are all still shielding the node while it no longer owes the run any work. With `creator_cooldown_blocks` defaulting to 0 there is no friction on repeating it. Rewards already zero the reserved node so no emission is gained, but for a machine that was going offline for maintenance anyway, this is a free "does not count as downtime" window.

Would it make sense for a host-initiated `autokick` to shorten the shield rather than grant the full buffer? Or a non-zero default cooldown? Not urgent, since nothing is stolen — just noting the incentive exists.

### 4. The `Seed == nil` drop changes pre-existing non-training behaviour

The new branch in `ComputeNewWeights`:

```go
if preservedParticipant.Seed == nil {
    am.LogWarn("ComputeNewWeights: Dropping preserved-only participant without seed", ...)
    continue
}
```

A preserved-only participant that submitted no seed previously entered `ActiveParticipants` and is now dropped. The end state converges, since `onSetNewValidatorsStage` already skipped nil-seed participants when adding epoch members — but the participant now disappears from `ActiveParticipants` entirely, which affects anything else reading that list.

This looks intentional and probably correct, it is just unrelated to training. Could it get its own line in the PR description? It would be easy for someone later to assume it came with the reservation work.

### 5. Two smaller notes

- **Prune invariant has a single enforcement point.** `SettledShardRetentionBlocks >= 2*EpochLength + ReleaseBufferBlocks` is validated only when `training_enabled` is true, and only via `MsgUpdateParams.Validate()` — `SetParams` itself does not validate. I probed it with retention below that floor: after `pruneClosedTrainshards` removes the shard, the reserved weight vanishes from the epoch view, which silently restores full rewards and drops the penalty shield.

  To be clear, this is not reachable on a real chain — mainnet `epoch_length = 17280` puts the floor at 34580, and the upgrade handler backfills it. I only reached it by setting params directly in a test. But for an invariant that maps this directly to money, would you consider asserting it in `pruneClosedTrainshards` as well, refusing to prune a shard whose close height is still within two epoch lengths? Defense in depth against a future code path that sets params without validating.

- **Small-network capacity is more restrictive than the caps suggest.** Because of the per-model "leave one free" guard, the effective maximum per model is `min(cap, capacity - 1)`. On a model with only two or three nodes, training is effectively unavailable regardless of how the bps caps are set. That is the right invariant to have, but a researcher hitting `insufficient free training capacity` on a small model will not guess why raising the cap did not help. Worth a line in the docs.

Nice work on this overall !

vbgd0:
Reviewed the non-generated diff (~4.6k lines: keeper/trainshard*, msg servers, bitcoin_rewards, chainvalidation, confirmation_poc, collateral hooks, params, upgrade handler). Overall the design is careful — deterministic candidate selection via hashed sort keys, frozen `reserved_until_height` so later param changes never rewrite history, per-block quotas with defer-instead-of-starve in the EndBlocker, and fail-closed store lookups. CI is green. Findings below, roughly by severity.

### 1. Fully-reserved host can be silently dropped from the next epoch (chainvalidation.go)

`mergeReservedNodesIntoPreserved` creates preserved-only participants with `Seed: nil`, and the new guard in `ComputeNewWeights` then drops any preserved-only participant whose seed is still nil after the `GetRandomSeed` loop:

```go
if preservedParticipant.Seed == nil {
    am.LogWarn("ComputeNewWeights: Dropping preserved-only participant without seed", ...)
    continue
}
```

A host whose nodes are *all* reserved does no PoC and produces no store commit, so it exists only as a preserved-only entry. If it has no random seed for the upcoming epoch, it is removed from the active set entirely — frozen weight, routing, everything — which contradicts invariant 6 ("being reserved never causes penalties"). Two questions:

- Is seed submission guaranteed for a fully-reserved host (i.e. submitted by the dAPI independent of PoC participation)? If it's tied to PoC activity, this drop will fire in practice.
- The guard also changes behavior for *ordinary* preserved-only participants that previously were added without a seed. What case motivated it — was it fixing a concrete crash downstream? If so, worth a comment; if not, consider keeping the participant and only excluding it from seed-dependent sampling.

### 2. Devshard settle shield is much broader than the other penalty paths (msg_server_settle_devshard_escrow.go)

The shield set is `CollectEpochReservedHostsForModel(...)` — any host with **any** node reserved for the model at **any** point in the epoch. A host serving a model with 8 nodes that had one node reserved for a few blocks skips miss/invalid aggregation for the whole epoch. You already wrote the precise helper for this — `CollectEpochFullyReservedHostsForModel` (interval-checkpoint coverage of all model nodes) — but it is dead code: nothing references it. Looks like the settle path was meant to use it. Either wire it in or delete it and comment why the broad shield is intended.

### 3. `epochBlockRange` treats an unknown epoch as (0, MaxInt64) (trainshard.go)

Every reservation in the store then "overlaps" that epoch. Combined with the unguarded `upcomingEpoch.Index-1` in `getInferenceServingNodeIds` (the merge call is guarded by `Index > 1`, this one is not), epoch 0/1 underflows to MaxUint64 → epoch not found → all reservations included. Harmless today (no shards can exist at genesis) but it's a latent trap for every future caller that passes a stale or future epoch index. Suggest returning a `found bool` and skipping the walk.

### 4. Reward forfeiture is epoch-granular — confirm it's intended (bitcoin_rewards.go)

`SettleAccounts` collects `ReservationScopeReward` **without** a height filter, so a node reserved for even one block of an epoch forfeits its entire epoch reward (its share going undistributed → governance). The shield paths (`claim_rewards`, `confirmation_poc`, effective-validation) are all height-aware; only the reward path is all-or-nothing per epoch. If "you lose the epoch you enter training in" is the intended economics, a comment saying so would prevent someone later "fixing" the asymmetry — if not, `TotalsAt`-style prorating is already available.

### 5. Autokick dedup can swallow real kicks (msg_server_autokick_trainshard_node.go)

The idempotency key is `(trainshard_id, request_id)` only. Two kicks that reuse a request_id for *different* nodes: the second silently no-ops and returns success, leaving the node reserved while the coordinator believes it was kicked (stage 2 then wipes and reprovisions on that assumption). Either include participant/node in the dedup key, or store the (participant, node) with the request and error on mismatch.

### 6. Smaller points

- `TrainingParams.Validate` doesn't reject `MaxActiveShards == 0`, `MaxActiveShardsPerCreator == 0`, or `MaxTotalReservedNodes == 0` while `TrainingEnabled` — training would be "enabled" yet every assemble fails with a confusing active-limit/capacity error. Cheap to validate alongside `MaxNodesPerShard`.
- The new collections (Trainshards, reservations, opt-ins, counters, cooldowns) are absent from genesis import/export. I see ClaimRecipients etc. share that gap, so this may be a known project-wide stance — but an exported-state restart mid-run drops active reservations and shard records silently. Worth at least a tracking issue.
- Self-kick (`msg.Creator == msg.Participant`) lets a host pull its node out of a run at any time with no on-chain accounting beyond the AUTOKICKED status/reason. Fine for v0, but if host reliability is ever going to matter for selection, the data to distinguish "host bailed" from "creator kicked" is only in the free-text reason.
- The upgrade handler backfills node statuses / clears zero-TTL opt-ins for state that can only exist if a pre-release of this feature ran on the target chain. If that's for an internal testnet, fine; on a chain that never had the feature these are no-op walks, just noting they're not dead weight by accident.

Static review; I did not run the chain test suite locally (CI covers it and is green).

x0152:
Thanks @Ryanchen911 and @vbgd0!

What I keep as is in v0 by design:
- Economics is not part of v0 yet. No payments, no compensation on chain
- Reward loss works per epoch, and the shield too. It is symmetric: no income for the epoch, no duty for the epoch
- Self-kick stays. A host needs a way to report a broken gpu, for example. Cooldown and shield length are governance params
- Seeds come from the dapi, not from poc. So a fully reserved host still has a seed and stays active. The Seed == nil check is not related to training, I will add it as a separate line in the pr description
- epochBlockRange keeps the open range for an unknown epoch. It cannot happen today, and without epoch data we prefer to shield a host, not punish it
- Genesis export/import is missing for many collections, not only for this one. Better to fix it in a separate issue

vbgd0:
Second pass at `330a55fd5` (+ `b108da23d`). Everything from the first round is either fixed or answered — thanks for the fast turnaround. Verified in the diff:

- Autokick dedup now stores the node under `(shard, request_id)` and refuses a reused id aimed at another node (`ErrTrainshardAutokickRequestReused`) ✅
- Zero caps rejected in `TrainingParams.Validate` and backfilled in the upgrade handler ✅
- `settledShardRetention` floor in the EndBlocker, dead `CollectEpochFullyReservedHostsForModel` removed, epoch-granular forfeiture/shield documented at both sites ✅
- Seeds come from the dAPI, so the `Seed == nil` guard can't hit a fully-reserved host — fine, and the note in the PR description will save the next reader the same detour.

One new item, introduced by `b108da23d`:

### Existing authz-mode hosts get no grant for the two new operation-key messages

`MsgRefreshTrainingNodeOptIn` and `MsgAutokickTrainshardNode` were added to `InferenceOperationKeyPerms`, which is what `custom_gentx` / `grant-ml-ops-permissions` grant cold→warm at onboarding. Hosts already on mainnet run in authz mode (the dAPI wraps in `MsgExec`, `tx_manager.go:833`) and hold grants only for the list as it was when they joined — so on every existing host the trainshard daemon's opt-in via `/admin/v1/tx/send` fails with an authz "not found" and the node never becomes eligible. This is exactly the v0.2.12 → v0.2.13 story for `MsgRespondDealerComplaints`; `grantRespondDealerComplaintsAuthz` in `app/upgrades/v0_2_13/upgrades.go` is the template (identify pairs by the `MsgStartInference` grant, reuse its expiration). The v0.2.16 handler should backfill both message types the same way, otherwise stage 2 only works for hosts that re-run the grant command by hand. (#1618's run guide step 3, "check the opt-in landed", is where operators would discover it.)

Small optional: the `handled == ""` legacy branch in `AutokickTrainshardNode` exists for KeySet-format rows that can only come from a pre-release build on a testnet. If no such chain is carried forward, it can go; if one is, worth a one-line comment naming it so nobody deletes the branch under it later.

Static review; CI green.

inference-chain/proto/inference/inference/trainshard.proto
```diff
@@ -0,0 +1,92 @@
...
+  string participant = 1;
+  string node_id = 2;
+  string model_id = 3;
+  int64 poc_weight = 4;
+  TrainshardNodeStatus status = 5;
+  int64 released_at_height = 6;
+  string release_reason = 7;
+  // reserved_until_height freezes the release buffer end at release time so
+  // epoch-level reads never depend on the current release_buffer_blocks value
+  int64 reserved_until_height = 8;
```
DimaOrekhovPS, line 46: what if training requires more time?

and why is this height different for each node and not smth global for each node of a trainshard?
is it because new nodes can join and old can drop during the training process?

inference-chain/x/inference/keeper/keeper.go
```diff
@@ -135,14 +135,20 @@ type (
...
+		TrainshardActiveIndex            collections.KeySet[uint64]
+		TrainshardExpiryIndex            collections.KeySet[collections.Pair[int64, uint64]]
+		TrainshardClosedIndex            collections.KeySet[collections.Pair[int64, uint64]]
+		TrainshardReservations           collections.Map[collections.Pair[string, string], uint64]
+		TrainshardReleaseIndex           collections.KeySet[collections.Triple[int64, string, string]]
+		TrainshardAutokickRequest        collections.Map[collections.Pair[uint64, string], string]
+		TrainingNodeOptIns               collections.Map[collections.Pair[string, string], int64]
+		TrainshardProposals              collections.Map[uint64, types.TrainshardProposal]
+		TrainshardProposalCounter        collections.Item[uint64]
+		TrainshardCreatorCooldown        collections.Map[string, int64]
```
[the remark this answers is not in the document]
DimaOrekhovPS: not critical, chain restart from a snapshot is not a routine event

inference-chain/proto/inference/inference/tx.proto
```diff
@@ -486,3 +493,67 @@ message MsgDeleteDevshardApprovedVersion {
...
+
+message MsgSettleTrainshard {
+  option (cosmos.msg.v1.signer) = "creator";
+  string creator = 1;
+  uint64 trainshard_id = 2;
+}
+
+message MsgSettleTrainshardResponse {}
+
+message MsgAutokickTrainshardNode {
```
DimaOrekhovPS, line 530: In my opinion the "Auto" is confusing. As I understood a researcher ordering the training can use this type of msg to stop a node out from participating in the trainshard if smth is wrong with that node. I don't see anything auto in the process, why no just MsgKickTainshardNode?

DimaOrekhovPS (approved):

DimaOrekhovPS (commented):
Targeting for 0.2.17

PR #1351, opened 2026-06-18
test: centralize bech32 test config to prevent sdk config races

x0152:
## Problem
Keeper tests were flaky with fatal error: concurrent map read and map write, failing CI intermittently (~1 in 3-5 runs).

## Why
Parallel tests (t.Parallel()) were creating keepers while repeatedly mutating global sdk.Config bech32 prefixes.
One test could write the config map while another was reading it during address decoding, which caused a Go runtime crash.

## Fix
Use a shared testutil.EnsureBech32Config() helper guarded by sync.Once, and route test setup through this helper instead of direct sdk.GetConfig().SetBech32Prefix* writes.
This change is test-only (production config is still initialized once at startup).

## Stack trace

```bash
fatal error: concurrent map read and map write

goroutine 2226 [running]:
internal/runtime/maps.fatal(...)
	runtime/panic.go:1046 +0x20
github.com/cosmos/cosmos-sdk/types.(*Config).GetBech32AccountAddrPrefix(...)
	gonka-ai/cosmos-sdk@v0.53.3-ps19/types/config.go:162
github.com/cosmos/cosmos-sdk/types.AccAddressFromBech32(...)
	gonka-ai/cosmos-sdk@v0.53.3-ps19/types/address.go:199 +0x48
github.com/productscience/inference/x/inference/keeper.Keeper.SetMaintenanceState(...)
	inference-chain/x/inference/keeper/maintenance.go:54 +0x5c
github.com/productscience/inference/x/inference/keeper.msgServer.ScheduleMaintenance(...)
	inference-chain/x/inference/keeper/msg_server_schedule_maintenance.go:116 +0x89c
github.com/productscience/inference/x/inference/keeper_test.TestLifecycle_NoTransitionsAtWrongHeight(...)
	inference-chain/x/inference/keeper/maintenance_test.go:503 +0x2c8
testing.tRunner(...)
	testing/testing.go:1934 +0xc8
created by testing.(*T).Run in goroutine 1
	testing/testing.go:1997 +0x364
```

### DIFF
--- a/inference-chain/app/tally_integration_test.go
+++ b/inference-chain/app/tally_integration_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/productscience/inference/app"
+	"github.com/productscience/inference/testutil"
 	inferencetypes "github.com/productscience/inference/x/inference/types"
 )
 
@@ -192,11 +193,7 @@ func TestTallyBugReproduction(t *testing.T) {
 func createTestApp(t *testing.T) *app.App {
 	t.Helper()
 
-	// Configure SDK for gonka addresses
-	config := sdk.GetConfig()
-	config.SetBech32PrefixForAccount("gonka", "gonkapub")
-	config.SetBech32PrefixForValidator("gonkavaloper", "gonkavaloperpub")
-	config.SetBech32PrefixForConsensusNode("gonkavalcons", "gonkavalconspub")
+	testutil.EnsureBech32Config()
 
 	db := dbm.NewMemDB()
 	logger := log.NewNopLogger()
--- a/inference-chain/app/upgrades/v0_2_8/upgrades_test.go
+++ b/inference-chain/app/upgrades/v0_2_8/upgrades_test.go
@@ -152,7 +152,7 @@ type testMocks struct {
 }
 
 func setupTestKeeper(t *testing.T, ctrl *gomock.Controller) (keeper.Keeper, sdk.Context, testMocks) {
-	sdk.GetConfig().SetBech32PrefixForAccount("gonka", "gonka")
+	keepertest.EnsureBech32Config()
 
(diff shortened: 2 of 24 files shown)


Ryanchen911:
A couple of small things I wanted to run by you, no strong opinions either way:

1. Might the race still be reachable from the other writers?
The sync.Once guards InferenceKeeperWithMock, but there are still ~40 unconditional sdk.GetConfig().SetBech32PrefixForAccount("gonka", "gonka") calls scattered across sibling test files in the same x/inference/keeper package (e.g. msg_server_test.go:26, devshard_settlement_test.go, msg_server_settle_devshard_escrow_test.go).

Today none of those call t.Parallel(), so they run in a separate phase from the guarded parallel test and don't race — so this PR is definitely correct for the crash we're seeing. My only worry is that the moment someone adds t.Parallel() to one of those tests (or a new parallel test hits an unguarded writer), the same concurrent map read and map write could come back. Would it be worth routing all of these through one shared guarded helper (or a TestMain) as a follow-up, so we don't have to remember this landmine later? Totally fine to keep it out of this PR — just flagging for a possible follow-up.

2. Minor heads-up on the pubkey prefix
The value set here is "gonka"/"gonka", whereas genesistransfer.go:85 and restrictions.go:68 set "gonka"/"gonkapub". With sync.Once, whichever helper the binary reaches first now "wins" the pubkey prefix permanently, instead of the previous last-writer-wins behavior. The crash path (account decoding) is unaffected since the account prefix is "gonka" everywhere, so this is almost certainly a non-issue — just flagging in case any test asserts on the validator/pubkey bech32 prefix.

Thanks for chasing down a genuinely annoying flake!

x0152:
@Ryanchen911  Agree, thanks!
I've applied this broadly and replaced all places where Bech32 config was being set directly

Ryanchen911:
Looks good, thanks for you work, I think this can be approved definitely!

PR #1355, opened 2026-06-22
fix(devshard): wire StalenessChecker into production hosts to stop non-payment

ouicate:
## Summary
A devshard session makes the user the sole sequencer, so the executor host's only economic leverage is **withholding its state-root signature** when the user censors the executor's `MsgFinishInference` (the tx that credits the host for delivered work). That defense — the `StalenessChecker`, reachable only via `host.WithGrace` — was never wired into the production host. `HostManager` built every host with `checker == nil` and `hostOptions` omitted `WithGrace`, so the host signed **every** state root, including the final settlement root, even when its delivered-work finish was censored. This PR wires `host.WithGrace` into `hostOptions`, sourced from the escrow-frozen `InferenceSealGraceNonces`, so the executor withholds once a censored finish goes stale.

## Problem
Both production constructors passed `nil` for the checker and `hostOptions` never added `WithGrace`:

- `host.NewHost(sm, m.signer, m.engine, escrowID, group, nil, m.hostOptions(...)...)` in both `create` and `recoverStoredSession` ⇒ `h.checker == nil`.
- `signIfAccepted` only withholds when `h.checker != nil`, so the host signed unconditionally.
- `Cost` is credited **only** by `MsgFinishInference` (`applyFinishInference`, which requires status `Started`). Censoring it leaves `HostStats[slot].Cost == 0`, and `Missed` is never incremented (only an explicit timeout does that), so the victim isn't even flagged.
- At settlement, the Finalizing→Settlement drain folds non-terminal inferences with no status check, on-chain payout derives from `HostStats.Cost`, the remainder (`escrow.Amount − totalPayout`) refunds to the creator, and `VerifyDevshardSettlement` has no `Cost > 0` lower bound.

Net effect: a malicious allowlisted user takes the streamed inference output, never sequences the finish, and settles with `Cost == 0` — the host computes for free and the user is refunded the reserved compute cost (paying only per-nonce fees). Confirmed by an empirically failing host test before this change.

## Solution
`hostOptions` now takes the escrow-frozen grace and installs `host.WithGrace(InferenceSealGraceNonces)`. The grace is the governance-controlled, escrow-frozen seal grace (normalized to a floor of 20), which sits well above the honest sequencer's one-nonce inclusion latency, so honest-but-slow users do not trigger spurious withholding while a censored finish denies the user the 2/3+ settlement quorum.

### DIFF
--- a/decentralized-api/internal/devshard/manager.go
+++ b/decentralized-api/internal/devshard/manager.go
@@ -323,7 +323,7 @@ func (m *HostManager) create(escrowID string) (*transport.Server, error) {
 	}
 
 	h, err := host.NewHost(sm, m.signer, m.engine, escrowID, group, nil,
-		m.hostOptions(escrow.EpochID)...,
+		m.hostOptions(escrow.EpochID, config.InferenceSealGraceNonces)...,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("create host: %w", err)
@@ -507,7 +507,7 @@ func (m *HostManager) recoverStoredSession(escrowID string) (*transport.Server, 
 	}
 
 	h, err := host.NewHost(sm, m.signer, m.engine, escrowID, meta.Group, nil,
-		m.hostOptions(meta.EpochID)...,
+		m.hostOptions(meta.EpochID, meta.Config.InferenceSealGraceNonces)...,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("create host: %w", err)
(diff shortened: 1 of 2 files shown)


a-kuprin:
While problem is real, but overall analysis was wrong and fix is in the wrong direction.
Discussed at H1 https://hackerone.com/reports/3819408

Fix is at https://github.com/gonka-ai/gonka/pull/1368

PR #1356, opened 2026-06-22
fix(inference): derive devshard settlement quorum from distinct validators

ouicate:
## Summary
A devshard escrow's settlement quorum was derived from the number of signed **slots**, not the number of distinct **validators**. Slots are assigned by weighted sampling **with replacement**, so one validator can legitimately occupy many — or all — of an escrow's 16 slots, and the per-slot signature payload commits only to `StateRoot + EscrowId + Nonce` (not the slot id). A single validator that owned `>= 11` slots could therefore reuse **one** signature across 11 slot ids and single-handedly satisfy the `11-of-16` quorum, forging settlement with no genuine multi-party agreement. This PR makes quorum count **distinct validators** and rejects escrow creation when the sampled slots resolve to too few distinct validators.

## Problem
- Quorum was `DevshardQuorumFor(len(escrow.Slots))` and the verification loop counted one vote per **slot** (`slotVotes++`), deduping only on `sig.SlotId`.
- The signed payload (`DevshardStateSignatureContent`) omits `slot_id`, so one 65-byte signature recovers to the same address for every slot that validator owns and matches `escrow.Slots[sig.SlotId]` each time.
- Escrow creation applied **no** distinct-validator floor and **no** per-validator slot cap, and sampled from model-local weights only (no abstention scaling), so a single-validator or dominant-validator model group could be fully owned by one party.

A single dominant/sole validator could thus settle an arbitrary final state, deny honest-minority slots any payout, and write per-validator epoch work-stat counters (`InferenceCount` / `MissedRequests` / `ValidatedInferences`) that feed the downtime reward gate. Confirmed by an empirical test in which one key met the `11-of-16` quorum with a single reused signature.

## Solution
Two changes:

- **`VerifyDevshardSettlement`** now derives the required quorum from the number of **distinct validators** that own slots (`DevshardQuorumFor(len(distinctValidators))`) and tallies **one vote per distinct slot-validator** (keyed by the slot's expected cold address, so a warm-key signature folds into that validator's single vote). A validator can never contribute more than one vote regardless of how many slots it owns.
- **`CreateDevshardEscrow`** rejects escrows whose sampled slots resolve to fewer than `DevshardMinDistinctSlotValidators` (2) distinct validators, so an escrow can never be created that a single party could settle alone.

### DIFF
--- a/inference-chain/x/inference/keeper/devshard_settlement.go
+++ b/inference-chain/x/inference/keeper/devshard_settlement.go
@@ -20,6 +20,10 @@ import (
 const (
 	DevshardGroupSize   = 16
 	DevshardQuorumSlots = 2*DevshardGroupSize/3 + 1
+	// DevshardMinDistinctSlotValidators is the minimum number of distinct
+	// validators an escrow's slots must resolve to at creation, so no single
+	// validator can reach settlement quorum alone.
+	DevshardMinDistinctSlotValidators = 2
 	// DevshardSettlementPhase is the phase byte appended to the state root preimage.
 	// The chain hardcodes 0x02 (Settlement) so only fully-finalized devshard states
 	// can pass verification. States at phase Active (0x00) or Finalizing (0x01)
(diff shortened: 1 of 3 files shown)


a-kuprin:
Non applicable, discussed at
https://hackerone.com/reports/3819413

PR #1357, opened 2026-06-22
fix(inference): refund creator when unsettled devshard escrow is pruned

ouicate:
## Summary
If a devshard escrow is never settled, the pruner (at `epoch + 2`) split the **entire** locked amount equally among the unique slot validators — the creator received **no refund**, and the integer-division remainder was **permanently stranded** in the module account. With no cancel/refund message, any failure to settle (lost quorum, validator withholding, client crash, or a hash-drift settlement rejection) became a total creator loss and a value-transfer to the validator group. This PR refunds the full locked amount to the creator on prune.

## Problem
- `distributeUnsettledEscrow` computed `share = escrow.Amount / uniqueValidators` and paid each unique slot validator; **no branch** sent anything to `escrow.Creator`, and `escrow.Amount % uniqueValidators` was never swept (and if `amount < uniqueValidators`, `share == 0` stranded the whole amount).
- The only devshard escrow txs are create and settle — there is **no** `CancelDevshardEscrow` / `RefundDevshardEscrow`, so "settle (needs quorum) or forfeit via prune" was the entire option set.
- Settlement is the only on-chain proof that validators performed (and were credited for) work; an unsettled escrow has no such proof, so distributing the balance to validators paid for unproven work.

## Solution
Replace the distribution with `refundUnsettledEscrow`: return the **full** `escrow.Amount` to `escrow.Creator` (bank memo `devshard_escrow_unsettled_refund`). The unsettled path never disbursed any funds — settlement, which pays validators and refunds the remainder, sets `escrow.Settled` and is therefore skipped by the pruner — so the full amount is still held by the module account and is returned to the payer intact, with nothing stranded.

### DIFF
--- a/inference-chain/x/inference/keeper/devshard_pruning.go
+++ b/inference-chain/x/inference/keeper/devshard_pruning.go
@@ -2,6 +2,8 @@ package keeper
 
 import (
 	"context"
+	"fmt"
+	"math"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/productscience/inference/x/inference/types"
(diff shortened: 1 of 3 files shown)


gmorgachev:
Hi @ouicate !
It's directly contradict the idea of `devshard` concept. The creator is responsible for creating escrow and it's proposer. Only creator can settle and only creator has incentive to do so 
The distribution back to creator allows to just make work in devshard, never settle and get full refund

PR #1361, opened 2026-06-23
adding new workflow to only build amd path for linux 

maria-mitina:
For testnet purposes only - created a separate workflow to build only amd64 binaries and publish on race-releases. Otherwise time to build is 60+min, we will be able to return to 10-13 min.


DimaOrekhovPS:
resolved here:
https://github.com/gonka-ai/gonka/pull/1362

Issue #1341, opened 2026-06-12
[BUG] devshard: data race on inflight receiptTime between send goroutine and escalation scheduler (go test -race fails on main)


redstartechno:
## Summary
`go test ./cmd/devshardctl/ -race -run TestRunInference` fails on current `main` (8bd883ba3) with 15 data race reports across 13 tests. The dominant race (11 of the 15 reports) has production code on both sides: the send goroutine writes `inf.receiptTime` inside `receiptOnce.Do` (`redundancy.go:1491`) while the orchestrator's escalation scheduler reads `inf.receiptTime.IsZero()` without synchronization (`redundancy.go:2251`, in `escalationForInflight` ← `nextEscalationTrigger` ← `awaitRace` ← `RunInference`).

## Motivation
`time.Time` is a multi-word struct, so the Go memory model gives this read no guarantees — the scheduler can observe a torn or stale value and mis-schedule (or skip) the receipt-timeout escalation. The comment block at `redundancy.go:1501-1512` documents a previous scheduler race of exactly this shape on `sendTime`, fixed by stamping it synchronously before the goroutine spawns; `receiptTime` has the same consumer but cannot be stamped synchronously (it is set when the receipt arrives), so it needs actual synchronization. The races also block ever enabling `-race` in CI for devshard. Found while race-testing an unrelated branch; reproduces identically on unmodified `main`.

## Impact
- Who is affected: hosts running the devshard gateway (request orchestration / escalation path); developers (the package cannot pass `go test -race`).
- Network-wide or limited: limited — single participant, non-chain.
- Likelihood: intermittent at runtime (timing-dependent); deterministic under the race detector — every `TestRunInference_*` run reports it.
- Severity: Impact limited × likelihood intermittent → low-to-medium. Practical worst case is a missed or mistimed receipt-timeout escalation (tail latency / stall on a silent primary — the exact regression class the `sendTime` fix in the comment describes).
- Affected components: `devshard/cmd/devshardctl` (`Redundancy.RunInference`, escalation scheduling).

## Detailed description

Reproduce (Linux, Go 1.24.2, commit 8bd883ba3):

```
cd devshard
go test ./cmd/devshardctl/ -race -count=1 -run TestRunInference
```

[...]

PR #1332, opened 2026-06-10
Add Docker-backed devshard E2E test suite

aikuznetsov:
## Summary

Adds a Docker-backed E2E test suite for `devshard` and wires it into the existing Go build/test CI path.

The suite runs real integration validation with:

- `mock-chain`
- Postgres
- three `devshard-host` participants
- `devshardctl` as the OpenAI-compatible HTTP entry point

Tests are written in Go and use `testcontainers-go` to own service lifecycle, networks, ports, logs, restarts, and volumes. `docker-compose.e2e.yml` only starts the Go test runner container.

## What Changed

- Added `devshard/e2e` tests for:
  - happy path completions and settlement contract validation
  - OpenAI-compatible streaming/SSE behavior
  - streaming and non-streaming cache isolation
  - SQLite volume restart/recovery scenarios
- Added shared E2E test utilities under `devshard/e2e/testutil`.
- Added E2E-only Docker image builds:
  - `devshard-mock-chain:local`
  - `devshard-host:local`
  - `devshardctl:local`
- Added `devshard/Makefile` E2E targets:
  - `make -C devshard e2e`
  - `make -C devshard e2e-test`
  - `make -C devshard e2e-images`
- Added `devshard/docker-compose.e2e.yml` for the Go test runner container.
- Added JUnit output via `gotestsum` at:
  - `devshard/build/test-results/e2e/junit.xml`
- Wired devshard E2E into `Go Build and Test / Build and Test Devshard`.
- Added `devshard/e2e/README.md`.

### Fix flaky unit tests
It also fixes flaky unit tests uncovered during CI hardening:
- Stabilizes `devshardctl` state-divergence participant blocking test by waiting for the async escrow block to be observed before sending the next request.
- Stabilizes runtimeconfig epoch transition listener tests by making the test fakes release transitions deterministically while keeping the ordered assertions.

## Notes

- E2E tests are guarded by `DEVSHARD_E2E=1`; the Compose runner sets this automatically.
- Testcontainers lifecycle logs are muted, while debug mode still shows E2E request/response logs and settlement contracts.
- E2E Docker networks are uniquely named per test run to avoid stale-network collisions after interrupted runs.
- `host.docker.internal:host-gateway` is configured for Linux CI runner compatibility.

## How To Run

Build images and run the full E2E suite:

```sh
make -C devshard e2e
```

## Readme.md

https://github.com/gonka-ai/gonka/blob/2230e307f336500bcfc374947a8ff37e4593c8c0/devshard/e2e/README.md

PR #1366, opened 2026-06-25
Ak/pixelplex refactoring into r2 (HA multi-instance devshard, decoupling refactor, and testenv integration suite)

a-kuprin:
## Summary

This PR lands an architecture change to support **multiple devshard instances** for high
availability, together with a structural refactor that extracts shared chain/runtime code
into `common/`, decouples **devshardd** from **decentralized-api**, and introduces
**edge-api** as a chain-facing read path. It adds a **Docker-backed integration test
suite (`testenv`)** that validates the full node stack without Testermint, removes **LCD
REST** from the gateway chain transport in favour of **gRPC**, and seeds the HA rollout
design with docs and operational fixes.

It also hardens HA storage (fail-closed Postgres + boot migrate), gateway/chain defaults,
validation leases, admin path containment, and edge verify/debug exposure defaults.

---

## Motivation

Today, a single `decentralized-api` process and tightly coupled devshard wiring block
horizontal scaling. To run **N versiond / devshardd instances** behind a sticky router
on **shared Postgres**, we need:

1. **devshardd** as a standalone binary (not embedded in dapi).
2. **Shared chain client code** in `common/` (queries, tx, runtimeconfig).
3. A **gateway** (`devshardctl`) that talks to decentralized-api for runtime params and to the
   chain over a single gRPC transport.
4. **Integration tests** that exercise the real compose topology without the full
   Testermint genesis cluster.

This PR implements that foundation and proves it with citest scenarios S1–S6, G1–G4,
and adversarial A1–A4.

---

### Multi-instance HA (versiond + devshardd)

- **Multiple `versiond` hosts**, each supervising a **devshardd** child, behind
  **versiond-router** (nginx consistent-hash on session id).
- **Shared Postgres** for payload storage and cross-instance lease semantics (required
  for HA; SQLite remains single-instance).
- With `DEVSHARD_HA=1` or `DEVSHARD_REQUIRE_POSTGRES=1`, storage is **Postgres-only**
  (no file/SQLite fallback under Postgres outage). Boot migrates local SQLite sessions
  and file payloads into Postgres.
- Join multi-versiond overlay requires `DEVSHARD_POSTGRES_PASSWORD` (no hardcoded
  password in compose).
- Gateway chat and long-poll paths validated end-to-end through the router.

See [high-availability-architecture.md](./high-availability-architecture.md) and
[proposals/high-availability.md](./proposals/high-availability.md).

### Decoupling and new packages

| Area | Change |
|------|--------|
| **`common/`** | Shared `chain` gRPC client, `chain/tx` broadcast helpers, `runtimeconfig` server/client |
| **devshardd** | Standalone process under `devshard/cmd/devshardd/`; runtime params via dapi long-poll |
| **decentralized-api** | Slimmed nodemanager/runtimeconfig delegation; devshardd no longer embedded |
| **edge-api** | Chain read APIs via `common/chain.Client` (HA chain proxy path) |
| **devshardctl** | OpenAI-compatible gateway; gRPC-only chain queries and escrow tx |

### Chain transport: REST → gRPC (gateway)

**devshardctl** no longer uses LCD REST (`NewRESTBridge`, `DEVSHARD_CHAIN_REST`) for
escrow create/settle or participant reads. All gateway chain I/O goes through
`common/chain` gRPC — the same transport **devshardd** already uses.

- mock-chain **LCD `:1317` / `restface` removed** from testenv
- compose sets `DEVSHARD_CHAIN_GRPC` only on gateway
- static gate: `TestG4_NoRESTChainClientsInGatewayProduction`
- Optional chain gRPC TLS via `CHAIN_GRPC_TLS=true|1|yes` (default remains plaintext)
- Gateway uses one shared chain gRPC client for runtime params and commitment reconciliation

Details: [testenv/docs/chain-transport-consolidation.md](../testenv/docs/chain-transport-consolidation.md).

---

## Integration testing: `testenv` vs [PR #1332](https://github.com/gonka-ai/gonka/pull/1332)

Two complementary Docker test suites; they should be **merged or coordinated** over time.

| | **testenv** (this PR) | **PR #1332 — `devshard/e2e`** |
|--|----------------------|-------------------------------|
| **Scope** | Full **node stack**: gateway + multi-instance versiond + decoupled devshardd + mock-dapi + mock-chain + mock-openai | **devshard protocol** behaviour: completions, settlement, streaming, SQLite/Postgres recovery |
| **Topology** | 2× versiond, sticky router, shared Postgres, devshardctl gateway | 3× devshard-host, devshardctl entry, mock-chain, Postgres |
| **Framework** | `devshard/testenv/citest` + compose gen (`gencompose`) | `testcontainers-go` + `devshard/e2e` |
| **CI** | `make -C devshard ci-testenv-unit` / `ci-testenv-integration` | `make -C devshard e2e` in Go Build and Test |
| **Goal** | Deploy/join-style wiring, HA routing, params long-poll, epoch switch, adversarial faults | Protocol contracts, cache isolation, volume restart |

**testenv** answers: *does the production-shaped stack boot and route correctly?*  
**#1332** answers: *does devshard protocol logic settle and recover correctly?*

Operator runbook: [testenv-v2.md](./testenv-v2.md). Scenario index:
[testenv/docs/scenarios.md](../testenv/docs/scenarios.md).

---

### Runtime and gateway

- PixelPlex merge + refactor for multi-versiond HA topology
- `devshardctl` effective chain gRPC from compose env (`DEVSHARD_CHAIN_GRPC`,
  `DEVSHARD_NODE_MANAGER_ADDR`)
- Gateway chat (stream + non-stream) over router → devshardd → mock-openai
- Params long-poll: mock-chain → mock-dapi → `GetRuntimeConfig` (lane C)
- Escrow create/settle/dispute signing uses default chain IDs when unset
  (`gonka-mainnet`; override with `DEVSHARD_CHAIN_ID` for testnet/testenv)
- Startup settlement rehydrates non-resident escrows so reconcile can settle after restart
- Unknown model `access_mode` requires an API key (does not fall open to public)
- Admin create/import `storage_path` must be relative to the gateway base storage
  directory (absolute paths and `..` rejected)
- Runtime-config `ServedAt` unset is encoded as unix `0` on the wire; negative values
  decode as unset

### Validation leases and payloads

- Validation lease results only apply when the submitting instance still owns a pending
  lease; leases older than the lease TTL are abandoned (no submit / no result write)
- File payload paths sanitize escrow id segments and stay under the configured base
  directory

### Edge / proxy

- `/v1/verify-*` and `/v1/debug/*` are private by default (403). Publish with
  `EDGE_API_EXPOSE_OPTIONAL_ROUTES=true` (put auth on nginx if exposed)

### testenv v2

- `mock-chain` (gRPC + RPC + admin), `mock-dapi`, `mock-openai`
- `cmd/gencompose` — config-driven multi-versiond compose
- Citest harness: S1–S6 stack, G1–G4 gRPC transport, A1–A4 adversarial, O1 observability
- CI workflow: `.github/workflows/devshard-testenv.yml`

### Docs and HA seeding

- [high-availability.md](./proposals/high-availability.md) — target HA decomposition
- [release-0.2.13-v2-r2.md](./release-0.2.13-v2-r2.md) — Postgres requirement for HA,
  `session_version` status field
- [storage-design.md](./storage-design.md), [rolling-update.md](./rolling-update.md) cross-links

### Cleanup

- Remove accidental committed binaries from tracking; extend `.gitignore`
- Delete mock-chain LCD REST surface and `bridge/rest.go` gateway path
- Minor flaky-test and governance-param fixes carried on the branch

### Operator env checklist

| Variable | Notes |
| --- | --- |
| `DEVSHARD_HA` / `DEVSHARD_REQUIRE_POSTGRES` | Fail-closed Postgres for multi-instance |
| `DEVSHARD_POSTGRES_PASSWORD` | Required for join versiond overlay |
| `DEVSHARD_CHAIN_ID` | Override when not mainnet |
| `CHAIN_GRPC_TLS` | Opt-in TLS to chain gRPC |
| `EDGE_API_EXPOSE_OPTIONAL_ROUTES` | Opt-in public verify/debug routes |

### Known follow-ups (not in this PR)

- Warm-key revoke after bind / long-lived cache
- Chat response cache: share body for identical prompts, but attribute escrow and request
  ids from the new request
- Settle success should wait for DeliverTx confirmation, not only SYNC CheckTx
- Dispute tx sequence mismatch retry and codec registry init safety
- Escrow rotation: do not retire active escrows until temp bridge capacity is ready
- In-flight identical chat streams: join one ML generation (replay prefix + live tail)
  instead of starting a second
- Escrow private keys at rest in `gateway.db` and admin state redaction

---

### Unit (no Docker)

```bash
make -C devshard ci-testenv-unit
```

Covers `common/runtimeconfig`, `devshard/chainoracle`, `devshard/runtimeparams`,
`devshard/testenv/...` packages, and `decentralized-api/nodemanager`.

### Integration (Docker)

```bash
make -C devshard ci-testenv-integration
```

| Suite | Command | Scenarios |
|-------|---------|-----------|
| Stack | `make -C devshard/testenv citest-stack` | S1–S6 |
| gRPC transport | `make -C devshard/testenv citest-grpc-transport` | G1–G4 |
| Adversarial | `make -C devshard/testenv citest-adversarial` | A1–A4 |

Key scenarios:

- **S2** — session stickiness across versiond upstreams
- **S3/S4** — governance long-poll and epoch switch
- **S5/G3** — gateway chat without LCD REST
- **S6** — behaviour when one versiond is stopped
- **A1–A4** — ML fault injection, stale escrow, revoked warm key

### Manual smoke

```bash
cd devshard/testenv
make build-devshardd gen-compose up
# gateway :18081, router :18080
```


a-kuprin:
commit https://github.com/gonka-ai/gonka/pull/1366/changes/7b2c7b4dd946d37c32108103dad1a3cdfbdd6d25 fixes https://github.com/gonka-ai/gonka/issues/1341
Now 

```
go test ./cmd/devshardctl/ -race -count=1 -run TestRunInference
```

pass

devshard/testenv/citest/a3_stale_escrow_test.go
```diff
@@ -0,0 +1,35 @@
...
+		}
+	})
+
+	escrowID := harness.GetGatewayEscrowID(t, client, eps.GatewayHTTP)
+	id, err := strconv.ParseUint(escrowID, 10, 64)
+	if err != nil {
+		t.Fatalf("parse escrow_id %q: %v", escrowID, err)
+	}
+
+	harness.Step(t, "settle escrow %d on mock-chain while 2× versiond stack is up", id)
```
anatoly-kuz-mntn, line 52: In `devshard/testenv/citest/a3_stale_escrow_test.go`, the test settles the escrow around line 32, but the assertion around line 34 only verifies the resulting mock-chain state. That proves the test hook/mutation worked, but it does not strongly verify how the production-shaped stack reacts to a stale escrow condition end to end.

Suggestion: after forcing the escrow into the stale/settled state, also assert the observable behavior from the gateway/router path, for example that a follow-up chat/finalize attempt fails or recovers in the expected way, and that the stale escrow is not reused incorrectly.

For example, after mutating the mock-chain escrow state, assert the behavior through the gateway path as well:

```go
gatewayURL := eps.GatewayHTTP
adminKey := harness.TestenvAdminAPIKey
model := config.PrimaryModelID(cfg)

req := harness.ChatCompletionRequest{
    Model: model,
    Messages: []harness.ChatMessage{
        {Role: "user", Content: "citest a3 stale escrow after forced settlement"},
    },
    MaxTokens: 32,
}

harness.Step(t, "gateway should not continue using stale settled escrow %d", id)
status := harness.PostGatewayChatExpectFailure(t, client, gatewayURL, adminKey, req)
require.NotEqual(t, http.StatusOK, status)
```

devshard/testenv/citest/a4_bad_warm_key_test.go
```diff
@@ -0,0 +1,48 @@
...
+	client := harness.HTTPClient()
+	mockDapi := harness.MockDAPIFromConfig(cfg)
+	requireHosts(t, cfg, 2)
+
+	granter := cfg.Hosts[0].Address
+	warm := cfg.WarmGrantee.Address
+	requireNonEmpty(t, granter, "host[0] address")
+	requireNonEmpty(t, warm, "warm_grantee address")
+
+	harness.Step(t, "revoke warm key %s for granter %s via /testenv/grantees", warm, granter)
```
anatoly-kuz-mntn, line 46: In` devshard/testenv/citest/a4_bad_warm_key_test.go`, the test revokes the warm key around line 28, but the assertion around line 33 only checks that the mock-chain grantee state changed. That proves the test hook worked, but it can still pass even if devshardd, devshardctl, or the router behave incorrectly after the warm key is revoked.

Suggestion: after revoking the warm key, also exercise an observable gateway or tx path and assert the intended runtime outcome. For example, verify that a follow-up gateway chat/finalize/settlement attempt is rejected cleanly, that the affected host is deactivated, or that a visible error/metric/log signal is emitted. This would make the test cover the production-shaped stack’s reaction to the fault, not only the fault injection itself.

devshard/testenv/citest/s5_gateway_chat_test.go
```diff
@@ -0,0 +1,58 @@
...
+	harness.WaitGETOK(t, client, eps.RouterHTTP+"/"+cfg.Versiond.VersionName+"/healthz", 5*time.Minute, "devshardd health via router", stack)
+
+	gatewayURL := eps.GatewayHTTP
+	model := config.PrimaryModelID(cfg)
+	adminKey := harness.TestenvAdminAPIKey
+
+	nonStreamReq := harness.ChatCompletionRequest{
+		Model: model,
+		Messages: []harness.ChatMessage{
+			{Role: "user", Content: "citest s5 non-stream"},
```
anatoly-kuz-mntn, line 39: **Missing settlement contract coverage in the new production-shaped stack**

`devshard/testenv/citest/s5_gateway_chat_test.go` validates that non-streaming and SSE chat requests reach mock-openai, but it stops at the chat response. It does not finalize the session or assert the settlement output.

**Suggestion:** add a dedicated S-scenario that extends S5 beyond “chat reaches mock-openai” and verifies the settlement contract through the new stack. For example:

```go
func TestS7_GatewayChatFinalizeSettle(t *testing.T) {
    harness.SkipUnlessEnv(t, "TESTENV_CITEST")
    harness.RequireDocker(t)

    stack, cfg, eps := harness.BootS1Stack(t, "citest-s7-*")
    client := harness.GatewayChatClient()
    t.Cleanup(func() {
        if t.Failed() {
            harness.DumpComposeLogs(t, stack, "devshardctl", "versiond-0", "versiond-1", "mock-chain", "mock-dapi")
        }
    })

    harness.WaitS1Healthy(t, stack, eps)
    harness.WaitGatewayChatReady(t, client, eps.GatewayHTTP, 3*time.Minute, stack)

    gatewayURL := eps.GatewayHTTP
    adminKey := harness.TestenvAdminAPIKey
    model := config.PrimaryModelID(cfg)

    req := harness.ChatCompletionRequest{
        Model: model,
        Messages: []harness.ChatMessage{
            {Role: "user", Content: "citest settlement contract path"},
        },
        MaxTokens: 32,
    }

    resp := harness.PostGatewayChatCompletion(t, client, gatewayURL, adminKey, req)
    harness.RequireMockOpenAIContent(t, resp.Choices[0].Message.Content)

    escrowID := harness.GetGatewayEscrowID(t, client, gatewayURL)

    // New helper/API wrapper around the devshardctl finalize endpoint.
    finalized := harness.FinalizeGatewaySession(t, client, gatewayURL, adminKey, escrowID)

    require.Equal(t, escrowID, finalized.EscrowID)
    require.NotEmpty(t, finalized.StateRootAndProtocolVersion)
    require.NotEmpty(t, finalized.Signatures)
    require.NotEmpty(t, finalized.HostStats)

    // New helper/API wrapper that submits the finalized settlement payload
    // through the same mock-chain/mock-dapi path used by this production-shaped stack.
    settleResp := harness.SettleDevshardEscrow(t, client, eps, finalized.RawJSON)

    require.Equal(t, 0, settleResp.Code)
    require.Contains(t, settleResp.EventTypes(), "devshard_escrow_settled")
}
```

devshard/testenv/citest/s6_versiond_stop_test.go
```diff
@@ -0,0 +1,59 @@
...
+	version := cfg.Versiond.VersionName
+	sessionA, upstreamA, sessionB, upstreamB := harness.FindDistinctStickySessions(t, client, eps.RouterHTTP, version)
+	require.NotEqual(t, upstreamA, upstreamB)
+
+	stoppedHost := harness.HostIDForUpstream(cfg, upstreamA)
+	require.NotEmpty(t, stoppedHost, "map upstream %q to host id", upstreamA)
+	harness.Step(t, "session %q → %s; session %q → %s; stopping %s",
+		sessionA, upstreamA, sessionB, upstreamB, stoppedHost)
+
+	stack.StopService(t, stoppedHost)
```
anatoly-kuz-mntn, line 40: **Missing restart/persistence coverage at the production-shaped level**

`devshard/testenv/citest/s6_versiond_stop_test.go` stops one versiond instance and checks routing behavior, but it does not restart versiond/devshardd and continue the same session afterward.

Suggestion: extend this scenario, or add a new one, that exercises:
```
start session -> send chat -> restart one versiond -> continue same session
              -> restart all versiond instances -> continue same session
              -> assert gateway chat still works
              -> assert session/nonce/state does not regress
```
That would verify the new versiond -> devshardd -> router -> gateway topology survives restarts without losing session state or regressing nonce/state handling.

anatoly-kuz-mntn (commented):
@a-kuprin  PTAL at comments

devshard/testenv/citest/s5_gateway_chat_test.go
a-kuprin: I think we can extend this test coverage in further releases, or at separate pr.
You are free to add it.

Here the purpose was to scaffold e2e for devshardd decoupled from dapi and test primarily the multi-versiond architecture.
Also settlement is covered at testermint tests, so we don't miss it.

I agree we should move it to this e2e environment

devshard/testenv/citest/a4_bad_warm_key_test.go
a-kuprin: Added at commit https://github.com/gonka-ai/gonka/pull/1366/commits/559533e9e512a4be8b0735d5d47eae491c0cc0c0

devshard/testenv/citest/s6_versiond_stop_test.go
a-kuprin: Added at commit https://github.com/gonka-ai/gonka/pull/1366/commits/559533e9e512a4be8b0735d5d47eae491c0cc0c0

There will be separate PR with feature for rolling update: https://github.com/gonka-ai/gonka/blob/ak/pixelplex-refactoring-into-r2/devshard/docs/rolling-update.md

There will be also added scenarios for versiond graceful one-by-one restart

devshard/testenv/citest/a3_stale_escrow_test.go
a-kuprin: Added at commit https://github.com/gonka-ai/gonka/pull/1366/commits/559533e9e512a4be8b0735d5d47eae491c0cc0c0

edge-api/Makefile
```diff
@@ -0,0 +1,54 @@
...
+ifeq ($(BLST_PORTABLE),1)
+export CGO_CFLAGS := $(CGO_CFLAGS) -D__BLST_PORTABLE__
+endif
+
+define DOCKER_BUILD
+	@echo "--> building edge-api docker image"
+	@echo "    	platform: $(PLATFORM)"
+	@DOCKER_BUILDKIT=1 docker build \
+		--platform $(PLATFORM) \
+		-f edge-api/Dockerfile \
```
snevolin: This target is invoked from the root Makefile via `make -C edge-api` when we want to build the Testermint environment. With this call the working directory is already "edge-api/", so the whole path will be `edge-api/edge-api/Dockerfile`, which is wrong. Change it to `-f Dockerfile \` to make this work.

edge-api/Makefile
```diff
@@ -0,0 +1,54 @@
...
+		. \
```
snevolin: Also to address the `inference-chain/`, `common/` and so on paths from the parent directory you should use `..` as a context.

edge-api/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+all: build-docker
+
+build-docker:
+	$(eval PLATFORM=$(DOCKER_PLATFORM))
+	$(eval DOCKER_TAG=ghcr.io/product-science/edge-api:$(VERSION))
+	$(DOCKER_BUILD)
+	@if [ "$(SET_LATEST)" = "1" ]; then \
+		echo "Setting latest tag..."; \
+		docker tag $(DOCKER_TAG) ghcr.io/product-science/edge-api:latest edge-api:latest; \
```
snevolin: docker tag can add only one tag at a time. To add two tags you should call it twice:
`docker tag $(DOCKER_TAG) ghcr.io/product-science/edge-api:latest; \
docker tag $(DOCKER_TAG) edge-api:latest; \`

edge-api-router/Makefile
```diff
@@ -0,0 +1,17 @@
...
+
+build-docker:
+	DOCKER_BUILDKIT=1 docker build \
+		--platform linux/amd64 \
+		-f Dockerfile \
+		. \
+		-t ghcr.io/product-science/edge-api-router:$(VERSION)
+	@if [ "$(SET_LATEST)" = "1" ]; then \
+		echo "Setting latest tag..."; \
+		docker tag ghcr.io/product-science/edge-api-router:$(VERSION) ghcr.io/product-science/edge-api-router:latest edge-api-router:latest; \
```
snevolin: The same as in edge-api. To add two tags at a time you should call `docker tag` twice:

`docker tag ghcr.io/product-science/edge-api-router:$(VERSION) ghcr.io/product-science/edge-api-router:latest; \`


`docker tag ghcr.io/product-science/edge-api-router:$(VERSION) edge-api-router:latest; \`

devshard/testenv/citest/s5_gateway_chat_test.go
aikuznetsov: Sounds good! I agree it makes sense to keep this PR focused. Let's add the end-to-end settlement coverage in a follow-up PR.

aikuznetsov (approved):

testermint/src/main/kotlin/LocalInferencePair.kt
snevolin: You added in this PR gRPC connectivity between devshardctl/edge-api and the chain node (the gateway used to talk over REST), but you didn't add the wiring for GRPC in testermint.

You should add `" DEVSHARD_CHAIN_GRPC=\$NODE_HOST:9090" +`

snevolin: Also you need to change the node's gRPC address from localhost:9090 to 0.0.0.0:9090 in init-docker-genesis.sh:

sed -Ei 's/^address = ".*:9090"$/address = "0\.0\.0\.0:9090"/g' "$STATE_DIR/config/app.toml"

Add it right before `cosmovisor run start &`

a-kuprin: Yep, running testermints now and debugging to let the pass

x0152:
> Settle success should wait for DeliverTx confirmation, not only SYNC CheckTx

I tried to understand the potential impact

Anyway, the fix looks minimal: copy the create-path polling pattern after broadcastTx and use the same confirmation logic for settle

common/storage/payloads/factory.go
```diff
@@ -0,0 +1,99 @@
...
+		return pgStore, pgStore.Close, nil
+	}
+
+	slog.Info("payload storage: using postgres with file fallback", "host", pgHost)
+	hybrid := NewHybridStorage(pgStore, fileStore, pgRetryInterval())
+	return hybrid, hybrid.Close, nil
+}
+
+func haModeEnabled() bool {
+	return envTruthy("DEVSHARD_HA") || envTruthy("DEVSHARD_REQUIRE_POSTGRES")
```
snevolin: The same check is duplicated in devshard (devshard/storage/ha.go) and I also have to cpy it in versiond to manage rolling overlap. But what if instead keeping the storage mode in three variables (DEVSHARD_HA, DEVSHARD_REQUIRE_POSTGRES and PGHOST) we would use only one: DEVSHARD_STORAGE_MODE = sqlite | hybrid | postgres (default hybrid)? It would be clearer.

a-kuprin:
> I tried to understand the potential impact
> 
> Anyway, the fix looks minimal: copy the create-path polling pattern after broadcastTx and use the same confirmation logic for settle

I was analyzing PR with AI tools and have skipped some concerns and this skips was added to PR description. Yes it could be done by copying the create-path polling pattern.

The impact is that it is theoretically possible that CheckTx is ok when accepted by mempool, but transaction wasn't really applyed across the chain.

common/storage/payloads/factory.go
a-kuprin: Yes, I think it should be adopted

a-kuprin: Addressed at commit https://github.com/gonka-ai/gonka/pull/1366/commits/5241e42237e15362f6f645d7e13fd4e962d1d863

Added to common using `DEVSHARD_STORAGE_MODE`, default is `auto`
If `PGHOST` is set `auto` -> `hybrid`
If `PGHOST` isn't set `auto` -> `sqlite`
Check it.

For high-availability mode `DEVSHARD_STORAGE_MODE=postgres` and this will migrate sqlite sessions and payloads to postgres

a-kuprin:
Important note for https://github.com/gonka-ai/gonka/pull/1366/commits/b414b8d7e701656eea04fe4610c30837a33e9fb8 commit.

There are already released v1,v2,v3 devshard versions, that used sqlite or hybrid mode.

Multiple instances of versiond (for high-availability) that are introduced at this PR require to use Postgres. It means that we just can't run multiple versiond instances for previous devshard versions.

This checks were automated at the commit.
versiond-router checks if version is v1,v2,v3 it routes always to single VERSIOND_LEGACY_HOST
If VERSIOND_HOSTS has multiple values, versiond-router adds **`Devshard-Ha: true`** header in requests, so devshardd can check it and check if `DEVSHARD_STORAGE_MODE` is set to **`postgres`**, if mode is not set request will fail!

v4 devshardd can migrate from sqlite existing session to Postgres if `DEVSHARD_STORAGE_MODE=postgres`

I've added deploy and test-plan, also test-plan is automated at S8 e2e testenv test

x0152:
> > I tried to understand the potential impact
> > Anyway, the fix looks minimal: copy the create-path polling pattern after broadcastTx and use the same confirmation logic for settle
> 
> I was analyzing PR with AI tools and have skipped some concerns and this skips was added to PR description. Yes it could be done by copying the create-path polling pattern.
> 
> The impact is that it is theoretically possible that CheckTx is ok when accepted by mempool, but transaction wasn't really applyed across the chain.

I mean If it fails at DeliverTx, the escrow never settles and hangs, if I'm not mistaken. Could we fix it right now?

a-kuprin:
> > > I tried to understand the potential impact
> > > Anyway, the fix looks minimal: copy the create-path polling pattern after broadcastTx and use the same confirmation logic for settle
> > 
> > 
> > I was analyzing PR with AI tools and have skipped some concerns and this skips was added to PR description. Yes it could be done by copying the create-path polling pattern.
> > The impact is that it is theoretically possible that CheckTx is ok when accepted by mempool, but transaction wasn't really applyed across the chain.
> 
> I mean If it fails at DeliverTx, the escrow never settles and hangs, if I'm not mistaken. Could we fix it right now?

Yes, it's not a lot of work to fix. But I will out of computer for some hours now

a-kuprin:
> I mean If it fails at DeliverTx, the escrow never settles and hangs, if I'm not mistaken. Could we fix it right now?

Fix is at https://github.com/gonka-ai/gonka/pull/1366/commits/f44a6feac8cf7eb5389c048296c77f585c7ba851 commit

x0152 (approved):

PR #1373, opened 2026-06-28
(fix)Close inference module pool drain: cap ActualCost to escrow + validator-share clawback 

ouicate:
## Summary

Closes the full **x/inference module pool-drain** surface around invalidation and validator work-sharing. This combines and **supersedes #1250**:

1. **Validator-share clawback on invalidation** (originally #1250): when an invalidated inference is refunded, the validator/executor work-shares credited for it are clawed back, and validation-credit guards are aligned with the invalidated state.
2. **Cap `ActualCost` to the escrow actually collected** (new): enforce the tokenomics invariant `ActualCost <= EscrowAmount` at every site that moves value out of the shared module pool.

Together they guarantee that **no more value can leave the module pool for an inference than was escrowed for it** — on executor payment, validator `ShareWork`, and invalidation `IssueRefund`/clawback alike.

## Root cause

`EscrowAmount` is derived from `MaxTokens` at `StartInference`; `ActualCost` is derived from `CompletionTokenCount` at `FinishInference` and is only bounded by the global `MaxAllowedTokens`, never by `MaxTokens`.

- The **finish-first** path (`setEscrowForFinished`) already clamped cost to escrow.
- The **start-then-finish** path (`ProcessFinishInference`) persisted the raw `ActualCost` **uncapped**. A start with `MaxTokens=1` (tiny escrow) + a finish with `CompletionTokenCount=MaxAllowedTokens` persists an `ActualCost` orders of magnitude larger than the escrow collected.

That inflated value was then read directly by value-moving code:
- `shareWorkWithValidators` → `ShareWork(..., inference.ActualCost)` over-credits validators.
- `refundInvalidatedInference` → `IssueRefund(..., inference.ActualCost)` over-refunds (and the clawback/executor debit would be inconsistent with the refund).

Both over-pay out of the shared module account, draining the pool. #1250 fixed *who* is debited on invalidation but not *how much* `ActualCost` can be; this PR fixes the magnitude at the root and at every consumption site.

## Changes

**Cap (new)**
- `calculations.ProcessFinishInference`: clamp persisted `ActualCost` to `EscrowAmount` on the start-then-finish path (mirrors `setEscrowForFinished`).
- `types.Inference.CappedActualCost()`: single source of truth for `min(max(ActualCost,0), EscrowAmount)`. Makes already-persisted (legacy) inflated rows safe **at spend time**, so no state migration is required.
- `shareWorkWithValidators`, `refundInvalidatedInference`, and the #1250 clawback basis `invalidatedInferenceClawbackAdjustments` now consume `CappedActualCost()`, so the refund and the clawback remain mutually consistent and escrow-bounded.
- `ActualCostWithinEscrowInvariant` + `module.RegisterInvariants`: crisis-module tripwire asserting `ActualCost <= EscrowAmount` for started inferences; finish-first transients (no escrow collected yet) are skipped as benign.

**Clawback (from #1250)**
- `refundInvalidatedInference` refactored to load + apply clawbacks of the shares credited for the now-invalidated inference; refund is attempted before mutating balances.
- Validation credit guards aligned with the invalidated state.


qdanik:
@ouicate 
This is the legacy part of inference. The common logic now is present in devshard module. I believe we don't need to fix this issue.

ouicate:
@qdanik agreed devshard is the common path going forward — and its settlement already enforces the same `cost ≤ escrow` rule this PR is about: `SettleDevshardEscrow` rejects `totalPayout > escrow.Amount` (`msg_server_settle_devshard_escrow.go:138-140`). This PR just applies that same rule to the older per-inference path, which today still bypasses it.

A few reasons it's still live rather than dead code:
- All four handlers are still registered and reachable on-chain — `MsgStartInference` / `FinishInference` / `Validation` only require active-participant status, and `MsgInvalidateInference` is authorized through the group policy (`permissions.go:111`, `123-126`); none of them is behind a devshard gate. `MsgSetDevshardRequestsEnabled` only
  flips the `DevshardRequestsEnabled` param — it doesn't touch these handlers, so it can't be what's protecting this path.
- The standard OpenAI endpoint still drives them: `server.go:100`  (`POST /v1/chat/completions → postChat`) submits `MsgStartInference` / `MsgFinishInference` (`post_chat_handler.go:412`, `:936-958`). Devshard is a separate route group under `/v1/devshard`.
- They move funds through the same `x/inference` module account used for devshard escrows and reward payouts (`SendCoinsFromAccountToModule(..., types.ModuleName, ...)` on both sides), so an inflated `ActualCost` over-refunds/over-credits that shared pool and can starve devshard refunds/payouts — it isn't an isolated legacy balance.

Executors supply `CompletionTokenCount` at finish — it's bounded only by the global `MaxAllowedTokens`, never by the `MaxTokens` that sized the escrow, and it isn't covered by any verified signature (executor sig is disabled by policy) — so a malicious host can persist `ActualCost ≫ EscrowAmount`. Value-conservation shouldn't depend on executors being honest;
catching dishonest ones is what validation/invalidation is for.

The fix is intentionally low-risk: `CappedActualCost()` clamps at spend time (no state migration — already-persisted rows are made safe where they're spent) and the crisis invariant asserts `ActualCost ≤ EscrowAmount`. If you'd rather treat this path as retired, I'm happy to instead gate these handlers behind a disable switch (or remove them) so they can't move funds — either option closes it.

x0152:
Hi, thanks for pr! 

As already mentioned, this path is legacy and doesn't need to be fixed. Devshards is the current path for inference, while this path is no longer used

ouicate:
Thanks for the review @x0152. Did you happen to see my previous reply? I tried to explain why this still should be fixed.

The concern isn't whether devshard is the preferred path going forward—it's that this legacy inference flow is still reachable and is still able to move funds from the same `x/inference` module account. As long as those handlers remain enabled, the bug is exploitable.

The core issue is that `ActualCost` can exceed the escrow collected for an inference. Because that value is later used for validator rewards and refunds, an attacker can withdraw more from the shared module pool than was ever escrowed. Based on the current pool balance, this represents roughly **14,000 GNK** that can be drained.

If the intention is that this path is truly retired, then I'd completely agree that the correct fix is to disable or remove the handlers so they can no longer move funds. But in its current state they're still registered, reachable, and connected to the shared module account, so leaving the invariant broken means the pool remains exposed.

I'm happy to adapt the PR if the preferred direction is to explicitly disable the legacy path instead of fixing it, but I don't think leaving the vulnerable code reachable is a safe option.

PR #1376, opened 2026-06-29
Bridge block sync

### LINKED ISSUE #1358: Inbound bridge votes lost after upgrade simulation (node unavailable for some minutes for upgrade) — no retry, deposits stuck BRIDGE_PENDING
## Summary

Upgrade simulation (paused only the Gonka `node` container which halts the chain) while bridge (Geth) and `api` stay up. Inbound bridge deposits are detected and queued on unpause, but **validator votes can be lost silently** with no automatic recovery. Deposits can remain **`BRIDGE_PENDING`** indefinitely (majority never reached, no mint).

Observed on **gonka-testnet-4** (3 validators) during a controlled `docker pause node` test with Sepolia USDT inbound.

Full write-up in repo: `docs/bridge-inbound-pause-test-report.md`

## Test artifact

| Field | Value |
|-------|--------|
| Sepolia block | `11116928` |
| Receipt index | `199` |
| Receipts root | `0x89909ccf6494c61516e0e0bfe45867dc96a566b500566b39ad3462d5ab38068f` |
| USDT contract | `0x7169d38820dfd117c3fa1f22a697dba58d90ba06` |
| Gonka recipient | `gonka1qdwgkcxww42sxmcm9gk0trvugev2g0248upwyq` |

**On-chain result:** `BRIDGE_PENDING`, 1 validator (`gonka1zwcutshh…`), `totalValidationPower: 21459`. Wrapped USDT balance unchanged at **2**.

## Observed behavior

1. **During pause:** Chain height stalled (~12 min); API logged stale block time. Bridge Geth timed out on `GET /v1/bridge/addresses`.
2. **On unpause (~16:39:20–24):** Bridge burst-posted block `11116928` to all three APIs. Join validators logged `Processing receipt` with **no** `Error processing bridge exchange`.
3. **On-chain:** Join votes **not** recorded. Seed voted later (~16:46:27); only one validator on-chain — insufficient for majority mint.

### 1. API bridge queue: process once, no retry

`getNextBlock()` deletes the block from `pendingBlocks` before processing. Failed votes are not re-queued. The 5-minute ticker only drains remaining queue entries.

`decentralized-api/internal/server/public/bridge_handlers.go` — `getNextBlock`, `processReceipt`

### 2. `MsgBridgeExchange` uses `SendTransactionAsyncNoRetry`

Bridge votes bypass the halt-aware retry path used by PoC/validation:

- `updateChainHalt()` halt flag is **ignored**; broadcast still attempted during stale/halted chain.
- `broadcastMessage` returns `nil` error even when `resp.Code != 0` (CheckTx failure only logged as `Broadcast failed immediately` in tx_manager).

`decentralized-api/cosmosclient/cosmosclient.go` — `BridgeExchange`  
`decentralized-api/cosmosclient/tx_manager/tx_manager.go` — `SendTransactionAsyncNoRetry`, `broadcastMessage`

### 3. Geth bridge does not replay delivered blocks

Each finalized Ethereum block is POSTed once to `/admin/v1/bridge/block`. HTTP 200 (`Block queued`) advances the cursor regardless of whether the on-chain vote succeeded. No reconciliation for `BRIDGE_PENDING` txs.

`bridge/script.sh` (`--bridge.postblock`)

## Likely failure mode at unpause

Unordered txs use `timeout = latestBlockTime + 60s`. After a long pause, `latestBlockTime` is stale → timeout can be **in the past** → CheckTx rejection. Bridge handler still logs `Processing receipt` and drops the block from queue.

## Impact

- Inbound bridge deposits can stall at `BRIDGE_PENDING` forever
- `Processing receipt` log is misleading (success appearance on failure)
- No self-heal without manual re-post or manual `bridge-exchange` CLI

## Suggested fixes (priority order)

1. Surface broadcast failures in `BridgeExchange` (`resp.Code != 0` / `classifyBroadcastResponse`)
2. Use retry/halt-aware tx path for bridge votes (`SendTransactionAsyncWithRetry` or equivalent)
3. Do not remove block from queue until vote broadcast succeeds (or keep failed-receipt retry set)
4. *(Optional)* Reconciliation loop: local validator submits missing votes for known `BRIDGE_PENDING` receipts
5. *(Optional)* Geth cursor ack only after processing success (API contract change)

## Verification

```bash
inferenced q inference bridge-transaction \
  --origin-chain ethereum --block-number 11116928 --receipt-index 199 \
  --node http://localhost:8000/chain-rpc/ -o json

docker logs api 2>&1 | grep -iE 'Broadcast failed|node block time is stale|Error processing bridge'
```

GLiberman:
### Endpoint Topology (production)

Geth talks to **two different servers** (confirmed by `local-test-net/docker-compose.bridge.yml` and `deploy/join/docker-compose.yml`):

| Operation | Server | URL | Geth flag |
| --- | --- | --- | --- |
| Post block | **admin** (`:9200`) | `POST /admin/v1/bridge/block` | `--bridge.postblock` |
| Get bridge addresses | **public** (`:9000`) | `GET /v1/bridge/addresses` | `--bridge.getaddresses` |
| Handshake (last block) | **public** (`:9000`) | `GET /v1/bridge/block/latest` | `--bridge.getlastblock` |

Both servers share the same `*BridgeQueue` instance (injected in `main.go`), so the admin POST handler writes `latestBlocks`/SQLite while the public GET handler reads it. Because they share state, the GET handler **must** read `latestBlocks` through a lock-guarded accessor.

---

### 1.1 The Receipt-Only Finality Pipeline

- **Decoupled Sync**: `bridge-geth` runs in a custom `ReceiptSync` mode where blocks are not executed and the traditional `blockchain.CurrentHeader()` stays near genesis. The node tracks chain progress entirely through the consensus-layer Engine API, feeding headers into the `Skeleton` syncer.
- **Finality-Aware Posting**: The downloader retrieves block bodies and receipts but does **not** forward them to the API immediately. It asserts that blocks are at or below the finalized skeleton head (`progress.Finalized`), matching the canonical chain path.
- **Handshake Alignment**: When Geth starts, it queries the API's `/v1/bridge/block/latest` endpoint to retrieve the last block the Cosmos ledger has processed, using this value to calculate and set its downloading start cursor (`origin = clamp(C, tail-1, latest-1)`).

### 1.2 Strict Sequential Processing Invariant

- **The Chain-Phase Rule**: The decentralized API is a gateway to a Cosmos ledger that enforces block receipt order. Releasing transactions out-of-order or skipping blocks is fatal.
- **Queue-As-Buffer**: If a block is skipped due to network latency, the API queues subsequent blocks but **blocks execution** until `latest+1` arrives, then drains in perfect order.
- **Confirm-Before-Advance (commit 2)**: `latest` advances **only after every receipt of a block is confirmed recorded on-chain** (module-state read). A drain failure leaves the block buffered, `latest` unchanged, and the POST still returns `200` — retried on the next POST. This replaces the old "broadcast = committed" / `500`-on-failure model.

### 1.3 Recovery vs. Self-Healing Restart

- **In-Memory Range History**: Geth groups block payloads into batches/ranges as they are downloaded. It stores the last *N* finalized ranges of these payloads in memory (default 4).
- **Geth-Side Self-Healing**: Geth queries the API's latest block before sending a range starting at `blockX`.
  - **If API is slightly behind** (within Geth's *N*-range memory): Geth re-transmits the missing range(s) from memory block-by-block, restoring API state.
  - **If API is deeply behind** (beyond Geth's memory capacity): Geth logs a critical error (`log.Crit`) and exits. The system supervisor (e.g., Docker, Cosmovisor) restarts Geth. On restart, Geth's startup handshake queries the API's actual progress and seamlessly re-seeds its sync origin to resume downloading from the correct block height.

### 1.4 Schema Unification

- **Legacy Cleanup**: Historically, the API status endpoint returned a bloat of redundant keys (`latestProcessedBlock`, `lastBlock`, `blockNumber`, `block`) to support older Geth versions.
- **Standardization**: We unify both Geth and the API to use **only** the `blockNumber` key (plus `chainId` for context) across all endpoints. Geth's `GetLastConfirmedBlock` is refactored to parse only `blockNumber`.

### 1.5 Initial State & Handshake Error Fallback

- **Fail-Safe Architecture**: The synchronization loop must be resilient against API-side downtime or uninitialized states. Geth must **never** block or freeze its downloader because of failures in the continuity check.
- **Graceful Fallback**: If the API is in its initial state (returning no block number, `ok == false`) OR if Geth fails to reach the API (transport error, connection timeout, 5xx server error, `err != nil`), Geth must **gracefully fall back to its previous legacy behavior**:
  - **At Startup**: Geth calculates its sync origin as `latest - 1` (normal previous behavior) instead of blocking or erroring.
  - **During Sync Loop**: Geth logs a warning and posts block ranges directly to the API (`sendRangeDirectly`) without enforcing continuity checks, managing range memory, or triggering restarts.
- **Automatic Bootstrapping Handoff**: Once Geth successfully posts its first block range through the fallback path, the API automatically bootstraps its database state. For all subsequent queries, the handshake will return `ok == true` and Geth will seamlessly transition to enforcing full continuity.

### 1.6 Two-Layer Recovery Model (Geth side)

Geth recovers from API/cosmos lag through **two complementary layers**. They are not redundant — each covers a different failure mode and time scale:

- **Layer 1 — Origin re-seed (`receiptSyncOrigin`, cycle start)**: At the start of every ReceiptSync cycle, Geth queries the API's last confirmed block `C` and sets `origin = clamp(C, skeletonTail-1, latest-1)`, so it re-downloads from `C+1`. This is the §1.1 handshake alignment and is the **only** mechanism that recovers a gap **across a restart** (a fresh process re-derives `origin` from `C`). It also re-fetches the missing bodies/receipts from devp2p execution peers.
  - **`latest` is the beacon/skeleton head** (from `d.skeleton.Bounds()`), **not** the API's `block/latest` (`C`). They move independently, which is exactly why we keep the explicit `C` query for re-seeding.
  - **Cosmos-down behavior is FALLBACK, not pause** (Option B): if the API is unreachable at cycle start, `receiptSyncOrigin` returns `latest-1` (proceed as before) instead of erroring. This keeps the downloader from freezing and lets Layer 2 own in-process recovery.
- **Layer 2 — In-RAM range cache (`CheckContinuityAndSend`, send time)**: After a segment is downloaded, Geth resends shallow gaps directly from a rolling cache of the last *N* finalized ranges (default 4), without waiting for a devp2p re-download. This is a **fast path / optimization** over Layer 1 for the common case where the API lost its volatile `pendingBlocks` buffer on restart (so `C` regressed by ≤ *N* ranges).

### 1.7 Configurable In-RAM Cache Window

The number of finalized ranges Geth keeps in memory is operator-configurable via `--bridge.cacheranges` (Geth) / `BRIDGE_CACHE_RANGES` (wrapper), **default 4**.

- **Sizing rationale**: at ~12–20 min of synchronous processing per segment, 4 ranges ≈ 48–80 min of in-RAM recovery window, which exceeds the realistic API/cosmos lag. A gap wider than the window is treated as an emergency: deliberate `log.Crit` → supervisor restart → Layer 1 re-seed.
- **Off switch**: continuity is disabled by leaving `--bridge.getlastblock` **unset** (existing `IsContinuityConfigured() == false` fallback → legacy direct posting, no restart).
- **Minimum window**: when continuity *is* configured, `--bridge.cacheranges` must be `>= 1`. If a value `< 1` is supplied, clamp to `1` and `log.Warn`.

---

### 2.1 The POST Contract: Receipt-ACK, not Commit

- **Geth ignores the POST result for flow control.** `importBlockReceiptResults` logs a warning and **continues** on a failed `CheckContinuityAndSend`; it never retries the same segment synchronously. Gating the POST on a synchronous Cosmos commit buys nothing on Geth's side.
- **`sendRangeDirectly` stops at the first non-200.** A `500` on a block that was actually *received fine* leaves the **rest of that segment unsent**, manufacturing gaps in what the API ever sees.
- **Contract:** the POST response means **"block received into the queue"** (`200`), **"malformed input"** (`400`), or **"I cannot accept this block right now — back off"** (`503`, queue full). It **never** reports Cosmos commit status.

### 2.2 Confirm-Before-Advance (never skip; resend is safe)

- **`BridgeExchange` is fire-and-forget today.** It uses `BroadcastTxSync` (CheckTx only, not inclusion) and `broadcastMessage` **swallows `resp.Code != 0`**. A CheckTx rejection — or an accepted-but-never-included tx — can report **success**, letting the API advance `latest` over a receipt that never landed: a **silent skip**.
- **Protection is against *skipping*, not *resending*.** Re-sending an already-applied receipt is harmless (§2.3). The API advances `latest` **only after the receipt is confirmed recorded on-chain**, and may freely re-send on any uncertainty.
- **No pre-check.** Send optimistically, then **read the outcome**.

### 2.3 On-Chain Idempotency & Dedup-as-Success

- The chain enforces exactly-once *effect* (dedup by content; mint once per `BRIDGE_PENDING`).
- A re-submission by the same validator returns `"validator has already validated this transaction"`. The API treats **"our validator already recorded this receipt"** as **success**, not an error.

### 2.4 Index-Independent Confirmation (validators may run `tx_index = "null"`)

- Validators may run `tx_index = "null"`. `WaitForTx` / `SyncNoRetry` fail-fast under `null` and cannot distinguish on-chain success from failure — **rejected** as the confirmation authority.
- **Rule:** confirmation uses a **module-state read** via the **already-deployed** `BridgeTransaction` query (`GetBridgeTransactionsByReceipt`): match full receipt content, check **our validator** is in `Validators`. No chain upgrade required. An `O(1)` dedicated query is a future optimization gated on a chain upgrade.

### 2.5 Back-Pressure instead of Silent Drop

- A full per-chain buffer returns **`503`** so Geth's `sendRangeDirectly` stops and resends later. We never drop a block while reporting it received (`200`).

### 2.6 Retry Model (drain on next POST; no background worker)

- The drain runs **inside `AddBlock`** and always restarts at `latest+1`. There is **no** background ticker/worker.
- A block that fails to commit stays in the queue; retried on the **next POST**. With an L1 that keeps producing blocks, Geth keeps POSTing, so retries keep flowing.

### 2.7 Concurrency: No Network I/O Under the Data Lock

- The drain performs broadcast + confirmation I/O per receipt. The data lock (`sync.RWMutex`) is taken only for quick in-memory reads/writes; all network I/O happens with the lock released. Draining is serialized by a **separate** `drainMu` so `GET /v1/bridge/block/latest` stays responsive.

---

### 3.1 Geth-Side Downloader, Continuity & Cache Lifecycle

Zoom: **`CheckContinuityAndSend`** — handshake, gap math, in-RAM resend vs `log.Crit`, and *which* blocks to send. The `SendRange` / `SendFilter` / `SendDirect` / `ResendMemory` nodes all end in "POST block-by-block"; per-POST `200`/`503` behavior is expanded in §3.4.

```mermaid
flowchart TD
    %% Define Nodes
    Start([Segment of results downloaded]) --> FetchAddrs[Fetch bridge contract addresses once]
    FetchAddrs --> PreparePayloads[Prepare BlockRequest payloads for all blocks in segment]
    PreparePayloads --> QueryAPI[Query GET /v1/bridge/block/latest]

    %% Far Right Branch: Handshake/Transport Failure (Graceful Fallback)
    QueryAPI -- "Transient Transport Error OR Handshake Failure\n(Graceful Fallback)" --> SendDirect[Send segment block-by-block directly]
    SendDirect --> Done([Segment Sync Complete])

    %% Success Branch
    QueryAPI -- "Success: Got apiLatest" --> Compare{Compare apiLatest with expected blockX - 1}

    %% Left Branch: API is behind (Check memory or abort for restart)
    Compare -- "apiLatest < expected\n(API is behind)" --> CheckMemory{Is apiLatest + 1\nin memory?}

    CheckMemory -- "No: Deep Gap\n(Unrecoverable)" --> LogCrit[log.Crit: Gap exceeds Geth memory]
    LogCrit --> Exit([os.Exit 1: Abort & Restart Node])

    CheckMemory -- "Yes: Recoverable" --> ResendMemory[Resend missing blocks from memory]
    ResendMemory --> SendRange[Send segment block-by-block via POST]

    %% Middle Branch: Sequential / Normal Path
    Compare -- "apiLatest == expected\n(Sequential)" --> SendRange

    %% Right Branch: API is ahead (Filter & Send only if needed)
    Compare -- "apiLatest >= blockX\n(API is ahead)" --> FilterRange{Are there blocks > apiLatest?}
    FilterRange -- "No: API fully ahead" --> Done
    FilterRange -- "Yes: Partial overlap" --> SendFilter[Send remaining blocks > apiLatest]

    %% Memory Caching (Only for successful handshake paths that actually sent blocks)
    SendRange --> CacheRange[Save segment to recentRanges memory]
    SendFilter --> CacheFilter[Save sent blocks to recentRanges memory]
    CacheRange --> Done
    CacheFilter --> Done
```

### 3.2 Memory Caching vs. Node Restart

| Path / Outcome | Trigger Condition | Actions Taken | Purpose |
| --- | --- | --- | --- |
| **Save Segment to Memory** (`CacheRange` / `CacheFilter`) | A block range (or filtered partial range) is successfully sent to the API, and the API handshake was successful (sequential, recovered from memory, or partially ahead). | 1. Post the newly sent blocks to the API.<br>2. Append the sent blocks as a range to Geth's in-memory `recentRanges` slice.<br>3. Evict the oldest range if `len(recentRanges) > BridgeMaxCachedRanges` (default 4). | Rolling window of the last *N* finalized ranges so Geth can self-heal and resend blocks if the API briefly drops or lags. |
| **Supervisor Restart** (`log.Crit`) | The API is healthy and online, but its progress (`apiLatest`) lags behind Geth's expected block height by more than the *N* ranges stored in Geth's memory (or the cache does not contiguously cover the gap). | 1. Log a critical error (`log.Crit`) specifying the gap size (go-ethereum's `log.Crit` terminates the process). | Triggers the node supervisor to restart Geth. On startup, Geth queries the API's actual progress to re-bootstrap its skeleton downloader origin (Layer 1), correcting the gap. |

> [!IMPORTANT]
> **No Caching on Legacy Fallback or Full Skip**
> * **Graceful Fallback**: If the API is offline or returns a transport error, Geth posts the range directly using `SendDirect` and **bypasses** all memory caching.
> * **API Fully Ahead**: If the API is fully ahead of the current segment (`apiLatest >= endBlock`), Geth skips sending entirely and **bypasses** memory caching.

### 3.3 API-Side Commit Pipeline (accept + drain)

The POST does the **accept** phase and returns; the **drain** phase commits from `latest+1`, advancing `latest` only when every receipt of a block is confirmed.

```mermaid
flowchart TD
    POST([POST /admin/v1/bridge/block]) --> Accept[ACCEPT phase: acquire data lock]
    Accept --> Valid{Valid blockNumber & fields?}
    Valid -- No --> R400([Return 400: malformed])
    Valid -- Yes --> Dup{blockNumber <= latest?}
    Dup -- "Yes: duplicate" --> Ret200d([Return 200: received])
    Dup -- "No" --> Gap{blockNumber > latest + 1?}
    Gap -- "Yes: out of order" --> Cap{Buffer full for chain?}
    Cap -- "Yes" --> R503([Return 503: BACK-PRESSURE, never drop])
    Cap -- "No" --> Buf[Buffer block] --> Ret200([Return 200: received])
    Gap -- "No: == latest + 1" --> Buf2[Buffer block] --> Ret200

    Ret200 -. triggers .-> Drain
    Ret200d -. triggers .-> Drain

    subgraph Drain[DRAIN phase: drainMu held, data lock NOT held during I/O]
        D1{block latest+1 buffered?}
        D1 -- No --> Stop([Stop draining])
        D1 -- "Yes" --> Send[Broadcast each receipt: no pre-check]
        Send --> Conf[Confirm via existing BridgeTransaction query: our validator recorded? NOT WaitForTx]
        Conf -- "recorded / already-validated = success" --> AllR{All receipts confirmed?}
        Conf -- "not yet / real error" --> LogStop[Log only; leave block; latest unchanged; stop]
        AllR -- No --> Send
        AllR -- "Yes" --> Adv[Advance latest+1: SQLite + memory; delete block]
        Adv --> D1
    end
```

**Startup (Phase 1, unchanged from commit 1):** on API boot, `NewBlockQueue` loads `latestBlocks` from SQLite once (`LoadAllBridgeLatestBlocks`). Fresh chains buffer ≥ 6 blocks before bootstrapping `latest = minBlock - 1`. All subsequent duplicate/out-of-order checks use the in-memory map — no SQLite reads on the POST hot path.

### 3.4 Geth POST loop — detail inside §3.1's send nodes (`200` / `503`)

Not a second top-level Geth chart — a **zoom into `sendRangeDirectly`**, i.e. what happens inside every "Send segment block-by-block via POST" box in §3.1 (`SendRange`, `SendFilter`, `SendDirect`, `ResendMemory`). §3.1 answers *what* to send; this answers *how each POST is interpreted*.

Also adds cycle context §3.1 omits: Layer 1 re-seed, segment download, Layer 2 cache write, and `importBlockReceiptResults` swallow-on-continue.

Commit 2 change here only: POST means `200` = received into queue (keep sending the segment); stop the segment on `503` / other non-`200` during accept — not on Cosmos commit failure.

```mermaid
flowchart TD
    Cyc([ReceiptSync cycle start]) --> Seed[receiptSyncOrigin: re-seed origin from API C — Layer 1]
    Seed --> DL[Download bodies + receipts for segment]
    DL --> Chk[CheckContinuityAndSend]
    Chk --> Sendr[sendRangeDirectly: POST each block in order]
    Sendr --> Resp{POST status?}
    Resp -- "200: received" --> Nextb[Next block in segment]
    Resp -- "503 / non-200" --> StopSeg[Stop sending this segment]
    Nextb --> Sendr
    Nextb --> Cache[SaveRangeToMemory — Layer 2]
    StopSeg --> Swallow[importBlockReceiptResults: log warn, CONTINUE]
    Cache --> NextCyc([Next cycle])
    Swallow --> NextCyc
    NextCyc --> Cyc
```

### 3.5 Why decoupled POST + confirmed commit

1. **Receipt-ACK POST**: Geth swallows the response and `sendRangeDirectly` stops at the first non-200. Returning `200` on receipt lets Geth POST the **entire** segment. The queue holds the full run and the drain cascades once `latest+1` clears — eliminating the "missing middle blocks" gap a mid-segment `500` would create.
2. **Confirmed commit**: `latest` is the contract with Geth's handshake. Advancing it only after on-chain confirmation guarantees resuming from `latest+1` is always correct, with **no skipped receipt**. Combined with on-chain dedup, re-sends are safe and idempotent.
3. **Back-pressure, not drop**: the single place we say "no" to Geth (`503`) is when the buffer is full — exactly when we want Geth to pause and resend later, never losing a block.


0xMayoor:
hit a permanent stall in the drain while testing this. a 0-value transfer to a bridge address (`transfer(bridge, 0)`) ends up posted with `Amount="0"`, and `ValidateBasic` rejects amount=0 — so that block never confirms, `latest` never moves past it, and everything after just buffers. permissionless, costs gas.

reproduced the contract side and the drain side directly (0-balance burn → the erc20 log; and that log as `Amount="0"` → drain wedges, on both the reject and the 60s confirm-timeout). 
the geth relay in between forwards the amount unfiltered as far as i can see — didn't run your geth binary itself.

is amount=0 supposed to be filtered upstream, or should the drain skip a receipt that can't ever validate?

GLiberman:
> hit a permanent stall in the drain while testing this. a 0-value transfer to a bridge address (`transfer(bridge, 0)`) ends up posted with `Amount="0"`, and `ValidateBasic` rejects amount=0 — so that block never confirms, `latest` never moves past it, and everything after just buffers. permissionless, costs gas.
> 
> reproduced the contract side and the drain side directly (0-balance burn → the erc20 log; and that log as `Amount="0"` → drain wedges, on both the reject and the 60s confirm-timeout). the geth relay in between forwards the amount unfiltered as far as i can see — didn't run your geth binary itself.
> 
> is amount=0 supposed to be filtered upstream, or should the drain skip a receipt that can't ever validate?

Good catch, missed that ethereum let transaction to be passed with zero amount. Filter will be applied on geth node

x0152 (approved):
Reviewed the current flow for other permanent-stall paths, didn't find additional realistic cases. I think adding tests would be great

note: if a validator drops out of the active set, its own node can locally stall on a new tx until it's active again next epoch - not permanent, other validators unaffected

Issue #1222, opened 2026-05-21
[P1] Int overflow


tcharchian:
The goal of this is to have in place after this a standard way of handling possible overflows, have it implemented consistently across the entire codebase and to have a check (preferably a static check, an AI persona if necessary as a backup) that flags anything that doesn't use the established pattern

olegsuhoparov:
Opened a surgical first PR against main: #1379.

It ports the already-accepted #1100/#1101 overflow fixes to main and adds two small guards for payout uint64->int64 conversion and validation totalWeight accumulation.

I intentionally left broad static analysis and #1017 supply-cap semantics out of scope so this remains reviewable.

Mayveskii:
Hi @olegsuhoparov 

  Wanted to flag that #1017 (Bitcoin reward supply-cap overflow guards) has been aligned with the overflow-handling direction from this meta-issue.

  Changes on branch `fix/bitcoin-rewards-supply-cap-overflow` (commit `9e4ebf74e`):

  - Added a reusable `checkedAddUint64` helper in `inference-chain/x/inference/keeper/bitcoin_rewards.go`, placed next to the `checkedAddInt64` pattern introduced in #1379
  .
  - Replaced the inline `math.MaxUint64` overflow checks in the supply-cap reduction loop and in `CalculateParticipantBitcoinRewards` with the new helper.
  - Upgraded the overflow guard logs from `Warn` to `Error`.
  - Added explicit unit tests for the overflow path:
    - `TestCheckedAddUint64`
    - `TestGetBitcoinSettleAmounts_SupplyCapReductionNoWrap`

  Since #1379 explicitly left the #1017 supply-cap semantics out of scope, #1017 adopts the same helper/logging/test pattern independently rather than conflicting with #1379.

olegsuhoparov:
Thanks for aligning #1017 with the overflow-handling direction from #1379.

That split makes sense to me: #1379 intentionally stays focused on settlement/claim/validation overflow paths and leaves the Bitcoin supply-cap semantics to a separate PR.

One thing to keep in mind is merge order: #1379 changes the `CalculateParticipantBitcoinRewards` fallback from `MaxUint64` to `MaxInt64`, so if #1379 lands first, #1017 will likely need a small rebase/adaptation around the fallback-path rationale and tests.

[...]

PR #1379, opened 2026-06-30
fix(inference): standardize checked integer handling in settlement and validation paths

olegsuhoparov:
## Summary

This PR keeps the overflow scope intentionally surgical for #1222.

- port the accepted #1100 settlement handling to main:
  - use `bits.Add64` for `RewardCoins + WorkCoins`
  - skip settlement only when the sum is genuinely zero
  - cap the `bitcoin_rewards.go` producer fallback at `MaxInt64`, not `MaxUint64`
- port the accepted #1101 validation widening to main:
  - make `ShouldValidate` use `uint64` powers
  - remove `uint32` narrowing in chain/DAPI callers
  - keep negative-weight rejection via `safeUint64FromInt64`
- add checked payout conversion before payment calls:
  - reject `uint64` amounts that do not fit the `int64` payment API
- add checked `totalWeight` accumulation:
  - return arithmetic overflow as an explicit error, not as “epoch data not found”
- add focused boundary tests for the changed behavior

## Out of scope

- broad repo-wide integer overflow audit
- static analyzer / CI guard
- #1017 supply-cap loop and event semantics
- unrelated casts outside settlement, validation, and payout boundaries

## Tests

cd inference-chain && go test ./x/inference/types ./x/inference/calculations ./x/inference/keeper
cd decentralized-api && go test ./internal/validation ./internal/startup

Notes:
- `GetTotalCoins()` still saturates to `MaxInt64` because it is an aggregate/helper boundary and this matches #1100.
- Payout conversion does not saturate. If a stored `uint64` amount cannot fit the `int64` payment API, the claim returns an error before bank sends and the settle record remains for intervention/retry.
- `getEpochGroupWeightData` keeps `found` separate from `error`: missing sub-epoch data can still be skipped, but arithmetic overflow is propagated.
- Static checks are intentionally left for a follow-up PR so this one stays reviewable.
- If a stored uint64 payout amount does not fit into the int64 bank payment range, ClaimRewards returns before any bank send and the settle record remains for operator intervention; retry only helps after state correction or migration.

### DIFF
--- a/decentralized-api/internal/startup/reward_recovery.go
+++ b/decentralized-api/internal/startup/reward_recovery.go
@@ -8,6 +8,8 @@ import (
 	"decentralized-api/internal/seed"
 	"decentralized-api/internal/validation"
 	"decentralized-api/logging"
+	"math"
+	"math/bits"
 	"sync/atomic"
 	"time"
 
@@ -137,7 +139,10 @@ func (c *RewardRecoveryChecker) AutoRewardRecovery() {
 	}
 
 	settleAmount := settleAmountResp.SettleAmount
-	totalAmount := settleAmount.RewardCoins + settleAmount.WorkCoins
+	totalAmount, carry := bits.Add64(settleAmount.RewardCoins, settleAmount.WorkCoins, 0)
+	if carry != 0 {
+		totalAmount = math.MaxUint64
+	}
 	logging.Info("[AutoRewardRecovery] Found settle amount for participant", types.Claims,
 		"address", address,
 		"rewardCoins", settleAmount.RewardCoins,
@@ -146,7 +151,7 @@ func (c *RewardRecoveryChecker) AutoRewardRecovery() {
 		"epochIndex", settleAmount.EpochIndex)
 
 	// Check if we have unclaimed rewards (totalAmount > 0 indicates pending rewards)
-	if totalAmount <= 0 {
+	if totalAmount == 0 {
 		logging.Info("[AutoRewardRecovery] No unclaimed rewards found", types.Claims, "address", address, "totalAmount", totalAmount)
 		return
 	}
--- a/decentralized-api/internal/validation/inference_validation.go
+++ b/decentralized-api/internal/validation/inference_validation.go
@@ -107,9 +107,9 @@ func (s *InferenceValidator) shouldValidateInference(
 	shouldValidate, message := calculations.ShouldValidate(
 		seed,
 		inferenceDetails,
-		uint32(inferenceDetails.TotalPower),
-		uint32(validatorPower),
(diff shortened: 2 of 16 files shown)


olegsuhoparov:
Follow-up note: while tightening the overflow tests, we found an existing unchecked uint64 accumulation in bitcoin reward denominator calculation (`CalculateParticipantBitcoinRewards`: `totalFullWeight` / `totalPoCWeight`). I removed a test that accidentally exercised and normalized that wrap path.

This PR intentionally leaves that bitcoin reward accounting fix out of scope; it should be handled as a separate follow-up so #1379 remains focused on the settlement/claim/validation overflow paths.

redstartechno:
I checked this against `upgrade-v0.2.16` @ `c33c9eaa5`, where settlement changes are landing now. The PR conflicts there in 5 files, so I mapped what is still missing on the train. The targeted tests on this head (`Overflow|Atomicity|Safe|Extreme|SettleAmount|ShouldValidate` across `keeper`, `types`, `calculations`) all pass.

**Already on the train** (landed in #1267, "Upgrade v0.2.14", 2026-07-23, largely identical to this PR):
- `accountsettle.go`: `bits.Add64(WorkCoins, RewardCoins)` with the carry check in `SettleAccounts`
- `types/settle_amount.go`: `carry != 0 || sum > math.MaxInt64`
- `bitcoin_rewards.go`: the `MaxInt64` cap instead of the `^uint64(0)` sentinel, including the same "capping to MaxInt64" log line
- `calculations/should_validate.go`: powers as `uint64` with `decimal.NewFromUint64`
- `startup/reward_recovery.go`: the carry-guarded `totalAmount`
- `keeper/safecast.go` exists, with `safeUint64FromInt64` and friends

**Still only in this PR:**
- `msg_server_claim_rewards.go`: the train still converts with bare `int64(escrowPayment)` (line 90) and `int64(settleAmount.GetRewardCoins())` (line 110). This PR routes both through `safeInt64FromUint64`, which the train's `safecast.go` doesn't have yet. It also has the `safeUint64FromInt64` conversions of the model weights in the new `getEpochGroupWeightData`.
- The accompanying tests (`claim_rewards_weight_overflow_test.go`, the atomicity additions).

`RewardCoins` is already capped at `MaxInt64` on the bitcoin-reward path, and `WorkCoins` comes from earned coin amounts, so neither should exceed `MaxInt64` through normal settlement. (The `GetTotalCoins` cap bounds only their sum, not each operand of those two casts.) So what's left reads as defense in depth rather than a live bug. If it's still wanted, rebasing onto `upgrade-v0.2.16` and keeping only the `msg_server_claim_rewards.go` part (plus its tests) would make a small, conflict-free PR.

PR #1381, opened 2026-06-30
fix(gateway): propagate proxy request ID to runner for FinalCost accounting

nsvdev:
## Problem

`GET /devshard/{escrow}/v1/requests/{id}` (the FinalCost / request accounting endpoint) always returns `404 not found` for v2 gateway requests. This breaks request detail expansion in the OpenBroker dashboard — `final_cost_json` is never populated for any v2 request.

**Root cause:** `handleStreaming` and `handleNonStreaming` both call:
```go
p.redundancy.RunInference(context.Background(), ...)
```
A fresh `context.Background()` carries no request ID, so `RunInference` mints a **new** ID for the runner that differs from the proxy's ID placed in the `X-Request-Id` response header. Since `request_accounting` is indexed by the **runner's** ID, every FinalCost lookup using the proxy's ID returns 404 — permanently, regardless of how long you wait.

## Fix

Propagate **only the request ID** (not cancellation) from the proxy context into the detached background context:

```go
logging.PropagateRequestID(context.Background(), r.Context())
```

`logging.PropagateRequestID` copies just the request ID value. `WithRequestID` is already a no-op when an ID is present, so `RunInference` reuses the proxy's ID instead of creating a new one. The runner still executes on a detached context — client disconnect does not cancel it — preserving the existing behaviour.

## Changes

`devshard/cmd/devshardctl/proxy.go` — 2 call sites (`handleStreaming` + `handleNonStreaming`) + `devshard/logging` import.

## Verification

`go build ./cmd/devshardctl/...` passes. After deploying, `GET /v1/requests/{proxy_request_id}` returns the accounting record immediately for all fresh requests.

### DIFF
--- a/devshard/cmd/devshardctl/proxy.go
+++ b/devshard/cmd/devshardctl/proxy.go
@@ -16,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"devshard/logging"
 	"devshard/state"
 	"devshard/types"
 	"devshard/user"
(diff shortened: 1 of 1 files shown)


qdanik (approved):

a-kuprin:
@nsvdev I have retergeted your PR. Please resolve conflict

PR #1391, opened 2026-07-03
enforce inference tokens floor

qdanik:
## Summary

Floors `min_tokens`/`max_tokens` to **64** on every inference request — enforced identically on executor, validator, and gateway. Closes **H1 #3833622**: the validation similarity-floor passes arbitrarily short completions regardless of model, so a substitute model could clear validation (model substitution → reward extraction). ≥64 tokens gives the logprob distance metric enough signal to discriminate.

## What changes

- `common/completionapi`: `MinTokensFloor = 64` + `EnforceTokenBudgetFloor` — `max_tokens`/`max_completion_tokens` ≥64, `min_tokens` ≥64 clamped to `max_tokens`. Applied on both execution (`ModifyRequestBodyWithLogprobsMode`) and validation (`ExecuteValidation`).
- `devshard/cmd/devshardctl`: same floor in the request-filter pipeline (`applyTokenBudgetFloor`).
- `stop_token_ids` stripped everywhere: with `min_tokens > 0` always on, vLLM masks stop-token logits and an out-of-vocab stop id CUDA-asserts the node.

## Safety

Executor, validator, and gateway apply the same floor → validation stays deterministic. Change bounded to requests asking for <64 output tokens.


devshard/cmd/devshardctl/request_filters_config.go
```diff
@@ -22,6 +22,8 @@ const MaxRequestNestingDepth = 32
 
 // Per-parameter bounds wired into the catalog. Values match supported-params.md.
 const (
+	MinTokensFloor = 64
```
a-kuprin: Is it enough? Or better to limit by at least 100 tokens?

a-kuprin:
Anyway it is scheduled to 0.2.15, so it will need rebasing after v4, and there will no be decentralized-api path.

But what should we not to forget - is to limit also on devshardd host (as devshardctl can be malicious).

It should be released in following order:
1. Gateway with this limit
2. devshardd v* with same limit

And limit should be placed in commons.

p.s.:
@qdanik 
Is it working correctly and floor limit, realy limits it on vllm side?

decentralized-api/internal/validation/inference_validation.go
```diff
@@ -947,6 +947,7 @@ func (s *InferenceValidator) validateWithPayloads(inference types.Inference, inf
 	requestMap["stream"] = false
 	requestMap["skip_special_tokens"] = false
 	delete(requestMap, "stream_options")
+	completionapi.EnforceTokenBudgetFloor(requestMap)
```
a-kuprin: Maybe also we should check if output tokens < floorLimit to throw validation error immediately?

devshard/cmd/devshardctl/request_filters_config.go
qdanik: 64 is enough

a-kuprin:
@qdanik 

Non-streaming reduced-`max_tokens` retries can land below the floor and get rejected by the host, but that path is planned for removal on `gateway-v4`, so not worth fixing here.

But there still **must-fix issue**: `MsgStartInference` can reserve with `MaxTokens < MinTokensFloor`, then the host payload check fails and the inference stays pending until seal. Please reject `MaxTokens < completionapi.MinTokensFloor` in `applyStartInference` so under-floor reservations never enter state.

a-kuprin (approved):

PR #1394, opened 2026-07-04
fix: guardian tiebreaker weightless votes and zero-weight power cap

maria-mitina:
## Summary

This branch contains two related inference-chain fixes discovered while reproducing guardian tiebreaker behavior on gonka-testnet-4.

### 1. Guardian tiebreaker with weightless guardian votes (`3c07314b7`)

Guardian PoC tiebreaker votes were dropped by the per-model voting-power filter before `guardianProtection` ran. When a guardian could not run the model under validation (no per-model voting weight), their tiebreaker vote was ignored and split votes could not be resolved.

**Fix:** `guardianProtection` reads the raw validation list; the weight-filtered list still feeds majority math unchanged.

### 2. Power capping degenerates when zero-weight hosts are in settlement (`7f3c00e95`)

During a botched 4B→7B model transition on testnet, epoch 15 settlement produced:
- 3 join hosts with positive 7B consensus weight (~81k total)
- 1 guardian with `consensus=0` (mined 4B, sole eligible group was 7B)

`CalculateOptimalCap` counted the zero-weight guardian in the participant headcount (`N=4`). When the first positive weight triggered the 30% threshold with `sumPrev=0`, the formula produced `cap=0`, zeroing every host:

```
Universal power capping applied originalTotalPower=81702 cappedTotalPower=0
```

**Fix:** Only participants with `Weight > 0` participate in cap threshold math (`positiveCount` replaces headcount). If `cap <= 0` while `totalPower > 0`, skip capping as a safety net. Zero-weight participants still pass through unchanged in the apply step.

**Regression test:** `TestApplyPowerCappingForWeights_SkipsZeroWeightParticipants` (epoch-15 scenario).

## Context

- Zero-weight hosts remain in `activeParticipants` because they passed PoC on a non-eligible model; aggregation overwrites their weight to 0 but does not remove them.
- Delegation affects validation voting power, not `ActiveParticipant.Weight` used in power capping.
- Collateral was not involved (`grace_period_end_epoch=180`).

## Test plan

- [x] `go test ./x/inference/keeper/ -run TestApplyPowerCappingForWeights_SkipsZeroWeightParticipants`
- [x] `go test ./x/inference/module/ -run TestApplyPowerCapping`
- [ ] `go test ./x/inference/module/ -run Guardian` (tiebreaker tests from `3c07314b7`)
- [ ] Re-run guardian tiebreaker orchestration on testnet after weights recover

Made with [Cursor](https://cursor.com)


maria-mitina:
by mistake

PR #1396, opened 2026-07-04
Devshard pre-0.2.14 microrelease

a-kuprin:
## Summary

Three independent improvements for standalone `devshardd` and session lifecycle:

1. **Passive ML-node cache with dapi fallback** — inference keeps working when decentralized-api is down or upgrading.
2. **Per-escrow storage backend stickiness** — each session stays on the backend that created it; new escrows can use Postgres while legacy SQLite sessions drain in place.
3. **Frozen `validation_rate` at escrow create** — validation sampling rate is lane A (escrow-row constant), not live governance at bind.

---

### 1. `5de9f9c4a` — Added cached MLNodes for devshardd mlnode-manager

Standalone `devshardd` acquires ML nodes only via gRPC to dapi NodeManager. When dapi is unreachable (upgrade, crash, network), every inference failed even if ML nodes were still up.

This commit adds a **devshardd-only** passive cache and gRPC-first fallback. **dapi is unchanged** (backward compatible with current NodeManager RPCs).

Hot-path `Observe` of a known node uses atomic updates only (no write lock, no prune). Cache TTL defaults to 10m (`MLNODE_CACHE_TTL`).

**Cold start:** if dapi is already down when `devshardd` starts and no acquire has ever succeeded, the cache is empty and fallback cannot help until at least one successful acquire has been observed.

---

### 2. `89f48c917` — db backend is sticked per session (per escrow)

Previously `NewStorage` picked **one** backend for the whole process. Legacy SQLite sessions could pin the process to SQLite until every escrow drained and the process restarted, blocking new Postgres sessions.

This commit makes the backend **sticky per escrow**:

- `NewStorage` returns a per-session router (`HybridStorage`).
- **New** escrows: Postgres when `PGHOST` is set, otherwise SQLite.
- **Existing** escrows: always served by the backend that physically holds them (from each backend’s persistent index, not an ephemeral route table).
- A store can serve legacy SQLite escrows and new Postgres escrows at the same time; SQLite drains in place as sessions settle/prune while new work goes to Postgres.
- `.pg-bound` means “Postgres currently holds at least one session,” not “Postgres was ever used.” It is cleared when Postgres drains to zero so a fully pruned store can boot SQLite-only again without manual cleanup.
- No silent fallback: a Postgres outage fails new Postgres-destined creates; it does not create the same escrow in SQLite (avoids forked append logs).

Docs updated in `devshard/docs/storage-design.md`.

---

### 3. `c9e6cc450` — validation_rate now is passed with escrow start as a constant parameter

`validation_rate` is moved to **lane A**: snapshotted onto the escrow row at `MsgCreateDevshardEscrow` and loaded into `SessionConfig` from the escrow only.


x0152 (approved):

gmorgachev:
merged to 0.2.14 branch in https://github.com/gonka-ai/gonka/pull/1417, together with memory lead 
let's make devshard release directly from upgrade-v0.2.14 branch

PR #1399, opened 2026-07-04
perf(devshard): shrink gateway memory footprint and add profiling

libermans:
Reduce devshardctl gateway RSS by not holding inactive escrows in memory, bounding the chat response cache, and avoiding full inference-map deep copies on the common read paths. Adds pprof/memstats for diagnosis. Runtime lifecycle:
- Boot loads only devshards marked active; inactive ones are skipped instead of being hydrated (and re-queried on chain) at startup.
- Non-resident devshards are served on demand: read-only debug/status routes hydrate a transient local-SQLite-only runtime (no chain/host clients), and settlement rehydrates a transient full runtime; both are released right after use.
- All disable paths (admin deactivate, rotation-without-settle, settle) now retire the runtime from memory drain-safely, only after in-flight requests complete. State accessors (avoid copying the full inference map):
- Add StateMachine.Config, SnapshotStateNoInferences, SnapshotInferences, and InferenceStatusCounts; use them at status/config/single-inference call sites instead of SnapshotState. Endpoints:
- /v1/state is now summary-only; the full inference dump moves to a new admin/debug endpoint /v1/debug/inferences (no pagination).
- /v1/debug/state reports inference status counts; single-inference lookup uses GetInference.
- Add /v1/debug/memstats and register net/http/pprof under /debug/pprof/, both gated behind admin auth. Chat response cache:
- Fix unbounded growth: entries keyed by request-body hash were only expired lazily on lookup, so unique requests lived until restart. Add a periodic expiry sweep and a total-bytes cap (DEVSHARD_CHAT_CACHE_MAX_BYTES).

### DIFF
--- a/devshard/cmd/devshardctl/escrow_rotator.go
+++ b/devshard/cmd/devshardctl/escrow_rotator.go
@@ -409,9 +409,43 @@ func (g *Gateway) settleDevshardOnChain(ctx context.Context, id string, req admi
 		rt.active.Store(false)
 	}
 	g.mu.Unlock()
+	wasResident := ok
 	if !ok {
-		log.Printf("devshard_settle_failed escrow=%s stage=runtime_lookup error=%q", id, "devshard is not active")
-		return nil, fmt.Errorf("devshard %s is not active", id)
+		// Non-resident devshard (inactive/settled): rehydrate a full runtime
+		// (with chain access) solely to build and broadcast this settlement,
+		// then release it. Concurrent settlement of the same id is guarded by
+		// the caller: auto/reconcile paths hold settlementInFlight[id] via
+		// scheduleAutoSettlement, and the chain rejects any duplicate settle
+		// broadcast, so no additional in-flight guard is taken here (taking
+		// one would deadlock the auto path, which already holds it).
+		cfg, known, cfgErr := g.lazyRuntimeConfig(id)
+		if cfgErr != nil {
+			log.Printf("devshard_settle_failed escrow=%s stage=lazy_config error=%q", id, cfgErr.Error())
+			return nil, cfgErr
+		}
+		if !known {
+			log.Printf("devshard_settle_failed escrow=%s stage=runtime_lookup error=%q", id, "devshard is not active")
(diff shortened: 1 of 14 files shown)


libermans:
@qdanik somehow I can't add you to the reviewer list

qdanik:
@libermans there is only one minor issue, not a blocker.

`handleAdminSettleDevshard` doesn't clear `SettlementPending` on success (only `scheduleAutoSettlement `does). Since `reconcilePendingSettlements` re-queues `!Active && SettlementPending` devshards on boot, an escrow settled via the admin endpoint while that flag is set gets re-broadcast (chain-rejected) every restart and never self-heals. Pre-existing gap, worth a one-line clear on the success path.

qdanik (approved):

devshard/cmd/devshardctl/escrow_rotator.go
```diff
@@ -463,5 +497,12 @@ func (g *Gateway) settleDevshardOnChain(ctx context.Context, id string, req admi
 		return nil, err
 	}
 	log.Printf("devshard_settle_submitted escrow=%s tx_hash=%s settler=%s", id, result.TxHash, result.Settler)
+	// A settled escrow is terminal: drop the resident runtime so its memory
+	// (state machine, inference map, SQLite handles) is released now rather
+	// than lingering until the next restart. Transient runtimes are closed by
+	// their own deferred cleanup above.
+	if wasResident {
+		g.retireRuntime(id, "settled")
+	}
```
a-kuprin, line 507: As @qdanik mentioned `handleAdminSettleDevshard` doesn't clear `SettlementPending` flag

It would be best to centralize it here on successful `settleDevshardOnChain` before return

```go
g.clearSettlementPending(id) \\add it

if wasResident {
  g.retireRuntime(id, "settled")
}
```

a-kuprin (approved):

a-kuprin:
@qdanik 
> handleAdminSettleDevshard doesn't clear SettlementPending on success 

Fixed

PR #1402, opened 2026-07-05
# perf(devshard): port escrow runtime lifecycle + gateway memory optimizations from dl/gateway-v2-mem

libermans:
## Summary

Cherry-picks three commits from `dl/gateway-v2-mem` (#1399) to reduce devshardctl gateway RSS, adapted where this branch's architecture diverges:

- **`feat(devshard): settle depleted escrow after in-flight requests drain`** (#1314, clean pick) — depleted escrows are marked settlement-pending and settled once the last active request drains; pending settlements are reconciled on restart.
- **`fix(devshard): retire runtime escrow`** (#1349, clean pick) — deactivation paths drop the runtime from memory drain-safely, releasing its state machine, inference map, and SQLite handles.
- **`perf(devshard): shrink gateway memory footprint and add profiling`** (from #1399, conflicts resolved) — boot loads only active devshards; non-resident escrows are served via transient rehydration (read-only for debug/status, full for settlement); `/v1/state` becomes a bounded summary with the inference dump moved to `/v1/debug/inferences`; adds `/v1/debug/memstats` + admin-gated pprof; fixes the unbounded chat response cache (periodic expiry sweep + total-bytes cap, `DEVSHARD_CHAT_CACHE_MAX_BYTES`).

## Adaptations for this branch

The memory commit predates this branch's sealed-inference architecture (live records are evicted to SQLite on seal), so it was not applied verbatim:

- `SnapshotStateNoInferences` referenced `RevealedSeeds`, a field that doesn't exist in this branch's `EscrowState` — auto-merged cleanly but wouldn't compile; removed. Its shallow copy shares `SealedAcc`, which is safe (only ever replaced wholesale, never mutated).
- `/v1/debug/inferences` uses sealed-aware `ExportAllInferenceRecords` instead of the PR's live-only `SnapshotInferences`, so sealed records still appear in the dump.
- `/v1/debug/state` keeps this branch's phase field and live/sealed count split while adopting the new no-deep-copy accessors (`InferenceStatusCounts`, `LatestNonce`, `Balance`).
- Dropped the PR's `handleInference` (`X-Inference-Id`) handler — this branch removed that endpoint and nothing routes to it.
- Test adaptations: `session_close_test.go` and `flush_snapshot_test.go` updated to this branch's `storage.Storage` interface and test helpers (`newTestStateMachine`, `RuntimeTestVersion`); `TestHandleState_IncludesSealedInferences` now asserts sealed records via `/v1/debug/inferences` and that `/v1/state` stays bounded.

## Breaking change

`/v1/state` no longer includes the `inferences` map — consumers must read `/v1/debug/inferences` (admin/debug, unpaginated).

## Not included

- `b87e51b` (suspicious-host tracking) — this branch already carries a diverged version of that work.
- The four intermediate commits (MiniMax route, param clamping) — independent, no dependency from the picks (verified: only shared file is `escrow_rotator.go`, disjoint functions).
- Known pre-existing gap flagged in #1399 review: `handleAdminSettleDevshard` doesn't clear `SettlementPending` on success, so an admin-settled escrow is re-broadcast on every restart. Worth a follow-up one-liner.

## Testing

- Full `go test ./...` passes (all packages, including the picks' new tests: `gateway_inactive_access_test.go`, `flush_snapshot_test.go`, `session_close_test.go`).
- `go vet ./...` clean.

### DIFF
--- a/devshard/cmd/devshardctl/gateway_store.go
+++ b/devshard/cmd/devshardctl/gateway_store.go
@@ -1089,12 +1089,16 @@ func (s *GatewayStore) UpsertDevshard(devshard GatewayDevshardState) error {
 
 func (s *GatewayStore) upsertDevshardTx(tx *sql.Tx, devshard GatewayDevshardState, now string) error {
 	createdAt := now
-	_ = tx.QueryRow(`SELECT created_at FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&createdAt)
+	if err := tx.QueryRow(`SELECT created_at FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&createdAt); err != nil && err != sql.ErrNoRows {
+		return fmt.Errorf("lookup created_at for devshard %s: %w", devshard.ID, err)
+	}
 	// Preserve the existing settlement_pending marker so an unrelated upsert
 	// never silently clears a queued settlement; a brand-new row falls back
 	// to the value carried on devshard.
 	settlementPending := gatewayBoolToInt(devshard.SettlementPending)
-	_ = tx.QueryRow(`SELECT settlement_pending FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&settlementPending)
+	if err := tx.QueryRow(`SELECT settlement_pending FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&settlementPending); err != nil && err != sql.ErrNoRows {
+		return fmt.Errorf("lookup settlement_pending for devshard %s: %w", devshard.ID, err)
+	}
 	if _, err := tx.Exec(`
 		INSERT OR REPLACE INTO gateway_devshards (
 			id, private_key_hex, private_key_env, model, storage_path, active, created_at, updated_at, protocol_version,
--- a/devshard/cmd/devshardctl/gateway_store.go
+++ b/devshard/cmd/devshardctl/gateway_store.go
@@ -1118,6 +1118,44 @@ func (s *GatewayStore) upsertDevshardTx(tx *sql.Tx, devshard GatewayDevshardStat
 	return nil
 }
 
+// GetDevshard returns the registry record for a single devshard. The second
+// return value is false when no row exists for the id. It is used by lazy
+// hydration to look up the config of a non-resident devshard without loading
+// the entire registry.
+func (s *GatewayStore) GetDevshard(id string) (GatewayDevshardState, bool, error) {
+	id = strings.TrimSpace(id)
+	var devshard GatewayDevshardState
+	var active int
+	var settlementPending int
+	err := s.db.QueryRow(`
+		SELECT id, private_key_hex, private_key_env, model, storage_path, active, created_at, updated_at, protocol_version,
+		       rotation_role, rotation_epoch, settlement_pending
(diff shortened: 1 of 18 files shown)


a-kuprin:
PR mentions https://github.com/gonka-ai/gonka/pull/1399
Let's close this known gap for `handleAdminSettleDevshard`
Suggestion here: https://github.com/gonka-ai/gonka/pull/1399/changes#r3533876458

devshard/state/machine.go
```diff
@@ -494,13 +494,72 @@ func (sm *StateMachine) Balance() uint64 {
...
+	s.WarmKeys = make(map[uint32]string, len(src.WarmKeys))
+	maps.Copy(s.WarmKeys, src.WarmKeys)
+
+	return s
+}
+
+// SnapshotInferences returns a deep copy of just the inference map. Use it for
+// endpoints that render the full inference list without paying to copy the
+// rest of the escrow state.
+func (sm *StateMachine) SnapshotInferences() map[uint64]*types.InferenceRecord {
```
a-kuprin: Let's remove it as there is no callers, and the idea of the PR was to remove inference snapshotting

devshard/cmd/devshardctl/gateway_store.go
```diff
@@ -1081,11 +1090,16 @@ func (s *GatewayStore) UpsertDevshard(devshard GatewayDevshardState) error {
 func (s *GatewayStore) upsertDevshardTx(tx *sql.Tx, devshard GatewayDevshardState, now string) error {
 	createdAt := now
 	_ = tx.QueryRow(`SELECT created_at FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&createdAt)
+	// Preserve the existing settlement_pending marker so an unrelated upsert
+	// never silently clears a queued settlement; a brand-new row falls back
+	// to the value carried on devshard.
+	settlementPending := gatewayBoolToInt(devshard.SettlementPending)
+	_ = tx.QueryRow(`SELECT settlement_pending FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&settlementPending)
```
a-kuprin: Here we silent sqlite errors. It's because we want silent error when there is no rows.

But it's better to silent only `sql.ErrNoRows`:

```go
if err := tx.QueryRow(`SELECT created_at FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&createdAt); err != nil && err != sql.ErrNoRows {
    return fmt.Errorf("lookup created_at for devshard %s: %w", devshard.ID, err)
}

settlementPending := gatewayBoolToInt(devshard.SettlementPending)
if err := tx.QueryRow(`SELECT settlement_pending FROM gateway_devshards WHERE id = ?`, devshard.ID).Scan(&settlementPending); err != nil && err != sql.ErrNoRows {
    return fmt.Errorf("lookup settlement_pending for devshard %s: %w", devshard.ID, err)
}
```

a-kuprin:
> Known pre-existing gap flagged in https://github.com/gonka-ai/gonka/pull/1399 review: handleAdminSettleDevshard doesn't clear SettlementPending on success, so an admin-settled escrow is re-broadcast on every restart. Worth a follow-up one-liner.

Now it is included

a-kuprin (approved):

PR #1403, opened 2026-07-05
count background race finalizers in the drain barrier

qdanik:
### Problem
A speculative race returns to the client as soon as the winner finishes and hands its still-pending losers to a background finalizer. That finalizer runs on `context.Background()` for up to `SecondaryWaitAfterWinner` (5 min): it refunds reserved cost (`HandleTimeout` → `Balance +=`) and persists loser signatures (`ProcessResponse` → `AddSignature`). It holds no `activeRequests` slot.

The drain barrier that decides when an escrow is "quiet" counts only `activeRequests`. So the instant the winning handler returns and that counter hits zero, drain-driven auto-settle (#1314) and retire (#1349) fire — while the finalizer for that same request is still running. Two failures follow:

- **Wrong settlement amount.** Settlement reads the balance snapshot while a refund is mid-flight, so a terminal on-chain settle goes out for the wrong amount. `sm.mu` keeps each read/write atomic but does not order them.
- **Lost persistence.** Retire closes the per-runtime SQLite store while the finalizer is still writing, dropping loser signatures (or leaving a reopened zombie handle).

It needs a specific interleaving (the last foreground request being a race with pending losers, coinciding with depletion/rotation/retire), so it is rare — but it is a terminal settlement on a wrong amount plus silent durability loss on a production lifecycle path, and the retire/close tests cannot catch it: they set `activeRequests` by hand and never run a real finalizer.

### Fix
Track finalizers in a separate `pendingFinalizers` counter on the runtime; the "quiet" barrier now waits for both counters before it settles and before it closes the store.

- `goTrackedFinalizer` wraps every background finalizer and increments the counter synchronously before the goroutine spawns, so a winning handler that returns immediately after can never observe the runtime as quiet mid-finalize.
- `releaseRuntime` and `releaseFinalizer` each re-check the other counter after decrementing; whichever reaches zero last fires the deferred settle/retire, and dedup makes a double-fire harmless.
- Every settle / retire / finalize gate now goes through `hasPendingWork()`.

### Tests
`gateway_finalizer_barrier_test.go`: settle defers past a pending finalizer, retire/close defers past a pending finalizer, the synchronous-start contract, and exactly-once settle on a concurrent two-counter drain. All four fail before the fix. Green under `-race` (`-count=20`); full package green without `-race`.

The `TestRunInference_* -race` failures are the pre-existing `inf.receiptTime` race, unrelated to this change.


a-kuprin (approved):
I don't see logical issues in the PR.

But I'd consider to make a more clear naming, now it becomes too complicated to understand purpose of variables for further code reading.

**Consider rename:**

`activeRequests` →  `activeUserRequests`

`pendingFinalizers` — collides with escrow finalize, so also should be renamed
**Gateway:** `pendingRaceCleanup`
**Helper:** `goTrackedRaceCleanup`
**Gate:** `escrowHasBackgroundWork()`

a-kuprin (approved):

PR #1409, opened 2026-07-06
fix(devshard): bound the gateway chat response cache

ouicate:
### Problem

The gateway keeps a single process-wide `chatResponseCache` (`devshard/cmd/devshardctl/response_cache.go`) that dedups chat completions, keyed by a SHA-256 of `model + request body` with a 1h TTL. Expired entries are only ever reclaimed lazily inside `Get` — i.e. when the *same key* is looked up again.

Real chat traffic has near-unique request bodies (different messages/ids), so a given key is essentially never queried a second time. Its expired entry is therefore never revisited and never deleted, and there is no size cap. The map grows by one full cached response body per distinct request and never shrinks: an unbounded memory leak that scales with total traffic × uptime, independent of how many entries are actually live.

### Fix

Cap the cache at `defaultChatResponseCacheMaxEntries` (2048) and evict when a new key would exceed it. Eviction runs under the existing lock in two phases:

1. Reclaim every expired entry (the common case for a unique-body workload — usually sufficient on its own).
2. Only if still at capacity with live entries, drop the oldest half in a single pass (ordered by `ExpiresAt`, which is insertion time + constant TTL, i.e. insertion order).

Dropping half amortizes the O(n) sweep across ~maxEntries/2 inserts instead of paying it on every `Set` once hot, and always keeps the newest entries so a just-cached response survives an immediate retry. Overwriting an existing key
skips eviction (the count doesn't grow). This matches the bounded-map idiom already used by the rate limiter and gossip dedup.

### Tests

`response_cache_test.go`: set/get round-trip with body-copy isolation, expired entry dropped on access, hard bound under 500 unique inserts (never exceeds max, newest survives), expired-reclaimed-before-live eviction priority, and
overwrite-does-not-evict.

### DIFF
--- a/devshard/cmd/devshardctl/response_cache.go
+++ b/devshard/cmd/devshardctl/response_cache.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"io"
 	"net/http"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


qdanik:
@ouicate similar fix was provided in PR #1402. 

The byte cap in #1402 targets actual RSS more directly and is already reviewed, so I'd suggest consolidating on #1402 and closing this as superseded.

PR #1410, opened 2026-07-06
feat(devshard): expose gateway chat cache size metric

ouicate:
### Problem

The gateway chat cache exports no metrics, so operators can't see how many responses are cached or how occupancy tracks traffic. Every other gateway subsystem (limiter, capacity, participant limiter, host connections) already exposes gauges via `gatewayMetricsCollector`; the cache was the only unobservable one.

### Fix

Add a `chatResponseCache.Len()` accessor and a new gauge `devshard_gateway_chat_cache_entries` to `gatewayMetricsCollector` — declared in `Describe`, emitted in `Collect` when `gateway.chatCache != nil`. No labels
(single process-wide cache). Purely additive, read-only, reuses the existing collector wiring.

### Tests

`TestGatewayMetricsCollectorReportsChatCacheEntries` seeds two entries and asserts the gauge reads 2 via the registry, following the existing `requireMetricGaugeValue` pattern. 

### Note

Companion to #1409 (which bounds the same cache). Independent of it: this PR only adds the accessor + gauge, so it applies cleanly on its own, and #1409 keeps the reported count bounded.

### DIFF
--- a/devshard/cmd/devshardctl/gateway_test.go
+++ b/devshard/cmd/devshardctl/gateway_test.go
@@ -2696,6 +2696,21 @@ func TestGatewayMetricsCollectorIncludesHostConnectionSnapshots(t *testing.T) {
 	requireMetricGaugeValue(t, families, "devshard_host_transport_connections", map[string]string{"address": "10.1.2.3", "state": "hold_after_close"}, 4)
 }
 
+func TestGatewayMetricsCollectorReportsChatCacheEntries(t *testing.T) {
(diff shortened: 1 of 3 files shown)


qdanik:
@ouicate Please, take a look into PR #1402 and adjust your changes based on changes in #1402. 

One note: #1402 adds a Stats() accessor returning (entryCount, totalBytes) that overlaps this PR's Len(). Since #1402 bounds the cache by BYTES, if it lands first, rebasing onto Stats() and also exposing a bytes gauge (devshard_gateway_chat_cache_bytes) would match what actually caps the cache.

PR #1415, opened 2026-07-07
fix(poc): retry early-proof once, then fail closed on unavailability

maksimenkoff:
The early-share guard's prefix-proof check silently passed a participant whenever their own published early rootHash could not be proven, because any non-typed error was treated as a transient network hiccup and the guard failed open. This is exploitable: an attacker who publishes a random rootHash for their early commit (with no backing MMR data) will always return HTTP 400 "snapshot not found" for the shared-leaf query. That response is not one of the three typed proof errors, so isPermanentProofError returns false and checkPrefix returns (true, "") — the participant clears the prefix check without ever proving they held the data they committed to. A participant cannot reasonably have a persistent transient failure against their own published commitment: they either kept the data (and can prove it) or they did not (and cannot). The one legitimate case for a single-shot failure is a genuine network flake, which a retry covers.

Changes:
- checkPrefix now retries the early-proof fetch once on non-permanent errors, then treats any remaining error as a hard failure and votes no with reason "early_proof_unavailable".
- Permanent proof errors continue to fail immediately without retry.
- Final-proof handling is unchanged: transient failures there still fail open, since the main validation path already vets the final commitment.

Tests:
- Replaced the "transient error does not fail participant" case with two new cases: persistent transient error across both attempts -> vote no, asserting exactly 3 fetcher calls (final + early + 1 retry); transient error on first attempt, success on retry -> no vote, also asserting 3 fetcher calls.
- Added flakyProofFetcher, a per-root call-counted fake that fails the first early-root call with a configured transient error and succeeds afterwards.
- Existing permanent-error, matching-prefix, vector-mismatch, and low-share cases are unchanged and still pass.

### DIFF
--- a/decentralized-api/poc/early_guard.go
+++ b/decentralized-api/poc/early_guard.go
@@ -375,22 +375,23 @@ func (g *EarlyShareGuard) checkPrefix(ctx context.Context, pf proofFetcher, stag
 		return true, ""
 	}
 
-	earlyProofs, err := pf.FetchAndVerifyProofs(ctx, work.url, ProofRequest{
+	// One retry on transient failures so a genuine hiccup doesn't fail the participant.
+	earlyReq := ProofRequest{
 		PocStageStartBlockHeight: stage,
 		ModelId:                  work.modelId,
 		RootHash:                 dec.earlyRoot,
 		Count:                    dec.earlyCount,
 		LeafIndices:              []uint32{sharedLeaf},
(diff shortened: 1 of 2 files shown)


gmorgachev:
thanks for bug fix! @maksimenkoff 
after discussion, same issue resolved in https://github.com/gonka-ai/gonka/pull/1416

PR #1419, opened 2026-07-07
devshard: gateway v2 cleanup (escrow rotation review fixes)

a-kuprin:
Small follow-up to escrow-rotation recovery on `dl/gateway-v2`.

## Commits

- **use structured HTTP status for chain tx 404 detection** — replace substring `isNotFoundError` with typed `chainHTTPError` + status code check.
- **make escrow DB retries respect context cancellation** — `withDBRetry` exits promptly on `ctx.Done()` instead of sleeping through shutdown.
- **store escrow commitment CreatedAt as time.Time** — drop string parse/format in the store and rotator.
- **centralize escrow rotation PrivateKeyEnv trimming** — one helper instead of duplicated trim logic.
- **omit zero validation fields from settlement JSON** — `omitempty` for unset validation rate; drop incorrect test expectation that `0` disables validation.

## Test plan

- [ ] `go test ./cmd/devshardctl -run 'TestRESTChainTxClient_GetTxEscrowID|TestReconcileCommitments|TestCreateRotationEscrow'`
- [ ] `DevshardTests.devshard gateway auto-seals inferences after grace timeout`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/devshard/cmd/devshardctl/chain_tx_rest.go
+++ b/devshard/cmd/devshardctl/chain_tx_rest.go
@@ -369,6 +369,28 @@ func (c *RESTChainTxClient) getJSON(ctx context.Context, path string, out any) e
 	return c.getJSONFromBaseURL(ctx, c.baseURL, path, out)
 }
 
+// chainHTTPError is returned by getJSONFromBaseURL on non-200 responses.
+type chainHTTPError struct {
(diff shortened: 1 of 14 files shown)


a-kuprin:
duplicate

PR #1421, opened 2026-07-07
fix(mlnode): scheduler-heartbeat liveness for vLLM runner + damped proxy health probe

baychak:
## Problem
vLLM `/health` is the wrong liveness signal in both directions, and mlnode relies on it twice:

1. **Real hang, `/health` 200.** A TP worker deadlocks on the shm collective: scheduler frozen, zero throughput — but `AsyncLLM.check_health` only reads the `errored` flag. The node silently serves nothing until someone notices (5–22 min in production).
2. **Busy engine, `/health` timeout.** A saturated engine misses the 2 s deadline: `runner.is_available()` reports dead → a healthy node gets restarted (observed 8+/day); the proxy prober flips on a single miss → the compatibility server stops, dropping in-flight relays (23 flaps/2.3 h).

## Fix
- **`runner.py`**: liveness = `vllm:iteration_tokens_total_count` heartbeat — advances on every engine step, freezes only on a stall. Per-instance; re-baselines after engine restarts; aggregates multi-registry scrapes; a failed scrape can't pass for idle; `ConnectTimeout` = saturation (grace path), refused = dead. `MLNODE_HANG_GRACE_SEC` (default 120, `0` disables).
- **`proxy.py`**: prober timeout 2 → 10 s; unhealthy only after 3 consecutive failures, instant recovery.

## Validation
21 unit tests. 4×B300 Kimi-K2.6 fleet, several days: false restarts → **0** (incl. a full hour at max_num_seqs saturation), real stalls detected in ~2 min and self-healed via the existing restart path, proxy flaps → 0. Detection layer only — the underlying vLLM stall is addressed in gonka-ai/vllm#58.

### DIFF
--- a/mlnode/packages/api/src/api/inference/vllm/runner.py
+++ b/mlnode/packages/api/src/api/inference/vllm/runner.py
@@ -1,4 +1,5 @@
 import os
+import re
 import subprocess
 import time
 import requests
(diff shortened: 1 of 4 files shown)


baychak:
Superseded by #1531, which carries this change cherry-picked onto `main` together with its counterpart and the RC image pin — collected there at @vbgd0's request so the release branch gets everything in one place. Review history stays here.

PR #1423, opened 2026-07-07
Bound gateway runtime-build fan-out to avoid LCD 429 crash-loop

qdanik:
## Summary

- Fixes a startup crash-loop: on a node with many devshards, `buildGatewayRuntimes` built every runtime in parallel with no concurrency limit, so the simultaneous chain LCD/REST queries could trip a `429` — which is not a soft error and fails startup, so the container restarts and repeats the same fan-out.
- Bounds the runtime-build fan-out with a semaphore. The limit is `DEVSHARD_MAX_CONCURRENT_RUNTIME_BUILDS` (default **16**, must be ≥ 1); an invalid value logs and falls back to the default.
- Sizes the shared bridge HTTP client's `MaxIdleConnsPerHost` to the same limit so the bounded fan-out reuses keep-alive connections instead of churning them (Go's default is only 2).
- Per-index result collection and the existing error handling (soft-skip `ErrEscrowNotFound` / missing private key, `firstFatal` otherwise) are unchanged.

## Why

`buildGatewayRuntimes` launched one goroutine per devshard into a buffered channel with no limiter, so every escrow's builder hit the chain LCD at once. Each build is itself a sequential chain of ~`3 + 2·group_size` LCD calls (`GetEscrow`, `GetSessionBindParams`, `GetValidationThreshold`, and `GetHostInfo` + `VerifyWarmKey` per group slot), so a node with N escrows issues an N-wide burst against the shared `node:1317` LCD. A `429` from that burst is not one of the soft errors the loop tolerates, so it becomes `firstFatal`, startup returns an error, the container exits, restarts, and re-runs the identical fan-out — a crash loop.

The real ceiling is the shared LCD's capacity, which is a deployment property (it also serves validation/PoC and the api container), not a constant derivable in code — the repo pins no rate limit for `node:1317`. So the limit is operator-overridable, with a conservative default of 16: a small keep-alive connection pool to a shared internal service, large enough to hide startup latency (32 escrows → 4 batches → seconds, not minutes) and small enough to stay a minor slice of the LCD. The bridge previously used Go's default transport (`MaxIdleConnsPerHost = 2`), so beyond ~2 concurrent builds each extra request churned a fresh TCP connection; pinning the pool to the limit removes that churn.

## Tests

- `TestBuildGatewayRuntimesBoundsBuilderConcurrency` — peak simultaneous builder invocations ≤ the resolved limit (fails at N with the old unbounded fan-out).
- `TestBuildGatewayRuntimesBoundedFanoutSurvivesRateLimitingLCD` — an LCD that 429s above the limit no longer fails startup; fails if the semaphore is removed.
- `TestBuildGatewayRuntimesHonorsConcurrencyEnvOverride` — `DEVSHARD_MAX_CONCURRENT_RUNTIME_BUILDS=4` bounds concurrency to 4; fails if the env value is not wired into the semaphore.
- `TestResolveMaxConcurrentRuntimeBuildsDefaultsWhenUnset` / `TestResolveMaxConcurrentRuntimeBuildsParsesOverride` — default when unset; parses a valid override; falls back on `0` / negative / non-numeric.
- `TestBuildRuntimeBridgeClientPinsIdleConnPool` — the bridge client's `MaxIdleConnsPerHost` / `MaxIdleConns` equal the limit and the 10s timeout is preserved.

`go test ./cmd/devshardctl/` is green; the new tests are also green under `-race`. The package's broad `-race` run additionally flags pre-existing `TestRunInference_*` / `TestWaitForFirstToken*` data races that are unrelated to this change.

## Test plan

- [x] `gofmt` / `go vet` clean on the changed files
- [x] `go test ./cmd/devshardctl/ -run 'TestBuildGatewayRuntimes|TestResolveMaxConcurrent|TestBuildRuntimeBridgeClient' -count=1` green
- [x] New tests green under `-race`
- [x] Each new guard verified to fail when its logic is removed (unbounded fan-out; ignored env override)
- [ ] `/run-integration` (Testermint) green

### DIFF
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -305,8 +305,50 @@ func buildRuntime(cfg RuntimeConfig, chainREST, defaultModel string, perf *PerfT
 	return rt, nil
 }
 
+// defaultMaxConcurrentRuntimeBuilds caps parallel runtime builds at startup: a
+// small keep-alive pool against the shared node LCD, big enough to hide startup
+// latency, small enough to avoid a 429 storm. Overridable per deployment.
+const defaultMaxConcurrentRuntimeBuilds = 16
+
+// resolveMaxConcurrentRuntimeBuilds is the startup build fan-out limit,
+// overridable via DEVSHARD_MAX_CONCURRENT_RUNTIME_BUILDS (must be >= 1).
+func resolveMaxConcurrentRuntimeBuilds() int {
+	raw := strings.TrimSpace(os.Getenv("DEVSHARD_MAX_CONCURRENT_RUNTIME_BUILDS"))
+	if raw == "" {
+		return defaultMaxConcurrentRuntimeBuilds
+	}
+	n, err := strconv.Atoi(raw)
+	if err != nil || n < 1 {
+		log.Printf("invalid DEVSHARD_MAX_CONCURRENT_RUNTIME_BUILDS=%q (must be >= 1), using default %d", raw, defaultMaxConcurrentRuntimeBuilds)
+		return defaultMaxConcurrentRuntimeBuilds
+	}
+	return n
+}
+
+// buildRuntimeBridgeClient returns the HTTP client for chain-LCD queries, sized
+// so the bounded build fan-out reuses keep-alive connections instead of churning
+// them (the default MaxIdleConnsPerHost is only 2).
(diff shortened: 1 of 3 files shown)


a-kuprin (approved):

gmorgachev:
I'd expect no 429 when direct connection to node:1317, without going thought nginx

PR #1432, opened 2026-07-09
perf(poc/artifacts): SMST deferred hashing + copy-on-write historical proofs

kAIPraxisBot:
Two independent SMST improvements to the PoC artifact store, each a separate commit. Both keep committed roots and proofs **byte-identical** — no consensus impact.

1. **Deferred hashing** — ~1.6–3.4× faster ingest.
2. **Copy-on-write historical proofs** — O(N) rebuild → O(depth), and removes the rebuild-DoS limiter.

---

# 1. Deferred hashing (commit 1)

`insertAt` re-hashed each node on the root→leaf path on every insert, so a shared upper node was re-hashed once for every insert that passed through it. This defers hashing — insert records structure and counts only (`node.hash = nil`), and hashes are filled lazily, once, when the root is needed (`GetRoot` / snapshot / rebuild) via `ensureHashed`. Same tree, same Merkle function — only the order in which nodes are hashed changes.

**Byte-identical:** `TestDeferredHashFingerprint` folds all flush roots plus sampled proofs over a 200k-insert workload (depth expansion + negative nonces) into one SHA-256 fingerprint. Identical on the base and this PR:

```
base   (per-insert): 1caeb0ad4ac558b705c3c0bf60a6fa56e25f2392195e485a82a4e82ae7a68ea4
this PR (deferred):  1caeb0ad4ac558b705c3c0bf60a6fa56e25f2392195e485a82a4e82ae7a68ea4
```

**Ingest** (single build, stride-100 monotonic nonces = worst porosity; `TestSMSTBuildProfile`). RAM is identical before/after — deferral changes *when* nodes are hashed, not the node structure:

| N | ingest base | ingest deferred | speedup | heap (both) |
|--:|--:|--:|--:|--:|
| 100k | 1.18 s | 0.62 s | 1.9× | 66.5 MB |
| 1M | 13.14 s | 5.45 s | 2.4× | 671.8 MB |
| 10M | 3m 14 s | 56.6 s | 3.4× | 6.5 GB |

---

# 2. Copy-on-write historical proofs (commit 2)

Historical snapshot proofs rebuilt the whole tree from the artifact log on **every request** (O(N) SHA-256), which a validator could turn into a rebuild-DoS by cycling distinct counts — guarded until now by a per-validator distinct-count quota. This makes the live tree copy-on-write and serves committed-count proofs from retained snapshots in **O(depth)**.

- **`insertCOW`** rewrites only the root→leaf path and shares every untouched subtree, so a captured root stays valid. Roots/proofs are **byte-identical** to `Insert` (verified by the full suite passing with the store on `insertCOW`).
- The store captures a retained snapshot at **every committed (flush) count** and serves its proofs from shared nodes — O(depth). Snapshots cost ~0 extra RAM (structural sharing).
- **Restart-safe:** `recover()` re-captures a snapshot at each committed count in one copy-on-write replay, so historical proofs stay O(depth) after a restart.
- A proof at a **non-committed count is rejected** outright (its root can never match the on-chain commitment), so a proof request can no longer trigger a rebuild. The rebuild path stays only as a defensive cold-start fallback.
- With the rebuild-DoS surface gone, the **per-validator distinct-count limiter is removed** (`poc_snapshot_limiter.go`, ~174 lines incl. tests).

**Measured** (before = O(N) rebuild-for-proof; after = COW retained snapshot; `TestCOWRnD`):

| N | proof @2/3 before → after | DoS: 10 distinct counts, before → after |
|--:|--:|--:|
| 100k | 819 ms → 8 µs | 3.81 s → 74 µs |
| 500k | 4.19 s → 10 µs | 19.6 s → 60 µs |
| 1M | 7.82 s → 7 µs | 40.4 s → 102 µs |

At 10M the rebuild is ~55 s — over the 20 s request timeout; the COW proof stays ~µs. Ingest via `insertCOW` costs +2–16% (path-copying allocates more), shrinking with scale, and is offset by the deferred hashing above.

**Tests:** `TestCOWStoreProofPaths` (live / retained-historical / post-restart), the differential fingerprint, and the concurrent-readers test are green under `-race`.

**Note on scope:** `snapshot_cache.go` / the rebuild path are intentionally kept as a defensive cold-start fallback rather than deleted, so this commit is roughly LOC-neutral overall while removing the limiter subsystem and the rebuild-DoS class. Fully collapsing the rebuild path is a possible follow-up.


a-kuprin (approved):

DimaOrekhovPS:
merged as part of https://github.com/gonka-ai/gonka/pull/1447

PR #1436, opened 2026-07-10
fix(validation): normalize logprob distance by validator width, not executor's

qdanik:
## Problem
`customDistance` (logprob validation) normalized the distance by the **executor's** `top_logprobs` width, while the numerator (`positionDistance`) sums over the **validator's** tokens. The executor is untrusted and controls the width of its own response's `top_logprobs`; padding that width inflates the denominator and shrinks the normalized distance, so an invalid inference can slip under the validation threshold. Reported as **H1 #3853145**.

## Fix
Normalize by the **validator's** `top_logprobs` width instead — trusted, since the validator re-runs the request with a forced `top_logprobs` count taken from the on-chain prompt. This matches the numerator's summation basis, so the executor can no longer influence the denominator. A guard on empty `validationLogprobs` keeps the function panic-safe on its own.

### DIFF
--- a/common/completionapi/request.go
+++ b/common/completionapi/request.go
@@ -15,9 +15,7 @@ import (
 const ForcedTopLogprobs = 5
 
 type ModifiedRequest struct {
-	NewBody                  []byte
-	OriginalLogprobsValue    *bool
-	OriginalTopLogprobsValue *int
+	NewBody []byte
 }
 
 func ModifyRequestBody(requestBytes []byte, defaultSeed int32) (*ModifiedRequest, error) {
@@ -42,7 +40,6 @@ func ModifyRequestBodyWithLogprobsMode(requestBytes []byte, defaultSeed int32, l
 		requestMap["logprobs"] = true
 	}
 
-	originalTopLogprobsValue := getOriginalTopLogprobs(requestMap)
 	// Pin top_logprobs to the protocol constant on both the original and the validation request; a larger client-supplied value must not pass through.
 	requestMap["top_logprobs"] = ForcedTopLogprobs
 
@@ -83,9 +80,7 @@ func ModifyRequestBodyWithLogprobsMode(requestBytes []byte, defaultSeed int32, l
 	}
 
 	return &ModifiedRequest{
-		NewBody:                  modifiedRequestBytes,
-		OriginalLogprobsValue:    originalLogprobsValue,
-		OriginalTopLogprobsValue: originalTopLogprobsValue,
+		NewBody: modifiedRequestBytes,
 	}, nil
 }
 
@@ -208,33 +203,4 @@ func getOriginalLogprobs(requestMap map[string]interface{}) *bool {
 	log.Printf("Original request logprobs = %v", logprobsValue)
 	trueValue := true
 	return &trueValue
-}
-
-func getOriginalTopLogprobs(requestMap map[string]interface{}) *int {
-	topLogprobsValue, ok := requestMap["top_logprobs"]
-	if !ok {
-		return nil
-	}
-
-	if topLogprobsValue == nil {
-		return nil
-	}
-
-	if topLogprobsValueInt, ok := topLogprobsValue.(int); ok {
-		return &topLogprobsValueInt
-	}
-
-	if topLogprobsValueBool, ok := topLogprobsValue.(bool); ok {
-		if topLogprobsValueBool {
-			one := 1
-			return &one
-		} else {
-			zero := 0
-			return &zero
-		}
-	}
-
-	// Discard any non-integer value
-	log.Printf("Original request top_logprobs = %v", topLogprobsValue)
-	return nil
 }
--- a/common/completionapi/request.go
+++ b/common/completionapi/request.go
@@ -11,6 +11,9 @@ import (
 	"common/logging"
 )
 
+// ForcedTopLogprobs is the top_logprobs the validator forces; must equal the devshard gateway's TopLogprobsForcedValue so executor and validator widths are comparable (H1 #3853145).
+const ForcedTopLogprobs = 5
+
 type ModifiedRequest struct {
 	NewBody                  []byte
 	OriginalLogprobsValue    *bool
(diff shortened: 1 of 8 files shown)


DimaOrekhovPS (approved):

common/validation/validation.go
```diff
@@ -278,7 +280,7 @@ func positionDistance(
 		}
 	}
```
a-kuprin, line 283: The executor controls `originalLogprobMap` entirely, so it controls `minOriginalLogprob1`. A single entry at `-1e308` — plain, valid JSON that decodes fine into float64 — makes the extrapolation `2*min1 - min2` overflow to `-Inf`. Then `denom` is `+Inf` (not `NaN`, so the guard doesn't fire), term is `Inf/Inf = NaN`, and the term is skipped.

So executor still can pass validation by controlling `originalLogprobMap`

a-kuprin:
For hygiene 
`mlnode/packages/benchmarks/src/validation/utils.py` reimplements this metric (token_distance2, commented "Matches Go customDistance/positionDistance") and still normalizes by the executor's width

common/completionapi/request.go
```diff
@@ -40,9 +43,8 @@ func ModifyRequestBodyWithLogprobsMode(requestBytes []byte, defaultSeed int32, l
 	}
 
 	originalTopLogprobsValue := getOriginalTopLogprobs(requestMap)
```
a-kuprin: PR fixes here what already worked in same case but because of error.
Anyway we should fix the previous error:

`ModifyRequestBody` starts with:

```go
var requestMap map[string]interface{}
json.Unmarshal(requestBytes, &requestMap)
```

For encoding/json into interface{}, every JSON number becomes `float64`. Not `int`.

Then:

```go
func getOriginalTopLogprobs(requestMap map[string]interface{}) *int {
	// ...
	if topLogprobsValueInt, ok := topLogprobsValue.(int); ok {
		return &topLogprobsValueInt
	}
	// ...
	// Discard any non-integer value
	log.Printf("Original request top_logprobs = %v", topLogprobsValue)
	return nil
}
```

For any real JSON body ("top_logprobs": 2, 20, 5, …):

Value in the map is float64(20).
.(int) fails.
.(bool) fails.
Function returns nil (and logs “Original request top_logprobs = 20”).

We should check for `float64` or convert topLogprobsValue to `int`

common/completionapi/request.go
```diff
@@ -40,9 +43,8 @@ func ModifyRequestBodyWithLogprobsMode(requestBytes []byte, defaultSeed int32, l
...
-	if originalTopLogprobsValue == nil || *originalTopLogprobsValue < 5 {
-		requestMap["top_logprobs"] = 5
-	}
+	// Pin top_logprobs to the protocol constant on both the original and the validation request; a larger client-supplied value must not pass through.
+	requestMap["top_logprobs"] = ForcedTopLogprobs
```
a-kuprin, line 44: Actually I didn't got, why this is better to pin `top_logprobs` to `5` ?

Why user has no possibility to request more `top_logprobs` ?

common/validation/validation.go
```diff
@@ -213,17 +213,19 @@ func customDistance(
 	for i := range originalLogprobs {
 		o := originalLogprobs[i]
 		v := validationLogprobs[i]
-		posDistance, err := positionDistance(o.TopLogprobs, v.TopLogprobs)
+		// Ignore executor top_logprobs beyond the validated width so a padded width can neither dilute the divisor nor perturb the fallback (H1 #3853145).
+		originalTopLogprobs := o.TopLogprobs
+		if len(originalTopLogprobs) > len(v.TopLogprobs) {
+			originalTopLogprobs = originalTopLogprobs[:len(v.TopLogprobs)]
+		}
```
a-kuprin, line 220: Agent says that the truncation here isn't needed and doesn't add security but can hurt on honest case:

## Honest case, one position

Pinned request: `top_logprobs = 5`.  
Validator returns 5 entries. Executor (vLLM) returns **6**: top-5 plus the sampled token that fell outside the top-5.

**Validator top-5**

| token | logprob |
|---|---|
| `1` | −0.10 |
| `2` | −2.0 |
| `3` | −3.0 |
| `4` | −4.0 |
| `77` | −5.2 |

**Executor (honest, slightly different numbers)**

| index | token | logprob |
|---|---|---|
| 0 | `1` | −0.11 |
| 1 | `2` | −2.1 |
| 2 | `3` | −3.1 |
| 3 | `4` | −4.1 |
| 4 | `5` | −5.0 |
| 5 | `77` | −5.3 | ← real match for validator’s `77`

Scoring walks **validator** tokens and looks them up in the executor map.

**Without truncation**

- Looks up `77` → finds −5.3 → tiny distance  
- Similarity stays high (~0.9908 in the full run)

**With truncation** (`[:5]` → drop index 5)

Executor map only has `1…5`. `77` is gone.

- Looks up `77` → miss → **fallback** estimate (extrapolated from executor mins)  
- That fallback is much worse than −5.3 → larger distance  
- Similarity drops (~0.9836) → **fail** at 0.99 threshold

Same tokens, same honest model; only the prefix cut threw away a real match because it sat past index 4.

---

## Do you need to remove truncation?

**For security / H1: no.** Width-independence already stops the padding attack. Truncation is not what defends you.

**For correctness / false fails: yes, you should drop it** if you care about honest nodes that legitimately return `top_k+1` (or any wider list that still contains the validator’s tokens). Keeping it means occasional unfair invalidations near the threshold.

**When you could leave it:** only if you are sure production always has equal widths (request hard-pinned to 5 **and** both sides never emit extras) **and** you accept residual risk from vLLM’s +1 behavior. That is a weak reason to keep a harmful line.

Practical recommendation: remove the truncation hunk; keep the width-independence change. Low risk, removes the false-negative path, no real security loss.

qdanik: The initial idea is to only check the top 5 top_logprobs.

Even if an honest host can generate more than 5 top_logprobs, it will be reduced to five and compared to the top_logprobs validator's host list.

common/completionapi/request.go
qdanik: looks like `OriginalTopLogprobsValue` is a dead code

common/validation/validation.go
qdanik: fixed

common/completionapi/request.go
qdanik: Should be investigated and delivered in another PR. I'll create an issue soon

a-kuprin (approved):

PR #1437, opened 2026-07-10
Rolling updates for versiond-managed binaries (blue/green + drain, part 1 of #1367)

snevolin:
## Summary

This PR adds **rolling updates for versiond-managed binaries**: when governance publishes a
new binary under the **same version name** (name stays, e.g. `v0.2.13`, only the `sha256`
changes), versiond now performs a **blue/green swap with drain** instead of the current
stop-then-start. In-flight work on the old `devshardd` child is allowed to finish; the new
child receives traffic only after it is actually ready.

It implements **Part 1 (Track A)** of the
[rolling-update plan](https://github.com/snevolin/gonka/blob/sn/versiond-rolling-updates/devshard/docs/rolling-update.md)
and is the first stage of the HA proposal
[#1367](https://github.com/gonka-ai/gonka/discussions/1367)
("rolling updates first" per maintainer feedback in the discussion).

> **Base branch:** `ak/pixelplex-refactoring-into-r2` (#1366), not `main` — the feature
> builds directly on the HA foundation from that PR (standalone `devshardd`, shared
> Postgres, per-child ports, `versiond-router`). Intended as a stacked PR.

---

## Motivation

Today a governance sha change is handled by `downloadAndSwap` as *stop-then-start on the
same port*: the old child gets `SIGTERM` and is `SIGKILL`ed after 5s, before the new binary
has proven it can serve. Any inference longer than ~5s (long SSE `/chat/completions`,
validation) is cut mid-flight.

For operators this is not cosmetic: a dropped in-flight inference means failed/unanswered
validations (Executor Reputation), and mid-epoch unavailability feeds into cPoC
confirmation ratio. The operator requirement (rolling-update.md §0) is:

1. requests already accepted by the old instance may **finish**;
2. the old instance is stopped **only after** the new one is ready and reachable;
3. once the new instance is reachable, **new** requests go to it while the old one drains.

---

## How a swap works now

```text
poll detects same name, new sha256
        │
        ▼
download into per-sha dir  bin/<name>/<sha>/     ← old child's binary untouched on disk
        │
        ▼
start NEW child on a NEW port                    ← old child keeps serving
        │
        ▼
wait for NEW child readiness (HTTP /ready, not just TCP)
        │
        ▼
atomic route swap: version → NEW port            ← in-flight requests stay on OLD conn
        │
        ▼
old child → "draining" (out of route table, visible in /healthz, no crash-restart)
        │
        ▼
poll /drain/status until inflight == 0  OR  VERSIOND_DRAIN_TIMEOUT
        │
        ▼
SIGTERM (long WaitDelay ≥ DEVSHARD_SHUTDOWN_GRACE) → SIGKILL only as backstop
```

If the new child never becomes ready, the swap **aborts and the old child keeps serving**
(retry on the next poll) — a failed rollout no longer takes the version down.

---

### versiond (`versioned/`)

| Area | Change |
|------|--------|
| **Blue/green + drain** | `downloadAndSwap` rewritten; per-name `current` + `draining[]` children; per-generation port allocation with release; draining children excluded from routes and from the crash-restart loop |
| **Readiness gate** | `waitForReady` probes `VERSIOND_READY_PATH` (default `/ready`); replaces the TCP-only probe and its "routing anyway" fallback; swap aborts (old keeps serving) on readiness failure |
| **Per-sha installs + GC** | binaries live in `bin/<name>/<sha256>/` (write-once; safe for a draining child's crash-restart); superseded install dirs are GC'd, keeping desired/live/draining plus a small retain cushion; in-progress downloads are never GC'd |
| **Storage guard** | overlap is enabled only with `PGHOST` set **and** no leftover SQLite-owned sessions in the data dir (probed via `_meta.db`); otherwise versiond falls back to today's stop-then-start — SQLite single-instance setups keep their current behavior |
| **Legacy compatibility** | binaries released before this work run unchanged: preflight (`--print-*` flags) falls back to the governance slot name on a recognized "unsupported flag" error (fail-closed on timeouts/signals); `/ready` falls back to `/healthz` → TCP for the default path only; missing `/drain/status` (404/405/501) shortens the drain wait instead of stalling 15m |
| **Oracle input validation** | version names must be simple path components and unique; sha256 must be 64 hex chars (mirrors chain-side `DevshardEscrowParams.Validate`); an invalid response fails the poll and leaves running children untouched |
| **Observability** | `/healthz` entries now include `status` (incl. `draining`), `sha256` and `binary_version` per child — also groundwork for the §1.8 router-evacuation track |

### devshardd (`devshard/cmd/devshardd/`)

| Endpoint / knob | Behavior |
|-----------------|----------|
| `GET /ready` | 200 only after full init **and** the chain event subscription is active (flips back on disconnect) |
| `POST /drain` | belt-and-braces: reject new work, finish existing |
| `GET /drain/status` | `{ready, draining, inflight}`; inflight counts all non-lifecycle HTTP requests (incl. full SSE stream duration) and is exported as Prometheus gauge `devshardd_lifecycle_inflight_requests` |
| `DEVSHARD_SHUTDOWN_GRACE` | replaces the hard-coded 5s `server.Shutdown` window (default `10m`); versiond's kill grace honors it (`max(VERSIOND_DRAIN_KILL_GRACE, DEVSHARD_SHUTDOWN_GRACE)`) |

### New configuration (all defaulted, no compose changes required)

| Env var | Default | Meaning |
|---|---|---|
| `VERSIOND_READY_PATH` | `/ready` | readiness path the supervisor probes |
| `VERSIOND_READY_TIMEOUT` | `60s` | max wait for the new child before aborting the swap |
| `VERSIOND_DRAIN_PATH` | `/drain` | path versiond POSTs to put the old child into drain mode |
| `VERSIOND_DRAIN_STATUS_PATH` | `/drain/status` | path versiond polls for the old child's in-flight count |
| `VERSIOND_DRAIN_TIMEOUT` | `15m` | max wait for the old child to go idle before `SIGTERM` |
| `VERSIOND_DRAIN_POLL_INTERVAL` | `1s` | in-flight poll cadence |
| `VERSIOND_DRAIN_KILL_GRACE` | `30s` | wait after `SIGTERM` before `SIGKILL` |
| `DEVSHARD_SHUTDOWN_GRACE` | `10m` | devshardd graceful-shutdown window on `SIGTERM` |

### Test infrastructure

- **Testermint**: `VersiondTests` gains a same-version binary update scenario — a slow
  request keeps running on the old child through the swap, continuity probes assert no
  routing gap, and the draining child is observed to exit. `local-test-net`'s
  `testapp-server` now builds **per-name stamped** binaries (`-X main.Version=<slot>`), as
  required by the protocol-embed check on this base branch.
- **Chain gRPC wiring for devshard flows on fresh builds**: testermint now passes
  `DEVSHARD_CHAIN_GRPC` to `devshardctl` (its `chain_rest` is deprecated on this base
  branch), and `init-docker-genesis.sh` binds the genesis node's gRPC to `0.0.0.0:9090`
  (join init already did this) so `edge-api`/`devshardctl` can reach it from peer
  containers. Without these, devshard/edge-api testermint flows only pass on stale cached
  images.
- **Go e2e** (`versioned/e2e`): `TestSameNameNewSHA_RollingUpdateDrainsOld` — long request
  survives a governance sha change while a concurrent request is served by the new binary;
  testapp got `/ready`, `/drain`, `/drain/status` and graceful shutdown to act as a
  drain-aware child.
- Unit suites extended: swap abort paths (new child not ready → old keeps serving),
  drain idle/timeout branches, GC retention, preflight matrix (stamped / legacy /
  mismatch / timeout), oracle validation, port allocation.

### Docs

[rolling-update.md](https://github.com/snevolin/gonka/blob/sn/versiond-rolling-updates/devshard/docs/rolling-update.md)
updated to match the implementation: final env defaults, the legacy-compatibility
fallbacks, GC behavior, and the `WaitDelay` / shutdown-grace interaction.

---

## Compatibility with deployed nodes

| Scenario | Result |
|----------|--------|
| new versiond + already-released devshardd (no `--print-*`, no `/ready`) | works via legacy fallbacks; first swap away from a legacy child uses the short drain grace (one-time transitional effect) |
| old versiond + new devshardd | works (`DEVSHARD_BINARY_LOG_VERSION` unset → link-stamp fallback) |
| SQLite single-instance | unchanged behavior (stop-then-start; overlap never enabled) |
| Postgres-backed nodes | full blue/green + drain |

No chain changes: versions still come from governance `DevshardEscrowParams.ApprovedVersions`
via the existing dapi `/versions` oracle; the wire contract is unchanged.

---

### Unit (no Docker)

```bash
cd versioned && go test -race ./...        # incl. swap/drain/GC/preflight/oracle suites
cd devshard  && go test -race ./cmd/devshardd/...
make -C devshard ci-testenv-unit
```

All green (also under `-race`).

### Docker e2e

```bash
make -C versioned e2e
```

All scenarios pass, including `TestSameNameNewSHA_RollingUpdateDrainsOld`: the in-flight
slow request completes on the old binary, new requests are served by the new one, the
draining child exits cleanly.

### Testermint

```bash
make local-build build-docker
make run-tests TESTS="VersiondTests"
```

Note: devshard/edge-api testermint flows require freshly built images **plus** the chain
gRPC fixes included here (`DEVSHARD_CHAIN_GRPC` + genesis `0.0.0.0:9090` bind); on stale
cached images the pre-existing gaps are masked.

---

## Out of scope / follow-ups

- **Track B** (§1.8): `versiond-router` host evacuation — upstream `down` + reload, drain
  poll, operator runbook. The `/healthz` draining visibility and `/drain/status` added here
  are the signals that track will consume.
- Back-pressure cap on concurrently draining children under rapid sha churn.
- Download size limits (defense-in-depth), exposing per-child inflight through versiond
  `/healthz` for the §1.8 runbook.


devshard/cmd/devshardd/server.go
```diff
@@ -11,15 +11,30 @@ import (
...
+		}
+		return c.JSON(http.StatusOK, status)
+	})
+	e.POST("/drain", func(c echo.Context) error {
+		lifecycle.StartDrain()
+		return c.JSON(http.StatusOK, lifecycle.Status())
+	})
+	e.GET("/drain/status", func(c echo.Context) error {
+		return c.JSON(http.StatusOK, lifecycle.Status())
+	})
```
a-kuprin, line 48: It's a publicly accessible devshardd server, but new endpoints are for administration devshardd reload.
Especially `/drain` cannot be public.

Consider making another echo server with administration endpoints and expose these endpoints at non-public port

a-kuprin:
**30s (DrainKillGrace)**: “we don’t know inflight; don’t stall 15m; give a short cushion then stop asking.”
**10m (DEVSHARD_SHUTDOWN_GRACE)**: “once we have sent SIGTERM, let long requests finish.”


Docs table line that says `VERSIOND_DRAIN_KILL_GRACE` = **“wait after SIGTERM before SIGKILL”** is slightly misleading for current code: for devshardd, post-SIGTERM kill grace is really `max(DrainKillGrace, DEVSHARD_SHUTDOWN_GRACE)`, and `DrainKillGrace` alone is the legacy pre-SIGTERM wait.

`DrainKillGrace` is used for backward compatibility when versioned is holding old devshardd, that do not have drain support. I think we should align them: let's make DrainKillGrace also 10m, and fix the docs to make it more precise

a-kuprin:
`waitForChild(c, timeout)` — used on swap/drain: wait until done or timeout. No outer cancel.
`waitForChildContext(ctx, c, timeout)` — used only in Manager.Shutdown: wait until done, timeout, or ctx cancelled.

`waitForChildContext` has 10s timeout, that doesn't match `DEVSHARD_SHUTDOWN_GRACE`

But as I understand versioned graceful stop or devshardd host stop is deffered for next PR

versioned/internal/process/manager.go
```diff
@@ -267,9 +337,10 @@ func (m *Manager) Reconcile(ctx context.Context, desired []oracle.Version) error
 		c.cancel()
 	}
 	for _, c := range toStop {
-		waitForChild(c, 5*time.Second)
+		waitForChild(c, m.childStopTimeout())
 	}
```
a-kuprin: When governance removes version why we are not waiting to drain: `drainAndStop` ?

versioned/internal/process/manager.go
```diff
@@ -19,55 +24,108 @@ import (
...
-	m.nextPort++
-	m.assignedPorts[name] = port
-	return port
+}
+
+// releasePort releases a child port after the child process exits.
+// Must be called with m.mu held.
+func (m *Manager) releasePort(port int) {
+	delete(m.allocatedPorts, port)
 }
```
a-kuprin, line 193: nextPort only increases (tests assert no immediate reuse). Long-lived versiond with many generations eventually walks off the valid port range. Prefer free-port scan

versioned/internal/process/manager.go
```diff
@@ -636,11 +955,97 @@ func childEnv(binaryLogVersion string) []string {
...
+	}
+	hasSQLite, err := hasDevshardSQLiteSessions(filepath.Join(m.cfg.DataDir, versionName))
+	if err != nil {
+		slog.Warn("rolling overlap disabled: cannot probe devshard sqlite sessions", "version", versionName, "error", err)
+		return false
+	}
+	if hasSQLite {
+		slog.Warn("rolling overlap disabled: devshard sqlite sessions are still present", "version", versionName)
+		return false
+	}
```
a-kuprin, line 1314: having `PGHOST` is not enough

At latest [`ak/pixelplex-refactoring-into-r2`](https://github.com/gonka-ai/gonka/tree/ak/pixelplex-refactoring-into-r2) it is introduced `DEVSHARD_HA` and `DEVSHARD_REQUIRE_POSTGRES` flags

```go
func HAModeEnabled() bool {
	return envTruthy("DEVSHARD_HA") || envTruthy("DEVSHARD_REQUIRE_POSTGRES")
}
```

The point is that even when `PGHOST` is set it still runs old sessions with sqlite, and if postgres is not available can fallback to Sqlite.

But for rolling update or high-availability mode for devshardd, we need shared storage (Postgres in our case). You are checking it with `hasDevshardSQLiteSessions`, but it's not enough as fallback is possible.

It's better to check against `HAModeEnabled`, in this mode sqlite sessions are forced to migrate to postgres and sqlite fallback never happens. So checking `HAModeEnabled` is enough and you do not need to check `hasDevshardSQLiteSessions`.

Please rebase /merge over latest `[ak/pixelplex-refactoring-into-r2](https://github.com/gonka-ai/gonka/tree/ak/pixelplex-refactoring-into-r2)` and adopt `HAModeEnabled`

a-kuprin:
Also at [`ak/pixelplex-refactoring-into-r2`](https://github.com/gonka-ai/gonka/tree/ak/pixelplex-refactoring-into-r2) branch was introduced testenv.
It aims to replace heavy testermint tests.

Consider adding a testenv scenario (e.g. S7), for testing that in-flight session keep living while version swap, node replacement, downing node. It could be used for following cases:
- Real devshardd lifecycle (/ready, /drain, inflight) under Postgres / multi-versiond
- Swap while a real chat/SSE or session is in flight
- mock-dapi can flip /versions sha without a gov proposal

versioned/internal/process/manager.go
```diff
@@ -475,35 +708,60 @@ func logInstalledVersionMismatch(scope, versionName, desiredArchiveHash string,
...
+	m.draining = make(map[string][]*child)
 	m.downloading = make(map[string]struct{})
 	m.routes.Store(map[string]string{})
 	m.mu.Unlock()
 
 	for _, c := range children {
 		slog.Info("shutting down", "version", c.version.Name)
-		waitForChild(c, 10*time.Second)
+		if err := waitForChildContext(ctx, c, m.childStopTimeout()); err != nil {
+			return err
```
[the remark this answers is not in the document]
snevolin: Fixed. The HTTP-server now keeeps it's own 10 second shutdown context. Also we're logging shutdown errors.

devshard/cmd/devshardd/server.go
snevolin: Done. Good catch. Moved /ready, /drain, /drain/status to a separate admin echo server.

versioned/internal/process/manager.go
snevolin: Right, good catch. Removal now goes through the same drainAndStop path as swaps: the route is dropped immediately (new requests 404), the child is moved to the draining set with restart=false, and drained asynchronously.

versioned/internal/process/manager.go
snevolin: Done. assignPort now scans from BasePort for the lowest free port, reusing released ones and skipping versiond's own listen port (parsed from config.ListenAddr()), so the range stays bounded by the number of live+draining children instead of walking off. Dropped the monotonic nextPort entirely.

versioned/internal/process/manager.go
snevolin: It would be better to add a special flag (i.g. --print-storage-mode) to devshardd and not to violate the DRY principle.

a-kuprin: Maybe env var is enough, as versioned is a real environment where devshardd is executed ? So we just don't need `--print-storage-mode`

devshardd is not its own container — it runs as a child process inside the versiond container. Child env is built from os.Environ(), so env vars set on the versiond container are inherited by devshardd

snevolin: Yes, right, but we shouldn't expect every devshardd binary an ability to correctly recognize DEVSHARD_STORAGE_MODE. Passing env variables is the operator's intention (I want to make the cluster work with Postgres), but not the ability of cluster components to work that way. They might not support it yet. Or they might have some problems operating with Postgres, it's their responsibility. The only way to ensure they can operate in overlap mode is to ask them.

a-kuprin: It is not required to work with postgres.
Tiny deployments could use sqlite and still should be able to do so.

So this is deployment config, not in-binary-build-in requirement

snevolin: Yes, Postgres is not required, and tiny SQLite deployments must keep working. And agree the mode is deployment config, not built into the binary. I changed the behaviour, versiond no longer reads the storage env at all. It asks the devshardd itself with --print-storage-mode: the binary applies its own mode.Resolve() to the inherited env and prints the resolved mode (sqlite / hybrid / postgres). Blue/green overlap is allowed only when both sides answer postgres, the running instance of devshardd and the incoming one.

devshard/cmd/devshardd/lifecycle.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (s *lifecycleState) middleware(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+		if isLifecycleBypassPath(c.Request().URL.Path) {
+			return next(c)
+		}
+		if s.draining.Load() {
+			return echo.NewHTTPError(http.StatusServiceUnavailable, "devshardd is draining")
+		}
+		s.addInflight(1)
```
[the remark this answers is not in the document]
snevolin: Fixed in 89c0b1eb8 by incrementing the counter before the draining check.

versioned/internal/process/manager.go
```diff
@@ -403,27 +492,68 @@ func (m *Manager) downloadAndSwap(ctx context.Context, v oracle.Version, sha str
...
 	delete(m.processes, v.Name)
-	m.startChild(ctx, v)
+	m.draining[v.Name] = append(m.draining[v.Name], old)
+	newChild.child.restart = true
+	m.processes[v.Name] = newChild.child
+	m.rebuildRoutes()
 	m.mu.Unlock()
 
+	slog.Info("swapped child route; old child draining", "version", v.Name, "old_port", old.port, "new_port", newChild.child.port)
+	m.requestDrain(old)
```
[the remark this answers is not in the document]
snevolin: Fixed in b879f3a9. Now the route's table internal representation is a special obejct instead of a plain address.

versioned/internal/process/manager.go
```diff
@@ -622,25 +946,148 @@ func (m *Manager) runChild(ctx context.Context, c *child) {
...
+}
+
+func (m *Manager) rollingOverlapAllowed(versionName string) bool {
+	name := strings.ToLower(m.cfg.BinaryName)
+	if name != "devshard" && name != "devshardd" {
+		return true
+	}
+	if strings.TrimSpace(os.Getenv("PGHOST")) == "" {
+		return false
+	}
```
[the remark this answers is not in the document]
snevolin: This was valid for the original PGHOST-only overlap mechanism, but the current implementation no longer relies on only PGHOST.

versioned/internal/process/manager.go
```diff
@@ -208,17 +319,17 @@ func (m *Manager) Reconcile(ctx context.Context, desired []oracle.Version) error
...
 	// Phase C (lock): apply decisions -- start ready children, mark downloads, stop removed.
 	m.mu.Lock()
-	for _, v := range toStart {
-		if _, already := m.processes[v.Name]; already {
+	for _, a := range toStart {
+		if _, already := m.processes[a.version.Name]; already {
 			continue // another reconcile started it
 		}
-		m.startChild(ctx, v)
+		m.startChild(ctx, a.version, a.sha256, m.installBinPath(a.version.Name, a.sha256), true)
```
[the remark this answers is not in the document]
snevolin: FIxed in d7e9afe6b. A version is now blocked from starting while a previous version of the same name is still draining, so two instances can never open the same data directory.

versioned/internal/process/manager.go
```diff
@@ -24,50 +29,142 @@ import (
...
+		}
+		m.allocatedPorts[port] = struct{}{}
 		return port
 	}
-	port := m.nextPort
-	m.nextPort++
-	m.assignedPorts[name] = port
-	return port
+	slog.Error("no child ports available", "base_port", m.cfg.BasePort, "max_port", maxChildPort)
+	return invalidChildPort
```
x0152: I think this should return an error here, because -1 may be treated as a valid port

versioned/internal/process/manager.go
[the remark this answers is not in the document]
x0152: > FIxed in d7e9afe6b

Hi! Could you point me to the commit with this fix? cannot find in this branch

versioned/internal/process/manager.go
```diff
@@ -127,27 +239,22 @@ func (m *Manager) Reconcile(ctx context.Context, desired []oracle.Version) error
...
 		isDownloading bool
 		child         *child
 	}
 	var snapshots []versionSnapshot
 
 	for _, v := range desiredSet {
-		versionDir := filepath.Join(m.cfg.BinDir, v.Name)
-		binPath := filepath.Join(versionDir, m.cfg.BinaryName)
 		if overrideSrc, isOverride := m.cfg.Overrides[v.Name]; isOverride {
+			binPath := filepath.Join(m.cfg.BinDir, v.Name, m.cfg.BinaryName)
```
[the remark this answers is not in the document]
snevolin: Added two conditions before the "already running" early return plus one additional hash check.

x0152:
The idea is right, some hardening might be needed, but it's non-blocking

x0152 (approved):

versioned/internal/process/manager.go
```diff
@@ -193,9 +300,11 @@ func (m *Manager) Reconcile(ctx context.Context, desired []oracle.Version) error
 		}
 
 		// Not running.
-		matches, metadata, diskBinaryHash, stateErr := installedVersionMatches(snap.versionDir, snap.binPath, desiredHash)
+		versionDir := m.installDir(snap.version.Name, desiredHash)
+		binPath := filepath.Join(versionDir, m.cfg.BinaryName)
+		matches, metadata, diskBinaryHash, stateErr := installedVersionMatches(versionDir, binPath, desiredHash)
```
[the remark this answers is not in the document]
snevolin: Fixed in dba49dfb, versiond now checks the old flat install before downloading. If the binary and install.json are valid, it copies them to the new per-SHA directory. The old files are kept for older versiond instances using the same shared volume.

versioned/internal/process/manager.go
```diff
@@ -554,22 +883,22 @@ func (m *Manager) runChild(ctx context.Context, c *child) {
...
+		cmd.Env = childEnv(preflight.binaryLogVersion, adminAddr)
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 		cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
 		// Cancel sends SIGKILL by default. Override to send SIGTERM for graceful shutdown.
 		cmd.Cancel = func() error {
 			return cmd.Process.Signal(syscall.SIGTERM)
 		}
-		cmd.WaitDelay = 5 * time.Second // SIGKILL after 5s if SIGTERM didn't work
+		cmd.WaitDelay = m.childStopTimeout()
```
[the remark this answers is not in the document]
snevolin: I replaced the competing timers with a small process state machine:

Running -> Terminating -> Killing -> Exited

Stop() sends SIGTERM. After the grace period, it sends SIGKILL if needed. Done() closes only after cmd.Wait() has reaped the process.

Manager.Shutdown now stops all children first. If its context expires, it force-stops the remaining children but still waits for them to exit.

I also added tests for graceful exit, forced kill, and process reaping.

versioned/internal/process/manager.go
```diff
@@ -579,13 +908,35 @@ func (m *Manager) runChild(ctx context.Context, c *child) {
 			return
 		}
 
-		// Wait for the child to start accepting connections before routing traffic.
-		if !waitForPort(ctx, c.port, 10*time.Second) {
-			slog.Warn("child did not start listening in time, routing anyway", "version", c.version.Name)
+		if !waitForReady(ctx, c.lifecyclePort(), m.cfg.ReadyPath, m.cfg.ReadyTimeout) {
```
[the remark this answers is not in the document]
snevolin: Fixed in 4f66fc262.

A child is now routed only when both checks pass:
- admin /ready returns 200;
- public /healthz returns 2xx.

Both checks use the same readiness timeout and are repeated together. Legacy children without the admin API keep the existing fallback behavior.

I also added tests for an unavailable public listener and for readiness changing between probe attempts.

versioned/internal/process/manager.go
snevolin: Agree. I changed assignPort to return (int, error) instead of -1.

snevolin:
Superseded by #1490, which targets devshard-0.2.14-v4 and has an updated description for the current sn/versiond-rolling-updates branch.

Issue #1199, opened 2026-05-19
Reproducible sampling for inference validation


tcharchian:
Review the existing reproducible / deterministic sampling work for inference validation and prepare a careful path toward adding it to MLNode versions.

This task is related to the known inference validation vulnerability described in the inference validation proposal. The proposed direction is a two-stage validation system with a cheap sequence check before the existing distribution check.

The goal is to take over the existing experiments and base implementation, review them carefully for vulnerabilities, and define a safe non-enforcing rollout path before any strict validation behavior is enabled.

## Context

Relevant materials:

* Inference validation proposal, section “Proper Fix: Two-Stage Validation System”:
    https://github.com/gonka-ai/gonka/blob/main/proposals/inference-validation/inference-validation.md#proper-fix-two-stage-validation-system
* Existing deterministic sampling / validation document and base implementation by @tamazgadaev:
    https://github.com/gonka-ai/vllm/blob/tg/detemrinistic_sampling_dump/docs/DETERMINISTIC_SAMPLING_VALIDATION.md

Please use these documents as the source of truth for the proposed design, implementation status, artifact format, implementation plan, tests, and known limitations.  ￼

## Task

Review the existing proposal and deterministic sampling implementation, then prepare a safe path toward gradual MLNode integration.

This task is not primarily about writing new code immediately. The important part is to carefully review the approach, check for vulnerabilities, and define how to introduce it softly into MLNode versions while collecting data and avoiding strict enforcement at first.

## Review scope

Please review:

1. The two-stage validation design from the inference validation proposal.
2. Deterministic sampling validation document.
3. The existing branch / base implementation.
4. Current implementation status.
5. What can be added to MLNode versions softly.
6. What should remain non-enforcing for now.
7. What data should be collected before enabling stricter validation.

## Review points

1. Confirm alignment with the proposed fix

Check whether the current implementation matches the two-stage validation direction described in the proposal.

Please focus on whether the implementation correctly supports:

* sequence / sampling replay validation;
* existing distribution validation;
* the intended order of checks;
* the intended protection against the known validation weakness.

2. Review seed, RNG, and replay logic

Check whether the seed generation, RNG initialization, and replay logic are implemented consistently with the:

* Inference validation proposal, section “Proper Fix: Two-Stage Validation System”:
    https://github.com/gonka-ai/gonka/blob/main/proposals/inference-validation/inference-validation.md#proper-fix-two-stage-validation-system
[...]

PR #1448, opened 2026-07-14
validation/detsample: Go chain-validator for reproducible sampling (Stage-1 replay + Stage-2 distance)

bonujel:
## What this is

The gonka half of the two-stage inference-validation design in gonka-ai/gonka#1199: the Go **chain-validator** that replays reproducible sampling artifacts, plus the shared cross-language conformance vectors that prove it is byte-for-byte identical to the vLLM-side Python validator.

This PR is:

* Ready for review
* Covers both stages of the #1199 two-stage design: Stage-1 sampling-replay (per-position + sequence) and the Stage-2 distance metric
* Non-enforcing, not wired into the consensus path
* Code and tests only, all under `decentralized-api/internal/validation/detsample/`
* Paired with the vLLM-side PR gonka-ai/vllm#62 (`deterministic-sampling-clean` → `main`), which carries the executor, the Python validator, and the GPU evidence

It is the gonka half of a symmetric pair: vLLM produces reproducible sampling artifacts and a Python validator; this package is the Go validator the chain runs. Both reduce the *same* signed logprobs to the *same* sampled token from the *same* seed — verified by a common vector file.

## How it works

The chain validates an inference in two stages.

* **Stage 1 — Sampling-Replay (this PR, complete).** The cheap, exact check: the validator re-derives the chain-bound RNG seed and *replays* the signed logprobs through a deterministic **decimal** pipeline, checking that the same seed reduces to the same sampled token, bit for bit — at both the single-position level (`VerifyPosition`) and the sequence level (`VerifySequence`).
  * It never recomputes logprobs — batch composition alone flips 0–8% of honest positions, even after the executor weight fix — it replays the *signed* ones.
* **Stage 2 — distance (metric here).** Covers the positions Stage 1 structurally can't (greedy / temperature-0 and unbounded support). This package provides the **metric** — `MAEDistance`, MAE over the top-K support — which separates honest from a wrong model by ~40–65×. The **threshold** that turns it into a verdict is a policy call (Decision D) and is not in this package.

**Scope of this PR.** Stage 1 is complete (per-position + sequence replay); Stage 2 contributes the distance metric only, not its threshold. Signature verification of the artifact (Decision A) and wiring into consensus are out of scope. In short: this makes honest executor output *replayable and provably consistent across languages*, not yet *enforceable*.

### RNG and pipeline

`rng.go`

* `Sha256CounterRNG`: portable counter-mode RNG, `u64 = first_8_bytes(SHA256(seed ‖ counter_be_u64))`.
* `Uint64Below` (unbiased `[0,n)` via rejection sampling) and `SampleCategoricalWeights` over integer weights, which raises on an all-zero weight vector instead of a silent fallback.

`pipeline.go`

* `LogprobsToWeights` / `DecimalSampleFromLogprobs`: the decimal pipeline via `cockroachdb/apd/v2`.
* Fixed lexicographic token-ID string order; temperature → softmax → `top_k`/`top_p`/`min_p` filter → renormalize → integer weights summing to exactly `2^16` (65536).
* Exactness is what lets executor and validator reduce identically: the float path sums to `[65530, 65541]` and false-rejects 14–77% of honest positions; the decimal total is exactly `2^16`, giving 0/128 false-rejects (E1).

### Seed derivation

`seed.go`

* `DeriveChainBoundSeed(userSeed int64, inferenceID string)`: Go side of gonka-ai/vllm#56, mirroring the Python `derive_chain_bound_seed` byte-for-byte.
* Domain-tagged (`gonka-deterministic-sampling-v1`), byte-length-prefixed framing, fail-closed: rejects empty, `>256`-char, and non-printable-ASCII inference IDs.
* The validator derives the seed and never trusts request-controlled material.

### Verdict — single position and sequence

`verify.go`

* `VerifyPosition(Position) Result`: single-position replay with a three-valued `Verdict {honest, fraud, inconclusive}`.
* A mismatch is *fraud* only when the validator could faithfully replay. An unsupported contract version or seed domain, a greedy (temp-0) position, a non-positive or unparseable temperature, or the validator's own replay error is *inconclusive*, never fraud — in a slashing system a false-fraud is far costlier than a missed detection.

`sequence.go`

* `VerifySequence(SequenceRequest, []SequencePosition) SequenceResult`: sequence-level Stage-1 orchestration — the full cheap-check stage.
* Composes the per-position seed `base|pos`, loops `VerifyPosition`, and aggregates: any Fraud → Fraud; else at least one Honest → Honest; else Inconclusive.
* Version, seed-domain, greedy, and unbounded-support gates run before any replay. Mirrors the Python `verify_sequence`.

### Stage-2 distance metric

`distance.go`

* `MAEDistance(executor, validator)`: Stage-2 distance as MAE over the signed top-K support, with sorted-key summation for byte-exact cross-language equality.
* `Stage2MAEFraudThreshold` is a documented placeholder (Decision D); nothing here enforces. Mirrors the Python `mae_distance`.

### Cross-language conformance (S3)

`testdata/conformance_vectors.json` + `conformance_vectors_test.go` + `sequence_conformance_test.go`

* The shared vector file: RNG reference, `Uint64Below` sequences, float→string canonicalization, seed-derivation reference, pipeline cases, and now sequence-level and distance cases too.
* The vLLM-side PR generates and replays the *same* JSON. This proves Python↔Go byte-parity from the primitives up to the sequence verdict and the distance value — for example, an MAE of `0.5000000000000001` matches exactly — closing the previously-untested cross-language gap (S3).

## Testing

`go test ./internal/validation/detsample/` is green. Coverage:

* `rng_test.go`, `pipeline_test.go`
* `seed_test.go` — domain-separation and fail-closed invariants
* `verify_test.go` — three-valued verdict paths
* `sequence_test.go` — aggregation across honest / fraud / inconclusive / unbounded / greedy / version / missing-logprobs
* `distance_test.go`
* `reject_test.go` — rejection-sampling paths
* `conformance_vectors_test.go`, `sequence_conformance_test.go` — the shared vectors

## Still open — do not merge for enforcement

1. **Decision A — signed-artifact contract (keystone).** Which fields bind into `responseHash`, which tokens' logprobs are signed (the filter's kept set, not a fixed K), canonicalization, filter-edge rules, and a version descriptor. `SupportedContractVersion = "1.0.0"` is a placeholder pending A; everything downstream depends on it.
2. **Wiring into consensus.** `inference_validation.go` does not call `detsample` yet. Turning it on is gated on Decision A and enforcement gating (Decision D).
3. **Seed hardening (Decision C).** `DeriveChainBoundSeed` binds `(user_seed, inference_id)`. `inference_id` (via gonka-ai/vllm#56) removes the executor-choosable hole, but `user_seed` stays request-controlled, leaving residual offline grinding. Bind inputs the executor cannot pick (executor address / block hash) or commit `user_seed` on-chain.
4. **Stage-2 threshold (Decision D).** The metric (`MAEDistance`) is settled; the threshold value, epochs, and cross-architecture CI gate are policy.
5. **Tokenizer determinism.** If the seed binds tokenized prompt material, tokenizer or engine drift changes the seed and false-rejects. Decide: pin the tokenizer/engine on-chain, or bind `prompt_hash`.

Note: RNG semantics (Decision B) are resolved to per-position independent seeding (`base|pos`), matched by the executor on the vLLM side.

## Companion PR

The vLLM side — the executor sampling fix, the Python validator, and the GPU evidence — is gonka-ai/vllm#62.

Refs: gonka-ai/gonka#1199


redstartechno:
I ran this locally against the current train (WSL, head `4448b0bc3` merged onto `upgrade-v0.2.16` @ `c33c9eaa5`; the branch is 29 commits behind it).

- `go test -race ./internal/validation/detsample/...` passes (all 7 `TestVerifySequence*`/`TestVerifyPosition*` plus the rest of the package), and `go vet` is clean.
- **One rebase issue that git doesn't flag.** Since this branch was cut, the train dropped `devshard` from `decentralized-api/go.mod` (there is no `devshard` require or replace there any more). This branch still carries `devshard v0.0.0-00010101000000-000000000000` in its `require` block, and the merge auto-resolves by keeping it. Build and tests still pass, but after the merge `go mod tidy` fails:

  ```
  devshard@v0.0.0-00010101000000-000000000000: malformed module path "devshard": missing dot in first path element
  ```

  On the train alone, `go mod tidy` runs fine. Rebasing onto the train and keeping only the `github.com/cockroachdb/apd/v2` move from indirect to direct (the one dependency change this PR actually needs) avoids that.

I did not re-run the cross-language side (gonka-ai/vllm#62's `scripts/gen_conformance_vectors.py` against these vectors), so bit-parity is as claimed, not independently checked here. vllm#62 is also still unreviewed, which seems worth knowing since the two are meant to land as a pair.

Issue #1450, opened 2026-07-14
Dynamic pricing: price pinned at min by integer truncation; capacity proxy miscalibrated per model


len5ky:
Reporting two verified defects in the dynamic-pricing keeper. Everything below
was checked against tag `release/v0.2.13` (the mainnet chain version) and
against live mainnet state via `node3.gonka.ai` on 2026-07-14. I run a
high-volume reseller on top of Gonka inference, so how the per-token price
responds to load matters directly — that's why I went through this code path
carefully.

Two separate problems, one small and mechanical, one structural. The mechanical
one is the more urgent: as of today, **every model's price on mainnet
is pinned at 1 ngonka, and the utilization-driven update cannot raise it at any
utilization under current params — only a param, code, or state intervention can.**

---

### The code

`CalculateModelDynamicPrice` computes the new price as a decimal and truncates
toward zero — in both branches:

[`inference-chain/x/inference/keeper/dynamic_pricing.go#L205-L216`](https://github.com/gonka-ai/gonka/blob/release/v0.2.13/inference-chain/x/inference/keeper/dynamic_pricing.go#L205-L216)

```go
} else {
    // Above stability zone - increase price (with cap)
    utilizationExcess := utilization.Sub(upperBound)
    adjustmentFactor := one.Add(utilizationExcess.Mul(elasticity))

    // Apply maximum increase cap (2% per block)
    if adjustmentFactor.GreaterThan(maxIncreasePerBlock) {
        adjustmentFactor = maxIncreasePerBlock
    }

    newPriceDec := decimal.NewFromUint64(currentPrice).Mul(adjustmentFactor)
    newPrice = uint64(newPriceDec.IntPart())          // <-- truncation
```

The maximum increase per block is ×1.02 (elasticity 0.05 × max excess 0.40).
With integer prices, `floor(p × 1.02) > p` requires **p ≥ 50**. For any price of
49 ngonka or below, the increase branch is a no-op at *any* utilization:

```
p = 1:   1 × 1.02 = 1.02  → IntPart → 1
p = 49: 49 × 1.02 = 49.98 → IntPart → 49
p = 50: 50 × 1.02 = 51.0  → IntPart → 51   (first price that can move up)
```

Decreases keep working all the way down to the floor clamp (truncation *helps*
them: 2 × 0.98 = 1.96 → 1), and the `MinPerTokenPrice` clamp
[...]

PR #1451, opened 2026-07-14
fix(x/inference): ceil dynamic price increases so sub-50 prices can rise

len5ky:
# fix(x/inference): ceil price increases so sub-50 prices can rise

Part 1 of 2 for #1450 — ceil the dynamic-pricing increase branch so a price
that has decayed below 50 ngonka can rise again under load.

## Fix

Increase branch only (`CalculateModelDynamicPrice`,
`inference-chain/x/inference/keeper/dynamic_pricing.go`):

```go
newPrice = uint64(newPriceDec.Ceil().IntPart())
```

Any block above the stability zone now raises the price by at least the integer
quantum (1). For `p ≥ 50` behavior is nearly identical (ceil vs floor differ by
≤1); exact cap multiples are unchanged (`1000 × 1.02 = 1020.0` → 1020). The
decrease branch stays floored: its floor guarantees downward progress
(`floor(2 × 0.98) = 1`), and ceiling it would create a mirror-image *downward*
trap where a small price could never fall (`ceil(2 × 0.98) = 2`).

<details><summary>Why this is needed — analysis excerpt</summary>

`CalculateModelDynamicPrice` computes the increased price as
`uint64(newPriceDec.IntPart())`. `IntPart()` truncates toward zero. The maximum
increase per block is ×1.02 (elasticity 0.05 × max excess 0.40), so
`floor(p × 1.02) > p` requires **p ≥ 50**:

```
p = 1:   1 × 1.02 = 1.02  → IntPart → 1
p = 49: 49 × 1.02 = 49.98 → IntPart → 49
p = 50: 50 × 1.02 = 51.0  → IntPart → 51   (first price that can move up)
```

So the increase branch is a no-op at *any* utilization once a price decays below
50. The `MinPerTokenPrice` clamp runs *after* truncation, so with
`min_per_token_price = 1` it does not rescue anything. Any stored price in 1–49
is a one-way trap: it can fall in but never climb out through utilization alone.
The only other writers of the price map (the grace-period initializer and the
missing-price → `base_per_token_price` fallback) don't apply to an already-priced
model after the grace boundary, so escaping the trap otherwise requires a param
change, a code change, or a state migration.

**How mainnet got there.** After the grace boundary, with
`base_per_token_price = 100`, `stability_zone_lower_bound = 0.40` and demand below
the stability zone, the price decays at up to −2%/block — and truncation makes it
faster than the continuous math suggests: 100 falls below the p=50 trap line in
26 blocks (≈2 min) and reaches the floor of 1 within 74 blocks (≈6 min; below 50,
`floor(0.98p)` loses a full 1 every block). `grace_period_end_epoch = 90` and the
chain is past epoch 300, so this happened long ago.

**Mainnet evidence.** `GET .../inference/all_model_per_token_prices` on
`node3.gonka.ai` returns `"price": "1"` for **all 8 models** (incl.
`moonshotai/Kimi-K2.6`, `zai-org/GLM-5.2-FP8`, `MiniMaxAI/MiniMax-M2.7`), verified
live 2026-07-14. `RecordInferencePrice` locks this price onto every inference at
start/finish, so all mainnet inference settles at 1/100th of the intended base
price and cannot recover under any demand until intervention.

**Why ceil, not round-half-up:** `round(1 × 1.02) = round(1.02) = 1`. Rounding
only shrinks the trap to `p < 25`; it does not eliminate it. Only ceil guarantees
a strict +1 for every sub-cap increase.

</details>

## Tests

`dynamic_pricing_test.go`, table-driven, added cases:
- `currentPrice=1, U=1.0 → 2` (trap escape; was 1 pre-fix)
- `49 → 50`, `50 → 51`, `1000 → 1020` (cap intact), `1000 @ U=0.61 → 1001`
  (sub-quantum rounds up), `0 → 1` (literal-zero stored price clamps to min).
- A monotonicity guard (`newPrice > oldPrice` on every increase case) and an
  independent quantum property (any `0 < oldPrice < 50` under the 2% cap must
  rise by exactly one quantum) — both derived from the trap arithmetic, not from
  the keeper's own expression, so a regression to plain floor or round-half-up
  fails the suite.

## Consensus impact

`UpdateDynamicPricing` runs for every model in `BeginBlock`, so this changes
on-chain price evolution and is **consensus-breaking**. It must ship in a
coordinated network upgrade. After this fix, prices climb from their current
floor values under demand (≈+1/block initially at 1 ngonka); if the team prefers
a reset to `base_per_token_price` at upgrade time, that would be a separate
param/state decision outside this PR.

## Out of scope (disclosed)

- **uint64 conversion wrap at astronomical prices** — `uint64(...IntPart())`
  is undefined once the decimal exceeds int64. Pre-existing; unchanged by this
  patch (the cast was already there).
- **No utilization > 1.0 clamp test** — at `U = 1.0` the factor is exactly the
  cap (1.02), so the `> maxIncreasePerBlock` clamp binds only above `U = 1.0`;
  that strict-clamp path is not separately exercised.
- **Decrease-branch truncation exceeding the −2% cap at tiny prices** —
  `floor(2 × 0.98) = 1` is a −50% step, far past the relative −2% cap.
  Pre-existing; the decrease branch is intentionally left floored (see above).

### DIFF
--- a/inference-chain/x/inference/keeper/dynamic_pricing.go
+++ b/inference-chain/x/inference/keeper/dynamic_pricing.go
@@ -213,7 +213,8 @@ func (k *Keeper) CalculateModelDynamicPrice(ctx context.Context, modelId string,
 		}
 
 		newPriceDec := decimal.NewFromUint64(currentPrice).Mul(adjustmentFactor)
-		newPrice = uint64(newPriceDec.IntPart())
+		// Sub-integer increases must round up, otherwise small prices can never rise.
+		newPrice = uint64(newPriceDec.Ceil().IntPart())
 
 		k.LogInfo("Price increased - above stability zone", types.Pricing,
 			"modelId", modelId, "utilization", utilization.String(), "excess", utilizationExcess.String(),
--- a/inference-chain/x/inference/keeper/dynamic_pricing_test.go
+++ b/inference-chain/x/inference/keeper/dynamic_pricing_test.go
@@ -20,6 +20,7 @@ func TestCalculateModelDynamicPrice(t *testing.T) {
 		name                string
 		utilization         float64
 		currentPrice        uint64
+		setZeroPrice        bool // force-store a literal 0 current price (harness skips SetModelCurrentPrice when currentPrice==0)
 		stabilityLowerBound float64
 		stabilityUpperBound float64
 		priceElasticity     float64
@@ -27,6 +28,7 @@ func TestCalculateModelDynamicPrice(t *testing.T) {
 		basePerTokenPrice   uint64
 		expectedOldPrice    uint64
 		expectedNewPrice    uint64
+		expectIncrease      bool
 		expectError         bool
 	}{
 		{
@@ -66,6 +68,7 @@ func TestCalculateModelDynamicPrice(t *testing.T) {
 			basePerTokenPrice:   100,
 			expectedOldPrice:    100,
 			expectedNewPrice:    101, // 100 * (1 + (0.80-0.60) * 0.05) = 100 * 1.01 = 101
+			expectIncrease:      true,
 			expectError:         false,
 		},
 		{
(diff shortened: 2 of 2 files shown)


qdanik:
@len5ky we're currently handling an inference through the gateway. This way is deprecated for now.

redstartechno (commented):
Reviewed this against `main` (`379bebced`) and ran the tests locally, since this PR has never had a CI run — more on that at the end.

## The arithmetic checks out

`CalculateModelDynamicPrice` (`inference-chain/x/inference/keeper/dynamic_pricing.go:213`) truncates with `IntPart()`, the max increase factor is `1 + (1 - 0.60) × 0.05 = 1.02`, and the `minPrice` clamp runs *after* truncation (`:224-229`), so with `MinPerTokenPrice = 1` it cannot lift a truncated result. `floor(p × 1.02) > p` does require `p ≥ 50`, and there is no upper price clamp that `Ceil()` could now collide with. The described trap is real in the code as it stands.

## Local test evidence

`go test ./x/inference/keeper/ -run TestCalculateModelDynamicPrice` (go1.24.2):

- on this branch: **ok** (0.16s)
- new test file applied on unpatched `main`: **FAIL** — `Minimum_price_escapes_increase_truncation_trap` (expected 2, actual 1), `Sub-50_price_increases_at_capped_boundary` (expected 50, actual 49), `Sub-quantum_increase_rounds_up`

So the tests do discriminate against the old code. Full `./x/inference/keeper/` and `./x/inference/calculations/` are green on the branch (15.8s / 3.7s), `go vet` clean.

## One correction to the impact claim

The PR body says `RecordInferencePrice` "locks this price onto every inference at start/finish, so all mainnet inference settles at 1/100th of the intended base price". On current `main` I do not think that holds any more:

- `RecordInferencePrice` (`dynamic_pricing.go:271`) has **no non-test caller** in the tree; `msg_server_start_inference.go` / `msg_server_finish_inference.go` no longer exist, and the commit that wired those calls (`fdb29d8a3`) is not an ancestor of `main`.
- The devshard escrow path charges from a governance param instead: `msg_server_create_devshard_escrow.go:101` copies `ep.TokenPrice` (`DevshardEscrowParams`, default `1` at `types/params.go:103`), which never reads the dynamic price map.

That is presumably what @qdanik meant above. But I don't think it makes the PR moot, because the dynamic price is still computed and still published:

- `UpdateDynamicPricing` still runs in `BeginBlock` every block (`module/module.go:214`) and writes the per-model price to consensus state.
- The public pricing endpoint overrides the legacy `UnitsOfComputePerToken × UnitOfComputePrice` quote with the dynamic price whenever dynamic pricing is on (`decentralized-api/internal/server/public/get_pricing_handler.go:56-62`, data from `GetAllModelPerTokenPrices` at `:154`), and also publishes capacity and utilization.

So the observable damage is a **wrong advertised price** (and a price trajectory that can only decay), not mis-settled funds — plus a landmine if the chain-side price path is ever re-enabled. Rewriting that paragraph in those terms would make this much easier to accept: it stops the PR from resting on a settlement claim a maintainer can immediately refute, and it still justifies the fix.

## On the fix itself

`Ceil()` on the increase branch does guarantee `newPrice ≥ oldPrice + 1` for any `0 < p < 50` under an above-zone utilization, so the trap is closed. Two notes:

1. It trades the trap for a cap violation at tiny prices: `p = 1` goes to `2`, a +100% block against a documented "max 2% per block". You disclose the mirror-image case for the decrease branch, but the increase-side comment at `:215` and the `maxIncreasePerBlock` comment at `:174` still read as if the relative cap is absolute. Worth one line saying the cap is relative and is superseded by the ±1 integer quantum below ~50 ngonka.
2. The root cause is the coarseness of the integer ngonka price unit, not the rounding mode. Two alternatives that keep the relative cap exact: store the price in a scaled fixed-point unit, or set `MinPerTokenPrice` above the trap line (≈50 at the current elasticity/bounds) so the truncation is never reached. Both are bigger than this patch; I'd still go with `Ceil()` as the minimal fix, but it's worth stating that choice explicitly since maintainers will have to live with the cap asymmetry.

## Two process notes

- **This PR has zero check runs.** Not "failing" — never started. We hit the same thing on our own #1556, which sat approved for 40 days with no CI; it is the first-time-contributor workflow-approval gate, and only a maintainer can release it. Worth asking for that directly, because right now nothing here has been verified by CI at all.
- **Base branch.** This targets `main`, but releases are cut on train branches (`upgrade-v0.2.16` → #1535) and `main` has not moved since 2026-08-17. A consensus-breaking change may well need to sit on the train instead. We lost ~50 days on a PR stranded on a base that had quietly been retired, so I'd ask the maintainers which base they want before this ages any further.

Not approving — I'm a fellow contributor, not a maintainer — but for what it's worth the analysis holds up and the tests are honest.

PR #1459, opened 2026-07-16
fix(devshard/user): guard lazy finalizeClients init against data race

redstartechno:

`Session.getFinalizeClients` performs an unsynchronized lazy initialization of the shared `s.finalizeClients` field, racing when the same session is hit concurrently. This produces a data race (flagged by `-race`) and can yield a torn/partial client list.

In `devshard/user/session.go`, `getFinalizeClients` does a check-then-write outside any lock — `if s.finalizeClients != nil { return ... }` … else `s.finalizeClients = make(...)` — while every other `Session` field is guarded by `s.mu` (a `sync.Mutex`). It is reached from `CollectSignatures`, which is invoked both from `Finalize` **and directly from the public HTTP endpoint** `handleCollectSignatures` (`POST /v1/debug/signatures/collect`, `cmd/devshardctl/proxy.go`). Two concurrent requests against one session race on the field. `go test -race` reproduces it deterministically (trace below).

Guard the lazy init with the session's existing mutex:

```go
s.mu.Lock()
defer s.mu.Unlock()
```

Verified this is deadlock-free: the sole caller (`CollectSignatures`) does not hold `s.mu` at the call site (it acquires the lock only later), and the returned slice is assigned once and never mutated, so it is safely published under the lock.

Low. Adds one mutex acquisition on a path that already synchronizes elsewhere; matches the struct's uniform locking discipline (same pattern as neighboring methods). No new field, no lock-ordering change. Deadlock ruled out by confirming no caller holds `s.mu` when reaching this method.

Added `TestUser_GetFinalizeClientsConcurrent`: 16 goroutines released simultaneously, each calling `getFinalizeClients`, asserting a consistent result. It reports a **DATA RACE** on the current code under `-race` and **passes cleanly with the fix**.

- `devshard/user/session.go`
- `devshard/user/session_test.go` (new test)

## Testing & evidence

`GOTOOLCHAIN=go1.24.2 go test ./user/... -race` and `go vet ./user/...`, native.

### Before
```
WARNING: DATA RACE
Write at 0x00c0001478e0 by goroutine 16:
  devshard/user.(*Session).getFinalizeClients()
      devshard/user/session.go:1245 +0x96
Previous read at 0x00c0001478e0 by goroutine 25:
  devshard/user.(*Session).getFinalizeClients()
      devshard/user/session.go:1253 +0x224
...
--- FAIL: TestUser_GetFinalizeClientsConcurrent (0.03s)
    testing.go:1490: race detected during execution of test
FAIL	devshard/user	0.084s
```

### After
```
ok  	devshard/user	5.412s   # full package under -race
# go vet ./user/...  -> clean
```

### DIFF
--- a/devshard/user/session.go
+++ b/devshard/user/session.go
@@ -1234,8 +1234,11 @@ func (s *Session) hasQuorum(nonce uint64, threshold uint32) bool {
 // getFinalizeClients returns a client list with admission control stripped.
 // Built lazily and cached on s.finalizeClients. Each client that implements
 // a ClearAdmission method gets a shallow copy with admission disabled;
-// others are used as-is.
+// others are used as-is. Safe for concurrent use: CollectSignatures can run
+// simultaneously from Finalize and the debug collect-signatures endpoint.
 func (s *Session) getFinalizeClients() []HostClient {
+	s.mu.Lock()
+	defer s.mu.Unlock()
 	if s.finalizeClients != nil {
 		return s.finalizeClients
 	}
--- a/devshard/user/session_test.go
+++ b/devshard/user/session_test.go
@@ -1157,3 +1157,35 @@ func TestFinalize_SignatureStatus_InsufficientQuorum(t *testing.T) {
 		require.Less(t, finalEntry.SigWeight, uint32(4))
 	}
 }
+
+// TestUser_GetFinalizeClientsConcurrent verifies the lazy finalizeClients
+// init is safe under concurrent access. CollectSignatures (which triggers
+// the init) is reachable simultaneously from Finalize and from the public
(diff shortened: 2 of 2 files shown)


a-kuprin (approved):

redstartechno:
Closing: the `gateway-v4` force-push on 2026-08-30 (`da9a69848`) rewrote `Session.getFinalizeClients` to build the client list fresh on every call, and the cached `s.finalizeClients` field no longer exists. The unsynchronized lazy initialization this PR guarded is gone by design, so there is nothing left to rebase — the remaining test would only exercise code that cannot race any more.

Thanks for the review, @a-kuprin.

PR #1460, opened 2026-07-16
fix(devshard/host): return challenge receipt without blocking on execution

redstartechno:

`ChallengeReceipt` blocks on the full inference before returning the receipt, despite its contract being fire-and-forget. For any multi-second/minute inference, the verifier's bounded receipt RPC times out first, so a **live executor is wrongly voted as timed out**.

In `devshard/host/host.go`, `ChallengeReceipt` calls `executeAsync`, which is `_, _ = h.RunExecution(ctx, job)` — synchronous, no `go`. But:

- The doc on `ChallengeReceipt` says it "signs an executor receipt, and **triggers async execution**".
- The `ExecutorClient` contract in `timeout.go` says it "returns a signed receipt … **Also triggers execution** so the inference actually completes" — receipt is the response; execution is a side effect, not a prerequisite.
- The function is literally named `executeAsync`; its sibling `validateAsync` already runs in a background goroutine.

Concrete harm: the verifier wraps the call in `VerifyTimeout` (default 3m). `VerifyRefusedTimeout` treats a client error as "executor unreachable → accept the timeout". A live executor whose long inference blocks the receipt response blows that RPC deadline and is wrongly timed out. The receipt (the liveness proof) never arrives even though the executor is healthy.

Launch the execution in the background so the receipt returns immediately:

```go
go h.executeAsync(context.WithoutCancel(ctx), job)
```

`context.WithoutCancel` is required: the caller passes the request context, which echo cancels once the handler returns — a plain `go` with that context would kill the inference the contract requires to complete. WithoutCancel preserves observability values while surviving the RPC return (in-repo precedent: `mlnode/retry.go`).

Safety: the dedup guard `h.executing[inferenceID]` is set under `h.mu` **before** the job is returned, so the goroutine cannot double-execute; `RunExecution` already logs all its failures via `observability.FailReceiptOrphan` (nothing newly swallowed); the goroutine ends when execution returns and the map entry is cleared on defer (no leak).

- Execution now completes slightly later relative to the receipt. This is by design and correct: `VerifyRefusedTimeout` rejects on the receipt alone, and `VerifyExecutionTimeout` is a separate, later check anchored at `ConfirmedAt + ExecutionTimeout` — no path requires `MsgFinishInference` present before the receipt returns.
- `context.WithoutCancel` also strips the request **deadline** from `engine.Execute`. This matches the established `validateAsync` / `context.Background()` background-execution norm, and real engines self-bound their runtime — noted here for transparency.

Added `TestHost_ChallengeReceipt_ReturnsBeforeExecutionCompletes` with a gated (`blockingEngine`) execution seam: asserts `ChallengeReceipt` returns the receipt while the engine is still busy, then that execution still triggers and `MsgFinishInference` reaches the mempool afterward. It **fails on current code** (receipt blocks, 2.03s timeout) and **passes with the fix**.

- `devshard/host/host.go`
- `devshard/host/host_test.go` (new test + seam)

## Testing & evidence

`GOTOOLCHAIN=go1.24.2 go test ./host/ -race`, `go test ./host/ ./transport/ ./user/`, `go vet`, native.

### Before
```
--- FAIL: TestHost_ChallengeReceipt_ReturnsBeforeExecutionCompletes (2.03s)
    ChallengeReceipt blocked on execution: receipt not returned while engine is busy
FAIL	devshard/host
```

### After
```
ok  	devshard/host      # -race
ok  	devshard/transport
ok  	devshard/user
# go vet -> clean on all three
```

### DIFF
--- a/devshard/host/host.go
+++ b/devshard/host/host.go
@@ -840,8 +840,14 @@ func (h *Host) signReceipt(req HostRequest) ([]byte, int64, *devshard.ExecuteReq
 
 // executeAsync runs inference and adds MsgFinishInference to the mempool.
 // Delegates to RunExecution which also caches the response body for reconnection.
+// The caller must not wait for the inference: it holds a request context whose
+// deadline is far shorter than a long generation, and the executor receipt is
+// already signed. Detaching from cancellation keeps MsgFinishInference on track
+// even after the challenging host hangs up.
 func (h *Host) executeAsync(ctx context.Context, job *devshard.ExecuteRequest) {
-	_, _ = h.RunExecution(ctx, job)
+	go func() {
+		_, _ = h.RunExecution(context.WithoutCancel(ctx), job)
+	}()
 }
 
 func (h *Host) ReleaseExecution(inferenceID uint64) {
--- a/devshard/host/host_test.go
+++ b/devshard/host/host_test.go
@@ -1235,6 +1235,43 @@ func (e *blockingInferenceEngine) Execute(ctx context.Context, req devshard.Exec
 	}
 }
 
+// ctxCapturingInferenceEngine blocks like blockingInferenceEngine, but also
+// records the ctx Execute received, so a test can inspect it (e.g. after the
+// request context is cancelled) instead of inferring detachment indirectly.
+type ctxCapturingInferenceEngine struct {
+	inner   *stub.InferenceEngine
+	started chan struct{}
+	release chan struct{}
+
+	mu  sync.Mutex
+	ctx context.Context
+}
+
+func newCtxCapturingInferenceEngine() *ctxCapturingInferenceEngine {
+	return &ctxCapturingInferenceEngine{
+		inner:   stub.NewInferenceEngine(),
+		started: make(chan struct{}, 1),
+		release: make(chan struct{}),
+	}
+}
+
+func (e *ctxCapturingInferenceEngine) Execute(ctx context.Context, req devshard.ExecuteRequest) (*devshard.ExecuteResult, error) {
+	// Publish ctx before signaling started, so the test can safely read it
+	// as soon as it observes started.
+	e.mu.Lock()
+	e.ctx = ctx
+	e.mu.Unlock()
+	e.started <- struct{}{}
+	<-e.release
+	return e.inner.Execute(ctx, req)
+}
+
+func (e *ctxCapturingInferenceEngine) executionCtx() context.Context {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	return e.ctx
+}
+
 func TestHost_ChallengeReceipt_DoesNotBlockOnExecution(t *testing.T) {
 	// Regression: ChallengeReceipt used to run the inference synchronously, so a
 	// slow execution held the receipt past the verifier's timeout and a live
@@ -1285,6 +1322,53 @@ func TestHost_ChallengeReceipt_DoesNotBlockOnExecution(t *testing.T) {
 	require.Eventually(t, func() bool {
 		return findMempoolFinish(h.MempoolTxs()) != nil
 	}, 5*time.Second, 10*time.Millisecond, "background execution must queue MsgFinishInference")
+}
+
+func TestHost_ChallengeReceipt_ExecutionSurvivesRequestCancel(t *testing.T) {
+	// Regression: ChallengeReceipt must detach execution from the request context
+	// (context.WithoutCancel), so a challenger that cancels/hangs up after
+	// receiving the receipt cannot abort the executor's in-flight inference.
+	hosts := []*signing.Secp256k1Signer{testutil.MustGenerateKey(t), testutil.MustGenerateKey(t), testutil.MustGenerateKey(t)}
+	user := testutil.MustGenerateKey(t)
+	group := testutil.MakeGroup(hosts)
+	config := testutil.DefaultConfig(len(hosts))
+	verifier := signing.NewSecp256k1Verifier()
+	sm, err := state.NewStateMachine("escrow-1", config, group, 10000, user.Address(), verifier, testutil.MustMemoryStore(t, "escrow-1", user.Address(), config, group, 10000))
+	require.NoError(t, err)
+	engine := newCtxCapturingInferenceEngine()
+	h, err := NewHost(sm, hosts[1], engine, "escrow-1", group, nil, WithGrace(10))
+	require.NoError(t, err)
+
+	diff := testutil.SignDiff(t, user, "escrow-1", 1, []*types.DevshardTx{testutil.StartTx(1)})
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	errCh := make(chan error, 1)
+	go func() {
+		_, _, cErr := h.ChallengeReceipt(ctx, 1, defaultPayload(), []types.Diff{diff})
+		errCh <- cErr
+	}()
+
+	select {
+	case <-engine.started:
+	case <-time.After(5 * time.Second):
+		t.Fatal("execution never started")
+	}
+	// executeAsync returns immediately after spawning execution, so
+	// ChallengeReceipt has already returned by the time Execute is running.
+	require.NoError(t, <-errCh)
+
+	execCtx := engine.executionCtx()
+	cancel() // synchronous: any derived context has Err() set once this returns
+	require.NoError(t, execCtx.Err(), "execution context must not observe the request cancellation")
+	require.Nil(t, execCtx.Done(), "execution context must be uncancellable")
+
+	// Execution is still alive; let it finish and confirm it actually
+	// completes, not just that it avoided the cancellation.
+	close(engine.release)
+	require.Eventually(t, func() bool {
+		return findMempoolFinish(h.MempoolTxs()) != nil
+	}, 5*time.Second, 10*time.Millisecond, "background execution must queue MsgFinishInference despite request cancel")
 }
 
 func TestWarmKey_HostFindsSlotByWarmKey(t *testing.T) {
--- a/devshard/host/host_test.go
+++ b/devshard/host/host_test.go
@@ -1209,6 +1209,84 @@ func TestHost_ChallengeReceipt_AlreadyFinished(t *testing.T) {
 	require.Equal(t, 1, engine.calls, "engine should not be called again")
 }
 
+// blockingInferenceEngine holds Execute open until released, so tests can observe
+// whether the caller waits for the inference to finish.
+type blockingInferenceEngine struct {
+	inner   *stub.InferenceEngine
+	started chan struct{}
+	release chan struct{}
+}
+
+func newBlockingInferenceEngine() *blockingInferenceEngine {
+	return &blockingInferenceEngine{
+		inner:   stub.NewInferenceEngine(),
+		started: make(chan struct{}, 1),
+		release: make(chan struct{}),
+	}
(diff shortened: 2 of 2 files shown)


redstartechno:
Note on the failed `Build and Test API Wrapper` check: this appears to be a pre-existing flaky test unrelated to this PR. The failure is:

```
mock.go:89: TryLock called more than once
--- FAIL: TestRegularPocScenario (0.00s)
FAIL    decentralized-api/internal/event_listener      11.075s
```

`TestRegularPocScenario` lives in `decentralized-api/internal/event_listener` and asserts a call count on the `mlnodeclient` mock (`mock.go:89`) — a timing-sensitive assertion. This PR only touches `devshard/host/host.go` and its test; the failing test constructs its own mock ML node client and does not exercise the host challenge-receipt path, so the new background goroutine here cannot reach it. The same job/step also failed the same day on an unrelated branch (`feature/empty-validator-set-fallback`, run 29468383638), which supports the flake reading.

A re-run should clear the check. Happy to dig further if you suspect any connection with this PR's change.

a-kuprin:
While this PR is correct I think there should be more changes.

It is hard to estimate tiny work and a real value of the change, while there are a lot more to do on the path.

Here what I have found reviewing this PR and digging a little into the path (the PR should, I think, **close at least 1st problem**).

### 1. Unify accept/execute paths

**Problem:** User SSE (`signReceipt`) and timeout challenge (`challengeReceiptLocked`) both do verify → sign receipt → dedup → maybe execute, but as duplicated logic that already diverges (ConfirmStart, cache replay, soft-fail semantics). Challenge also blocks the receipt RPC on full ML, unlike SSE.

**Task:** Extract one shared “accept as executor” helper; thin SSE vs challenge wrappers; challenge returns receipt without waiting on ML.

---

### 2. Gateway disconnect must not kill executor proof

**Problem:** Today, if the host→client (or gateway→host) stream dies in a way that cancels host→ML, generation can abort. Then the executor may never publish a clean `MsgFinishInference`, so the work isn’t accountable/validatable even though the host had accepted it. A drop by `devshardctl`/user must not be a reason the executor fails to produce execution proof.

**Task:** Decouple ML execution lifetime from the outbound user/gateway stream. After receipt/accept, run inference to completion (or controlled failure) under a host-owned context; always try to land `MsgFinishInference` (+ cache body) even if the client has gone. Treat stream write failure as delivery loss, not work cancellation.

---

### 3. Same-nonce reconnect for `devshardctl`

**Problem:** Host can re-serve receipt / `CachedResponseBody` on the same `StartInference`, but `devshardctl` never resends the same nonce—it opens new attempts. After an accidental drop, the user side can’t reattach to in-flight or finished work on that nonce.

**Task:** On disconnect/retry, allow `devshardctl` to resend the **same nonce** with the same payload/params (reject mismatch via prompt/params hash). Host validates equality, skips re-execute if already running/done, and returns receipt + live/cached stream (`CachedResponseBody` when complete).

---

### 4. HA / fault-tolerance across `devshardd` instances

**Problem:** In-flight inference state (payload, execution progress, cached response) lives in process memory. If a connection drops mid-serve, or `devshardd` reboots/fails over, another instance can’t resume: no durable payload, no way to reattach to an ML job still running, no disk `CachedResponseBody`. Work is lost or must be blindly restarted without shared proof.

**Task:** Persist, until the inference is finished/settled: payload + parameters (for same-hash reconnect), execution/receipt metadata, and `CachedResponseBody` (on disk). On another `devshardd`: if ML still running, reconnect to that job; else recreate from stored payload; serve reconnecting clients the live stream or cached body. Goal: drop/reboot/failover doesn’t erase the path to a validatable `MsgFinishInference`.

redstartechno:
Thanks for the detailed walk-through of the path — this is really useful context, and I agree the receipt-async change on its own is a thin slice of a larger problem.

**On #1 (unify accept/execute paths):** happy to take this on in this PR. The current change already covers the "challenge returns the receipt without waiting on ML" half; I'll extend it to extract a shared "accept as executor" helper and thin the SSE (`signReceipt`) and challenge (`challengeReceiptLocked`) sides down to wrappers over it, so the ConfirmStart / cache-replay / soft-fail divergence you called out gets removed rather than duplicated further. I'll push that as follow-up commits here.

**On #2–#4:** agree these are a larger fault-tolerance/HA effort — a host-owned execution context that always tries to land `MsgFinishInference` even after a client drop (#2), same-nonce reconnect in `devshardctl` (#3), and durable cross-`devshardd` state for reboot/failover (#4). Each reads as its own reviewable unit rather than something to fold into this PR, so tracking them under #1466 makes sense to me. I'm glad to pick up #2 as a follow-up once the unification here lands, if that ordering works for you.

One question before I refactor #1: do you have a preferred home/signature for the shared helper — keep it in `devshard/host` next to `host.go`, or somewhere both the SSE and challenge paths already import — so I match how you'd want it structured?

a-kuprin:
> I'm glad to pick up #2 as a follow-up once the unification here lands, if that ordering works for you.

Yes take it if you wish

a-kuprin:
> do you have a preferred home/signature for the shared helper — keep it in `devshard/host` next to `host.go`, or somewhere both the SSE and challenge paths already import — so I match how you'd want it structured?

Keep it at `devshard/host` as both call sites already live at the package

redstartechno:
Pushed the item #1 unification as a follow-up commit (`da404e855`), kept in `devshard/host` as you suggested.

**What it does:** the shared "accept as executor" core — `VerifyPayload` → sign receipt → `executing` dedup → replay check → mark-executing + build job — now lives in one helper, `acceptAsExecutorLocked` (caller must hold `h.mu`, which both call sites already do). `signReceipt` (SSE) and `challengeReceiptLocked` are thin wrappers: they keep only their own lookup (diff-by-nonce + `StartInference` scan vs `sm.GetInference` + `StatusPending`) and pass their divergent behavior in explicitly via `acceptParams`.

**Divergences you called out are now explicit parameters instead of drifting duplicates:**
- `addConfirmStart` — `MsgConfirmStart` is added on the SSE path only, as before.
- `softFailVerify` — SSE keeps the hard `VerifyPayload` error; challenge keeps the documented soft-fail (verifier has already checked the payload), with the rationale comment retained.
- `alreadyFinished` hook — replay sources stay separate: SSE replays from the `completedResponses` cache, challenge scans the mempool for an existing `FinishInference`.
- Execution trigger is unchanged: SSE still runs synchronously on the live request (the job carries the response writer for streaming), challenge keeps the detached `go executeAsync(context.WithoutCancel(...))` from the original commit.

**Evidence:** `go build`, `go vet`, and `go test ./host/ -race -count=1` all green (`ok devshard/host 2.007s`); the five pre-existing accept-path tests pass unchanged (the test diff is additions-only). Added `TestHost_AcceptDivergence_BadPayload`, which feeds the same mismatched payload to both paths and pins the divergence: SSE returns `ErrPayloadMismatch`, challenge soft-fails with a nil error and no receipt — the challenge-side bad-payload branch had no coverage before.

On #2 — will pick it up as a follow-up PR once this lands, as discussed.

redstartechno:
Recut onto current main (`da404e855 → 70fa6d78d`) after the 0.2.14 merge rewrote `devshard/host` and made the branch conflict. Two notes on what changed:

**The bug still exists on the new main, so the fix is re-applied in minimal form.** On v4, `ChallengeReceipt` calls `executeAsync` before returning, but `executeAsync` is now a plain synchronous `RunExecution` call — so a live executor's long inference still blocks the receipt past the verifier's 3-minute `VerifyTimeout`, and `VerifyRefusedTimeout` still treats that as "executor unreachable". The recut fix makes `executeAsync` match its name and doc comment:

```go
func (h *Host) executeAsync(ctx context.Context, job *devshard.ExecuteRequest) {
	go func() {
		_, _ = h.RunExecution(context.WithoutCancel(ctx), job)
	}()
}
```

It's now a 4-line production change: v4 already has the `h.executing` dedup guard and the receipt/job split (`challengeReceiptLocked`), so nothing else from the old branch is needed. `executeAsync` has exactly one caller (`ChallengeReceipt`); the SSE deferred path calls `RunExecution` directly and is unaffected.

**The earlier unification commit is dropped.** The `acceptAsExecutorLocked` helper from our previous iteration doesn't map onto v4's structure (v4 refactored `RunExecution` as the shared execution primitive instead). If you'd still like the `signReceipt`/`challengeReceiptLocked` accept-paths unified on the new structure, I'm happy to do that as a separate follow-up — but it seemed wrong to bundle a structural refactor into the recut of an unreviewed fix.

**Evidence** (Windows native + WSL for `-race`, devshard module):
- Fail-first: with the new regression test `TestHost_ChallengeReceipt_DoesNotBlockOnExecution` on unfixed v4 code, `ChallengeReceipt` blocks on execution → `FAIL devshard/host 5.123s`. With the fix: `--- PASS (0.01s)`.
- Full package: `ok devshard/host` build/vet clean; new test stable under `-race -count=30` (30/30).
- Dependent packages green: `devshard/protocol`, `devshard/transport`, `devshard/user`.

**Tradeoff note for transparency:** `context.WithoutCancel` means a client disconnect no longer propagates cancellation into `engine.Execute` — intentional, since the receipt is already signed and `MsgFinishInference` must still land, but it removes an incidental early-abort path. There is no engine-level deadline wrapping `Execute` on either the old or new code path; if you'd like one, that would make a good follow-up.

a-kuprin:
I have rebased it and scheduled to devshard v5

a-kuprin:
@redstartechno 
Please add one nit test hardening:

Test does not prove `WithoutCancel`. It uses `context.Background()` and a gate. It proves “receipt returns while Execute is in flight,” which is the bug. It does not prove that cancelling the request context leaves execution alive.

redstartechno:
Good catch — the previous test only pinned the async half. Pushed the hardening in `2ebe60b4f` (test-only; `devshard/host/host.go` is byte-identical to the previous head).

The new test `TestHost_ChallengeReceipt_ExecutionSurvivesRequestCancel` cancels the request context after execution has started, and asserts on the context the engine actually received:

```go
execCtx := engine.executionCtx()
cancel() // synchronous: any derived context has Err() set once this returns
require.NoError(t, execCtx.Err(), "execution context must not observe the request cancellation")
require.Nil(t, execCtx.Done(), "execution context must be uncancellable")
```

A small `ctxCapturingInferenceEngine` publishes the ctx `Execute` received (under a mutex, before signalling start), so the assertion is on the real execution context rather than on an indirect symptom. Since `cancel()` propagates synchronously and `context.WithoutCancel` yields a context whose `Done()` is nil, this needs no waiting or settle window — there is no timing element, and no way for it to pass vacuously on a loaded runner. The test then releases the engine and still requires `MsgFinishInference` to land, so it covers both the structural property and the fact that execution really does run to completion after the caller has gone.

**Evidence** (WSL, devshard module):

- Fail-first: with `context.WithoutCancel` removed from `executeAsync`, the new test fails with
  `Received unexpected error: context canceled` / `execution context must not observe the request cancellation`,
  while the existing `TestHost_ChallengeReceipt_DoesNotBlockOnExecution` still **passes** on that same broken code — exactly the gap you described.
- With the fix in place: `--- PASS`, and 30/30 under `go test ./host/... -run TestHost_ChallengeReceipt_ExecutionSurvivesRequestCancel -race -count=30`.
- The existing `DoesNotBlockOnExecution` is unaffected and still green under `-race`: the capturing engine is a separate type, `blockingInferenceEngine` is untouched.
- `go vet ./host/...` clean.

PR #1462, opened 2026-07-16
fix: bind attempt to bridge message hash

x0152:
Bind  attempt  into the BLS message hash to prevent cross-attempt partial-sig replay


a-kuprin:
We should withhold it and first prepare how to upgrade deployed bridge contract.

Also contract should be backward compatible and support old message format, for step-by-step deployment

PR #1469, opened 2026-07-17
feat(dapi): public mlnode metrics federation — /v1/mlnodes/metrics

clanster:
## What / why

Adds a public, cached, rate-limited endpoint `GET /v1/mlnodes/metrics` on dapi that federates Prometheus metrics from every ML node a participant operates into a single exposition. This gives the network and operators one scrape target per participant for GPU / host / vLLM health, without exposing the ML nodes themselves.

## Design

`observability.MLNodeMetricsHandler` enumerates the participant's ML nodes via the broker and concurrently scrapes each node's metrics exporter (not raw vLLM — allowlist filtering and normalization stay node-side). Results are merged per MetricFamily with a `node_id` label injected on every series, producing a promtool-clean exposition. The snapshot is cached (10 s TTL) and single-flighted, so any number of external scrapers costs the ML nodes at most one fan-out per TTL. Public-path rate limiting is enforced by a dedicated nginx zone shipped in this PR.

## Contract points

- **Three-state node semantics**: `mlnode_up{node_id} 1` = scraped OK; `mlnode_up{node_id} 0` = reachable but scrape failed or over-cap; a node answering 404 (metrics disabled, or an image predating the exporter) is **absent from the output entirely** — no zeros, no up-series.
- Per-node ceilings: 2 MiB body / 5000 series; an over-cap snapshot is rejected and the node reported down.
- Dedicated nginx `metrics_zone` (default 10 r/m, burst 30, per IP; tunable via `METRICS_RATE_LIMIT_RPM` / `METRICS_BURST`), so scraping consumers never compete with inference/API clients in the shared `api_zone`. CORS is enabled for browser consumers, and the endpoint is deliberately decoupled from the API kill switch.
- Kill switch: `DAPI_API__MLNODE_METRICS_DISABLED=true` disables the endpoint via env, no rebuild.

A few details deviate deliberately from the original design doc (pull-through cache instead of a background poller; endpoint path; per-node timeout). The ADR amendment covering these lands separately together with the operator documentation.

## Validation

- Unit tests included (`observability/mlnode_metrics_test.go`): family merge across nodes with correct `node_id` labels and up-state, 404-node absence, series-cap rejection; merged output re-parses cleanly.
- Live production validation on a production network node with a 15-node fleet (mixed models and image versions): federated output well inside the ceilings; a real node outage and its recovery observed with correct `mlnode_up` transitions; pre-exporter nodes correctly absent; exposition formats, gzip, and content negotiation verified against real Prometheus scrapes; rate-limit isolation confirmed (exhausting `metrics_zone` leaves `api_zone` endpoints unaffected).

## Risk

Low. New endpoints only — no changes to existing routes or their behavior. ML nodes are reached only from the dapi process; memory is bounded by the per-node ceilings and at most one snapshot rebuild per TTL.

## Rollout

Ships with v0.2.14. Safe with mixed fleets: the dapi-first upgrade order is tolerated by design — ML nodes without the exporter answer 404 and are simply absent from the output.

## Rollback

`DAPI_API__MLNODE_METRICS_DISABLED=true` (env only, no rebuild); the nginx `metrics_zone` independently throttles the public path. Full revert = revert this PR.


proxy/entrypoint.sh
```diff
@@ -445,6 +445,13 @@ GLOBAL_BURST=${GLOBAL_BURST:-5000}
 # Gonka API (Standard/Punisher)
 # Default: 10r/m + 600 burst = 1 hr recovery
 GONKA_API_RATE_LIMIT_VAL=${GONKA_API_RATE_LIMIT_RPS:-10}
+# Dedicated budget for the public mlnode metrics federation endpoint:
+# separated from api_zone so scraping consumers (Prometheus/Grafana) do not
+# compete with inference/API clients for the same bucket.
+METRICS_RATE_LIMIT_VAL=${METRICS_RATE_LIMIT_RPM:-10}
+METRICS_RATE_UNIT=${METRICS_RATE_UNIT:-m}
+METRICS_BURST=${METRICS_BURST:-30}
+METRICS_CONN_LIMIT=${METRICS_CONN_LIMIT:-20}
```
a-kuprin, line 454: About defaults. As this is per IP limits, we can even more limit them:

`METRICS_RATE_LIMIT_RPM` 6 (or even 3–4 r/m if scrape_interval ≥ 15–30s) is more then enough, according to cache TTL = 10 seconds
`METRICS_BURST` 5
`METRICS_CONN_LIMIT` 3

a-kuprin (approved):

Issue #1466, opened 2026-07-17
Better devshardd inference handling


a-kuprin:
While reviewing this PR: https://github.com/gonka-ai/gonka/pull/1460 I have found couple of problems/tasks, and listed them at PR comment https://github.com/gonka-ai/gonka/pull/1460#issuecomment-5001774184

1. Unify accept/execute paths (should be targeted by https://github.com/gonka-ai/gonka/pull/1460 PR)

Problem: User SSE (signReceipt) and timeout challenge (challengeReceiptLocked) both do verify → sign receipt → dedup → maybe execute, but as duplicated logic that already diverges (ConfirmStart, cache replay, soft-fail semantics). Challenge also blocks the receipt RPC on full ML, unlike SSE.

Task: Extract one shared “accept as executor” helper; thin SSE vs challenge wrappers; challenge returns receipt without waiting on ML.

2. Gateway disconnect must not kill executor proof

Problem: Today, if the host→client (or gateway→host) stream dies in a way that cancels host→ML, generation can abort. Then the executor may never publish a clean MsgFinishInference, so the work isn’t accountable/validatable even though the host had accepted it. A drop by devshardctl/user must not be a reason the executor fails to produce execution proof.

Task: Decouple ML execution lifetime from the outbound user/gateway stream. After receipt/accept, run inference to completion (or controlled failure) under a host-owned context; always try to land MsgFinishInference (+ cache body) even if the client has gone. Treat stream write failure as delivery loss, not work cancellation.

3. Same-nonce reconnect for devshardctl

Problem: Host can re-serve receipt / CachedResponseBody on the same StartInference, but devshardctl never resends the same nonce—it opens new attempts. After an accidental drop, the user side can’t reattach to in-flight or finished work on that nonce.

Task: On disconnect/retry, allow devshardctl to resend the same nonce with the same payload/params (reject mismatch via prompt/params hash). Host validates equality, skips re-execute if already running/done, and returns receipt + live/cached stream (CachedResponseBody when complete).

4. HA / fault-tolerance across devshardd instances

Problem: In-flight inference state (payload, execution progress, cached response) lives in process memory. If a connection drops mid-serve, or devshardd reboots/fails over, another instance can’t resume: no durable payload, no way to reattach to an ML job still running, no disk CachedResponseBody. Work is lost or must be blindly restarted without shared proof.

[...]

PR #1482, opened 2026-07-20
Devshard 0.2.14 v4

a-kuprin:
# Release guide: `devshard-0.2.14-v4`

Operator-facing notes for the v4 line: multi-instance HA, Postgres storage,
gRPC-only gateway chain transport, versionless observability, **same-name
binary rolling updates** (blue/green + drain), and rollout constraints for
mixed pre-v4 / v4 estates.

Detailed deploy verification: [v4-deploy-test-plan.md](./v4-deploy-test-plan.md).
Architecture: [high-availability-architecture.md](./high-availability-architecture.md).
Observability: [pr-versionless-observability.md](./pr-versionless-observability.md).
Rolling updates: [rolling-update.md](./rolling-update.md).

---

## Overview

v4 is the first version intended to run **N `versiond` / `devshardd` replicas**
behind **versiond-router** on **shared Postgres**, with sticky session routing
and validation-lease exclusivity. Pre-v4 approved versions stay **single-host**
(SQLite) until retired.

The gateway (`devshardctl`) talks to the chain over **gRPC only** — LCD REST
create/settle paths are gone.

Public observability is **versionless**: dashboards scrape
`/devshard/sessions|stats|metrics` without binding protocol version. Only the
escrow owner binds via signed chat. Legacy `/devshard/{version}/…` obs URLs keep
working via join-proxy internal rewrite.

When governance publishes a **new binary under the same version name** (name
unchanged, only `sha256` changes), `versiond` performs a **blue/green child
swap with drain** instead of stop-then-start: the new `devshardd` must pass
admin `/ready` and public `/healthz` before receiving traffic; already-accepted
requests (including SSE) finish on the old generation. Overlap requires both
binaries to report `postgres` via `--print-storage-mode`; SQLite / hybrid /
unknown modes fall back to exclusive stop/start.

---

## What's in this release

| Area | Change |
| --- | --- |
| **HA topology** | Multi-versiond + `versiond-router` (consistent-hash on session/escrow id); optional legacy pin for pre-HA versions |
| **Storage** | `DEVSHARD_STORAGE_MODE=postgres` fail-closed; boot migrate local SQLite → PG for **this** version’s data dir |
| **Rolling updates** | Same-name SHA swap: blue/green overlap + drain when both children report `postgres`; else stop/start. Private admin `/ready`, `/drain`, `/drain/status` |
| **Validation** | Shared Postgres leases (`devshard_validation_leases`); SQLite leases are no-ops |
| **Gateway transport** | Chain queries + escrow tx via gRPC; `--chain-rest` / `DEVSHARD_CHAIN_REST` removed |
| **Settle confirm** | Settle waits for DeliverTx (`GetTx`) after SYNC CheckTx — same pattern as create |
| **Failover** | Router retries another HA peer on first upstream 502 / connect failure |
| **Versionless obs** | Obs GETs never bind; owner chat binds; join rewrite + PG session lookup; `devshard_obs` rate limit |
| **Status field** | Gateway `protocol_version` → `session_version` (bind / settlement tag) |
| **Tier A `/v1` reads** | Served by **edge-api** on new proxies; same handlers still dual-served on **dapi** as deprecated |
| **testenv** | Named stack behavior tests plus G1–G4, A1–A4, and versiond rolling-update citest (see [testenv/docs/scenarios.md](../testenv/docs/scenarios.md)) |

---

### Gateway CLI: `--chain-rest` → `--chain-grpc`

`devshardctl` no longer accepts LCD REST for chain I/O.

| Before (pre-v4 / r1) | After (r2 / v4) |
| --- | --- |
| `--chain-rest` / `DEVSHARD_CHAIN_REST` (LCD, e.g. `:1317`) | `--chain-grpc` / `DEVSHARD_CHAIN_GRPC` (gRPC, e.g. `:9090`) |
| Optional `DEVSHARD_TX_QUERY_REST` | Unused — tx confirm uses gRPC `GetTx` |
| Admin / persisted `chain_rest` | Deprecated and **ignored**; use `chain_grpc` / env |

**Migration:**

1. Point gateway at chain gRPC: `--chain-grpc host:9090` or
   `DEVSHARD_CHAIN_GRPC=host:9090` (also accepts `NODE_GRPC_URL`).
2. Remove `--chain-rest` and `DEVSHARD_CHAIN_REST` / `DEVSHARD_TX_QUERY_REST`
   from compose, systemd, and scripts.
3. Optional TLS: `CHAIN_GRPC_TLS=true|1|yes` (default remains plaintext).
4. Override chain id when not mainnet: `DEVSHARD_CHAIN_ID` (default
   `gonka-mainnet`).

Join / testenv compose already seeds `DEVSHARD_CHAIN_GRPC` only.

### Gateway status: `protocol_version` → `session_version`

The gateway runtime status field **`protocol_version`** (always `"1"` from the
removed client routing enum) is replaced by **`session_version`**.

| Before | After |
|--------|-------|
| `protocol_version: "1"` | `session_version: "v2"` (example) |

**`session_version`** is the session bind tag from
`EscrowState.StateRootAndProtocolVersion` — the same value used for state-root
hashing and on-chain settlement (`state_root_and_protocol_version`). That is
what external tools should display or assert on, not the old `"1"` enum.

**Migration:** update any monitor, dashboard, or script that polled gateway
`/status` (or equivalent) for `protocol_version` to read **`session_version`**
instead.

Gateway devshard config/API: the old **`protocol_version`** request field is
**`route_prefix`** (e.g. `/devshard/v2`). The legacy HTTP mount
`/v1/devshard` is no longer supported; clients must use `/devshard/<version>/`.

### Storage backend: Postgres required for HA / multi-instance

To run **more than one** `versiond` / `devshardd` instance for the same HA
participant, you **must** use Postgres:

| Deployment | Backend |
|------------|---------|
| Single instance / local dev | SQLite **or** Postgres |
| **Multiple HA versiond replicas** | **Postgres required** (`DEVSHARD_STORAGE_MODE=postgres` + `PGHOST` / `PG*`) |

Why: validation leases and shared session state only work on a shared store.
SQLite is single-writer; its lease store is a no-op under multi-instance.

Set `Devshard-Ha: true` (injected by the router when multi-host HA) requires
literal `DEVSHARD_STORAGE_MODE=postgres` and `PGHOST` or the child returns
**503**.

See [storage-design.md](./storage-design.md) and
[rolling-update.md](./rolling-update.md).

### Rolling updates (same name, new sha256)

Governance still updates `DevshardEscrowParams.ApprovedVersions` the same way
(name + binary URL + sha256). What changes is **how versiond applies** a
same-name SHA change:

| Storage mode (both old + new `--print-storage-mode`) | Swap behavior |
| --- | --- |
| Exactly `postgres` | Blue/green: start new child on a new port → wait admin `/ready` + public `/healthz` → route new traffic to new generation → drain old (proxy leases then `/drain/status`) → SIGTERM |
| `sqlite`, `hybrid`, `auto`/unknown, legacy binary without the flag | Exclusive stop/start (no overlapping children) |

Operator knobs (all defaulted; see [rolling-update.md](./rolling-update.md)):

| Env | Default | Role |
| --- | --- | --- |
| `VERSIOND_READY_TIMEOUT` | `60s` | Abort swap if incoming child never becomes ready (old keeps serving) |
| `VERSIOND_DRAIN_TIMEOUT` | `15m` | Max wait for old proxy leases + lifecycle inflight |
| `VERSIOND_DRAIN_KILL_GRACE` | `10m` | Legacy no-status cushion / process kill backstop |
| `DEVSHARD_SHUTDOWN_GRACE` | `10m` | `devshardd` graceful HTTP shutdown after SIGTERM |

`versiond /healthz` shows per-child `status` (`running` / `draining`), `sha256`,
and `binary_version` during a swap. Whole-`versiond` host evacuation is a
**separate** layer (router drain) — not part of a child SHA swap.

Manual walkthrough: [v4-deploy-test-plan.md](./v4-deploy-test-plan.md) **§7**.

### Versionless observability (bind safety + canonical URLs)

Observability GETs no longer call `CreateSession`. Unbound escrow obs returns
**404**; the first signed owner `POST …/chat/completions` binds the chosen
version. Prefer canonical monitor paths:

| Preferred | Legacy (still works) |
| --- | --- |
| `/devshard/sessions/{id}/diffs` | `/devshard/{version}/sessions/{id}/diffs` (join proxy rewrite, no `Location`) |
| `/devshard/stats/shards/{id}` | `/devshard/{version}/stats/shards/{id}` |
| `/devshard/metrics` | `/devshard/{version}/metrics` |

| Path | Meaning |
| --- | --- |
| `/devshard/healthz` | versiond supervisor |
| `/devshard/{version}/healthz` | that child (not rewritten) |

With Postgres, versiond routes versionless session obs via `sessions.version`
(fan-out fallback if lookup disabled / SQLite). Join proxy rate-limits obs GETs
separately from chat:

| Env | Where | Default |
| --- | --- | --- |
| `DEVSHARD_OBS_RATE_LIMIT_RPS` | join proxy | 10 |
| `DEVSHARD_OBS_BURST` | join proxy | 20 |
| `VERSIOND_DISABLE_SESSION_LOOKUP` | versiond | unset (lookup on when `PGHOST` set) |

Manual walkthrough: [v4-deploy-test-plan.md](./v4-deploy-test-plan.md) §4.
Design note: [pr-versionless-observability.md](./pr-versionless-observability.md).

---

### Target topology

```text
clients / gateway (devshardctl)
       │
       ▼
 public edge proxy  (deploy/join `proxy` — nginx)
       │
       ├── /v1/ Tier A ──► edge-api  (or edge-api-router)
       └── /devshard/ ──► versiond-router  (sticky hash; first-502 failover)
                                 │
                                 ├── versiond-0 ──► devshardd (per approved version)
                                 ├── versiond-1 ──► …
                                 └── …
                                        │
                                        └── shared Postgres
```

**Deprecated dapi dual-serve:** New join proxies steer Tier A `/v1/` reads to
**edge-api** (see `EDGE_API_ROUTE_PATHS` in `proxy/entrypoint.sh`). The same
handlers (from `common/queryapi`) remain mounted on **decentralized-api** for
operators still running **pre-v4 / old proxy** configs that forward `/v1/*` to
dapi. Those dapi responses set `Deprecation: true` (and a `Link` successor hint);
prefer edge-api. Also still on dapi (deprecated): `/v1/bridge/block/latest` and
`/v1/supply/total` (not on edge-api Tier A).

| Variable | Role |
| --- | --- |
| `VERSIOND_HOSTS` | HA pool hostnames |
| `VERSIOND_LEGACY_HOST` | Single host for pre-HA SQLite versions (default: first of `VERSIOND_HOSTS`) |
| `VERSIOND_NON_HA_VERSIONS` | Version path segments pinned to legacy (join default: `v1 v2 v3`) |
| `VERSIOND_SERVICE_NAME` | Edge proxy upstream for `/devshard/` (set to `versiond-router` by HA overlay) |
| `EDGE_API_SERVICE_NAME` | Edge proxy upstream for Tier A `/v1/` (default `edge-api`; multi → `edge-api-router`) |
| `DEVSHARD_STORAGE_MODE=postgres` | Fail-closed shared storage for HA-capable versions |
| `DEVSHARD_POSTGRES_PASSWORD` / `PG*` | Shared DB |
| `DEVSHARD_CHAIN_GRPC` | Gateway chain gRPC (no LCD) |
| `KEY_NAME` (+ shared keyring) | **Same** on every HA replica of one participant |

### Join docker-compose (must bump for v4)

Production join stack lives under **`deploy/join/`**. Operators must pull/update
these files (or equivalent k8s manifests) when rolling v4 — image tags, the
**edge-api** service, edge **proxy** `EDGE_API_*` / `VERSIOND_SERVICE_NAME`, and
the HA overlays.

| File | Role |
| --- | --- |
| [`deploy/join/docker-compose.yml`](../../deploy/join/docker-compose.yml) | Base: node, api (dapi), **edge-api**, versiond, **proxy** (edge), explorer, … |
| [`deploy/join/docker-compose.versiond.yml`](../../deploy/join/docker-compose.versiond.yml) | HA: postgres + versiond2 + versiond-router; proxy → router |
| [`deploy/join/docker-compose.edge-api-multi.yml`](../../deploy/join/docker-compose.edge-api-multi.yml) | Optional: edge-api2/3 + edge-api-router; proxy → router |
| [`deploy/join/docker-compose.devshard-gateway.yml`](../../deploy/join/docker-compose.devshard-gateway.yml) | Optional gateway |

```bash
# Baseline join
cd deploy/join
docker compose -f docker-compose.yml up -d

# HA versiond (+ shared Postgres)
docker compose -f docker-compose.yml -f docker-compose.versiond.yml up -d

# Optional multi edge-api
docker compose -f docker-compose.yml -f docker-compose.edge-api-multi.yml up -d
```

**Base `edge-api` + edge `proxy` (from `docker-compose.yml`):**

```yaml
  edge-api:
    container_name: edge-api
    image: ghcr.io/product-science/edge-api:0.2.13
    environment:
      - EDGE_API_PORT=18080
      - CHAIN_GRPC_URL=node:9090
      - EDGE_API_OTEL_ENABLED=${EDGE_API_OTEL_ENABLED:-false}
      - OTEL_ENDPOINT=${OTEL_ENDPOINT:-}
      - OTEL_HEADERS=${OTEL_HEADERS:-}
    depends_on:
      - node
    restart: always

  proxy:
    container_name: proxy
    image: ghcr.io/product-science/proxy:0.2.13
    ports:
      - "${API_PORT:-8000}:80"
      - "${API_SSL_PORT:-8443}:443"
    environment:
      # … other proxy env …
      - EDGE_API_SERVICE_NAME=${EDGE_API_SERVICE_NAME:-edge-api}
      - EDGE_API_PORT=${EDGE_API_PORT:-18080}
    depends_on:
      # … node, api, versiond, explorer, …
      edge-api:
        condition: service_started
```

**HA overlay (`docker-compose.versiond.yml`) — points edge proxy at versiond-router:**

```yaml
  versiond-router:
    container_name: versiond-router
    image: ghcr.io/product-science/versiond-router:0.2.13
    environment:
      - VERSIOND_HOSTS=${VERSIOND_HOSTS:-versiond versiond2}
      - VERSIOND_PORT=8080
      - VERSIOND_LEGACY_HOST=${VERSIOND_LEGACY_HOST:-versiond}
      - VERSIOND_NON_HA_VERSIONS=${VERSIOND_NON_HA_VERSIONS:-v1 v2 v3}

  proxy:
    environment:
      - VERSIOND_SERVICE_NAME=versiond-router
      - VERSIOND_PORT=8080
```

**Multi edge-api overlay (`docker-compose.edge-api-multi.yml`):**

```yaml
  edge-api-router:
    container_name: edge-api-router
    image: ghcr.io/product-science/edge-api-router:0.2.13
    environment:
      - EDGE_API_HOSTS=${EDGE_API_HOSTS:-edge-api edge-api2 edge-api3}
      - EDGE_API_PORT=18080

  proxy:
    environment:
      - EDGE_API_SERVICE_NAME=edge-api-router
```

Bump image tags (`edge-api`, `proxy`, `versiond`, `versiond-router`, …) to the
v4 release tag when publishing. Full file contents: see the paths in the table
above (do not edit only env without refreshing compose from the release branch).

### Images / release tooling note (`edge-api`)

| Artifact | Local build | In `make release` | Upgrade zip (`make build-for-upgrade`) | GH Actions publish to GHCR |
| --- | --- | --- | --- | --- |
| `ghcr.io/product-science/edge-api` | `make edge-api` / `edge-api-build-docker` | Yes (`edge-api-release` → build + `docker-push`) | Yes → `public-html/v2/edge-api/edge-api-amd64.zip` (also uploaded by `publish_upgrade_binaries.yml` on `release/v*` tags) | **No** dedicated push workflow — use local `VERSION=<tag> make edge-api-release` (same as other join images) |
| `ghcr.io/product-science/edge-api-router` | `make edge-api-router-build-docker` | **No** — still not a `release` dependency | **No** | **No** |
| `ghcr.io/product-science/proxy` (edge proxy) | `make proxy-build-docker` | Yes (`proxy-release` → buildx `--push`) | **No** | Manual via `make release` / proxy Makefile |

CI today:

- [`.github/workflows/verify.yml`](../../.github/workflows/verify.yml) — `build-and-test-edge-api` (compile + `scripts/validate-edge-api.sh`); does **not** push images.
- [`.github/workflows/test-workflow.yml`](../../.github/workflows/test-workflow.yml) — builds via `make build-docker`, saves `edge-api.tar` / `edge-api-router.tar` as **job artifacts** for Testermint (retention 1 day); not a GHCR release.
- [`.github/workflows/docker-build.yml`](../../.github/workflows/docker-build.yml) — only decentralized-api + inference-chain (different image names under `ghcr.io/<owner>/…`).
- [`.github/workflows/publish_upgrade_binaries.yml`](../../.github/workflows/publish_upgrade_binaries.yml) — on `release/v*` tag push, uploads upgrade zips (including `edge-api-*.zip`) to `product-science/race-releases`; does **not** push GHCR images.

Publish the join image with:

```bash
VERSION=<tag> make edge-api-release
# or as part of the full join release:
VERSION=<tag> make release
```

### Routing rules (critical)

| Version | Upstream | Storage |
| --- | --- | --- |
| Listed in `VERSIOND_NON_HA_VERSIONS` (e.g. v1–v3) | **`VERSIOND_LEGACY_HOST` only** | Local SQLite |
| Any other (v4+, future) | Sticky hash across **all** healthy `VERSIOND_HOSTS` | Shared Postgres; `Devshard-Ha: true` |

Pre-v4 binaries do **not** implement fail-closed Postgres + boot migrate for
already-deployed data dirs. Do **not** put `< v4` in the multi-instance pool.
Flipping `DEVSHARD_STORAGE_MODE=postgres` does **not** migrate sessions owned by
old binaries — only the v4 (or newer) child’s own data dir can boot-migrate.

Debug headers: `X-Upstream-Addr`, `X-Versiond-Backend`
(`versiond_legacy` vs `versiond_ha_pool`).

### Rollout phases (summary)

**Phase A — Single-instance baseline**

1. Refresh `deploy/join/` compose; deploy v4 images (`proxy` / edge, `edge-api`,
   `versiond`, `versiond-router`, gateway, …).
2. Pin pre-HA versions: `VERSIOND_NON_HA_VERSIONS=v1 v2 v3`.
3. Bring up shared Postgres; set `postgres` mode on HA-capable children.
4. Approve/force **v4**; leave pre-v4 on the legacy host’s SQLite volumes.

**Phase B — Expand HA for v4+**

1. Add versiond-1…N with the same Postgres credentials and **same `KEY_NAME`**.
2. Keep NON_HA pin; confirm HA paths show `versiond_ha_pool` + multi upstream.
3. Exercise stickiness, kill-one-host failover (§3), the optional lease race (§2),
   and a same-name SHA rolling update on a Postgres HA version (§7).

**Phase C — Retire pre-v4 (later)**

Drain old version paths; remove from `VERSIOND_NON_HA_VERSIONS` only when idle
(or after an out-of-band migrate tool exists).

Full checklists and negative proofs (multi-host + sqlite → 503, migrate inventory):
[v4-deploy-test-plan.md](./v4-deploy-test-plan.md) §1.

### Operator test plans (same doc)

| Plan | Proves |
| --- | --- |
| §1 Test deployment | NON_HA pin, sqlite→HA-fail→migrate→HA, mixed binding |
| §2 Validation race | Same-key HA: one lease row per inference |
| §3 High availability | Kill versiond → survivors serve (first-502); restart rejoins |
| §4 Versionless observability | Unbound obs 404; owner chat binds; rewrite; PG route; rate limit |
| §5 Edge-api / deprecated dapi | New proxy → edge-api; old proxy → dapi dual-serve |
| §6 Escrow long-poll warm | Host-events cache; inference with inference-node down |
| §7 Rolling update | Same-name SHA: in-flight SSE survives; Postgres overlap; hybrid stop/start |

---

## Upgrade / rollout checklist

- [ ] Refresh **`deploy/join/`** compose from the release branch (base + versiond /
      edge-api overlays); bump image tags including **edge-api** and **proxy**
- [ ] Edge proxy: `EDGE_API_SERVICE_NAME` / `EDGE_API_PORT` set; HA overlay sets
      `VERSIOND_SERVICE_NAME=versiond-router`
- [ ] Publish / pull `ghcr.io/product-science/edge-api:<tag>` via
      `VERSION=<tag> make edge-api-release` (or full `make release`)
- [ ] Gateway: replace `--chain-rest` / `DEVSHARD_CHAIN_REST` with `--chain-grpc` /
      `DEVSHARD_CHAIN_GRPC`
- [ ] Shared Postgres up; `DEVSHARD_STORAGE_MODE=postgres` + `PGHOST` on HA children
- [ ] `VERSIOND_NON_HA_VERSIONS` set for every still-deployed pre-HA version
- [ ] HA replicas share one `KEY_NAME` / keyring for the participant
- [ ] Per-versiond SQLite volumes remain **distinct**; Postgres is shared
- [ ] Monitors read `session_version`, not `protocol_version`
- [ ] Prefer versionless obs URLs (`/devshard/sessions|stats|metrics`); confirm
      legacy `/devshard/{v}/…` still 200 with no public redirect
- [ ] Join proxy: set/tune `DEVSHARD_OBS_RATE_LIMIT_RPS` / `DEVSHARD_OBS_BURST` if needed
- [ ] Smoke: chat/settle on a NON_HA path and on v4 HA path; Tier A `/v1/` via edge-api
- [ ] Confirm old proxies (no `EDGE_API_SERVICE_NAME`) still reach Tier A via
      deprecated dapi dual-serve; new proxies should use edge-api
- [ ] Optional: run §2 lease race, §3 kill/restart, §4 versionless obs, and
      **§7 rolling update** (same-name SHA with a long stream) from the deploy
      test plan
- [ ] Confirm HA children report `postgres` via `devshardd --print-storage-mode`
      before expecting blue/green overlap on a governance SHA bump

---

### Escrow ID: `string` vs `uint64`

On this branch, devshard keeps **`escrowID` as `string`** in the bridge API,
session storage keys, HTTP paths, and devshard protocol messages. On-chain
`DevshardEscrow.id` remains **`uint64`**; adapters parse/format at the chain
boundary.

Unifying to **`uint64` throughout** is a protocol / wire contract change — bundle
with larger protocol bumps rather than a standalone migration.

---

## Related docs

| Doc | Use |
| --- | --- |
| [v4-deploy-test-plan.md](./v4-deploy-test-plan.md) | Deploy + seven operator test plans (§1–§7) |
| [pr-versionless-observability.md](./pr-versionless-observability.md) | Versionless obs design + unit/manual checklist |
| [high-availability-architecture.md](./high-availability-architecture.md) | Current runtime topology |
| [storage-design.md](./storage-design.md) | Storage mode selection |
| [rolling-update.md](./rolling-update.md) | Same-name SHA blue/green + drain (Track A) |
| [testenv/docs/chain-transport-consolidation.md](../testenv/docs/chain-transport-consolidation.md) | REST → gRPC migration detail |


devshard/user/session.go
```diff
@@ -1334,11 +1342,23 @@ func (s *Session) fetchSignature(ctx context.Context, hostIdx int, nonce uint64,
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
+	postRoot, ok := s.postStateRootForNonce(nonce)
```
Ryanchen911, line 1397: **[HIGH] Settlement can be permanently stuck (locked escrow funds) after snapshot-only recovery**

`fetchSignature` now bails when `postStateRootForNonce(nonce)` misses (line 1345, `if !ok { return false }`), and `postStateRootForNonce` returns false whenever `s.diffs` is empty (`if len(s.diffs) == 0 { return nil, false }`).

Consider a session that has finalized and persisted a snapshot at the final nonce with all `HostSyncNonce` cursors already equal to that nonce, and the gateway restarts **before** the settle tx is broadcast (upgrade, crash, or blue/green swap — routine in the new HA model). On recovery, `RecoverSession` takes the fast-path early return (`recover.go`, `replayFrom > meta.LatestNonce`): the replay loop that is the only post-snapshot populator of both `s.diffs` and `s.signatures` is skipped, and the backfill block loads nothing (`backfillFrom = minHostSyncNonce+1 > snapNonce`). The snapshot blob carries state but no signatures, and recovery never re-reads them from the store. So `s.diffs == []` and `s.signatures == {}`.

On the settle retry, both signature-collection paths no-op: `fetchSignature` returns false on the missing post-state-root, and `sendCatchUpWith` returns early because `diffsForHost` yields an empty slice on empty `s.diffs`. Accumulated weight stays 0, so Finalize's `PhaseSettlement` guard returns "insufficient signatures" **permanently** — the escrow (real USDT) cannot settle. The only recovery today is manually deleting the snapshot to force a full replay.

Note the sibling `processResponse` handles the identical "nonce beyond diffs" case with a live-compute fallback (line 459: `// Finalize path: nonce beyond diffs array, compute live.` → `s.sm.ComputeStateRoot()`), but `fetchSignature` has no equivalent.

**Suggested fix:** when `postStateRootForNonce` misses on the finalize/settlement nonce, fall back to `s.sm.ComputeStateRoot()` as `processResponse` already does (valid here because the SM is restored to the frozen final state). Alternatively, restore `s.signatures` from the store on the snapshot-only recovery path.

devshard/transport/client.go
```diff
@@ -712,11 +709,5 @@ func (c *HTTPClient) observeTransportFailure(path string, err error) {
 	if c == nil || c.config.Admission == nil || strings.TrimSpace(c.config.ParticipantKey) == "" {
 		return
 	}
-	// A cancelled request context is our own signal (client disconnect, race
-	// resolved, drain), not a fault of the host. Attributing it as a transport
-	// failure would quarantine a healthy host, so skip it.
-	if errors.Is(err, context.Canceled) {
-		return
-	}
 	c.config.Admission.ObserveTransportFailure(c.config.ParticipantKey, path, err)
```
Ryanchen911, line 728: **[HIGH] Removing the `context.Canceled` guard quarantines healthy hosts on normal race/drain cancellation**

This PR deletes the guard that previously sat at the top of `observeTransportFailure`:

```go
// A cancelled request context is our own signal (client disconnect, race
// resolved, drain), not a fault of the host. Attributing it as a transport
// failure would quarantine a healthy host, so skip it.
if errors.Is(err, context.Canceled) {
    return
}
```

`observeTransportFailure` now forwards **every** `http.Do` error (including `context.Canceled`) to `Admission.ObserveTransportFailure`. The redundancy layer routinely cancels speculative losers, stalled winners, and drained attempts via `inf.cancel()`, and client disconnects cancel the attempt context; these surface at `c.http.Do(req)` in `doPostRaw` as `context.Canceled`. For the `/chat/completions` path, `participantPathKind` == `"inference"`, so `ObserveTransportFailureForModel` does not early-return; `isEOFTransportFailure(context.Canceled)` is false, so it falls through to `applyQuarantineLocked(..., participantQuarantineProbe)` — probe-quarantining a **healthy** host for `transportFailureQuarantine` (default 30m) purely for losing a race or being drained.

**Impact:** a busy speculative-racing pool steadily quarantines healthy hosts (every slow speculative loser and every pre-header client disconnect), starving routing capacity. The maintainer's own removed comment documents that this cancellation happens in normal operation.

**Suggested fix:** restore the `context.Canceled` short-circuit at the top of `observeTransportFailure` (or filter cancellation before quarantining in `ObserveTransportFailureForModel`).

devshard/chainoracle/blocks/client/http.go
```diff
@@ -0,0 +1,532 @@
...
+	}
+	c.mu.RUnlock()
+
+	go func() {
+		for _, h := range replay {
+			select {
+			case <-ctx.Done():
+				c.unsubscribe(id)
+				return
+			case sub.ch <- h:
```
Ryanchen911: **[HIGH] "send on closed channel" panic: `fanout` closes a subscriber channel while the replay goroutine is still sending (process-crashing DoS)**

`Subscribe` registers the subscription in `c.subs` under `subMu`, releases the lock, then spawns a replay goroutine that drains the cached snapshot into `sub.ch` with a blocking send (line 238: `case sub.ch <- h:`) holding **no lock**. Concurrently, `fanout` (called from `store` on the ingest goroutine) iterates `c.subs` under `subMu` and, when a subscriber's 16-slot buffer is full, takes the `default` branch and executes `close(sub.ch)` (line 425).

Because the subscription is visible in `c.subs` while the replay goroutine runs, `fanout` can close `sub.ch` while the replay goroutine is blocked on `sub.ch <- h`. In Go, **sending on a closed channel panics**, and there is no `recover` anywhere in `devshard/chainoracle/`, so this crashes the whole process. The `unsubscribe` guard only prevents a double *close*, not send-after-close.

**Trigger:** a consumer calls `Subscribe(ctx, oldHeight)` where the (unbounded, unpruned) cache holds >16 headers ≥ `oldHeight`, then reads slowly. The replay goroutine blocks once the 16-slot buffer fills; a new header arrives → `store` → `fanout` → full buffer → `close(sub.ch)` → the blocked replay send panics → crash. The identical pattern exists in `observer/mock.go` (replay send vs `fanoutLocked` close).

**Suggested fix:** never `close` a subscriber channel from `fanout` while other goroutines may send. Have the sole owner (the `Subscribe` goroutine, on ctx cancellation) close it via a `done` channel / `sync.Once`; `fanout` should only stop delivering (mark dropped). Apply the same fix to `observer/mock.go`.

Ryanchen911:
## Review findings — 6 medium-severity items (calibrated)

Following the 3 inline HIGH comments, here are 6 medium items. Each notes the **precondition** under which it actually bites, since several are gated by deployment/topology.

---

**M1 — Recovery re-tags the signed state-root domain with `meta.Version` instead of the fixed protocol tag** · `devshard/user/recover.go` (and the mirror at `manager.go`)

`RecoverSession` builds the state machine with `state.WithVersion(recoveredVersion)` where `recoveredVersion = meta.Version` (the runtime/route name), replacing the base code which used the fixed protocol tag. The version string feeds the signed state-root preimage (`ComputeStateRootFromRestHash` mixes in `sha256(version)`), so replay recomputes each diff's root under the new tag and compares to the stored `rec.StateHash`.

**Precondition:** only bites when `meta.Version` differs from the protocol tag that signed the stored diffs — i.e. upgrading **existing on-disk/Postgres sessions** from a base build into this build. A clean v3/v4-parallel deployment that does not migrate live sessions avoids it. Where it does apply, replay fails with `ErrLocalStateUnrecoverable` and the escrow is skipped/never settled.
*Suggestion:* keep signing with the fixed protocol tag; treat `meta.Version` only as runtime-bind identity. If unification is intended, add a migration that verifies stored diffs still recompute under the new tag before switching.

---

**M2 — `At(height)` / `fetchAndVerifyAt` does not bind the returned header to the requested height** · `devshard/chainoracle/blocks/client/http.go`

After verifying chain-id and commit signatures (with `lastHeight=0`), it caches under `h.Height` and returns without checking `h.Height == height`. Since `CanonicalHeaderBytes` binds Height, a validly-signed header for a *different* height passes verification.

**Severity note:** currently closer to **Low** — the Proof/AppHash cross-check is a testenv simplification (`// check Proof.Value against the header's AppHash (testenv simplification)`) and `At()`/`Prove()` are not yet wired into a settlement-proof cross-check. Flagging as a latent hazard to fix before that wiring lands.
*Suggestion:* reject when `h.Height != height` after verify; cache only under the matching height.

---

**M3 — Schema migration (`ApplyPG`) has no cross-process lock** · `devshard/storage/migrate/pg.go`

In HA, N replicas share one Postgres and each runs `MigratePostgres` → `ApplyPG` at boot (still synchronous in `NewPostgres` as of `1527bef`; the recent commit decoupled only the index rebuild). Two replicas booting against an un-migrated DB can both read the same `MAX(id)`, both run the DDL, and both `INSERT INTO schema_migrations` the same id — the loser gets a unique-violation and that replica's boot fails.

**Precondition / impact:** typically a transient boot failure that clears on restart once a peer finishes, not necessarily a permanent crash-loop. Still worth hardening for fresh-cluster bootstrap and rolling updates that add a step.
*Suggestion:* wrap `ApplyPG` in a `pg_advisory_lock`, or make the insert `ON CONFLICT (id) DO NOTHING` and tolerate concurrent-DDL errors (42P07).

---

**M4 — New public edge-api service has no `BodyLimit`** · `edge-api/internal/server/server.go`

`server.New` installs only `Recover()` + tracing, then mounts the ported query handlers. `/v1/verify-proof` and `/v1/verify-block` `Bind` + JSON-decode the whole body into memory. The base dapi server applied a global 10 MB `BodyLimit`; this was not carried over.

**Precondition:** the default public proxy caps bodies at `client_max_body_size 10M` and doesn't expose the verify/debug routes by default, so the OOM only materializes when edge-api is reached directly (bypassing the proxy) or those optional routes are explicitly opened.
*Suggestion:* add `middleware.BodyLimit("10485760")` in edge-api `server.New` for defense-in-depth regardless of proxy topology.

---

**M5 — `escrow_id`-labeled gauges are never deleted** · `devshard/observability/metrics_lifecycle.go`

This PR removes `DeleteEscrowMetrics(escrowID)`, the only pruner for the `escrow_id`-labeled series `devshard_validation_queue_depth` and `devshard_mempool_size`. Teardown paths (`HandleSettlementFinalized`, `EvictBefore`) call `Host().Close()` but never drop the labels. Every escrow ever served leaves permanent series in the registry exposed at `/metrics`.

**Impact:** unbounded registry/memory growth and ever-larger scrape payloads over a long-running process; externally driven (one series pair per funded escrow).
*Suggestion:* reintroduce `DeleteLabelValues(escrowID)` for both vectors and call it from the teardown paths after `Close()`.

---

**M6 — HA compose overlay pins pre-HA `versiond:0.2.13`** · `deploy/join/docker-compose.versiond.yml`

Base `docker-compose.yml` pins `0.2.14-devshard-v4`, but the new HA overlays pin `0.2.13` (`versiond`, `versiond-router`, `edge-api`, `edge-api-router`). Because the overlay's `versiond` service uses `<<: *versiond` and merges on top, running `docker compose -f docker-compose.yml -f docker-compose.versiond.yml up -d` results in the `0.2.13` image winning. Confirmed by actually running the merge: both `versiond` and `versiond2` resolve to `0.2.13`, overriding base's `0.2.14-devshard-v4`.

**Impact:** the overlay simultaneously sets `DEVSHARD_STORAGE_MODE=postgres`, sticky routing, and a second replica — all v4 features — while running the pre-HA binary, so the advertised routing / blue-green behavior is not honored. Whether this can lead to double-settle is **not demonstrated** and stated only as an unverified downstream risk.
*Suggestion:* pin all overlay images to the same `0.2.14-devshard-v4` tag as base, or parameterize the tag from a single variable so base and overlay cannot drift.

devshard/user/session.go
a-kuprin: Good catch! But to make it clear, it is not related to upgrade or blue/green swap, as it is related to gateway (devshardctl).

The simple bug explanation is: if there was snapshot after Finalize on Settlement phase, and there was crush on settlement (or reboot), settlement will never completes, because empty diff prevents get `postRoot` at `fetchSignature`

Why the bug appeared:
`fetchSignature` used to take whatever bytes GET /signatures returned and put them straight into `s.signatures`. The fix at `93744bc2f` requires verifying those bytes over the canonical `StateSignatureContent{postRoot, escrowID, nonce}` and checking the recovered address (with warm-key fallback).

The fix is coming

deploy/join/docker-compose.versiond.yml
```diff
@@ -0,0 +1,97 @@
...
+# hash on escrow/session ID) in front of both. Sets proxy
+# VERSIOND_SERVICE_NAME=versiond-router.
+#
+#   docker compose -f docker-compose.yml -f docker-compose.versiond.yml up -d
+#
+# x-versiond anchor holds everything shared between versiond and versiond2;
+# only container_name and per-instance data volumes differ.
+
+x-versiond: &versiond
+  image: ghcr.io/product-science/versiond:0.2.13
```
DimaOrekhovPS: In the main `docker-compose.yml` the version is `0.2.14-devshard-v4`, but here it's `0.2.13`

Should be fixed for consistency?

devshard/user/session.go
a-kuprin: Fix is addressed at https://github.com/gonka-ai/gonka/pull/1482/commits/516df3927997bf26016e7b9e52564f216f196ebf. Please check it

## Changes

1. **Store reload on snapshot-only recovery** (`recover.go`)
On early return (`replayFrom > LatestNonce`), restore signatures via `store.GetSignatures(escrow, LatestNonce)`, so when we are recovering from snapshot and signatures are already there, we use them

2. **Live state-root fallback in `fetchSignature`** (`session.go`)
If `postStateRootForNonce` misses, verify against `s.sm.ComputeStateRoot()` (same pattern as `processResponse`), as was proposed

3. **Re-run Finalize when quorum is missing** (`escrow_rotator.go`)
Even if 1, 2 doesn't help we rerun `Finalize` if `!HasQuorumAt(nonce)`. Skip only when already settled **and** quorum is present

devshard/transport/client.go
a-kuprin: Nice catch, it was lost on merging 0.2.14 and v3 fixes to large refactoring

deploy/join/docker-compose.versiond.yml
a-kuprin: Yes, should be fixed

devshard/transport/client.go
a-kuprin: Fixed here https://github.com/gonka-ai/gonka/pull/1482/commits/a6e73db5a1655e1f0050dd379cd793ad2cde1c55

I was checking multiple times that nothing was lost on 0.2.14 remerging, but it was missed.

Also `/debug/pprof/` was missed on gateway, restored it

versiond-router/nginx.conf.template
```diff
@@ -0,0 +1,96 @@
...
+        proxy_request_buffering off;
+
+        # Inference can be slow; connect fails fast so sticky can retry another peer.
+        proxy_read_timeout 600s;
+        proxy_send_timeout 600s;
+        proxy_connect_timeout 2s;
+
+        # Reroute on first 502 / connect failure. No http_503 (drain stays sticky).
+        # No non_idempotent — do not double-fire POST chat bodies.
+        proxy_next_upstream error timeout http_502;
```
bonujel, line 88: **[MEDIUM] The new HA failover path makes `AppendDiff`'s non-idempotent diff INSERT reachable: a duplicate nonce fails the request (HTTP 500) and leaves in-memory state ahead of Postgres**

This comment is anchored on the new `proxy_next_upstream error timeout http_502` because that is what makes the problem *reachable* — the underlying `AppendDiff` INSERT predates this PR, but multi-instance HA (sticky router + 502 failover + shared Postgres) newly exposes it.

`AppendDiff` inserts the diff main row with a bare `INSERT` and no conflict handling (`devshard/storage/postgres.go:770`), while `devshard_diffs` has `PRIMARY KEY (epoch_id, escrow_id, nonce)` (`postgres_migrate.go:62`). Notably the *same* transaction is already tolerant of replay everywhere else — signatures use `ON CONFLICT … DO UPDATE` (`postgres.go:783`) and `latest_nonce` uses `GREATEST` (`postgres.go:792`). Only the diff main row is not.

**How the HA path triggers it:** diffs are assembled deterministically by the gateway sequencer, so the same nonce carries identical bytes on every replica. This router retries on `error / timeout / http_502` (line 88). When a 502 / connect-failure sends the same nonce to a second replica whose in-memory state hasn't yet caught up, both replicas write the same `(epoch_id, escrow_id, nonce)` and the second one hits the primary key.

**Two consequences:**

1. **A legitimate request returns HTTP 500.** The bare INSERT surfaces `SQLSTATE 23505` (`insert diff: duplicate key value violates unique constraint "devshard_diffs_epoch_N_pkey"`), even though the correct row already exists in shared PG (written by the first replica). The correct behavior here is idempotent success, not an error.

2. **In-memory state ends up ahead of Postgres.** `applyAndPersist` mutates in-memory state *before* persisting and does **not** roll back on failure: `sm.ApplyDiff` (`devshard/host/host.go:562`) + `mempool.RemoveIncluded` (`host.go:566`) run first, then `store.AppendDiff` (`host.go:582`). When AppendDiff fails, `LatestNonce` has already advanced but PG holds no such row. A retry of the same diff is then swallowed by the guard at `host.go:551` (`diff.Nonce <= currentNonce`), so the missing durable row is never backfilled — memory silently leads PG.

**Reproduced** against a real Postgres container and in a host-level unit test:

- Same-nonce `AppendDiff` twice → `insert diff: ERROR: duplicate key value violates unique constraint "devshard_diffs_epoch_7_pkey" (SQLSTATE 23505)`, with the correct row already durably present.
- Injecting an AppendDiff failure into `applyAndPersist` → in-memory `LatestNonce` advances `0 → 1` despite the failed persist, and a retry of the same nonce returns nil via the guard (durable gap never repaired).

**Suggested fix:** make the diff main row idempotent — `INSERT … ON CONFLICT (epoch_id, escrow_id, nonce) DO NOTHING` and treat 0-rows-affected as success (matching the signature / `latest_nonce` policy already in this transaction). Separately, persist-before-publish or roll back the in-memory mutation on AppendDiff failure so memory cannot lead durable state.

common/validation/validation.go
```diff
@@ -0,0 +1,499 @@
...
+	originalLogits := originalResponse.ExtractLogits()
+	validationLogits := responseValidation.ExtractLogits()
+	baseResult := BaseValidationResult{InferenceId: inferenceID, ResponseBytes: respBodyBytes}
+	// CompareLogits short-circuits to perfect similarity (1.0) when the ORIGINAL
+	// logits are empty, so an executor that stored a response with no logprobs
+	// would always pass. Reject only the asymmetric case (exactly one side empty):
+	// the executor's output cannot be verified against the validator's
+	// re-execution. Both-empty is left to CompareLogits so legitimate
+	// reasoning-burn empties (e.g. Kimi-K2.6, finish_reason=length) still match.
+	if (len(originalLogits) == 0) != (len(validationLogits) == 0) {
```
bonujel, line 473: **[MEDIUM] Both-empty logits pass validation with similarity 1.0**

This path rejects only the asymmetric empty case (exactly one side empty). Both-empty falls through to `CompareLogits` (see the comment at 467-468), and `CompareLogits` → `customSimilarity` → `customDistance` short-circuits: `customDistance` returns `0.0` when the original is empty (line 211), so similarity is `1.0` — a perfect pass. An inference whose stored response and the validator's re-execution both carry no logprobs passes with no content check at all.

For comparison, the mainnet DAPI validator is fail-closed on the same condition: `if len(originalLogits) == 0 || len(validationLogits) == 0 { return nil, errors.New("no logits found …") }` (`decentralized-api/internal/validation/inference_validation.go:985`) — either side empty rejects. Here both-empty passes. That may be intentional for the reasoning-burn case the comment describes; raising it so the difference is deliberate rather than a leftover from the split.

Not a trivial bypass — three things blunt the direct attack: the stored `responsePayload` must hash to the on-chain `ResponseHash`, the validator re-executes with logprobs force-enabled, and the asymmetric branch (line 469) catches a fake-empty original whose re-execution comes back non-empty. The gap is narrower: an executor that can deterministically drive the ML node to emit `content:[]` with empty logprobs (some `finish_reason=length` / reasoning-burn shapes on non-streaming JSON) gets similarity 1.0 for those inputs, and the validator agrees because it produces empty too. With any handle on the seed to steer toward such an input, an inference that should have produced real output passes without a substantive check, where mainnet would reject it.

Suggested fix: match mainnet and treat both-empty as invalid, or replace the bare `content:[]` allowance with an explicit, checkable "legitimate empty" shape (e.g. require `finish_reason=length` and `completion_tokens > 0`) instead of accepting any empty logits array as a perfect match.

devshard/chainoracle/blocks/client/http.go
a-kuprin: It was used only in tests (mock oracle) so I didn't pay much attention to this code part. But it is possible that some code pieces move at somewhere prod path.

The bad thing here that replay (block catch-up) and fanout (life blocks updates) both write to same cannel concurrently.

Life block updates should never start before catch-up is finished.

So fix should be in the other direction:
Restructure it **replay → gap → then live**, one sender until live, then fanout only for true new headers.

The fix is implemented at
https://github.com/gonka-ai/gonka/pull/1482/commits/09f9b5cb3c3b5e8fae4fe6b312eb8b797a9aa9fb

common/validation/validation.go
a-kuprin: It is intentional

Current semantics:

- **Asymmetric empty (XOR guard):** reject — same idea as mainnet’s “can’t compare” fail.
- **Both-empty**: warn + autopass via CompareLogits → similarity 1.0 — deliberate vs mainnet’s || fail-closed, for Kimi-K2.6 reasoning-burn (#1233).
- **4xx from re-exec**: already warn + autopass, matching mainnet.


Same “can’t prove cheat → don’t punish” posture as mainnet’s 4xx autopass. Keeping this for now and logging both-empty hits, as discussed at PR https://github.com/gonka-ai/gonka/pull/1283 that was merged to v4

a-kuprin: Added logging and comments on this: https://github.com/gonka-ai/gonka/pull/1482/commits/b8b86e4f86a1aa177e7c5f325f24c05527c40a33

deploy/join/docker-compose.versiond.yml
a-kuprin: Fixed here: https://github.com/gonka-ai/gonka/pull/1482/commits/6b0963954ef4d17b5ab897eb2ee3e5af343cae3f

versiond-router/nginx.conf.template
a-kuprin: I think this highlights more important issue on HA design.

It is possible that instances (sharing same db) has different in-memory state, and this in-memory state should be aligned.
So fix direction is wrong. I will address in a fix this more important issue with in-memory state divergence

bonujel: Agreed, aligning in-memory state across instances is the right root fix, and `ON CONFLICT` alone only papers over the symptom.

One thing that matters for that fix: right now `applyAndPersist` writes to memory before it persists (`ApplyDiff` → `AppendDiff`, no rollback on failure), so a replica can hold a nonce Postgres doesn't have. If alignment reconciles memory against the shared DB, PG can't be allowed to lag the state you're aligning to — so persist-before-publish (or rollback on persist failure) is probably needed regardless of how alignment is done.

Until that lands, the idempotent-INSERT part is still worth keeping as a floor so a legitimate request doesn't 500 on a duplicate nonce that's already durably correct. Not a replacement for the real fix, just avoids the user-visible failure meanwhile.

a-kuprin: Yes, It is already implemented in this fashion.

HA design always should be persist task first, then run it, if fails, rerun according persisted task

a-kuprin:
@Ryanchen911

> **M1 — Recovery re-tags the signed state-root domain with `meta.Version` instead of the fixed protocol tag** · `devshard/user/recover.go` (and the mirror at `manager.go`)

`meta.Version`, host `boundVersion`, route `/devshard/<name>/`, and the state-root domain tag are the same string by design (approved_versions.name).


> M2 — At(height) / fetchAndVerifyAt does not bind the returned header to the requested height · devshard/chainoracle/blocks/client/http.go

`chainoracle` here is just a mock for tests only. It will be verifying chain data accurately in a further release, where this data will be needed for protocol changes

> M3 — Schema migration (ApplyPG) has no cross-process lock · devshard/storage/migrate/pg.go

Under normal HA deploy, that process is restarted by the supervisor, and the retry sees migrations already applied and comes up clean. No durable corruption

> M4 — New public edge-api service has no BodyLimit · edge-api/internal/server/server.go

Traffic to `edge-api` is meant to go through the join proxy. `edge-api` has no host port publish — only docker-internal — so internet clients hit the proxy

> M5 — escrow_id-labeled gauges are never deleted · devshard/observability/metrics_lifecycle.go

Should be fixed, I will add a fix shortly

> M6 — HA compose overlay pins pre-HA versiond:0.2.13 · deploy/join/docker-compose.versiond.yml

Already fixed at: https://github.com/gonka-ai/gonka/commit/6b0963954ef4d17b5ab897eb2ee3e5af343cae3f

versiond-router/nginx.conf.template
bonujel: That's the right principle. One path that doesn't follow it yet is `applyAndPersist` itself — it runs before it persists: `sm.ApplyDiff` (host.go:562) and `mempool.RemoveIncluded` (566) mutate in-memory state first, then `store.AppendDiff` (582), and on AppendDiff failure it returns without rolling either back. In a quick unit test that fails the AppendDiff, `LatestNonce` still advances (0 → 1) and the retry is swallowed by the `diff.Nonce <= currentNonce` guard (551), so PG ends up behind memory with no re-run.

If the in-memory-alignment fix already covers this path, ignore me. Just flagging it in case it's a separate spot from whatever "persist first" is implemented in — this one still needs the order flipped (or a rollback on persist failure) to match.

a-kuprin: Fix is massive and needs a separate review, so I've added it to the PR
https://github.com/gonka-ai/gonka/pull/1496

There is one more question that is out-of-scope: "Crash recovery vs ML execution"
Possible inference processing filures and proposal how to make it fault-tolerant is discussed at Issue https://github.com/gonka-ai/gonka/issues/1466, that should be the follow-up changes

a-kuprin: > That's the right principle. One path that doesn't follow it yet is `applyAndPersist` itself — it runs before it persists: `sm.ApplyDiff` (host.go:562) and `mempool.RemoveIncluded` (566) mutate in-memory state first, then `store.AppendDiff` (582), and on AppendDiff failure it returns without rolling either back. In a quick unit test that fails the AppendDiff, `LatestNonce` still advances (0 → 1) and the retry is swallowed by the `diff.Nonce <= currentNonce` guard (551), so PG ends up behind memory with no re-run.
> 
> If the in-memory-alignment fix already covers this path, ignore me. Just flagging it in case it's a separate spot from whatever "persist first" is implemented in — this one still needs the order flipped (or a rollback on persist failure) to match.

I've ment it is already implemented, but isn't yet published. Here is the fix, please review it https://github.com/gonka-ai/gonka/pull/1496

a-kuprin:
@Ryanchen911 M5 is fixed here: https://github.com/gonka-ai/gonka/pull/1482/commits/f091187003a558fbc79d29b339b7ff416aff38ee

Ryanchen911:
Thanks for the quick and thorough turnaround on these — verified the fixes against the latest head (`3f68fde`):

- **HIGH — settlement stuck after snapshot-only recovery** (`516df39`): ✅ `fetchSignature` now falls back to `ComputeStateRoot()` for the current frozen nonce, `recover.go` reloads final-nonce signatures from the store, and settle re-runs Finalize when quorum is missing. Even more complete than suggested.
- **HIGH — `context.Canceled` quarantining healthy hosts** (`a6e73db`): ✅ guard restored, and nice catch also fixing the SSE clean-EOF-on-cancel path so a cancelled empty stream isn't scored as success.
- **HIGH — fanout closing subscriber channels (panic/DoS)** (`09f9b5c`): ✅ the Subscribe goroutine is now the sole closer and fanout only cancels; the same fix applied to `observer/mock.go`.
- **M5 — escrow-labeled gauges never pruned** (`f091187`): ✅ `DeleteEscrowMetrics` restored and wired into the teardown paths.
- **M6 — HA overlay pinning pre-HA `0.2.13`** (`6b09639`): ✅ all four overlay images bumped to `0.2.14-devshard-v4`.

Appreciate the fast fixes. 🙌

Makefile
```diff
@@ -73,11 +74,19 @@ versiond-build-docker:
...
+
+edge-api-router-build-docker:
+	@make -C edge-api-router build-docker SET_LATEST=1
+
 testapp-server-build-docker:
 	@echo "Building testapp-server docker image ($(DOCKER_PLATFORM))..."
 	@docker build --platform $(DOCKER_PLATFORM) -t testapp-server:latest -f local-test-net/Dockerfile.testapp-server .
 
-release: decentralized-api-release inference-chain-release tmkms-release proxy-release proxy-ssl-release bridge-release versiond-release
+release: decentralized-api-release inference-chain-release tmkms-release proxy-release proxy-ssl-release bridge-release versiond-release edge-api-release
```
snevolin: What about versiond-router?

a-kuprin: missed

a-kuprin: added at https://github.com/gonka-ai/gonka/pull/1482/commits/6ec0c66595f3f19d3cd0d986a3787569e7cced74

DimaOrekhovPS (approved):

gmorgachev (approved):

versiond-router/nginx.conf.template
mtvnastya: hi @bonujel! thanks for the review of the upgrade pr - your review fits under the Gonka bounty program conditions. Can you share your Gonka account address so that payment for this contribution could be proposed to the community?

feel free to post the address here or contact me via email anastasia.matveeva at productscience.ai

PR #1483, opened 2026-07-20
fix(devshard): align EvaluateValidationResponse with mainnet validation policy

a-kuprin:
Reopening https://github.com/gonka-ai/gonka/pull/1283 by @0xMayoor closed by old target removal

-----

EvaluateValidationResponse auto-passes in two cases where mainnet rejects.

Empty original logprobs short-circuits CompareLogits to 1.0 because customDistance returns 0 when the original is empty. Mainnet rejects when either side has no logits. An executor can serve a fabricated completion with no logprobs and every validator votes valid.

400 or 422 from the validator's own re-execution returns Valid:true before reading the body. The re-run uses executor-supplied enforced_tokens, so a malformed response pushes the backend into a 4xx and auto-wins. Mainnet has no 4xx auto-pass path at all.

Fix: reject the asymmetric empty-logits case (both-empty stays valid so reasoning-burn empties still pass), return Valid:false on 4xx. The 4xx is triggered by executor-controlled input so failing closed is the safer default.

Test fails on current code, passes with the fix.

PR https://github.com/gonka-ai/gonka/pull/1282 logs this path but keeps the behavior unchanged.


a-kuprin:
Rebased and merged at https://github.com/gonka-ai/gonka/pull/1487

PR #1490, opened 2026-07-20
Rolling updates for versiond-managed binaries (blue/green + drain, part 1 of #1367)

snevolin:
## Summary

This PR adds **rolling updates for versiond-managed binaries**. When governance
publishes a new binary under the same protocol version name and changes only its
`sha256`, `versiond` now performs a blue/green child replacement instead of
stop-then-start.

The new generation is downloaded and validated, started on separate ports, and
required to pass both private readiness and public health checks before the
route changes. New requests then use the new generation while requests already
accepted by the old generation retain a generation-specific proxy lease and
finish before the old process is stopped.

This implements **Part 1 / Track A** of the
[rolling-update plan](https://github.com/snevolin/gonka/blob/sn/versiond-rolling-updates/devshard/docs/rolling-update.md)
and the first stage of the HA proposal
[#1367](https://github.com/gonka-ai/gonka/discussions/1367).

> **Base branch:** `devshard-0.2.14-v4`.
>
> Whole-`versiond` host evacuation is a separate layer implemented by
> [#1489](https://github.com/gonka-ai/gonka/pull/1489). A governance SHA update
> does not remove `versiond` from the router pool.

---

## Motivation

Previously, a same-name SHA change stopped the current child before the incoming
binary had proved that it could serve traffic. Long inference and validation
requests could be terminated during that gap.

The required behavior is:

1. A request accepted by the old generation can finish on that generation.
2. The incoming generation receives no traffic until it is actually ready.
3. Once ready, all new requests use the incoming generation.
4. A failed rollout leaves the current generation serving.
5. Old and new processes overlap only when their storage mode makes overlap
   safe.

---

## Rolling-update flow

```text
oracle reports: same name, different sha256
        |
        v
validate oracle identity and resolve bin/<name>/<sha256>/
        |
        v
probe old and new --print-storage-mode
        |
        +-- either side is not exactly postgres --> exclusive stop/start
        |
        v
allocate separate public and admin ports
        |
        v
start NEW child while OLD keeps serving
        |
        v
require NEW admin /ready == 200 and public /healthz == 2xx
        |
        +-- timeout/failure --> stop NEW, keep OLD routed, retry later
        |
        v
atomically publish NEW generation target and retire OLD target
        |
        +-- stale, not-yet-acquired lookup retries against NEW
        +-- already accepted request keeps an OLD target lease
        |
        v
wait for OLD proxy leases to reach zero (or drain deadline)
        |
        v
POST OLD /drain, then poll OLD /drain/status until inflight == 0
        |
        v
SIGTERM OLD --> graceful wait --> SIGKILL only as backstop --> reap
```

The route table stores a concrete `Target`, not just an address. A target belongs
to one child generation: protocol name, SHA, ports, and process. Its lease spans
the complete proxied response, including the full lifetime of an SSE stream.
This closes the race between loading an old route and entering the child's own
lifecycle middleware.

---

### Child generation

```text
preparing -> starting -> running -> retiring -> draining -> stopping -> stopped
     |           |          |
     +-----------+----------+----> failed --retry--> starting
```

- `running` generations may be published in the route table.
- `retiring` means route admission is closed and the old target is waiting for
  acquired proxy leases.
- `draining` waits for child lifecycle inflight after proxy work is gone.
- `stopping` owns post-`SIGTERM` grace and process completion.

The manager keeps one current child plus zero or more draining generations per
version name. Generation transitions and route publication are serialized by
the manager lock, while lifecycle HTTP calls and long waits run outside it.

### OS process

```text
Running -> Terminating -> Killing -> Exited
```

Each process incarnation is owned by a small supervisor. `Stop()` sends
`SIGTERM` to the process group and arms the graceful timer. Expiry or an
explicit force moves it to `Killing`. `Done()` closes only after `cmd.Wait()`
has confirmed process exit and reap. Manager shutdown stops all children first,
then waits for all of them; a manager deadline forces outstanding children but
does not return while they remain unreaped.

---

### `versiond`

| Area | Current behavior |
|---|---|
| Blue/green generations | One serving generation and independently tracked draining generations per version name |
| Proxy target leases | Accepted requests remain attached to the exact old generation; stale unacquired lookups retry on the new route |
| Readiness gate | Incoming child must pass private admin `/ready` and public `/healthz`; both are rechecked before publication |
| Storage safety | Overlap is enabled only when both the running and incoming binaries report `postgres` through `--print-storage-mode` |
| Exclusive fallback | SQLite, hybrid, auto/unknown, legacy binaries, and probe errors use compatible stop-then-start behavior |
| Private lifecycle channel | Admin-capable children receive `DEVSHARD_ADMIN_ADDR=127.0.0.1:<port>`; lifecycle probes never traverse the public proxy |
| Port ownership | A bounded supervisor-managed pool allocates and reuses public/admin ports, skips the `versiond` listener, and returns exhaustion as an explicit start error |
| Per-SHA installs | Immutable `bin/<name>/<sha256>/` directories preserve binaries still used by draining or restartable children |
| Legacy install promotion | Verified flat-layout binaries are promoted atomically into the per-SHA layout without requiring an artifact download |
| Install GC | Keeps desired, live, draining, and recent complete installs; incomplete downloads are not treated as committed installs |
| Removed versions | Route removal is immediate, drain is asynchronous, and a same-name re-add waits for the old generation to finish |
| Process supervision | Explicit graceful, forced, and reaped states replace timeout-based assumptions around `exec.Cmd` |
| Preflight validation | Binary/protocol/admin/storage capabilities are probed with narrow legacy fallback and fail-closed transient errors |
| Oracle validation | Names are unique safe path components; SHA values are exactly 64 hexadecimal characters |
| Observability | `/healthz` exposes serving and draining generations with status, SHA, and embedded binary version |

### `devshardd`

Lifecycle controls move to a loopback admin listener managed by `versiond`:

| Endpoint / setting | Behavior |
|---|---|
| `GET /ready` | Returns 200 only after initialization, store recovery, listener startup, and active chain-event subscription |
| `POST /drain` | Rejects new non-lifecycle work while accepted requests finish |
| `GET /drain/status` | Returns readiness, drain state, and full-request inflight count, including SSE duration |
| `devshardd_lifecycle_inflight_requests` | Prometheus gauge for the same lifecycle inflight count |
| `DEVSHARD_SHUTDOWN_GRACE` | Configurable `server.Shutdown` budget; default `10m` |
| `--print-storage-mode` | Reports the effective storage backend using the application's real storage-mode resolver |
| `--print-admin-api-version` | Advertises support for the private lifecycle listener |

The public devshard listener continues to expose inference traffic, `/healthz`,
and metrics as before. `/ready`, `/drain`, and `/drain/status` are not public
versiond routes for admin-capable children.

### Configuration

All settings have defaults; existing compose files do not need new mandatory
environment variables.

| Environment variable | Default | Meaning |
|---|---:|---|
| `VERSIOND_READY_PATH` | `/ready` | Private child readiness path; public health must also pass |
| `VERSIOND_READY_TIMEOUT` | `60s` | Maximum incoming-child readiness wait before aborting the swap |
| `VERSIOND_DRAIN_PATH` | `/drain` | Private child drain command path |
| `VERSIOND_DRAIN_STATUS_PATH` | `/drain/status` | Private child inflight status path |
| `VERSIOND_DRAIN_TIMEOUT` | `15m` | Shared deadline for retired proxy leases and child lifecycle inflight |
| `VERSIOND_DRAIN_POLL_INTERVAL` | `1s` | Child inflight polling interval |
| `VERSIOND_DRAIN_KILL_GRACE` | `10m` | Legacy pre-SIGTERM cushion and child process kill backstop |
| `DEVSHARD_SHUTDOWN_GRACE` | `10m` | `devshardd` graceful HTTP shutdown budget |

For `devshardd`, post-`SIGTERM` process grace is
`max(VERSIOND_DRAIN_KILL_GRACE, DEVSHARD_SHUTDOWN_GRACE)`. For other managed
binaries it is exactly `VERSIOND_DRAIN_KILL_GRACE`.

---

## Compatibility and failure policy

| Scenario | Result |
|---|---|
| New `versiond` with an already released `devshardd` | Starts through narrow preflight/readiness compatibility fallbacks; storage mode is unknown, so overlap stays disabled |
| Old `versiond` with a new `devshardd` | Continues to work through the existing link-stamp fallback |
| Both generations report Postgres | Full blue/green overlap and drain |
| SQLite, hybrid, auto, or unknown storage | Stop-then-start; two processes never share the local data directory |
| Incoming child fails readiness or public health | Incoming child is stopped and the current route remains unchanged |
| Invalid oracle response | Reconcile poll fails closed and current children remain untouched |
| Missing legacy `/drain/status` | Waits the configured legacy cushion before `SIGTERM` because inflight is not observable |
| Port pool exhaustion | Start/swap fails explicitly; no partially configured child is executed or routed |

Legacy fallback is accepted only for recognized unsupported-flag or
unsupported-endpoint responses. Probe timeouts, signals, OOM-like failures, and
ambiguous execution errors fail closed and retry on a later reconcile poll.

---

### Unit and race coverage

- swap publication, readiness/public-health gates, abort paths, drain idle and
  timeout paths, and asynchronous version removal;
- proxy lease behavior before, during, and after route retirement;
- Postgres-only overlap and hybrid/unknown exclusive fallback;
- process FSM graceful exit, escalation, process-group signaling, and reap;
- port allocation/reuse/exhaustion and release on partial start failure;
- per-SHA install verification, GC, legacy promotion, and concurrent promotion;
- preflight capability matrix and oracle input validation;
- devshard lifecycle readiness, drain admission, inflight, metrics, and shutdown
  grace.

### `versioned` Docker e2e

`TestSameNameNewSHA_RollingUpdateDrainsOld` drives an oracle SHA change while a
long request is held on the old binary. It requires that the request finishes on
the old generation, concurrent new work reaches the new generation, and the old
process exits after drain.

```bash
make -C versioned e2e
```

### Full-stack `devshard/testenv`

The integration suite runs through the generated HA topology and public API:

- `TestVersiondRollingUpdateSameVersionSHA` proves Postgres-backed overlap,
  uninterrupted long-stream completion, per-host route swap, and old-generation
  exit;
- `TestVersiondRollingUpdateHybridFallback` proves the same SHA change uses
  stop-then-start and never exposes concurrent generations in hybrid mode.

```bash
make -C devshard/testenv citest-versiond-rolling-update
```

The test harness now lets Docker allocate published host ports, avoiding the
host-side free-port race between discovery and compose startup.

### Testermint

`VersiondTests` also contains a same-version update scenario for the legacy
full-node suite. Test binaries are stamped per protocol slot and the fresh-stack
gRPC wiring is configured explicitly instead of relying on cached images.

```bash
make local-build build-docker
make run-tests TESTS="VersiondTests"
```

---

## Documentation

[`devshard/docs/rolling-update.md`](https://github.com/snevolin/gonka/blob/sn/versiond-rolling-updates/devshard/docs/rolling-update.md)
documents the implemented Track A behavior, storage prerequisite, generation
and process lifecycles, compatibility fallbacks, configuration defaults, and
the distinction between child rolling update and whole-host evacuation.

---

## Out of scope

- Whole-`versiond` host evacuation, replacement, and supervisor upgrade:
  [#1489](https://github.com/gonka-ai/gonka/pull/1489).
- Kubernetes rollout mapping and manifests.
- Back-pressure policy for intentionally rapid governance SHA churn.
- Artifact download size limits as additional defense in depth.


a-kuprin (approved):
https://github.com/gonka-ai/gonka/pull/1437 was already reviewed and approved

PR #1491, opened 2026-07-20
fix(security): dial-time SSRF guard for participant InferenceUrl (#1470)

### LINKED ISSUE #1470: Security: Residual SSRF on InferenceUrl — DNS/rebind + validator redirect (incomplete fix after #505/#534)
## Summary

**Residual SSRF** after prior paid mitigations (v0.2.7 / v0.2.8):

| Prior fix | What it covered | What remains broken |
|-----------|-----------------|---------------------|
| [PR #534](https://github.com/gonka-ai/gonka/pull/534) | TA client **no redirects** (`NewNoRedirectClient`) | Direct dial to DNS→private IP still allowed |
| [PR #505](https://github.com/gonka-ai/gonka/pull/505) | `ValidateURLWithSSRFProtection` rejects **literal** private IPs | **No DNS resolve**; hostnames always pass |

Any participant-controlled `InferenceUrl` hostname (or public URL that **redirects**) can force **validator DAPIs** (and other sinks) to connect to loopback, link-local metadata (`169.254.169.254`), or RFC1918 during **mandatory** payload retrieval / PoC validation / TA executor forwarding.

This is incomplete remediation, not a greenfield finding.

---

## Severity (self-assessment)

**High** — protocol-mandated HTTP from validator hosts to attacker-controlled URL; host infrastructure compromise (IMDS, localhost admin, internal network). Not direct on-chain fund drain.

---

## Architecture / threat model

```
Attacker sets InferenceUrl (hostname / rebind / open redirect)
        │
        ▼
   Chain: Participant.InferenceUrl
        │
   ┌────┴─────────────────┬──────────────────┐
   ▼                      ▼                  ▼
Validator payload     TA → executor       PoC off-chain
retrieval (primary)   POST                proof HTTP
DEFAULT http.Client   NoRedirectClient    multi-worker
FOLLOWS REDIRECTS     still dials DNS IP
```

**Attacker preconditions**

- Can set/update `InferenceUrl` via `MsgSubmitNewParticipant` / unfunded variant while `OpenRegistrationPermission` allows, **or**
- Already has a **hostname** URL on-chain → **DNS rebinding** needs **no new tx**

**Victim:** any honest host running DAPI that validates / TAs / PoC-validates against that participant.

---

## Root cause (registration gate)

`inference-chain/x/inference/utils/signature_and_url_validation.go`:

```go
func ValidateURLWithSSRFProtection(fieldName, raw string) error {
    // scheme http/https + host non-empty
    if isLocalhost(host) { return err }
    ip := net.ParseIP(host)
    if ip != nil {
        if isPrivateIP(ip) { return err }
    }
    return nil // hostname path: NEVER resolves DNS
}
```

Called from:

- `types/message_submit_new_participant.go` `ValidateBasic`
- `types/message_submit_new_unfunded_participant.go` `ValidateBasic`

Unit tests only cover: public URL, `localhost`, `192.168.0.1` — **no hostname/DNS cases**  
(`signature_and_url_validation_test.go`).

### Matrix (registration)

| URL | Gate result | Notes |
|-----|-------------|--------|
| `http://127.0.0.1/` | REJECT | literal |
| `http://169.254.169.254/` | REJECT | literal |
| `http://localtest.me/` | **ACCEPT** | DNS → 127.0.0.1 |
| `http://ssrf.attacker.tld/` | **ACCEPT** | attacker DNS |
| `http://0x7f000001/` / decimal IP strings | **ACCEPT*** | `ParseIP` fails → hostname |
| `http://[::ffff:127.0.0.1]/` | REJECT | mapped IPv6 parsed |

\*Depends on OS resolver if those “hostnames” resolve.

---

## URL write path

`keeper/msg_server_submit_new_participant.go`:

- **Create** participant with `Url`
- **Update** existing: `existing.InferenceUrl = msg.Url` (same permission gate)

DAPI registers `PublicUrl` via `participant_registration.go`.

If registration later closes, create/update txs fail — **rebinding still works** for hostnames already stored.

---

### 1) Validator payload retrieval — **primary**

```
validateInferenceAndSendValMessage
  → retrievePayloadsWithRetry  // up to 10 attempts, long backoff
      → RetrievePayloadsFromExecutor
          → Participant.InferenceUrl from chain
          → BuildPayloadRequestURL(..., "v1/inference/payloads", id)
          → FetchPayloadsHTTP(payloadRetrievalClient, ...)
```

```go
// decentralized-api/internal/validation/payload_retrieval.go
var payloadRetrievalClient = &http.Client{
    Timeout: 30 * time.Second,
    // no CheckRedirect → default Go client follows up to 10 redirects
    // no DialContext IP ACL
}
```

Also sends validator identity headers (`X-Validator-Address`, signature, epoch) to whatever is dialed.

### 2) Transfer Agent → executor

```
getExecutorForRequest → GetRandomExecutor → executor.InferenceUrl
POST via NewNoRedirectClient  // redirects blocked (#534)
```

Direct **DNS → private IP** still dials.

### 3) PoC off-chain validator

`decentralized-api/poc/validator.go` — loads `InferenceUrl`, parallel proof HTTP workers.

### 4) Devshard

Reuses `FetchPayloadsHTTP` / `BuildPayloadRequestURL`.

### Not a fix: proxy sidecar

`proxy/sidecar` resolves DNS and skips private IPs only when building nginx **whitelist** — does **not** stop DAPI egress SSRF.

---

### A — DNS hostname (register/update)

1. Point `ssrf.attacker.tld` → `169.254.169.254` / internal IP / loopback  
2. `MsgSubmitNewParticipant{ Url: "http://ssrf.attacker.tld" }`  
3. Passes `ValidateBasic`  
4. When peers validate / TA / PoC-hit this host → SSRF

### B — DNS rebinding (no chain update)

1. Register hostname with public A record  
2. During validation window, flip A to private (low TTL)  
3. No on-chain re-validation at fetch time  

### C — Redirect residual (validator payload client only)

1. `InferenceUrl = http://attacker.example/open` (public, accepts registration)  
2. Attacker responds `302 Location: http://127.0.0.1:9200/admin/v1/config` (or IMDS)  
3. TA client will **not** follow; **`payloadRetrievalClient` will**  
4. Can turn this into **localhost admin/config impact on validators** without publishing admin ports  

---

## Lab proof (safe, local)

Using public DNS name `localtest.me` → `127.0.0.1`:

1. `http://127.0.0.1:PORT/...` → **REJECT** (literal)  
2. `http://localtest.me:PORT/...` → **ACCEPT** (hostname)  
3. Validator-style HTTP GET retrieves loopback service body (exfiltrates lab secret)

Runnable replica (research package): registration gate replica + e2e fetch against local victim service.

---

## Impact

| Impact | Detail |
|--------|--------|
| Confidentiality | Cloud IMDS, internal HTTP, localhost admin/config on validators |
| Integrity | Pivot into host networks; secondary compromise of DAPI/node |
| Availability | Hang/slow validators on attacker-controlled responses; retry amplification (×10) |
| Scale | Every honest validator that retrieves payloads becomes an SSRF client |

---

## Required remediation

1. **Dial-time deny-list** (mandatory): shared safe `http.Client` that resolves, rejects loopback/link-local/RFC1918/ULA, dials only public IPs  
2. Apply to **all** sinks: `payloadRetrievalClient`, PoC proof client, TA client (defense-in-depth)  
3. Disable redirects on payload client **or** re-check `Location` host/IPs with the same policy  
4. Registration-time resolve is optional extra; **not sufficient alone** (rebinding)  
5. Regression tests: hostname→127.0.0.1, hostname→169.254.169.254, public→302 private, IPv4-mapped IPv6, decimal/hex host strings  

### Sketch

```go
// DialContext: LookupIP → reject isPrivateIP → dial public only
// CheckRedirect: ErrUseLastResponse OR validate Location with same policy
```

---

## Evidence index

| Path | Role |
|------|------|
| `inference-chain/x/inference/utils/signature_and_url_validation.go` | Gate gap |
| `inference-chain/x/inference/utils/signature_and_url_validation_test.go` | Thin tests |
| `inference-chain/x/inference/types/message_submit_new_participant.go` | ValidateBasic |
| `inference-chain/x/inference/keeper/msg_server_submit_new_participant.go` | Create/update URL |
| `decentralized-api/internal/validation/payload_retrieval.go` | Client + sink |
| `decentralized-api/internal/validation/inference_validation.go` | Retry loop |
| `decentralized-api/internal/server/public/post_chat_handler.go` | TA NoRedirect + executor URL |
| `decentralized-api/poc/validator.go` | PoC sink |
| `inference-chain/app/upgrades/v0_2_8/upgrades.go` | Prior bounty / PR #505 #534 context |

---

## Disclosure

- Static analysis + **local lab only**; no production hosts targeted.  
- If public issues are not preferred, please close and accept via **HackerOne** ([gonka.ai report vulnerability](https://gonka.ai/docs/report-vulnerability/)); we can re-file privately with PoC artifacts.

Ryanchen911:
Fixes #1470.

## Summary

Closes the residual SSRF left after #505/#534. The registration gate only rejects **literal** private IPs and does no DNS resolution, so any participant-controlled `InferenceUrl` hostname passes — and DNS rebinding / attacker-controlled DNS can force validator/TA DAPIs to connect to loopback, cloud metadata (`169.254.169.254`), or RFC1918 during **mandatory** payload retrieval, PoC proof fetch, and TA→executor forwarding.

The real fix is at **dial time**: a `net.Dialer.Control` hook re-checks the resolved IP on **every** connection (including each redirect hop and dual-stack candidate), which is what defeats rebinding. A registration-time resolve alone cannot.

## What changed

**Dial-time guard** applied to all sinks that dial a participant-controlled `InferenceUrl`:

| # | Sink | Location |
|---|------|----------|
| 1 | Validator payload retrieval | `internal/validation/payload_retrieval.go` |
| 2 | PoC proof fetch | `poc/proof_client.go` |
| 3 | TA → executor forward | `internal/server/public/post_chat_handler.go` |
| 4 | devshard payload fetch | `internal/devshard`, `cmd/devshardd` |
| 5 | devshard transport dialer | `devshard/transport` |

New helpers: `decentralized-api/internal/httpguard` (reuses the chain's `utils.IsPrivateIP`) and `devshard/transport/ssrf.go` (a deliberate mirror, since the `devshard` module can't import the chain utils).

**Redirects**: the payload/proof clients refuse redirects (`http.ErrUseLastResponse`), so a public `InferenceUrl` cannot `302` a validator to a private target (attack chain C).

## Correctness: shared clients

The server and devshard HTTP clients are **shared** between the operator's own trusted ML nodes (`node.InferenceUrl()`, routinely on localhost/private ranges) and untrusted participant URLs. Blanket-guarding them would break local ML calls. The guard is therefore applied via **separate clients used only for the participant path** — the trusted local-node path is never guarded.

## Deployment switch (secure by default)

Local dev / docker-compose / testermint register docker-internal hostnames (e.g. `http://genesis-api:9000`) that resolve to private IPs, so a hard dial-time block would break them. A new switch — `DAPI_API__ALLOW_PRIVATE_INFERENCE_URLS` (and `DEVSHARD_ALLOW_PRIVATE_ADDRESSES` for standalone devshardd), **default false = blocked** — is enabled only in those environments (`docker-compose-base.yml`, `DockerGroup.kt`). Production templates leave it unset, so real validators stay protected.

## Registration gate

Intentionally left **DNS-free**: `ValidateBasic` is stateless and must be deterministic across validators — network I/O there would break consensus, and a registration-time lookup can't stop rebinding regardless. A code comment documents this. Also fixes an IPv4-mapped IPv6 gap in `isPrivateIP` so `::ffff:169.254.169.254` is caught.

## Tests

- Dial guard blocks: hostname→`127.0.0.1`, →`169.254.169.254`, public→`302`→private, IPv4-mapped IPv6, decimal/hex host encodings, RFC1918/ULA/link-local.
- `allowPrivate=true` permits all of the above (dev/test regression).
- The trusted local ML path is never guarded.
- Redirects are refused by the payload/proof clients.

Verified with `go build ./...` and `go test ./...` across `decentralized-api`, `devshard`, and `inference-chain` (only pre-existing Docker-dependent testcontainers tests are skipped locally for lack of a Docker host).


a-kuprin:
Registration only blocks literal private IPs / localhost on `InferenceUrl`. A hostname passes without DNS. Later, gateways/validators may dial that URL, so DNS can make the dial land on a private IP.

**Informational (not High/Medium):**

1. **Attacker chooses host/IP via DNS, not the path.** Your client always hits fixed routes like `/v1/inference/payloads/...` or `/devshard/{version}/...`. On IMDS or random internal services that is usually **404 / wrong handler**, not a chosen sensitive path.

2. **Data is not stolen.** The response stays on the dialing node. There is no solid exfil back to the attacker.

3. **No economic attack.** No escrow drain, no consensus break, no profitable protocol abuse.

4. **Self-limiting.** If the URL always points private, the “attacker” cannot really serve as a host/executor and gets missed/penalized. They are not a clean remote attacker with upside.

5. **Max impact** is roughly: “did something on this IP:port answer HTTP?” — noise/probe, not takeover.

**Disposition:** Treat as incomplete defense-in-depth on egress. Optional hardening (dial-time private-IP block) is fine; severity **Informational**.


**For an Informational / constrained-path issue, this PR is larger and more invasive than it needs to be — and it still misses a live sink.**

**~21 files / +622 for wiring + duplication**, not because the idea needs that surface

## Verdict on [#1491](https://github.com/gonka-ai/gonka/pull/1491)

**Technically valid approach, overscoped for the real risk.** CI is green (API /devshard / chain / lint). Dial-time `Control` + no-redirect on participant clients is the right *kind* of fix. Leaving registration DNS-free is correct.

But for an Informational / constrained-path issue, this PR is larger and more invasive than it needs to be — and it still misses a live sink.


### Smarter / smaller design

Two choke points cover almost everything that matters:

| Place | Covers |
|-------|--------|
| `common/validation.PayloadRetrievalClient` (guarded dial + no redirect) | All `FetchPayloadsHTTP` callers, including standalone `devshardd` |
| `devshard/transport.getTransport` `Control` | Gateway/host peer dials |

Optional: one env (`ALLOW_PRIVATE_…`) read at those two sites.

**Skip or defer:** classic TA/`post_chat`, PoC client churn, chain `ValidateBasic` essay (short comment enough), duplicated `isPrivateIP` (reuse one helper via `common` if both modules can import it).

That is roughly **3–6 files**, not 21.


### Bottom line
- **Valid as defense-in-depth**, CI-passing.
- **Not smart relative to impact:** too broad, some deprecated sinks, duplicate policy, and the important standalone payload default client is still unguarded.
- #1491 is oversized

a-kuprin (changes requested):
Oversized PR

Ryanchen911:
Thanks for the review — agreed on scoping to the real risk. Pushed 489b80f narrowing the PR to the two choke points you identified.

**Kept**
- Payload-retrieval client (guarded dial + no-redirect) — covers every `FetchPayloadsHTTP` caller, including standalone `devshardd`.
- `devshard/transport.getTransport` `Control` — gateway/host peer dials.

**Reverted**
- TA→executor forward client split (`post_chat_handler` / `server`)
- PoC proof-client `AllowPrivate` plumbing
- Long `ValidateBasic` comment → one-liner
- Unused exported `IsLocalhost` helper

Net is now ~11 source files (config + one env switch + the two guards + tests + deploy toggles), down from 21.

**On "standalone payload default client still unguarded":** it is guarded in this PR — the standalone `devshardd` validator dials payloads through `v.payloadHttpClient`, built via `httpguard.NewClient(...)` in `cmd/devshardd/main.go` and passed into `ValidateInferenceWithExecutor` (`cmd/devshardd/validation.go`). I went with constructor injection rather than a shared package-level default client because the two binaries source the toggle from different env vars (`DAPI_API__ALLOW_PRIVATE_INFERENCE_URLS` vs `DEVSHARD_ALLOW_PRIVATE_ADDRESSES`) and the ML-node path on the same struct must stay unguarded — so a single global default couldn't serve both. Happy to switch to a shared default client if you'd prefer that shape.

**On the duplicated `isPrivateOrLocalAddr`:** the `devshard` module has no dependency on `github.com/productscience/inference`, so the chain util can't be imported there. Left it as a documented mirror rather than adding a module dependency just for one predicate — let me know if you'd rather I add the dep or a small shared module.

Ryanchen911:
hi @a-kuprin , can you please check this PR again, thanks a lot!

a-kuprin:
@Ryanchen911 I've added it to v5 milestone. I believe it will be merged next week

a-kuprin:
@Ryanchen911 Please rebase. It should be targeted to [devshard-0.2.15-v5](https://github.com/gonka-ai/gonka/tree/devshard-0.2.15-v5)

Ryanchen911:
Rebased onto `devshard-0.2.15-v5`. The v5 refactor let me shrink this further and drop the duplication you flagged: **21 files → 14**.

**Both choke points are now exactly the two you named:**

- `common/validation.PayloadRetrievalClient` — now guarded + no-redirect. The single caller (`devshard/cmd/devshardd/inference/validate.go:107`) passes `nil`, so it lands on this default client.
- `devshard/transport.getTransport` — guard on the dialer.

**The duplicated `isPrivateIP` is gone.** On v5 the dependency graph changed: `devshard` → `common` → `inference-chain`. Previously the devshard module couldn't import the chain util, which is why I'd mirrored the predicate. Now the guard lives once in `common/httpguard` and both choke points share the same `utils.IsPrivateIP` as the registration gate.

**All decentralized-api changes are gone.** v5 moved payload retrieval into `common`, and dapi no longer imports `common/validation` or `devshard/transport` — so the config field, the `main.go` wiring, the `internal/httpguard` package and the `DAPI_API__ALLOW_PRIVATE_INFERENCE_URLS` flag were all dead code. `apiconfig/config.go` and `main.go` are now byte-identical to the target branch.

The flag is a single `DEVSHARD_ALLOW_PRIVATE_ADDRESSES` (default false = secure), wired into the three processes that actually dial: devshardd, devshardctl, and the Testermint in-container devshardctl exec. Production templates under `deploy/join/` leave it unset.

One note on scope, since you'd asked about the standalone path earlier: devshardctl also dials through `transport`, so it needs the same opt-out in citest/Testermint — without it the e2e stacks would be blocked by the guard.

I kept own-ML-node and self-hosted-endpoint dials unguarded, and verified each: the ML engine uses its own client (`inference/engine.go:68`), and chain RPC / public API go through `common/chain` and `NewChainPhaseGate`. Those legitimately point at private addresses in production (`node:9090`, `http://api:9000`). Only `HostInfo.URL`, which comes from the on-chain `Participant.InferenceUrl` (`bridge/grpc.go:117`), is guarded.

Tests: four modules build; `common/httpguard`, `common/validation`, `devshard/transport`, `devshard/protocol`, `devshard/cmd/devshardd/...`, `devshard/cmd/devshardctl`, the gencompose golden test, and the chain utils tests all pass. `devshard/transport` and `devshard/protocol` needed a `TestMain` opt-out — they run `httptest` on loopback through the real dialer. The storage packages fail locally only on `rootless Docker not found` (testcontainers), unrelated to this change.

Still happy to trim further if you'd prefer the devshardctl/Testermint wiring split into a separate PR.

PR #1492, opened 2026-07-21
devshard(fix): Decouple Postgres connect from session-index rebuild

a-kuprin:
### Summary

- Fixes hosts stuck returning 500s (`postgres storage is configured but
  unavailable: index existing sessions: repair session index for <escrow>:
  timeout`) where one slow/leftover escrow's index repair blocked all live
  sessions.
- Decouples Postgres **connect** (`PG_CONNECT_TIMEOUT`, 2s) from **session
  index rebuild** (`PG_INDEX_TIMEOUT`, new, 30s); rebuild runs asynchronously
  and Postgres is only promoted once its index is ready.
- Makes index repair batched and skip-aware (set-based `unnest` DELETE/INSERT
  chunked at 1000; no writes on the happy path).
- Turns mid-rebuild requests into retryable 503 `initializing` and wires storage
  readiness into `/ready`.

## Problem statement

Hosts running in Postgres storage mode (`PGHOST` set) were failing **all** new
session traffic with HTTP 500s like:

```text
init storage session: devshard postgres storage is configured but unavailable:
index existing sessions: repair session index for 31069:
timeout: context deadline exceeded: escrow 33355
```

Root cause was a coupling in how Postgres storage opened:

1. `NewPostgres` did connect + ping + migrate **and** a full session-index
   rebuild (`indexExisting`) under a single short context
   (`PG_CONNECT_TIMEOUT`, default 2s).
2. `indexExisting` reconciled `devshard_session_index` against
   `devshard_sessions` with **one `INSERT ... ON CONFLICT DO NOTHING` per
   escrow, on every boot/reconnect** — even when the index already matched.
   This is O(N) round-trips and easily blows a 2s budget on a slow/large DB.
3. When the rebuild timed out, `NewPostgres` failed, so the hybrid router stayed
   in **degraded sqlite-owned-only** mode and rejected every new escrow with
   `ErrStoragePostgresUnavailable`.
4. The background reconnect loop retried the same coupled open and kept failing
   with the same `repair session index for <old escrow>` timeout.

Net effect: one stuck/slow index repair on a leftover escrow (e.g. `31069`,
already gone on chain) blocked live escrows (`33355`, `33481`, …) indefinitely.

Secondary issues found in the original repair path:

- Stale-index deletes and repair inserts were per-row (N round-trips).
- The repair loop re-`INSERT`ed every session even when the index row already
  matched — pointless writes on the happy path.

---

### 1. Batched, skip-aware index repair

`indexExisting` was refactored into:

- `loadSessionsForIndex` — one `SELECT` of `devshard_sessions` (+ epoch-conflict
  check).
- `diffSessionIndex` — one `SELECT` of `devshard_session_index`; already-correct
  rows are recorded in `indexedOK` and **skipped**, mismatched/orphan rows go to
  a stale list.
- `deleteStaleSessionIndex` / `insertMissingSessionIndex` — set-based
  `DELETE ... USING unnest(...)` and `INSERT ... SELECT unnest(...)`, chunked by
  `postgresIndexRepairBatchSize = 1000` via `forEachEscrowEpochBatch`.

Happy path (index already consistent) now issues **zero** repair DML.

### 2. Connect timeout separated from index rebuild

- `NewPostgres` now returns after connect + ping + migrate; the session-index
  rebuild runs in a background goroutine bounded by a new
  `PG_INDEX_TIMEOUT` (default 30s), independent of `PG_CONNECT_TIMEOUT`.
- New readiness API on `*Postgres`: `WaitReady(ctx)`, `Ready()`, and internal
  `waitReadyForOp()`. `Close()` cancels the in-flight rebuild.
- Boot (`openPostgresReady`) and hybrid **reconnect** now `WaitReady` under
  `PG_INDEX_TIMEOUT` **before** promoting Postgres, so a handle is only promoted
  once its routing index is fully rebuilt (never with an empty index).
- The rebuilt `escrowIdx` / `knownEpochs` maps are published atomically under
  `s.mu` (no off-lock mutation from the goroutine).

### 3. Request gating + readiness signalling

- Escrow-keyed Postgres operations gate on `waitReadyForOp()` (bounded by
  `postgresOpTimeout`), so a request during rebuild **waits** instead of racing
  an empty index. On timeout it returns the new transient sentinel
  `storage.ErrStorageIndexRebuilding`.
- `devshard/server/routes.go` maps `ErrStorageIndexRebuilding` to
  **`initializing` → HTTP 503** (retryable), not a 500 `storage_err`.
- Storage readiness is surfaced through `HybridStorage.Ready()` →
  `ManagedStorage.Ready()` → `HostManager.StorageReady()`, and the admin
  `/ready` endpoint now returns 503 (with `"storage_ready": false`) until the
  index is ready.

### Behavior summary

| Situation | Before | After |
| --- | --- | --- |
| Slow/large index rebuild | Connect fails at 2s → degraded, rejects new escrows | Connect succeeds; rebuild runs up to `PG_INDEX_TIMEOUT`; promote only when ready |
| Happy-path reconnect | O(N) `INSERT ON CONFLICT` per escrow | 1 `SELECT` + set-based diff; zero DML when consistent |
| Request during rebuild | 500 `storage_err` | Waits; on timeout 503 `initializing` (retryable) |
| `/ready` while rebuilding | 200 (storage state ignored) | 503 with `storage_ready:false` |

### Tests

- `storage`: batched repair (missing/stale/kept), skip-indexed unit,
  `forEachEscrowEpochBatch` chunk-by-1000, `WaitReady` blocks until index,
  connect returns before index, reconnect waits for ready before promote
  (all pass under `-race`).
- `server`: `ErrStorageIndexRebuilding` → 503 `initializing` + metric reason.
- `cmd/devshardd`: `/ready` reflects storage readiness.

### New / changed env knobs

| Env | Default | Scope |
| --- | --- | --- |
| `PG_CONNECT_TIMEOUT` | `2s` | dial + ping + migrate only |
| `PG_INDEX_TIMEOUT` | `30s` (new) | async session-index rebuild / `WaitReady` |

### DIFF
--- a/devshard/cmd/devshardd/app.go
+++ b/devshard/cmd/devshardd/app.go
@@ -111,7 +111,7 @@ func buildApp(ctx context.Context, cfg runtimeConfig) (_ *devshardApp, err error
 	e := buildServer(lifecycle)
 	var admin *echo.Echo
 	if cfg.AdminAddr != "" {
-		admin = buildAdminServer(lifecycle)
+		admin = buildAdminServer(lifecycle, manager.StorageReady)
 	}
 	manager.Register(e.Group(""))
 	chainRuntime.chainEvents.OnReady(lifecycle.SetReady)
--- a/devshard/cmd/devshardd/lifecycle_test.go
+++ b/devshard/cmd/devshardd/lifecycle_test.go
@@ -14,7 +14,7 @@ import (
 func TestLifecycleReadyAndDrainStatus(t *testing.T) {
 	lifecycle := newLifecycleState()
 	e := buildServer(lifecycle)
-	admin := buildAdminServer(lifecycle)
+	admin := buildAdminServer(lifecycle, func() bool { return true })
 	e.GET("/work", func(c echo.Context) error {
 		time.Sleep(20 * time.Millisecond)
 		return c.String(http.StatusOK, "done")
@@ -62,7 +62,7 @@ func TestLifecycleDrainRejectsNewWork(t *testing.T) {
 	lifecycle := newLifecycleState()
 	lifecycle.SetReady(true)
 	e := buildServer(lifecycle)
-	admin := buildAdminServer(lifecycle)
+	admin := buildAdminServer(lifecycle, func() bool { return true })
 	e.GET("/work", func(c echo.Context) error {
 		return c.String(http.StatusOK, "done")
 	})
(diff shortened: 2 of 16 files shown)


x0152 (approved):
I think we can safely include it in the 0.2.14-v4 release

PR #1494, opened 2026-07-22
fix(broker): prevent panic submitting to shut-down worker

redstartechno:

`NodeWorker` can panic the whole `decentralized-api` process with `panic: send on closed channel`. The submit path uses `select { case w.commands <- cmd: ... default: ... }`, which protects against a **full** channel but not a **closed** one. When a node is removed while a reconcile pass is still running, `Shutdown()` closes `w.commands`; a submission racing that close sends on a closed channel and crashes the process — there is no recover on this path.

Code-traced from the broker's command flow: `Shutdown` closes `w.commands` while the reconcile loop can still hold the worker reference and submit. Nothing synchronizes the two. On the current base the production send is `reconcile()` at `broker/broker.go:1087` (`worker.submit(ctx, cmd, generation)`); the exported `Submit` has no production callers today, which is why the guard lives in the private `submit` rather than in the wrapper. The regression tests reproduce the crash deterministically on unguarded code (traces below), and the race detector independently flags the same window.

Adds a `closed bool` guarded by a `closedMu sync.Mutex`. `submit` checks the flag under the lock and returns `false` — the existing "not accepted" result callers already handle for the full-channel case — instead of sending on a closed channel. `Shutdown` sets the flag under the same lock before closing, and is idempotent: a second call returns instead of double-closing `w.shutdown`.

The lock is held across the channel send. That is only acceptable because the send is non-blocking (`select` with `default`), so the critical section is bounded; this is documented at the `submit` doc comment so a later refactor to a blocking send does not silently introduce a stall. As a side effect the lock also closes a latent `WaitGroup` Add/Wait race between submit and shutdown.

- A mutex acquisition is added to the submit path. It is uncontended except during shutdown, so the cost is a few nanoseconds; the non-blocking channel send dominates.
- Behavior change: submitting to a shut-down worker now returns `false` instead of panicking. `false` is the pre-existing "command not accepted" contract, so callers need no changes — and the previous behavior was a process crash.
- The added latency slightly widens a pre-existing timing window in `TestNodeWorker_QueueFull`, which this PR addresses directly (see below).

- `decentralized-api/broker/node_worker.go` (fix)
- `decentralized-api/broker/node_worker_test.go` (regression tests; plus the `TestNodeWorker_QueueFull` assertion change requested in review)

No API/proto/state changes; the broker's public surface is unchanged.

Fail-first, then fix, for each guard:

- **Submit-after-shutdown**: with the `closed` check removed, `TestNodeWorker_SubmitAfterShutdown` / `TestNodeWorker_ConcurrentSubmitAndShutdown` fail with `panic: send on closed channel` plus a `DATA RACE` between `submit` and `run`. Restored, they pass.
- **Idempotent shutdown**: with the `if w.closed` early-return removed from `Shutdown`, `TestNodeWorker_ShutdownIdempotent` fails with `panic: close of closed channel` at `node_worker.go:130`. Restored, it passes.

## Testing & evidence

Linux (WSL2), Go 1.25.9 linux/amd64, base `upgrade-v0.2.16` @ `7fb3d00d9`.

```
$ go build ./...                       # exit 0
$ go vet ./broker/...                  # clean
$ go test ./broker/... -race -count=1
ok  	decentralized-api/broker	30.859s

$ go test ./broker/ -race -count=10 \
    -run 'TestNodeWorker_QueueFull|TestNodeWorker_SubmitAfterShutdown|TestNodeWorker_ShutdownIdempotent'
ok  	decentralized-api/broker	11.854s      # 30/30

$ go test ./broker/ -race -count=20 -run TestNodeWorker_QueueFull
ok  	decentralized-api/broker	22.315s      # 20/20, repeated twice
```

### Review round

Three test changes requested in review, all in `node_worker_test.go`:

1. Added `TestNodeWorker_ShutdownIdempotent` (the description previously claimed it; the code was already idempotent, the test was missing).
2. `TestNodeWorker_QueueFull` relaxed from exact counts to bounds.
3. Dropped the `100ms` sleep in `TestNodeWorker_SubmitAfterShutdown` — with `closed` set before `Shutdown` returns, the result is deterministic.

On (2), the measured direction is the opposite of the one suggested, so the bounds went the other way. The worker's `run()` dequeues a command and then executes it, which frees a buffer slot while the submit loop is still running — so **more** than queue-size submissions can succeed, never fewer. Every observed failure is `submitted == 11, failed == 14` (sum 25). `submitted <= 10` would exclude exactly the value that drifts, leaving the test as flaky as the exact-count version: measured 6/20 and 3/20 failures under `-race` with that bound, versus 3/20 for the original assertions.

The invariant that holds is the mirror: the channel starts empty and only the worker drains it, so the first queue-size sends cannot fail.

```go
assert.GreaterOrEqual(t, slowCmdSubmitted, 10, "Queue-size submissions must always succeed")
assert.LessOrEqual(t, slowCmdFailed, 15, "At most the commands beyond queue size may fail")
assert.Equal(t, 25, slowCmdSubmitted+slowCmdFailed, "Every submission should either succeed or fail")
```

With those bounds: 40/40 pass under `-race -count=20` across two independent samples, 30/30 without `-race`.

Note this flake is `-race`-only; `.github/workflows/verify.yml` runs plain `go test ./... -v`, so it was never reddening CI either way.


redstartechno:
CI triage of the `build` (Multi-Platform Build) failure — infra flake, not related to this PR's diff:

- Failure: [run 29945973154](https://github.com/gonka-ai/gonka/actions/runs/29945973154/job/89011305151) died at Docker stage `[builder 11/14]` while linking the **inference-chain** binary: `link: mapping output file failed: no space left on device` — runner disk exhaustion.
- This PR touches only `decentralized-api/broker/` (2 files); the failing step compiles `inference-chain/cmd/inferenced`, which this PR does not modify. The job that actually builds and tests the changed component, **Build and Test API Wrapper**, is green, as are all other test jobs.
- Same signature has hit unrelated branches before: the identical `no space left on device` link failure was diagnosed on #1336 ([run 29056214849](https://github.com/gonka-ai/gonka/actions/runs/29056214849), 2026-07-09) alongside three unrelated same-day branches. The same workflow also failed yesterday on `ak/postgres-connect-timeout-fix` ([run 29815400340](https://github.com/gonka-ai/gonka/actions/runs/29815400340), wasmvm download hiccup) and passed on re-run 2.5h later.

Should clear on a maintainer re-run; nothing to change on the PR side.

redstartechno:
Rebased onto `main` at `379bebced`. #1536 refactored the submit path into a private `submit(ctx, cmd, generation uint64)` behind the exported `Submit`, which conflicted with this patch.

Two things changed as a result, both mechanical:

- The guard now lives in `submit()` rather than `Submit()`. That is the function performing the channel send, so it also covers the second caller — the deployment-generation path in `broker.go` — which `Submit()` alone would have missed. `generation` is threaded through the send unchanged.
- The two regression tests moved past the tests #1536 added; they are otherwise byte-identical.

`Shutdown()` is unchanged from the previous revision.

Re-verified on the new base (`-race`, Linux, Go 1.25.9):

```
go build ./broker/...                  # exit 0
go vet ./broker/...                    # clean
go test ./broker/... -race -count=1    # ok  decentralized-api/broker  31.076s
```

The two regression tests under `-race -count=10`: 20/20 pass, no races.

Fail-first still holds on this base. Removing only the guard body from `submit()` and leaving everything else in place:

```
WARNING: DATA RACE
Read at 0x00c001d43120 by goroutine 40:
  runtime.chansend()
  decentralized-api/broker.(*NodeWorker).submit()
      .../node_worker.go:111
Previous write at 0x00c001d43120 by goroutine 41:
  runtime.closechan()
  decentralized-api/broker.(*NodeWorker).run()
      .../node_worker.go:83
...
--- FAIL: TestNodeWorker_SubmitAfterShutdown (0.10s)
panic: send on closed channel [recovered, repanicked]
```

Restoring the guard returns both tests to PASS.

One unrelated observation while running this, in case it saves someone time: full-package `go test ./broker/... -race -count=10` intermittently fails in `GetHardwareNodes()` mock setup (`broker_test.go:25`, via `broker.go:655`), from a timer-driven `SyncNodesCommand` outliving a test. That reproduces on pristine `379bebced` with nothing from this PR applied, so it is pre-existing and out of scope here.

redstartechno:
Refreshed this branch onto the current `upgrade-v0.2.16` head (`7fb3d00d9`). GitHub had the PR marked as conflicting against a stale base ref; it shows as mergeable again now.

I used a merge rather than a rebase deliberately, so that @a-kuprin's `83a0407cb` train merge is preserved rather than discarded.

The change relative to the new base is unchanged, still limited to the two declared files:

```
decentralized-api/broker/node_worker.go      | 21 +++++++++++++++-
decentralized-api/broker/node_worker_test.go | 37 ++++++++++++++++++++++++++++
2 files changed, 57 insertions(+), 1 deletion(-)
```

Verification on the merged head `426ffaac0`, Go 1.25.9 (linux/amd64):

- `go build ./...` — exit 0
- `go vet ./broker/...` — clean
- `go test ./broker/... -race -count=1` — ok
- `TestNodeWorker_SubmitAfterShutdown` and `TestNodeWorker_ConcurrentSubmitAndShutdown` under `-race -count=10` — 20/20 passed

The bug is still live on the current base: `NodeWorker` on `7fb3d00d9` has no `closed`/`closedMu`, so the send-on-closed-channel window is intact. The guard sits in the private `submit()`, so it covers both callers — `Submit()` and the deployment-generation path at `broker.go:1087`.

### One thing worth disclosing

`TestNodeWorker_QueueFull` is a pre-existing flaky test, and this change slightly widens its window under `-race`.

The test submits 25 commands from a single goroutine into a 10-slot queue while the worker goroutine drains it, then asserts exactly 10 submitted and 15 failed. If the worker dequeues one command before the submit loop finishes, an 11th submit succeeds and the assertion fails with `expected 10, actual 11` / `expected 15, actual 14`. Any change in submit-path latency shifts that probability, and the mutex adds a small amount.

Measured over 20 iterations per leg, each a separate process, **with** `-race`:

- pristine `upgrade-v0.2.16`: 1/20 failures
- this branch: 6/20 failures — identical assertion values every time

**Without** `-race`, which is how CI runs it (`go test ./... -v`), 30 iterations per leg:

- pristine: 0/30 failures
- this branch: 1/30 failures
- whole `./broker/...` package: 5/5 passed on both legs

So the effect is confined to `-race` runs and should not affect CI, but I would rather flag it than have it turn up later.

I am happy to make that assertion timing-independent (asserting bounds rather than exact counts) either here or in a separate PR, whichever you prefer — I did not want to expand this PR's scope unasked.

The separately reported `GetHardwareNodes()` mock-setup flake in full-package `-race` runs did not reproduce in this round.

a-kuprin (approved):

a-kuprin:
## Logical issues

No bug in the guard itself. A few gaps around it:

**1. Tests do not match the PR text.** The body still cites `TestNodeWorker_ShutdownIdempotent`. The diff has `TestNodeWorker_SubmitAfterShutdown` and `TestNodeWorker_ConcurrentSubmitAndShutdown`. Idempotent shutdown is untested.

**2. `TestNodeWorker_SubmitAfterShutdown` depends on `sleep(100ms)`.** After `Shutdown()` with an empty queue, `wg.Wait()` can return before `run()` closes `commands`. The sleep exists to make the old code panic. With the flag, the sleep is unnecessary; the concurrent test is the one that actually covers the race.

**3. `TestNodeWorker_QueueFull` gets flakier.** Exact `10/15` counts assume the worker does not drain during the submit loop. The extra mutex changes timing; the author measured 1/20 → 6/20 failures under `-race`. That assertion should use bounds, ideally in this PR so CI `-race` does not become noisier.

**4. Pre-existing, not solved here:** `RemoveWorker` holds the work-group write lock across `Shutdown()`/`Wait()`. That blocks `GetWorker` for the whole drain, but the stale-pointer race is already closed by this PR.

## Worth doing before merge

- Add a real `Shutdown` twice test (the description already claims it).
- Relax `TestNodeWorker_QueueFull` to bounds so this mutex does not make it flakier.

Nice-to-have: drop the 100ms sleep now that `closed` makes post-shutdown submit deterministic.

a-kuprin:
A few test gaps to close before merge:

1. **Add `TestNodeWorker_ShutdownIdempotent`.** The description says this exists; the diff does not. A second `Shutdown()` must return without panicking on `close(w.shutdown)`.

2. **Relax `TestNodeWorker_QueueFull`.** It asserts exact `10` submitted / `15` failed while the worker is draining. The extra mutex on `submit` makes that race worse (you measured 1/20 → 6/20 under `-race`). Assert bounds (`submitted <= 10`, `failed >= 15`, `submitted+failed == 25`) instead of exact counts.

3. **Drop the `100ms` sleep in `TestNodeWorker_SubmitAfterShutdown`.** After `Shutdown()`, `closed` is already set, so `Submit` is deterministically `false`. The sleep was only needed to make the unfixed code panic. `TestNodeWorker_ConcurrentSubmitAndShutdown` already covers the race.

Nits, not blockers: the write-up overstates callers — on this base the only production send is reconcile’s `worker.submit(...)`, not a general status/queue path. Holding `closedMu` across the send is fine only because it is non-blocking; that should stay documented.

redstartechno:
Thanks — all three addressed in `e79bb3785`, test-only, `node_worker.go` byte-identical to the previous head.

1. **`TestNodeWorker_ShutdownIdempotent` added.** You were right that the description claimed a test that wasn't there. The guard itself already existed, so this only pins it: removing the `if w.closed` early-return from `Shutdown` makes the new test fail with `panic: close of closed channel` at `node_worker.go:130`.

2. **`TestNodeWorker_QueueFull` relaxed — but in the opposite direction to the one suggested.** `submitted <= 10` / `failed >= 15` excludes exactly the value that drifts, so it stays as flaky as the exact-count version. `run()` dequeues a command and then executes it, freeing a buffer slot while the submit loop is still going, so **more** than queue-size can succeed, never fewer — every observed failure is `submitted == 11, failed == 14`.

   Measured under `-race -count=20`: original exact assertions 3/20 failures; with `submitted <= 10` 6/20 and 3/20 on reruns; with the mirrored bounds 0/20 twice.

   The invariant that actually holds is that the channel starts empty and only the worker drains it, so the first queue-size sends cannot fail:

   ```go
   assert.GreaterOrEqual(t, slowCmdSubmitted, 10, "Queue-size submissions must always succeed")
   assert.LessOrEqual(t, slowCmdFailed, 15, "At most the commands beyond queue size may fail")
   assert.Equal(t, 25, slowCmdSubmitted+slowCmdFailed, "Every submission should either succeed or fail")
   ```

   Happy to flip it back if you read the intent differently, but the numbers above are what I get.

3. **Sleep dropped** from `TestNodeWorker_SubmitAfterShutdown`.

On the nits: the description overstated the caller surface and has been corrected — on this base the only production send is `reconcile()` at `broker/broker.go:1087`, and the exported `Submit` has no production callers, which is why the guard sits in the private `submit`. The non-blocking-send rationale for holding `closedMu` across the send stays documented at the `submit` doc comment.

Not addressed, as it is pre-existing and outside this PR: `RemoveWorker` holding the work-group write lock across `Shutdown()`/`Wait()`. Happy to take it as a follow-up if you want it.

Full run on `e79bb3785` (Go 1.25.9, WSL2): `go build ./...` exit 0, `go vet ./broker/...` clean, `go test ./broker/... -race -count=1` ok, the three targeted tests `-race -count=10` 30/30, `TestNodeWorker_QueueFull -race -count=20` 20/20 twice.

GLiberman:
This fixed two races in the previous implementation:
- The run loop could close commands while reconciliation was sending to it, causing panic: send on closed channel.
- submit() could call WaitGroup.Add() concurrently with Shutdown() calling Wait(), which is unsafe.
But it preserved the problematic lifecycle order:
- RemoveWorker still waited before removing the worker.
- The broker command loop remained blocked.
- Queued commands were still executed after removal had been requested.
- Result delivery could itself block on the broker queue.
- The node was deleted from broker state only after worker shutdown finished.
So the relationship is:
- #1494 repaired concurrency inside the old synchronous shutdown design.
- #1701 replaced that synchronous shutdown design because the design itself caused the production liveness incident.
- #1701 still preserves the synchronization invariant discovered by #1494, using stopping plus a mutex, but no longer closes commands or waits from RemoveWorker.

PR #1495, opened 2026-07-22
test(dapi): fix spurious TryLock failure in event_listener integration test

redstartechno:

The CI job **Build and Test API Wrapper** intermittently fails on unrelated PRs with:

```
mock.go:89: TryLock called more than once
--- FAIL: TestRegularPocScenario
```

in `decentralized-api/internal/event_listener`. The failure is spurious — it hits branches that don't touch this package at all.

Two independent occurrences on changes that cannot affect this test:

- PR #1460 CI on 2026-07-16 (diff touches only `devshard/host`) — triaged in https://github.com/gonka-ai/gonka/pull/1460#issuecomment-4995850897
- Same-day failure on `feature/empty-validator-set-fallback` (https://github.com/gonka-ai/gonka/actions/runs/29468383638), also unrelated to this package.

Root cause, traced through the code: `MockClient.WithTryLock` (`decentralized-api/mlnodeclient/mock.go`) probes the mock's mutex with a non-blocking `TryLock()` and `t.Fatal`s if it is already held. But the mutex it probes is the same one every mock method takes, and broker command dispatch is asynchronous by design: `reconcile()` submits commands into per-node `NodeWorker` channels (`broker/broker.go` → `worker.Submit`), and each node's `NodeWorker.run()` goroutine executes them later, briefly locking the mock (`InitGenerateV2`, `Stop`, `InferenceUp`, …). The test synchronizes with fixed sleeps before asserting, which is a heuristic, not a handshake — on a slow or contended CI runner the assertion can fire exactly while a worker goroutine legitimately holds the lock, and the test fatals even though nothing was "called twice".

Replaces `WithTryLock(t, f)` with a blocking `WithLock(f)` (`Lock`/`defer Unlock`) and renames the 11 call sites in `integration_test.go`. An assertion now waits out an in-flight worker call instead of failing on it, which also makes the counter reads deterministic rather than sleep-dependent. The unused `testing` import is dropped.

- Test/mock code only; no production behavior changes.
- Deadlock risk of a blocking lock was checked explicitly: none of the 11 callback bodies calls a `MockClient` method that takes the mutex (they only read exported counter fields), there is no nested `WithLock`, and a `require` failure inside the callback still releases the lock via the deferred `Unlock`.
- The removed TryLock probe carried no real signal: "mutex held at assertion time" is legitimate under the async dispatch design and never distinguished a product bug from scheduler timing. A truly wedged worker would now surface as a test timeout with a goroutine dump instead of a misleading fatal.

The spurious-failure mechanism (non-blocking probe racing an async worker) is removed entirely — there is no code path left that can produce the `TryLock called more than once` fatal. The affected suites pass repeatedly under the race detector (evidence below). The flake does not reproduce on an idle machine by nature (it needs a starved runner), which is consistent with it only ever appearing in CI.

- `decentralized-api/mlnodeclient/mock.go` (mock helper)
- `decentralized-api/internal/event_listener/integration_test.go` (call-site renames)

## Testing & evidence

Linux (WSL2), `GOTOOLCHAIN=go1.24.2`, base `8a35022be`.

### Before

Upstream CI occurrences of the spurious failure (unrelated diffs):

```
mock.go:89: TryLock called more than once
--- FAIL: TestRegularPocScenario
```
- https://github.com/gonka-ai/gonka/actions/runs/29468383638 (feature/empty-validator-set-fallback)
- PR #1460's "Build and Test API Wrapper" run, 2026-07-16 (devshard-only diff)

### After

```
$ go test ./internal/event_listener/ ./mlnodeclient/ -race -count=2
ok  	decentralized-api/internal/event_listener	38.539s
ok  	decentralized-api/mlnodeclient	1.435s

$ go vet ./internal/event_listener/ ./mlnodeclient/
(clean)
```


redstartechno:
Data point: this exact flake just failed CI on `main` itself — run https://github.com/gonka-ai/gonka/actions/runs/30409535828 (commit `182d0820c`, 2026-07-28) has `Build and Test API Wrapper` red with:

```
TestRegularPocScenario
mock.go:90: TryLock called more than once
```

(visible in that run's `api-junit-report` artifact). That's the third recorded occurrence of this signature, and the first directly on `main`. This PR's blocking `WithLock` change removes the race, so merging it would clear this class of red runs.

a-kuprin (approved):

aikuznetsov (approved):

redstartechno:
Another live occurrence, this time on a different open PR rather than on `main`.

CI for #1787 (an unrelated one-line testermint patch fix) went red on `Build and Test API Wrapper`:

```
--- FAIL: TestRegularPocScenario (0.00s)
    mock.go:91: TryLock called more than once
```

Run: https://github.com/gonka-ai/gonka/actions/runs/35130269007 (job 104909244963), 2026-09-16.

`mlnodeclient/mock.go:91` is the `t.Fatal("TryLock called more than once")` that this PR removes — the assertion helper is non-blocking, so a per-node broker worker legitimately holding the mutex fails the test instead of being waited out. That makes this the second confirmed victim: the first was `main` itself on 2026-07-28 (run 30409535828, noted above), and now it is costing a red check on an unrelated PR.

The change here is test-only (`WithTryLock` → blocking `WithLock`), all checks are green, and it has been approved by both reviewers since 2026-08-24. Happy to rebase it onto the current `upgrade-v0.2.16` head if that helps it land — just say the word.

redstartechno:
Follow-up on the run above, since it turned into a clean flake proof: I force-pushed #1787 with an amend that changed only the commit timestamp — identical tree, identical patch — and `Build and Test API Wrapper` came back green (run 35255663572). Same code, red then green, so the failure is the non-blocking `TryLock` assertion racing a broker worker, not anything in that PR.

redstartechno:
Closing as superseded: `44be40bec` on `upgrade-v0.2.16` ("fix test", 2026-09-22) replaces the non-blocking `TryLock` assertion in `mlnodeclient/mock.go` with a bounded wait, which removes the same spurious `TryLock called more than once` failure this PR targeted. The branch now conflicts with that change, and there is nothing left for it to add. Thanks to everyone who reviewed and approved it.

PR #1496, opened 2026-07-23
fix(devshard): HA: idempotent AppendDiff, stale-standby reconcile, and persist-first

akup:
## Summary

Fixes HA failover onto a **stale standby** `devshardd` that still held an older
in-memory escrow nonce while shared Postgres had already advanced. Catch-up no
longer fails with Postgres unique violations (`23505`) / HTTP 500 for
already-durable identical diffs, and persist failures can no longer leave
memory permanently ahead of the store.

End-state behaviour:

1. **Idempotent `AppendDiff`** — identical `(escrow, nonce)` replay succeeds;
   conflicting bytes → typed fork + `diff_fork_detected` (never overwrite).
2. **Lazy reconcile** — nonce gap → fast-forward durable rows from Postgres,
   then apply the tip (`reconcile_fast_forward`).
3. **Bounded persist retry** — transient PG blips (`diff_persist_retry`).
4. **Persist-first** — validate/preview → persist → commit live SM so a failed
   write leaves memory unchanged.

Design / checklist: [proposals/ha-diff-persist-consistency.md](./proposals/ha-diff-persist-consistency.md),
[proposals/ha-diff-persist-consistency-impl.md](./proposals/ha-diff-persist-consistency-impl.md).

Operator docs (no proposal links in product docs):  
[high-availability-architecture.md](./high-availability-architecture.md),
[release-0.2.14-v4.md](./release-0.2.14-v4.md),
[v4-deploy-test-plan.md](./v4-deploy-test-plan.md) §3.6.

---

## Motivation

Under sticky HA + shared Postgres, replica A can advance durable nonces while
standby B keeps a stale live session. After `proxy_next_upstream` failover,
catch-up on B hit a bare `INSERT` → `23505` even though the durable row was
correct. Separately, apply-then-persist left a memory-ahead hole when
`AppendDiff` failed.

---

## Changes

| Area | Change |
|------|--------|
| **Storage** | Idempotent identical `AppendDiff` (Postgres / SQLite / Memory); fork split; shared `AppendDiffWithRetry` (+ lock/unlock-callback variant so the host releases `h.mu` during backoff) |
| **Host** | Lazy reconcile on nonce gap; persist-first `ValidateDiff` → persist → `CommitValidated` |
| **Gateway** | Persist-first compose (`PreviewLocalBestEffort` → persist → `CommitValidated`); hold session lock during persist retry |
| **State** | `ValidateDiff` / `PreviewLocalBestEffort` trial-apply and return a `ValidatedDiff` handle; `CommitValidated` installs the precomputed post-state (single apply, no recompute) and flushes obs writes buffered during the trial (written once, and never for a diff that fails to persist) |
| **Metrics** | `diff_persist_retry_total`, `diff_fork_detected_total`, `reconcile_fast_forward_total` |
| **testenv** | `TestHAStaleStandbyCatchupIdempotent` (`testenvci`) |
| **Docs** | HA architecture, release notes, deploy test plan §3.6; storage-design wording |

---

## Crash recovery vs ML execution (out of scope here)

Persist-first and durable diff replay cover the **journal / state machine**
only. After a crash, recovery **replays diffs**; it does **not** re-run ML
automatically.

- Still `Pending` in SM, this host is executor, client retries with payload →
  can execute again. Dedup (`executing` / `completedResponses`) is **in-process**,
  not durable.
- Already `Finished` / timed out in later diffs → `signReceipt` will not start
  execution again.

Restoring interrupted inferences across drop / reboot / HA (decouple ML from
the user stream, same-nonce gateway reconnect, durable payload / receipt /
cached body) is **follow-up** work:
[gonka-ai/gonka#1466](https://github.com/gonka-ai/gonka/issues/1466).

### DIFF
--- a/devshard/cmd/devshardd/session/manager.go
+++ b/devshard/cmd/devshardd/session/manager.go
@@ -215,7 +215,6 @@ func (m *HostManager) HandleSettlementFinalized(escrowID string) error {
 	m.sessionsMutex.Lock()
 	srv, hadSession := m.sessions[escrowID]
 	delete(m.sessions, escrowID)
-	delete(m.resolutionFailures, escrowID)
 	m.sessionsMutex.Unlock()
 	if hadSession {
 		srv.Host().Close()
@@ -227,6 +226,9 @@ func (m *HostManager) HandleSettlementFinalized(escrowID string) error {
 		}
 		return err
 	}
+	// Negative-cache so a concurrent/next bind does not recoverStoredSession
+	// the just-settled row (getOrCreate also rejects non-active meta).
+	m.rememberResolutionFailure(escrowID, storage.ErrSessionNotActive, time.Now())
 	return nil
 }
 
@@ -314,7 +316,8 @@ func (m *HostManager) sweepExpiredResolutionFailuresLocked(now time.Time) {
 func isPermanentResolutionFailure(err error) bool {
 	return errors.Is(err, storage.ErrSessionVersionConflict) ||
 		errors.Is(err, storage.ErrSessionEpochConflict) ||
-		errors.Is(err, storage.ErrEpochPruned)
+		errors.Is(err, storage.ErrEpochPruned) ||
+		errors.Is(err, storage.ErrSessionNotActive)
 }
 
 func (m *HostManager) storeSessionIfAbsent(escrowID string, srv *transport.Server) *transport.Server {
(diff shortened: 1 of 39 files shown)


a-kuprin:
@Ryanchen911 @x0152 @0xMayoor @bonujel
Please review. It is required part of v4 release.

Discussed at https://github.com/gonka-ai/gonka/pull/1482#discussion_r3637781883

x0152 (approved):
fixes the problem as described, nothing blocking

Ryanchen911:
Looks good to me, thanks for the work!

bonujel:
Went through the diff-consistency part — looks solid. persist-first closes the memory-ahead-of-PG window and the reconcile path covers the stale-standby catch-up. Nice touch adding the `proto.Equal` identity check on top of `ON CONFLICT DO NOTHING` — plain DO NOTHING would swallow a same-nonce-different-payload write, this way identical replay passes but a real mismatch raises `ErrDiffFork`.

One small nit, more future-proofing: `CommitValidated`'s guard is nonce-only, which works because everything that advances state bumps the nonce — except `InjectWarmKeys`. All its call sites hold `h.mu` or run single-threaded at startup today, so no window, but might be worth a comment/race test pinning that so a future off-`h.mu` WarmKeys path doesn't reopen it.

PR #1499, opened 2026-07-25
feat(inference): gradually prune InferenceValidationDetails

Ryanchen911:
## Summary
Adds gradual per-block pruning for approximately 4 GiB of historical `InferenceValidationDetails` measured on mainnet. This is Follow-up A from #1287 and targets v0.2.16.

## Implementation
- add `inference_validation_details_pruned_epoch = 11` to `PruningState`
- run a dedicated `InferenceValidationDetails` pruner from EndBlock
- retain five epochs and delete at most 1,000 validation-detail entries per block
- preserve current/previous epoch data required by reward claims and validation-parameter queries

## Dependency
This branch is stacked on #1773, which fixes the generic pruner's cross-epoch budget, marker advancement, active-inference retry, and non-positive-limit behavior. Merge #1773 first.

## Tests
- pruning threshold boundary
- per-block maximum and marker advancement
- repeated-call idempotency
- generic multi-epoch backlog and failure regression coverage from #1773
- `go test ./x/inference/keeper ./x/inference/types ./app/upgrades/...`

The obsolete pre-devshard developer-stat state is handled separately by #1793, stacked on this PR.

Refs #1223 and #1287.


bonujel:
Following up on Copilot's inline note on `pruning.go:20` — I reproduced it against this branch (`f518107`), and it has a second consequence that I think turns it from a perf concern into a correctness one.

**TL;DR:** on a multi-epoch backlog the generic `Pruner` (a) deletes far more than `PruningMax` per block, and (b) advances the progress marker past epochs it hasn't finished draining, so the leftovers are never reclaimed. Neither is introduced by this PR, but this PR is the first consumer that runs entirely on that path.

### Reproduction

Mainnet-shaped backlog: every eligible epoch is non-empty, only the per-epoch counts vary. No artificially empty epochs. Uses this PR's own test helpers, so it drops straight into `pruning_test.go`:

```go
// Mainnet-shaped backlog: all eligible epochs are non-empty, counts just vary.
func TestInferenceValidationDetailsPruningBacklog(t *testing.T) {
	k, ctx := keepertest.InferenceKeeper(t)
	require.NoError(t, k.PruningState.Set(ctx, types.PruningState{}))

	counts := map[uint64]int{1: 5000, 2: 300, 3: 4000, 4: 200, 5: 3000, 6: 100}
	seeded := 0
	for e := uint64(1); e <= 6; e++ {
		setValidationDetails(ctx, k, e, counts[e])
		seeded += counts[e]
	}

	// current epoch 11, threshold 5 -> endEpoch = 6, so epochs 1..6 are all eligible.
	for block := 1; block <= 10; block++ {
		before := 0
		for e := uint64(1); e <= 6; e++ {
			before += countValidationDetails(ctx, k, e)
		}
		require.NoError(t, k.Prune(ctx, 11))
		after := 0
		for e := uint64(1); e <= 6; e++ {
			after += countValidationDetails(ctx, k, e)
		}
		st, _ := k.PruningState.Get(ctx)
		t.Logf("block %2d: deleted %5d, %5d remain, marker=%d",
			block, before-after, after, st.InferenceValidationDetailsPrunedEpoch)
	}

	left := 0
	for e := uint64(1); e <= 6; e++ {
		n := countValidationDetails(ctx, k, e)
		t.Logf("  epoch %d: %4d/%d left", e, n, counts[e])
		left += n
	}
	st, _ := k.PruningState.Get(ctx)
	require.Equal(t, 0, left, "%d/%d entries never reclaimed; marker parked at %d",
		left, seeded, st.InferenceValidationDetailsPrunedEpoch)
}
```

Result:

```
block  1: deleted  3600,  9000 remain, marker=0
block  2: deleted  3000,  6000 remain, marker=6
block  3: deleted     0,  6000 remain, marker=6
...
block 10: deleted     0,  6000 remain, marker=6

  epoch 1: 3000/5000 left
  epoch 2:    0/300 left
  epoch 3: 2000/4000 left
  epoch 4:    0/200 left
  epoch 5: 1000/3000 left
  epoch 6:    0/100 left

6000/12600 entries never reclaimed; marker parked at 6
```

(blocks 4–9 elided above; they are identical to block 3.)

**1. Block 1 deletes 3600 with `PruningMax = 1000`** — Copilot's point. The pruner's own log for that single block:

```
Starting pruning start_epoch=1 end_epoch=6 threshold=5 list=inference_validation_details
Items pruned for epoch epoch=1 pruned=1000
Items pruned for epoch epoch=2 pruned=300
Items pruned for epoch epoch=3 pruned=1000
Items pruned for epoch epoch=4 pruned=200
Items pruned for epoch epoch=5 pruned=1000
Items pruned for epoch epoch=6 pruned=100
```

(timestamps and the repeated `list=`/`subsystem=` fields trimmed.)

`prunedCount` is declared at `pruning.go:436` and never incremented, so `prunesLeft` is the full `PruningMax` for every epoch in the range.

**2. By block 2 the pruner considers itself finished with 48% of the backlog still in the store.** `Prune` advances the marker whenever *any* epoch in the range comes back empty (`pruning.go:468`), without checking whether earlier epochs are drained. The small epochs empty out first, come back `0` on the next block, and pull the marker past the big epochs that still hold data. Once `startEpoch > endEpoch` those leftovers are unreachable forever.

Worth stressing that (2) needs no zero-traffic epoch — any variance in per-epoch entry counts is enough, and mainnet certainly has that.

### Why the existing pruners never hit this

In steady state the marker keeps up, so `startEpoch == endEpoch` and the range holds exactly one epoch. With one epoch, "per-epoch budget" and "per-block budget" coincide, and there is no later epoch for the marker to jump to. Both defects only surface during multi-epoch backlog catch-up — which is exactly and only what this PR does, starting from marker `0`. (It's also what `initNewPruningState` in the v0.2.12 upgrade sidesteps, by seeding the new markers forward instead.)

So: not your code, but the PR's stated goal — bleeding off the ~4 GiB gradually — isn't reached as-is. Block 1 is a bulk delete, and most of the 4 GiB stays.

### Fix

Copilot's suggestion fixes both, which is worth spelling out: accumulate `prunedForEpoch` into `prunedCount` and stop iterating once `prunedCount >= PruningMax`. The only way an epoch is left partially drained is by hitting `prunesLeft`, and that is precisely when the budget is exhausted and the loop breaks — so the marker can never step over an unfinished epoch. One change, both symptoms.

One consequence to be aware of: with that in place the drain rate becomes a strict `PruningMax` per block, so whether `1000` clears the backlog in acceptable time becomes a real question — it may want a larger value for this pruner.

Since that change alters EndBlock behaviour for all ten pruners on a release branch, it probably shouldn't ride along in here. Happy to put up a small prerequisite PR (the accumulate/break fix plus regression tests with multiple eligible epochs) for this to rebase onto — just say the word. It likely wants a nod from whoever owns the pruning plan.

### Smaller notes

- `TestInferenceValidationDetailsPruningMaxLimit` can't catch (1): with `current=6, threshold=5` the range is the single epoch 1, so it asserts the per-epoch cap. The PR description calls it a "per-block max limit".
- The safety argument checks out — I traced every keeper reader. `ClaimRewards` is gated by `(currentEpochIndex - 1) != msg.EpochIndex` (`msg_server_claim_rewards.go:186`) and `GetInferenceValidationParameters` only reads the current and previous epoch group. Nothing in genesis. With threshold 5 the oldest epoch kept is `current-4`, while the newest epoch `ClaimRewards` will accept is `current-1`, so there is margin to spare.
- Field number `11` is safe: `reserved "subnet_pruned_epoch"` reserves the name only, no number range, and the regenerated descriptor puts the new field ahead of the `reserved_name` entry correctly.
- Worth a line somewhere that not seeding the new marker is deliberate here (the opposite of `initNewPruningState`), otherwise the next reader will file it as an oversight.

Ryanchen911:
Retargeted from `upgrade-v0.2.15` to `upgrade-v0.2.16`. Mainnet applied 0.2.15 at height 5316315, so the old base is no longer the shipping branch.

Rebased with `--onto` rather than switching the base in the UI: the branch was cut from 0.2.15, so a bare retarget would have dragged that history into the diff (172 files). As rebased it's unchanged at **5 files, +314/-76**, one commit.

Verified on `upgrade-v0.2.16`:

- Field number `11` on `PruningState` is still free on 0.2.16 — the message ends at `claim_recipients_pruned_epoch = 10`, and nothing was added there since.
- `go build ./...` clean in `inference-chain`
- `go test ./app/upgrades/...` passes for every version, `v0_2_10` through `v0_2_16`
- `go test ./x/inference/keeper/...` passes in full

**Still blocked on the generic `Pruner` defect, not on the base.** @bonujel reproduced it above: `prunedCount` is never accumulated (`pruning.go:436`), so the per-block cap is really per-epoch, and the progress marker can advance past epochs that aren't drained — stranding those entries permanently. That means this PR's stated goal of bleeding off the ~4 GiB gradually isn't reached as written: the first block is a bulk delete and roughly half the backlog is never reclaimed.

The fix is small (accumulate and break on budget exhaustion) but it changes EndBlock behaviour for all ten pruners, so it wants a nod from whoever owns the pruning plan before it goes up as a prerequisite PR. @gmorgachev you offered to validate against mainnet data — if the approach looks right to you I'll put that up and rebase this onto it. Worth noting the drain rate then becomes a strict `PruningMax` per block, so `1000` may want raising for this pruner specifically.

PR #1501, opened 2026-07-26
feat(mlnode): metrics exporter — GET /api/v1/metrics, schema v1

baychak:
Phase 1 of the ML-node monitoring plan: an env-gated Prometheus exporter on the mlnode API, so the dapi federation endpoint added in #1469 has something to scrape.

Additive: a new `api/metrics/` package, a `gpu/manager.py` addition, and a two-line router registration. No change to inference or PoC paths.

## What it exposes — `GET /api/v1/metrics`

| Source | Content |
|---|---|
| vLLM replicas | 17 allowlisted families, `replica` label, `*_created` noise stripped |
| NVML (pynvml) | 12 GPU fields with per-sensor gating — temp (incl. HBM), power, clocks, ECC DBE, PCIe replay |
| Host | CPU/memory via cgroup v2 with v1 fallback, HF-cache free space |
| XID | NVML event API (works unprivileged; dmesg does not) |
| Meta | schema/version info, per-source scrape timestamps |

Contract points:
- `GONKA_METRICS=full|off`, default `full`; `off` returns 404 so a node is indistinguishable from a pre-metrics image — which is what makes the dapi-first rollout order safe
- allowlist enforced in code, not config (schema v1)
- `model_name` normalised to the served name: live vLLM leaks the local HF-cache path in that label, and a filesystem path must never leave the node
- no placement labels (hostname/IP/ports/paths) — enforced by a unit test
- vLLM replicas started with `VLLM_NO_USAGE_STATS=1`

## Validation

- 181 unit tests green (16 new)
- Overhead **0.205%** of a core, measured differentially: 10 min baseline vs 10 min with 10 s scrapes; endpoint latency 41 ms
- NVML matrix complete (5/5): bare-metal privileged 4xA100-SXM4, docker without `--privileged`, rented container (RTX 3090), GDDR (no HBM/ECC series), HBM with ECC on (DBE series present) and with ECC disabled (series correctly absent). XID event API worked in all five; dmesg only on the bare host — hence the event-API choice
- Live on **14 production mlnodes** of network-000 (12x MiniMax-M2.7, 2x Kimi-K2.6, two different image versions), federated publicly. Full-fleet output: 2700 series, 308 KB — well inside the per-node ceilings in #1469

## Deploy note learned the hard way

Fork images carry model-specific forced args inside their patched `runner.py`/`proxy.py`. Never replace the mlnode source tree wholesale — overlay additively. Two resilience fixes here (guarded config-info collection, tolerated missing proxy generation counter) came out of exactly that.

## Rollback

`GONKA_METRICS=off` on the node, no rebuild.

Rebased onto v0.2.14; `mlnode/packages/api` is byte-identical between that and this branch base, so the diff is only the exporter.

### DIFF
--- a/docs/adr/mlnode-monitoring-metrics-v1.md
+++ b/docs/adr/mlnode-monitoring-metrics-v1.md
@@ -1,0 +1,176 @@
+# ADR: ML-node monitoring — metrics and interface contract v1
+
+Status: **accepted** (team-approved 2026-07-10)
+Related: [operator guide](../observability.md), [decision log](../decision-log.md)
+
+## Context
+
+ML-node monitoring architecture (v0.5): an exporter inside mlnode → a
+collector + public endpoint in dapi → consumers store data on their side.
+This ADR fixes the Phase 0 contract: the final allowlist, formats,
+off-behavior, numeric limits and the schema versioning rules.
+
+## Decisions
+
+### D1. Allowlist v1 (verified against live vLLM 0.20.0 and v0.23.0 sources)
+
+Verification performed 2026-07-08:
+
+- live `/metrics` of vLLM **0.20.0** (production ML node, Kimi-K2.6): every
+  family below present;
+- vLLM sources **v0.20.0 vs v0.23.0** (`vllm/v1/metrics/*.py`): metric name
+  sets identical (42/42), label names identical — no version skew for the
+  allowlist. A live 0.23 run is deferred to the Phase 4 E2E stand (Open-1).
+
+**vLLM P0** (labels: `model_name`, `replica`): `vllm:num_requests_waiting`,
+`vllm:num_requests_running`, `vllm:kv_cache_usage_perc`,
+`vllm:num_preemptions_total`, `vllm:time_to_first_token_seconds`,
+`vllm:request_queue_time_seconds`, `vllm:prompt_tokens_total`,
+`vllm:generation_tokens_total`, `vllm:request_success_total`
+(`finished_reason`: stop|length|abort|error|repetition — captured live).
+
+**vLLM P1**: `vllm:inter_token_latency_seconds`,
+`vllm:e2e_request_latency_seconds`, `vllm:prefix_cache_queries_total`,
+`vllm:prefix_cache_hits_total`, `vllm:request_prompt_tokens`,
+`vllm:request_generation_tokens`, `vllm:iteration_tokens_total`,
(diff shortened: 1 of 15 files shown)


baychak:
Thanks — six of the seven were right and are fixed in f7e522b8. One is a false positive; details below so the record is accurate.

## Fixed

The guide and the ADR described the design as proposed, while the federation side landed differently. Following either would have sent an operator to a URL that does not exist and had them set environment variables nothing reads. Corrected against the merged implementation:

- endpoint is `/v1/mlnodes/metrics` (exact-match nginx locations), not `/v1/metrics`
- kill switch is `DAPI_API__MLNODE_METRICS_DISABLED=true` — one switch, evaluated per request; the pair `DAPI_API__METRICS_COLLECTOR_ENABLED` / `DAPI_API__METRICS_ENDPOINT_ENABLED` does not exist
- rate limiting is the proxy's nginx `metrics_zone`, tunable via `METRICS_RATE_LIMIT_RPM` / `METRICS_BURST`, not Echo middleware

I also corrected two things the review did not flag, found while checking the above:

- the guide claimed a 45 s background poller with a ≤5 min buffer and per-sample timestamps. As built it is a pull-through cache: 10 s TTL, single-flighted rebuilds. Staleness therefore surfaces through `mlnode_source_scrape_timestamp_seconds` and `mlnode_up`, not sample timestamps.
- the three-state node semantics were undocumented, including that a 404 node is absent from the output entirely — the property that makes a dapi-first rollout safe.

Both ADR deviations are recorded as as-built notes rather than rewritten silently, so the reasoning stays visible.

Note on the numbers: the review quoted 10 r/m burst 30, which were the defaults in #1469. `upgrade-v0.2.15` now ships 6 r/m burst 5, so the docs point at the environment variables rather than repeating values that keep moving.

## Not a defect

> `test_metrics_exporter.py:254` … extra closing brace (`}}`) … makes the input invalid Prometheus text format

That line is inside an f-string, where `}}` is the escape for a single `}`. The fixture renders as:

```
vllm:request_success_total{engine="0",finished_reason="st\"op",model_name="…"} 1.0
```

One closing brace, valid exposition. Verified by rendering the string. Left unchanged.

baychak:
Superseded by #1531, which carries this change cherry-picked onto `main` together with its counterpart and the RC image pin — collected there at @vbgd0's request so the release branch gets everything in one place. Review history stays here.

PR #1505, opened 2026-07-27
feat(chain): fall back to CometBFT RPC when chain gRPC is unreachable

gmorgachev:
edge-api and devshardd failed every chain query against a node that ran with grpc disabled or :9090 unpublished. Direct gRPC stays primary for performance, but queries now switch to CometBFT RPC/ABCI on transport failure and retry the same request there, so callers see no error.

Recovery is lazy instead of restart-driven: after 30 minutes on RPC, one request probes gRPC, and any answer (including an application error) makes it primary again. A node that enables gRPC later is picked up without restarting the services around it.

### DIFF
--- a/common/chain/client.go
+++ b/common/chain/client.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/cosmos/cosmos-sdk/client/grpc/cmtservice"
 	blstypes "github.com/productscience/inference/x/bls/types"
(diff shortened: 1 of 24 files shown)


akup:
I’ve added follow-up PR to close some gaps on this fallback logic:
1) Gateway also uses connection with fallback
2) Aborted probes no longer count as proof of health
3) RPC host is not required (derives from GRPC_HOST if absent, instead of failing to boot)
4) Errors from fallback was silent
5) Added observability for follback path (primary path had observability)
6) QueryConn() rejects broadcasts

https://github.com/gonka-ai/gonka/pull/1507

DimaOrekhovPS (approved):
Copilot comments can be disregarded

akup (approved):

PR #1507, opened 2026-07-27
feat(chain): polishing fall back to CometBFT RPC

akup:
It is follow-ups for https://github.com/gonka-ai/gonka/pull/1505

There are following fixes, that makes it more complete:

1. Gateway also uses connection with fallback

2. Aborted probes no longer count as proof of health
According to Copilot finding: https://github.com/gonka-ai/gonka/pull/1505#discussion_r3655632131

3. RPC host is not required
RPC host environment variable is not required and dervide from gRPC, earlier it failed to boot without RPC host env var.

4. Errors from fallback now are returned

5. Added observability to fallback connection, as primary connection was observable

6. `QueryConn()` rejects broadcasts

### Aborted probes no longer count as proof of health
A probe that was canceled or hit its deadline was being read as "gRPC answered", flapping every query back onto a still-dead endpoint. Probe outcomes are now classified into three cases — answered, transport-down, and no-verdict. A no-verdict probe changes nothing and, crucially, does not consume the probe window, so the next request retries immediately instead of waiting another 30 minutes. Without that, a client that kept dropping connections could pin the service to RPC indefinitely.

### One endpoint policy for every service
`chain.NewWithQueryFallback` derives the RPC endpoint from the gRPC host (`:26657`) when none is configured and degrades to gRPC-only when it cannot, rather than refusing to boot — a fallback meant to keep queries working must not become the reason a deployment stops starting. edge-api, devshardd and the devshardctl gateway all go through it, so adoption is uniform instead of per-service. `edge-api.loadConfig` returns an error instead of calling `os.Exit`, which also makes it testable.
When a query fails on **both** transports the returned error names both causes, so a fully unreachable node no longer looks like a plain ABCI failure.

### The degraded path is observable
Previously only the direct conn was wrapped in an `ObservedConn`, so every `chain.grpc.query` span and status code disappeared the moment a service went degraded — you went blind exactly when you needed the data. Both conns are now
instrumented and distinguishable:
| Signal | On gRPC | On CometBFT RPC |
| --- | --- | --- |
| `chain.grpc.query` spans | `chain.transport=grpc` | `chain.transport=comet-rpc` |
| `chain.query.transport.active` gauge | `grpc`=1, `comet-rpc`=0 | `grpc`=0, `comet-rpc`=1 |
| Logs | — | one warning on the switch, then one per failed 30-min probe |

The gauge is the thing to alert on, and it reports from process start so an absent `comet-rpc` series means healthy rather than unreported. It uses the global OTel meter, so it is a no-op until a process installs a meter provider. The previously silent `probeFailed` path now logs, so a service that has been on RPC for a week is no longer indistinguishable from one that never fell back.

### `QueryConn()` rejects broadcasts
Cosmos's `client.Context.Invoke` special-cases `*tx.BroadcastTxRequest` and broadcasts it over CometBFT rather than querying. Since the fallback retries a failed request on the other transport, a `txtypes.NewServiceClient(c.QueryConn())`
would have been able to submit the same signed transaction twice. `BroadcastTx` is now rejected with `InvalidArgument` on the fallback conn — from the first call, not just once gRPC drops, since a broadcast that works until the day of an outage is the worst version of this bug. Tx-service *queries* such as `GetTx` still route normally. No caller does this today; the guard closes it before one does.

### DIFF
--- a/common/chain/client.go
+++ b/common/chain/client.go
@@ -178,6 +178,10 @@ func (c *Client) Conn() grpc.ClientConnInterface { return c.conn }
 // QueryConn returns the connection queries run on. It equals Conn() unless the
 // client was built with NewWithRPCFallback. Use it for query clients this
 // package does not expose directly.
+//
+// Queries only. Building a tx service client on this conn and broadcasting
+// through it is rejected rather than silently retried across transports; use
+// Conn() for transactions.
 func (c *Client) QueryConn() grpc.ClientConnInterface { return c.queryConn }
 
 // InferenceQueryClient returns a query client for the inference module.
--- a/common/chain/client.go
+++ b/common/chain/client.go
@@ -179,9 +179,9 @@ func (c *Client) Conn() grpc.ClientConnInterface { return c.conn }
 // client was built with NewWithRPCFallback. Use it for query clients this
 // package does not expose directly.
 //
-// Queries only. Building a tx service client on this conn and broadcasting
-// through it is rejected rather than silently retried across transports; use
-// Conn() for transactions.
+// Queries only. When this is a fallback conn, broadcasting a transaction
+// through it is rejected rather than silently retried across transports. Use
+// Conn() for transactions either way.
 func (c *Client) QueryConn() grpc.ClientConnInterface { return c.queryConn }
 
 // InferenceQueryClient returns a query client for the inference module.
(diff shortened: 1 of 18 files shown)


DimaOrekhovPS (approved):
Looks good, copilot comments can be disregarded

PR #1509, opened 2026-07-27
feat(devshard): serve versionless obs from dapi/edge-api

akup:
## Summary
- Move versionless public observability (`/devshard/sessions|stats|metrics`) out of **versiond** into shared **`common/devshardobs`**, dual-served on **dapi** and **edge-api**.
- Keep **versiond** as a pure versioned reverse proxy: `/devshard/{version}/…` is served as-is (no strip-rewrites), so clients can pin a previous stable version.
- Join proxy steers versionless obs to **edge-api** when `EDGE_API_SERVICE_NAME` is set (including `edge-api-router`), otherwise **dapi**.
- Versionless routing: PG escrow→version lookup when available, else fan-out newest→oldest over running versions from versiond `/healthz`; `/stats/shards` list aggregates across versions; `/metrics` pins primary.

## Test plan
- [x] `go test ./devshardobs/` in `common`
- [x] `go test ./testenv/citest/ -run TestProxyObsRouting` (EDGE_API empty → dapi; set → edge-api; `edge-api-router` upstream)
- [x] Host bind/obs: `TestObsGET_DoesNotBindSession`, `TestOwnerChat_BindsSession`, `TestStatsShardDetailSkipsForeignVersionSession`
- [ ] Manual: versioned `/devshard/{v}/stats/shards` reaches that child; versionless `/devshard/stats/shards` hits dapi or edge-api per env
- [ ] Manual (multi): with `EDGE_API_SERVICE_NAME=edge-api-router`, versionless obs upstream is the router


common/devshardobs/versions.go
```diff
@@ -0,0 +1,113 @@
...
+		return out, h.cacheErr
+	}
+
+	versions, err := h.fetch(ctx)
+	h.cached = versions
+	h.cachedAt = time.Now()
+	h.cacheErr = err
+	out := append([]string(nil), versions...)
+	return out, err
+}
```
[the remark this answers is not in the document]
akup: Fixed here:
https://github.com/gonka-ai/gonka/pull/1509/commits/7f7ebb81c286457ad224e04b7f3474b649dee00b

proxy/entrypoint.sh
```diff
@@ -617,37 +617,44 @@ else
...
-# Public obs (versionless + rewritten legacy) uses a tighter rate-limit zone.
-# Protocol POSTs and payloads stay on /devshard/{version}/... via the exempt
-# catch-all below (chat/SSE need high limits).
+# /devshard/ routing:
+#   /devshard/{version}/…     → versiond as-is (protocol + versioned obs)
+#   /devshard/healthz         → versiond supervisor
+# Versionless public obs lives under /v1/devshard/… (see API_VERSION_LOCATIONS):
+#   sessions|stats/shards|metrics → edge-api if EDGE_API_SERVICE_NAME set, else dapi
+#   other /v1/devshard/*          → rewrite → /devshard/v1/* → versiond
+# Protocol POSTs/payloads stay on the exempt catch-all (chat/SSE need high limits).
```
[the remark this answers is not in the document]
akup: Fixed here:
https://github.com/gonka-ai/gonka/pull/1509/commits/0a6d3520db8b0499faa3892905c227e8984ff043

edge-api/cmd/edge-api/main.go
```diff
@@ -54,13 +45,22 @@ func main() {
 		_ = shutdownObs(ctx)
 	}()
 
-	chainClient, err := chain.NewWithQueryFallback(cfg.ChainGRPCURL, cfg.ChainRPCURL)
+	chainClient, err := chain.New(cfg.ChainGRPCURL)
```
DimaOrekhovPS: Was it intentional to remove the fallback here?

GLiberman:
There will be def-ly issues right now with default compose settings - which I believe should be addressed:

1. [`postgresConfigured()` (line 71)](/gonka/common/devshardobs/lookup.go:71) treats PGDATABASE alone as a configured lookup. The default join compose always sets PGDATABASE=payloads while leaving PGHOST empty ([compose (line 79)](gonka/deploy/join/docker-compose.yml:79)). The lookup therefore attempts a Postgres ping, fails, and [`OpenFromEnv` (line 34)](/gonka/common/devshardobs/open.go:34) returns without constructing the fan-out handler.
Dapi then skips mounting observability ([main.go (line 272)](/gonka/decentralized-api/main.go:272)), while nginx defaults to routing these requests to dapi. The result is the existing 410 Gone deprecation catch-all ([server.go (line 189)](/gonka/decentralized-api/internal/server/public/server.go:189)) for every advertised versionless endpoint.
Lookup initialization failure must degrade to a nil lookup plus fan-out, as versiond did previously. Also, dapi’s payload database and devshard’s session database need separate configuration.

2. Handler defaults to http://versiond:8080 ([handler.go (line 169)](/gonka/common/devshardobs/handler.go:169)), but the multi-versiond overlay only points the outer proxy at versiond-router ([docker-compose.versiond.yml (line 94)](/gonka/deploy/join/docker-compose.versiond.yml:94)). It does not set DEVSHARD_VERSIOND_URL on dapi or edge-api.
Consequently, versioned traffic can fail over through versiond-router, while versionless observability remains pinned to the first versiond container and fails when it is down. The multi-edge overlay also supplies neither this URL nor the devshard session-database credentials.

edge-api/cmd/edge-api/main.go
akup: No definetely not. It is regression

akup: https://github.com/gonka-ai/gonka/pull/1509/commits/629ad707d9e57f4d707de7d75431c90b0afc393f
Regression from [1505](https://github.com/gonka-ai/gonka/pull/1505) fixed

PR #1514, opened 2026-07-28
fix(devshard): block inference admission during gateway finalize

redstartechno:

A TOCTOU race in the devshardctl gateway between `/v1/finalize` and inference admission: finalize checks `escrowHasBackgroundWork()` once, but `rt.active` stays `true` for the whole finalize call (it only flips in `markDevshardInactiveAfterFinalize`, after a 2xx), and `acceptsNewInferences` checks only `rt.active`. A chat completion can therefore be admitted and executed concurrently with an escrow finalize on the same runtime/session.

- The drain-barrier contract documented at the `activeUserRequests`/`pendingRaceCleanup` field comment in `gateway.go` says settle and store-close must wait until background work is quiesced. The finalize HTTP path enforces that with a one-shot check that nothing prevents from going stale: `beginFinalizeGate` did not exist, and admission (`reserveRuntimeForModel` → `acceptsNewInferences` → `reserveRuntimeLocked`, all under `g.mu`) has no visibility into an in-flight finalize.
- Reproduced in-process before the fix: with a finalize parked mid-flight inside `rt.handler.ServeHTTP`, `acceptsNewInferences` returned `ok=true` and pooled admission handed out the runtime (probe output below, captured on unpatched `gateway.go`).
- Two adjacent gaps compound it: the single-runtime finalize route (`handleSingleOnly`) had **no** quiescence check at all, and the direct `/devshard/{id}` chat path checked `acceptsNewInferences` early but then reserved unconditionally ~50 lines later (after parsing and the rate limiter), so an admission decision could also go stale against a finalize that started in between.

Adds a per-runtime `finalizing` gate (`atomic.Bool`) set atomically with the quiescence check inside one `g.mu` critical section (`beginFinalizeGate`). Since admission's gate-check + counter-increment also runs entirely under `g.mu`, the two decisive actions are now serialized: either admission wins and finalize sees active requests (409), or finalize wins and admission is refused with reason `finalize_in_flight`. The gate:

- refuses a second finalize while one is in flight (409 `already has a finalize in flight`) instead of queueing it on `finalizeMu` — a queued retry would otherwise run after a failed first finalize had already cleared the flag, reopening the race;
- is cleared via an unconditional `defer` on failure or panic so the runtime stays usable (on the `/devshard/{id}` route a successful finalize deactivates the runtime before the clear runs, so there is no `active && !finalizing` window);
- is applied to both finalize routes, and the direct chat path now re-checks admission and reserves in a single `g.mu` section (`reserveRuntimeIfAccepting`), mirroring the pooled path.

- **Deliberate behavior change:** single-runtime finalize can now return 409 while requests or race cleanups are draining. Previously it had no quiescence check. This matches the `/devshard/{id}` route's existing contract (same status, same body shape). Because e2e helpers finalize immediately after a completion — while a loser-side race cleanup may still be draining — `e2e/testutil.FinalizeSession` gained a bounded retry (409-only, 100ms interval, 2s budget) that documents the new contract; non-409 behavior is unchanged.
- A second concurrent finalize is now refused rather than serialized behind `finalizeMu`. Callers that relied on fire-and-forget double-submits will see a 409 on the loser; retrying after the winner completes behaves as before.
- The new admission-skip reason is `finalize_in_flight` (not `finalizing`) so metrics don't share a bucket with the session-phase reason `phase=finalizing`.
- `settleDevshardOnChain` is untouched: it already closes admission by flipping `rt.active` inside the same `g.mu` section as its own quiescence check, which is equivalent protection, and its finalize still serializes via `finalizeMu`.

Fail-first probe on unpatched `gateway.go` (new-symbol-free, deleted after capture), with a finalize parked mid-flight:

```
=== RUN   TestFailFirstFinalizeInFlightBlocksNewInferences
    zz_failfirst_test.go:41: acceptsNewInferences during in-flight finalize: ok=true reason=""
    zz_failfirst_test.go:44:
        	Error:      	An error is expected but got nil.
        	Messages:   	pooled admission must not pick a runtime whose finalize is in flight
--- FAIL: TestFailFirstFinalizeInFlightBlocksNewInferences (0.03s)
FAIL	devshard/cmd/devshardctl	0.112s
```

With the fix, seven new regression tests pin the behavior (admission refused during in-flight finalize on both routes, gate reopens after failed/panicking finalize, concurrent second finalize refused, failed finalize allows retry, single-route drain 409). All are channel-synchronized, no sleeps.

- `devshard/cmd/devshardctl/gateway.go` (+67/−3)
- `devshard/cmd/devshardctl/gateway_finalize_admission_test.go` (new, 238 lines)
- `devshard/e2e/testutil/requests.go` (+25/−1)

## Testing & evidence

- Windows native (`CGO_ENABLED=0`, go1.26.4): finalize suite 10/10 pass; full `./cmd/devshardctl/` shows only the three pre-existing Windows-path failures (`TestResolveAdminStoragePath`, `TestAdminImportDevshardRejectsAbsoluteStoragePath`, `TestResolveBaseStorageDirNormalizesLegacyStateDBPath`), verified identical on pristine main.
- Linux (WSL2): `go test ./cmd/devshardctl/ -run 'Finalize' -race -count=5` → ok; full package with `-race` → ok, no data-race warnings.
- `go vet ./cmd/devshardctl/ ./e2e/...` clean; new files gofmt-clean (pre-existing gofmt drift elsewhere in `gateway.go` left untouched).
- Docker-bound e2e suites were **not** run (no Docker on this machine); the `e2e/testutil` change is compile- and vet-verified only.

### Before

```
--- FAIL: TestFailFirstFinalizeInFlightBlocksNewInferences (0.03s)
    zz_failfirst_test.go:41: acceptsNewInferences during in-flight finalize: ok=true reason=""
FAIL	devshard/cmd/devshardctl	0.112s
```

### After

```
--- PASS: TestFinalizeInFlightBlocksNewInferences (0.03s)
--- PASS: TestSingleOnlyFinalizeInFlightBlocksNewInferences (0.00s)
--- PASS: TestSingleOnlyFinalizeRequiresNoActiveRequests (0.00s)
--- PASS: TestFailedFinalizeReopensAdmission (0.00s)
--- PASS: TestPanickingFinalizeReopensAdmission (0.00s)
--- PASS: TestConcurrentFinalizeIsRefused (0.00s)
--- PASS: TestFailedFinalizeAllowsRetry (0.00s)
--- PASS: TestGatewayHandleDevshardFinalizeRequiresNoActiveRequests (0.03s)
PASS
ok  	devshard/cmd/devshardctl	0.143s

go test ./cmd/devshardctl/ -race -count=1   (WSL)
ok  	devshard/cmd/devshardctl	44.085s
```

### DIFF
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -97,6 +97,12 @@ type devshardRuntime struct {
 	// pendingRaceCleanup counts background race cleanups (refund + loser-signature persistence) still in flight
 	pendingRaceCleanup atomic.Int64
 
+	// finalizing marks a runtime whose finalize request is in flight. It is set
+	// in the same g.mu critical section that observed the drain barrier as
+	// quiet, and admission reads it under that same lock, so no new inference
+	// can slip in between the quiescence check and the finalize handler run.
+	finalizing atomic.Bool
+
 	// settlementPending marks an escrow that has been deactivated and must
 	// be settled once its in-flight requests drain. settlementReason is
 	// written before the flag and read after it in the lock-free drain hook;
@@ -572,6 +578,9 @@ func (rt *devshardRuntime) retireClose(reason string) error {
 func (rt *devshardRuntime) acceptsNewInferences() (bool, string) {
 	if rt == nil || !rt.active.Load() {
 		return false, "inactive"
+	}
+	if rt.finalizing.Load() {
+		return false, "finalize_in_flight"
 	}
 	if rt.proxy == nil || rt.proxy.sm == nil {
 		return true, ""
@@ -1365,6 +1374,11 @@ func (g *Gateway) handleSingleOnly(w http.ResponseWriter, r *http.Request) {
 	g.mu.Unlock()
 	if len(runtimes) == 1 {
 		if r.URL.Path == "/v1/finalize" && r.Method == http.MethodPost {
+			if ok, reason := g.beginFinalizeGate(runtimes[0]); !ok {
+				http.Error(w, fmt.Sprintf(`{"error":{"message":"devshard %s %s"}}`, runtimes[0].id, reason), http.StatusConflict)
+				return
+			}
+			defer g.endFinalizeGate(runtimes[0])
 			g.finalizeMu.Lock()
 			defer g.finalizeMu.Unlock()
 			log.Printf("gateway_finalize_lock_acquired escrow=%s path=%s", runtimes[0].id, r.URL.Path)
@@ -1561,7 +1575,12 @@ func (g *Gateway) handleDevshard(w http.ResponseWriter, r *http.Request) {
 			logRequestStage(ctx, "gateway_devshard_limiter_bypassed_during_poc", "escrow", devshardID, "input_tokens", inputTokens, "reason", currentPoCPhaseReason())
 		}
 
-		g.reserveRuntime(rt, inputTokens)
+		if ok, reason := g.reserveRuntimeIfAccepting(rt, inputTokens); !ok {
+			logRequestStage(ctx, "gateway_devshard_unavailable", "escrow", devshardID, "reason", reason)
+			g.recordGatewayRequestOutcome(limitModel, "runtime_unavailable", runtimeSkipReasonKey(reason))
+			http.Error(w, fmt.Sprintf(`{"error":{"message":"devshard %s is unavailable for new inferences: %s"}}`, devshardID, reason), http.StatusConflict)
+			return
+		}
 		defer g.releaseRuntime(rt, inputTokens)
 		logRequestStage(ctx, "gateway_devshard_runtime_selected", "escrow", devshardID, "input_tokens", inputTokens)
 
@@ -1575,10 +1594,11 @@ func (g *Gateway) handleDevshard(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if innerPath == "/v1/finalize" && r.Method == http.MethodPost {
-		if rt.escrowHasBackgroundWork() {
-			http.Error(w, fmt.Sprintf(`{"error":{"message":"devshard %s has active requests"}}`, devshardID), http.StatusConflict)
+		if ok, reason := g.beginFinalizeGate(rt); !ok {
+			http.Error(w, fmt.Sprintf(`{"error":{"message":"devshard %s %s"}}`, devshardID, reason), http.StatusConflict)
 			return
 		}
+		defer g.endFinalizeGate(rt)
 		g.finalizeMu.Lock()
 		defer g.finalizeMu.Unlock()
 		log.Printf("gateway_finalize_lock_acquired escrow=%s path=%s", devshardID, r.URL.Path)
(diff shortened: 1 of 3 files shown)


redstartechno:
Follow-up on the disclosed testing gap ("Docker e2e not run locally"): I've since set up Docker Engine in WSL2 and ran the full devshard e2e suite (`make -C devshard e2e`) on this branch (`724b49ffa`).

**Result: 21/21 PASS in ~160s** (Linux/WSL2 Ubuntu 22.04, Docker Engine 29.5.3, Compose v5.1.4), including:

- `TestE2E_NonStreamingRejectsAfterFinalize` — the finalize-path behavior this PR touches
- both SQLite and Postgres restart-recovery groups
- all streaming/non-streaming and cache-isolation tests

A baseline run on `main` also passes 21/21, so no delta introduced by this branch. The e2e disclosure in the PR description can be considered closed.

a-kuprin (approved):

redstartechno:
Retargeted this from `gateway-v4` to `devshard-0.2.15-v5` and recut the branch onto it — force-pushed, so the diff is now against the live train rather than a retired branch.

Why: `gateway-v4` is fully contained in v5 (ahead 0, behind 847) and has no open PR to main, so this PR had no path to land where it was.

The defect is still present on v5: there is no `beginFinalizeGate`, no `finalizing` flag and no `reserveRuntimeIfAccepting`, and the check-then-act is still at `gateway.go:1868` — an inference can be admitted between finalize observing the runtime as quiet and the finalize handler running.

On the recut:

- `gateway.go` merged cleanly despite the file growing from ~1.7k to 4.7k lines since this PR was written. The only conflict was in `devshard/e2e/testutil/requests.go`, additive on both sides (v5 added debug/gossip helpers, this PR added the finalize-409 retry constants); both were kept.
- I checked the two `escrowHasBackgroundWork()` call sites that appeared since: `retireRuntimeLocked` (`:4201`) defers retirement to the drain hook, and `deactivateAndSettleDevshardByID` (`:4345`) runs after `active` is already false, which `acceptsNewInferences` refuses on. Neither needs the gate, so the scope of this PR is unchanged.

Verification on the new base, WSL2, `go1.25.9 linux/amd64`:

```
$ go test ./cmd/devshardctl/ -count=1
ok  	devshard/cmd/devshardctl	138.943s

$ go test ./cmd/devshardctl/ -run 'Finalize|Admission' -count=1 -v
--- PASS: TestFinalizeInFlightBlocksNewInferences (0.00s)
--- PASS: TestSingleOnlyFinalizeInFlightBlocksNewInferences (0.00s)
--- PASS: TestSingleOnlyFinalizeRequiresNoActiveRequests (0.00s)
--- PASS: TestFailedFinalizeReopensAdmission (0.00s)
--- PASS: TestPanickingFinalizeReopensAdmission (0.00s)
--- PASS: TestConcurrentFinalizeIsRefused (0.00s)
--- PASS: TestFailedFinalizeAllowsRetry (0.00s)
```

`go vet ./cmd/devshardctl/ ./e2e/testutil/` is clean. The approval predates the recut, so re-review is fair — the content is unchanged from what was approved apart from the conflict resolution above. If you would rather this waited for v6 or wanted it dropped, say so and I will close it.

PR #1515, opened 2026-07-28
fix(deploy): pin join compose to pre-v4 0.2.14 images

DimaOrekhovPS:
restore proxy/versiond/proxy-ssl to 0.2.14 and drop unreleased edge-api from the default stack.

### DIFF
--- a/deploy/join/docker-compose.yml
+++ b/deploy/join/docker-compose.yml
@@ -87,22 +87,9 @@ services:
       - "127.0.0.1:9200:9200"
     restart: always
 
-  edge-api:
-    container_name: edge-api
-    image: ghcr.io/product-science/edge-api:0.2.14-devshard-v4
-    environment:
-      - EDGE_API_PORT=18080
-      - CHAIN_GRPC_URL=node:9090
-      - EDGE_API_OTEL_ENABLED=${EDGE_API_OTEL_ENABLED:-false}
-      - OTEL_ENDPOINT=${OTEL_ENDPOINT:-}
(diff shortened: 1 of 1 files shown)


gmorgachev (approved):

GLiberman (approved):

redstartechno:
Heads-up: this change appears to have broken the `Build and Test edge-api (validation)` job on `main` — the first red `Go Build and Test` run on main is `182d0820c` (this merge): https://github.com/gonka-ai/gonka/actions/runs/30409535828

Cause: the `edge-api` service was removed from `deploy/join/docker-compose.yml`, but `deploy/join/docker-compose.edge-api-multi.yml` still defines `edge-api-router` with `depends_on: edge-api`. `scripts/validate-edge-api.sh` renders that file pair, which now fails:

```
$ docker compose -f deploy/join/docker-compose.yml -f deploy/join/docker-compose.edge-api-multi.yml config --quiet
service "edge-api-router" depends on undefined service "edge-api": invalid compose project
```

(Reproduced locally on a pristine `182d0820c` checkout.)

Since the intent here is a temporary pre-v4 pin, possible fixes are either making the `edge-api-multi` overlay self-contained (define the pinned `edge-api` service there) or skipping that render pair in `validate-edge-api.sh` while the rollback is in effect. Happy to send a small PR for whichever direction you prefer.

(Note: the other failing job in that run, `Build and Test API Wrapper`, is unrelated — it's the known `TestRegularPocScenario` TryLock mock flake, fix pending in #1495.)

PR #1516, opened 2026-07-28
fix(testermint): compile v0.2.14 rehearsal prep

milkstoner:
## Summary

- pass the versioned devshard route prefix when starting the pre-upgrade proxy
- read the settlement protocol version from `stateRootAndProtocolVersion`
- update the embedded patch hunk size for the added lines

## Why

`testermint/upgrade-rehearsal/previous-release-prep.patch` is applied to the v0.2.14 checkout during upgrade rehearsal preparation. The generated test used an outdated `startDevshardProxy` call and referenced `parsed.version`, so the preparation test no longer matched the v0.2.14 testermint API.

## Validation

- `git apply --check` against `release/v0.2.14` — passed
- applied the patch in a detached `release/v0.2.14` worktree — passed
- `git diff --check` on the applied result — passed
- verified the generated test contains `routePrefix = devshardVersionedRoutePrefix()` and `settlement.parsed.stateRootAndProtocolVersion`
- Gradle `testClasses` was not completed locally because the validation host has no JDK; repository CI is expected to provide the compile result

If accepted, please consider this contribution for the Gonka contributor reward program.

### DIFF
--- a/testermint/upgrade-rehearsal/previous-release-prep.patch
+++ b/testermint/upgrade-rehearsal/previous-release-prep.patch
@@ -3,7 +3,7 @@ new file mode 100644
 index 000000000..5a2d131fe
 --- /dev/null
 +++ b/testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt
-@@ -0,0 +1,121 @@
+@@ -0,0 +1,125 @@
 +import com.productscience.*
 +import org.assertj.core.api.Assertions.assertThat
 +import org.junit.jupiter.api.Tag
@@ -49,7 +49,11 @@ index 000000000..5a2d131fe
 +            modelId = defaultModel,
 +        )
 +
-+        val handle = genesis.startDevshardProxy(escrowId = escrowId, keyName = user.keyName)
++        val handle = genesis.startDevshardProxy(
++            escrowId = escrowId,
++            keyName = user.keyName,
++            routePrefix = devshardVersionedRoutePrefix(),
++        )
 +        var devshardRequests = 0
 +        var settlementVersion = ""
 +        try {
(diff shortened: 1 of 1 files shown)


redstartechno:
This patch is now blocking the active `upgrade-v0.2.16` train (#1535).

On the latest rehearsal run the `upgrade-rehearsal` job fails at step *Run old-checkout prep test* with a single Kotlin compile error:

```
e: .../old/testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt:67:51 Unresolved reference 'version'.
> Task :compileTestKotlin FAILED
```

Run: https://github.com/gonka-ai/gonka/actions/runs/32675567180/job/97282935722

Line 67 of the generated test is `settlementVersion = settlement.parsed.version` — exactly the reference this PR changes.

One thing has moved since this PR was opened: the rehearsal now resolves `PREVIOUS_RELEASE` to `release/v0.2.15`, not v0.2.14. The fix applies unchanged there. At `release/v0.2.15`, `DevshardSettlementData` still exposes the property as `stateRootAndProtocolVersion`, with `version` only as the `@SerializedName` JSON key:

```kotlin
// devshardctl finalize emits "version" (v1 wire compat); inferenced reads the same key.
@SerializedName("version")
val stateRootAndProtocolVersion: String,
```

so that rename is the load-bearing part of this PR.

For what it's worth, the other half (`routePrefix = devshardVersionedRoutePrefix()`) does not appear to be required at v0.2.15: `startDevshardProxy` there declares `routePrefix: String? = null` and falls back to `defaultDevshardRoutePrefix()`. The helper does resolve at that tag, so it should still compile — just noting it in case you would rather keep the patch minimal.

Posting the diagnosis only; I have not touched anything here.

PR #1517, opened 2026-07-29
Implement graceful versiond host evacuation and replacement

snevolin:
## Summary

This PR targets `upgrade-v0.2.15` and implements **Track B: graceful
evacuation, replacement, addition, and decommission of a whole `versiond`
host**, as described in:

- [the HA proposal](https://github.com/gonka-ai/gonka/discussions/1367);
- [`rolling-update.md` section 1.8](devshard/docs/rolling-update.md#18-versiond-router-draining-versiond-hosts-ha);
- rollout and acceptance-test items 7-8 in section 1.10.

It also gives `edge-api` the same readiness-first shutdown contract and ships a
safe, topology-aware v4-to-v5 upgrade path for existing join deployments.

The main changes are:

- replace the `versiond` and `edge-api` nginx pool routers with unprivileged
  HAProxy 3.2 images;
- discover replicas through DNS and admit them through active readiness checks;
- preserve stable escrow routing with consistent hashing per protocol version;
- make `SIGTERM` on `versiond` a bounded host drain that preserves accepted
  requests and complete SSE streams;
- make `edge-api` leave rotation before draining accepted chain queries;
- formalize host, child-generation, supervised-process, and `devshardd`
  lifecycle behavior as table-driven state machines;
- enforce shared Postgres storage for HA traffic at startup and request time;
- add a topology-aware upgrade driver with readiness gates, traffic isolation,
  functional rollback, and restart-safe PostgreSQL migration;
- add full-stack evacuation coverage and release/deployment contract tests.

Track A, the same-version SHA rolling update of an individual `devshardd`
generation, is described and reviewed in
[#1490](https://github.com/gonka-ai/gonka/pull/1490). Track B builds on those
generation and process-lifecycle primitives at the whole-host level.

---

## Runtime architecture

~~~text
clients / gateway
        |
        v
public edge proxy
        |
        +-- /v1/* --------> edge-api-router (HAProxy, round-robin)
        |                         |
        |                         +-- edge-api-0
        |                         +-- edge-api-1
        |
        +-- /devshard/* --> versiond-router (HAProxy, consistent hash)
                                  |
                                  +-- versiond-0
                                  |      +-- devshardd child for v1
                                  |      +-- devshardd child for v2
                                  |      +-- ... one child per approved version
                                  |
                                  +-- versiond-1
                                         +-- the same version set

HA-eligible devshardd generations share Postgres across versiond hosts.
Pre-HA SQLite versions remain pinned to their explicit legacy owner.
~~~

Each `versiond` supervises one `devshardd` child per approved protocol version;
the diagram does not imply a one-to-one `versiond`/`devshardd` relationship.

### Why the service routers use HAProxy

Before this change, two nginx OSS containers routed the service pools:

- `versiond-router` kept `/devshard/*` sessions sticky by escrow ID and selected
  HA or legacy backends by protocol version;
- `edge-api-router` distributed stateless `/v1/*` Tier A reads round-robin.

Their startup scripts rendered `VERSIOND_HOSTS` and `EDGE_API_HOSTS` into static
upstream lists. nginx could refresh the IP of a listed hostname, but it could
not discover a newly added replica or remove a departed one from that list. It
also had no active application-readiness contract, so a listening but starting,
degraded, or draining process could still receive new traffic.

The HAProxy routers implement the pool behavior directly in the data plane:

- `server-template` follows all addresses published under the shared
  `versiond-pool` and `edge-api-pool` DNS aliases;
- starting or stopping a replica changes DNS membership without a router
  configuration change;
- `init-state fully-down` keeps a newly discovered address out of rotation until
  its first successful check;
- versiond has one consistent-hash backend per declared protocol version;
- `hash-key addr` makes the ring independent of DNS answer order and HAProxy
  slot assignment, so router restarts do not re-home every escrow;
- edge-api retains round-robin balancing;
- established connections remain on their original backend when readiness is
  withdrawn, while new work is sent only to healthy replicas.

The HAProxy configuration preserves the relevant nginx behavior: SSE streaming,
forwarding headers, legacy-version pinning, bounded request bodies, and retries
that never replay non-idempotent application work.

---

### versiond router

Pool membership and route eligibility are separate questions:

1. Docker DNS supplies the current addresses under `VERSIOND_POOL_HOST`.
2. Each declared version receives its own backend and its own readiness query.
3. A host missing or losing one version leaves only that version's ring and can
   continue serving every other healthy route.

The versiond health check is intentionally two-stage so v5 HAProxy can safely
run during a mixed-version cutover or a rollback to v4:

~~~text
GET /readyz[?version=<v>]
    200 -> v5 readiness is satisfied
    404 -> pre-v5 capability; continue with the legacy route check
    other -> backend is DOWN

GET /healthz or GET /<v>/healthz
    200 -> backend is UP
    other -> backend is DOWN
~~~

A v5 `503` is authoritative and removes a starting or draining host. Accepting
the pre-v5 `404` never admits a process by itself: the real host or version route
must still return `200`.

`VERSIOND_VERSIONS` declares the routable version window. Once that list is
non-empty, an undeclared version fails closed with `503` instead of falling
through to a host by accident. Versions in `VERSIOND_NON_HA_VERSIONS` use the
same per-version check but remain pinned to `VERSIOND_LEGACY_HOST`, because their
SQLite sessions cannot fail over.

The request hash is derived inside HAProxy. Session paths use the escrow ID, so
every request for the same live session reaches the same `versiond`; non-session
paths hash by normalized path. Client headers cannot choose the backend.

### versiond endpoints

All readiness endpoints are on the traffic listener:

| Endpoint | Contract |
| --- | --- |
| `GET /healthz` | unchanged legacy JSON array of per-version child state |
| `GET /readyz` | host-level capacity for versionless or coarse routing |
| `GET /readyz?version=<v>` | a current, ready route for exactly `<v>` |

Child readiness is monitored continuously. A running child that loses its chain
subscription loses its readiness vouch and leaves the corresponding pool. The
host-level `Converged` condition latches after the full desired set has run once,
so an ordinary same-name child replacement does not evict every host at once.
Reconcile errors remain observable without hiding routes that are still serving.

### edge-api router

`edge-api-router` discovers `EDGE_API_POOL_HOST`, balances healthy replicas
round-robin, and checks `GET /readyz`. `edge-api` readiness is a bounded, cached
chain probe, so an instance that cannot reach the chain leaves rotation and
rejoins automatically when the dependency recovers.

### HA storage guard

The HA overlay sets `GONKA_HA=true`, which makes `devshardd` refuse to start
unless its storage is fail-closed Postgres. HAProxy also removes any
client-supplied `Devshard-Ha` header and stamps its own trusted value:

- an explicit HA deployment keeps the guard latched during partial outages;
- without that declaration, the safety fallback derives the header from the
  number of usable servers in the backend selected for the request;
- a pinned single-owner legacy backend strips the header.

`devshardd` validates the header again against its active storage. This catches
partial rollouts and accidental multi-host scaling even when the deployment flag
was omitted.

---

### versiond host evacuation

The operator uses the ordinary container lifecycle. For example, from
`deploy/join`:

~~~console
source ./config.env && \
docker compose -f docker-compose.yml -f docker-compose.versiond.yml \
  stop versiond2
~~~

The safe ordering is owned by `versiond`:

~~~text
operator / Docker       versiond                  HAProxy             devshardd
       |                   |                         |                     |
       | SIGTERM           |                         |                     |
       |------------------>| announcing              |                     |
       |                   | /readyz = 503            |                     |
       |                   | admission still open    |                     |
       |                   |                         | check fails         |
       |                   |                         | host leaves pool    |
       |                   | draining                |                     |
       |                   | admission closed       | no new requests     |
       |                   | wait proxy leases      |                     |
       |                   | POST /drain -------------------------------->|
       |                   | wait child idle                              |
       |                   | SIGTERM / reap ----------------------------->|
       |                   | stop HTTP and exit                           |
~~~

`VERSIOND_DRAIN_ANNOUNCE` defaults to five seconds. During that window the host
already fails readiness but still accepts work, giving HAProxy enough time to
withdraw it before admission closes. A host that never reached `serving` skips
the announcement and drains directly because it was never eligible for traffic.

Every proxied request acquires a host admission lease before forwarding. The
lease is released only when the handler finishes; for SSE that means the full
lifetime of the stream. A child generation has its own target lease, so a
rolling replacement cannot reap the exact process still serving that response.

New requests go to a surviving host. HA sessions can recover there from shared
Postgres. A legacy SQLite owner cannot be evacuated while
`VERSIOND_NON_HA_VERSIONS` is non-empty.

Temporary stop, replacement, addition, and permanent decommission are described
in the [host evacuation runbook](devshard/docs/versiond-host-evacuation.md).
Permanent membership is represented by the persisted Compose replica count,
for example `VERSIOND2_REPLICAS=0`, so a later Docker daemon restart cannot
silently recreate a decommissioned service.

### edge-api shutdown

`edge-api` follows the same readiness-first sequence:

~~~text
SIGTERM -> /readyz=503 -> announce -> close listener
        -> wait accepted queries -> force on budget/repeated signal -> exit
~~~

The default five-second `EDGE_API_DRAIN_ANNOUNCE` lets HAProxy remove the
replica before Echo closes the listener. Accepted queries then receive up to
`EDGE_API_SHUTDOWN_BUDGET` to complete. A repeated signal shortens announcement
or forces remaining connections, and Docker's stop grace remains the final
backstop.

### Budgets and escalation

`versiond` fixes one absolute deadline when shutdown begins. Announcement, poll
unwind, host leases, child drain, child process grace, and HTTP shutdown all
spend from that same budget; sequential phases cannot add independent maximums.

| Setting | Default | Meaning |
| --- | ---: | --- |
| `VERSIOND_DRAIN_ANNOUNCE` | `5s` | fail readiness while still accepting |
| `VERSIOND_HOST_SHUTDOWN_BUDGET` | `25m` | one internal absolute deadline |
| `VERSIOND_STOP_GRACE_PERIOD` | `30m` | Docker's outer `SIGKILL` reserve |
| `EDGE_API_DRAIN_ANNOUNCE` | `5s` | edge-api readiness announcement |
| `EDGE_API_SHUTDOWN_BUDGET` | `2m` | accepted edge-api request drain |
| `EDGE_API_STOP_GRACE_PERIOD` | `3m` | edge-api outer `SIGKILL` reserve |

Repeated `SIGTERM` is idempotent; `SIGINT` explicitly forces versiond teardown.
Budget expiry forces remaining HTTP work and children, then confirms process
reap before exit. The poll worker receives at most 10% of the remaining host
budget, capped at five seconds. If it ignores cancellation, the manager drain
barrier already prevents new generations and restarts, so shutdown continues.

---

### 1. versiond host lifecycle

~~~text
starting -> serving -> announcing -> draining -> stopping -> stopped
    |          |           |            |           |
    +----------+-----------+------------+-----------+-> forcing -> stopped
~~~

- `starting`: not ready and not accepting;
- `serving`: ready, accepting, and reconciling children;
- `announcing`: unready but still accepting while HAProxy removes the host;
- `draining`: admission and desired-state changes are closed;
- `stopping`: child and HTTP graceful shutdown is running;
- `forcing`: deadline or explicit escalation is closing remaining work;
- `stopped`: HTTP has stopped and every child has been reaped.

The transition table owns legality, admission policy, and readiness policy.
`starting -> announcing` is forbidden because a host that never served must not
open admission while shutting down.

### 2. Child generation lifecycle

~~~text
preparing -> starting -> running -> retiring -> draining -> stopping -> stopped
     |           |          |
     +-----------+----------+----> failed ----> starting
~~~

The generation FSM owns route publication, exact-target retirement, drain, and
restart. `retiring` removes that generation from new routing before lifecycle
drain starts. The manager rechecks the generation under its mutex immediately
before every process fork, so host drain and generation retirement form a real
start barrier rather than a timing assumption.

### 3. Supervised OS process lifecycle

~~~text
running --stop requested--> terminating --grace/force--> killing
   |                              |                         |
   +---------- process exited ----+-------------------------+--> exited
~~~

This FSM owns process-group `SIGTERM`, escalation to `SIGKILL`, and reap. Only
the process-exited event produced by `cmd.Wait` reaches `exited` and closes
completion.

### 4. devshardd request lifecycle

~~~text
starting --chain ready--> serving --chain disconnected--> disconnected
                              ^                               |
                              +----------- chain ready -------+

starting / serving / disconnected --drain requested--> draining
draining --any lifecycle event------------------------> draining
~~~

Readiness, admission, drain state, and the in-flight counter come from one table
under one mutex. `draining` is terminal and rejects new non-lifecycle work.

---

## Safe v4-to-v5 deployment transition

The steady-state readiness and drain contracts protect v5 operation, but an
existing v4 installation cannot safely reach that state through a blind
full-model Compose recreate. The deployment contracts are materially different:
v4 uses static nginx upstreams and has no `/readyz`; v5 uses DNS-discovered
HAProxy pools and active readiness. The HA PostgreSQL mount also changes from an
anonymous image volume to persistent bind-mounted `PGDATA`.

`deploy/join/upgrade-devshard-v5.sh` provides the compatibility bridge. The
operator runs one command from `deploy/join`:

~~~console
./upgrade-devshard-v5.sh
~~~

The driver sources `config.env`, checks its dependencies, and detects two
independent topology axes from the existing containers:

| Axis | Standard | HA / multi |
| --- | --- | --- |
| versiond | `versiond` | shared Postgres, `versiond2`, and `versiond-router` |
| edge-api | `edge-api` | `edge-api2`, `edge-api3`, and `edge-api-router` |

The standard Quickstart therefore remains a single-service upgrade and does not
create HA-only services. HA overlays are loaded only for installations that
already use them. Every replacement is targeted with `--no-deps`; unrelated
node, chain, proxy, bridge, Tmkms, explorer, and ML services remain running.

### Ordered cutover

~~~text
detect topology and source state
        |
        +-- capture immutable rollback images and all routable versions
        |
        +-- HA only: disk preflight -> atomic PostgreSQL migration
        |
        +-- versiond HA:
        |      isolate versiond2 in old nginx
        |      -> replace and wait ready/routes
        |      -> install HAProxy router
        |      -> replace legacy owner versiond
        |
        +-- edge-api multi:
               isolate edge-api2 in old nginx
               -> replace and wait ready
               -> install HAProxy router
               -> replace remaining replicas one at a time
~~~

The temporary nginx barrier removes the active replacement from the rendered
upstream and reloads nginx without terminating established connections. An
after-render hook persists that isolation across a router container restart.
Once the first v5 replica is healthy, HAProxy's active checks protect later
replacements.

Before cutover-sensitive replacements the script captures the immutable source
image ID, whether the service was running, and every version route that was
available. `EXIT`, `INT`, `TERM`, and `HUP` use the same compensation path. A
failed replacement before the readiness-aware router is installed restores the
captured image and original running/stopped state, or stops a newly introduced
service, then requires three consecutive functional probes. After HAProxy owns
an edge-api pool, an unready replacement is stopped while the surviving replicas
continue serving. Versiond rollback must restore every baseline version;
edge-api rollback must execute the chain-backed `/v1/versions` query; router
rollback must route the union of all replica baselines.

The mixed-version HAProxy check described above lets a restored v4 versiond
rejoin only when its real routes work. A v5 process that explicitly returns
`503` remains out of rotation. Temporary rollback tags are retained after a
failed operation and removed only after the complete upgrade succeeds.

### PostgreSQL migration and recovery

The v4 HA overlay inherits the Postgres image's anonymous
`/var/lib/postgresql/data` volume. v5 stores `PGDATA` under the stable
`DEVSHARD_POSTGRES_DATA_DIR` bind mount.

Before recreating PostgreSQL, the upgrade driver mounts the source read-only,
measures it with `du`, checks the target filesystem with `df`, and requires the
source size plus a 10% reserve. The entrypoint copies into a staging directory,
validates `PG_VERSION`, syncs the copy, and atomically publishes it. The source
volume remains unchanged as a physical recovery copy.

Startup fails closed when versiond artifacts imply an existing installation but
the source database is missing; it never silently initializes an empty HA
database. A recovery overlay can reattach an explicitly selected detached v4
volume and reuse the same restart-safe preflight and atomic migration. Partial
staging is reclaimable, completed staging is reusable, and the source is
detached only after bind-mounted `PGDATA` is verified.

The exact procedure and recovery boundaries are documented in the
[`0.2.15-v5` release guide](devshard/docs/release-0.2.15-v5.md).

---

## Failure policy

| Failure | Result |
| --- | --- |
| One host loses one version | removed only from that version's backend |
| Host disappears from DNS | its HAProxy slot becomes unresolved |
| New host appears in DNS | starts down and joins only after checks pass |
| Every eligible host is unready | router fails closed with `503` |
| Oracle/archive reconcile fails while old routes serve | routes remain available; failure stays observable |
| Shutdown budget expires | remaining work is forced and every child is reaped |
| Replacement fails before router cutover | captured source state is restored, or a new service is stopped |
| HAProxy-protected edge replica fails | unready replica is stopped; healthy pool remains available |
| PostgreSQL migration fails | both source and target are preserved; database remains stopped for recovery |

---

## Test coverage

The full-stack
[`TestVersiondHostEvacuation`](devshard/testenv/citest/versiond_host_evacuation_test.go)
uses two real `versiond` containers, HAProxy, shared Postgres, and a paused
OpenAI-compatible SSE response. It verifies:

1. stable escrow placement across a router restart;
2. readiness withdrawal before admission closes;
3. zero-tolerance continuity probes through the sticky path;
4. new requests avoiding the target while its accepted SSE stream completes;
5. session recovery on the Postgres-backed survivor;
6. graceful process exit without Docker's `SIGKILL` backstop;
7. DNS and pool withdrawal after the target stops;
8. a restarted host remaining out of rotation behind an oracle barrier;
9. readiness-gated rejoin and restoration of sticky placement.

Focused coverage includes:

- transition matrices, invalid edges, and ordering invariants for all four FSMs;
- shutdown races, repeated signals, force paths, stuck poll workers, and process
  reap under `go test -race`;
- child-readiness withdrawal, stale-vouch expiry, per-version routing, and
  undeclared-version refusal;
- deterministic hash-ring stability under adverse DNS ordering;
- HA header derivation from the selected backend and Postgres storage guards;
- real HAProxy render/runtime checks, including mixed v4/v5 readiness;
- edge-api chain readiness and graceful/forced shutdown;
- upgrade-driver topology, barrier, signal-compensation, rollback, and complete
  version-baseline tests;
- real-Docker PostgreSQL in-place migration, detached-volume recovery, and
  restartability tests;
- Compose image/health contracts and a pull-first smoke test of all four release
  images.

Relevant commands:

~~~console
make -C versiond-router test-render
make -C edge-api-router test-render
(cd versioned && go test -race ./... && go vet ./...)
(cd edge-api && go test -race ./... && go vet ./...)
(cd devshard && go test ./...)
make -C versioned e2e
make -C devshard/testenv citest-versiond-host-evacuation
bash deploy/join/upgrade-devshard-v5_test.sh
bash deploy/join/versiond-compose-config_test.sh
bash deploy/join/devshard-postgres-migration-preflight_test.sh
bash deploy/join/devshard-postgres-entrypoint_test.sh
bash deploy/join/legacy-router-upgrade-barrier_test.sh
bash deploy/join/release-image-smoke_test.sh --contract
~~~

Dedicated workflows run the host-evacuation acceptance test, real PostgreSQL
upgrade/recovery tests, and release-image smoke checks for the files that affect
those contracts.

---

## Deployment and documentation

- [v5 upgrade and recovery guide](devshard/docs/release-0.2.15-v5.md)
- [versiond host evacuation runbook](devshard/docs/versiond-host-evacuation.md)
- [rolling update design](devshard/docs/rolling-update.md)
- [HA architecture](devshard/docs/high-availability-architecture.md)
- [versiond-router contract](versiond-router/README.md)
- [edge-api-router contract](edge-api-router/README.md)

Important deployment settings:

- `VERSIOND_POOL_HOST` and `EDGE_API_POOL_HOST` resolve to every replica in the
  applicable pool;
- `VERSIOND_VERSIONS` declares every version the HA router may serve;
- `VERSIOND_LEGACY_HOST` is required while
  `VERSIOND_NON_HA_VERSIONS` is non-empty;
- `VERSIOND_STOP_GRACE_PERIOD` exceeds
  `VERSIOND_HOST_SHUTDOWN_BUDGET`;
- `EDGE_API_STOP_GRACE_PERIOD` exceeds the edge-api announce and shutdown
  budgets;
- the four v5 deployment services use the published
  `0.2.15-devshard-v5` image tag.

---

Supersedes #1489. This branch applies Track B to `upgrade-v0.2.15` and includes
the release transition needed to reach the new DNS-discovered, readiness-gated
runtime safely from existing v4 join deployments.


snevolin:
Closing this PR so the functionality can be split into smaller, independent PRs that can be reviewed, tested, and merged separately. The work will be resubmitted as focused changes.

snevolin:
This umbrella PR was closed so its functionality could be split into smaller, independently reviewable and mergeable changes.

The focused changes extracted from this work are now available as:

- [#1599 Replace the versiond router with HAProxy](https://github.com/gonka-ai/gonka/pull/1599)
- [#1606 Reconcile versiond routes from governance](https://github.com/gonka-ai/gonka/pull/1606)
- [#1604 Add graceful versiond host evacuation](https://github.com/gonka-ai/gonka/pull/1604)
- [#1600 Replace the edge API router with HAProxy](https://github.com/gonka-ai/gonka/pull/1600)
- [#1605 Add graceful edge API shutdown](https://github.com/gonka-ai/gonka/pull/1605)
- [#1601 Fail closed on unsafe HA storage](https://github.com/gonka-ai/gonka/pull/1601)
- [#1602 Track live PostgreSQL readiness](https://github.com/gonka-ai/gonka/pull/1602)
- [#1603 Persist and migrate devshard PostgreSQL data](https://github.com/gonka-ai/gonka/pull/1603)
- [#1607 Expose durable PostgreSQL storage identity](https://github.com/gonka-ai/gonka/pull/1607)
- [#1608 Validate shared PostgreSQL before HA mutations](https://github.com/gonka-ai/gonka/pull/1608)
- [#1609 Add the public HAProxy routing tier](https://github.com/gonka-ai/gonka/pull/1609)
- [#1610 Run versiond routers as an independent fleet](https://github.com/gonka-ai/gonka/pull/1610)
- [#1611 Add a fail-safe devshard HA host updater](https://github.com/gonka-ai/gonka/pull/1611)

All target `upgrade-v0.2.15` and include focused tests. The merge-order sections in dependent PRs identify which smaller capabilities must land first, while each concern remains independently reviewable.

PR #1518, opened 2026-07-29
fix(mlnodeclient): report MLnode HTTP failures instead of silently succeeding

Ryanchen911:
## What this is

`Stop` and `InferenceUp` discard the HTTP response and return only transport
errors, so **any** status the MLnode replies with is treated as success. The
MLnode does use status to signal refusal:

- `/api/v1/inference/up` → **409** when vLLM is already running or still
  starting (`mlnode/packages/api/src/api/inference/routes.py:26-29`), **500**
  when startup failed
- `/api/v1/stop` → only ever **200**

So a caller could be told a model was loaded when it was not, and the broker
could mark a failed launch healthy.

Pre-existing since `c3577272a` (#1348). Split out of the onboarding work in
#1296 at @bonujel's request: these calls sit on the broker's production PoC and
inference path, so they should get reviewed on their own rather than inside an
onboarding diff.

## Behavior change on a live path

`/inference/up` answering 409 now marks the node `FAILED` where it previously
reported success. A concrete way to reach that state, for anyone auditing the
blast radius:

mlnode's `/api/v1/stop` only calls `_async_stop()` when `is_running()` is true
(`routes.py:82-85`), but during startup `is_running()` is false while
`is_starting()` is true (`manager.py:162, 222`). So a stop-then-up against a
*starting* node leaves the startup task alive, `/inference/up` returns 409, and
`InferenceUpNodeCommand` records `FAILED`.

It self-heals through the status query worker and the reconciler, and the new
outcome is the truthful one — the old silent success is the bug. Calling it out
because it is a live-path change, not because it needs reverting.

## Changes

- Check the response status on `Stop` and `InferenceUp`.
- Add `StatusError`, so callers can separate transient refusals
  (409 / 429 / 408 / 5xx) from deterministic ones (400 / 404 / 422) without
  pattern-matching error strings. `IsStatusError` / `IsTransientStatus` are the
  helpers; `Transient()` keeps the policy in one place.
- Bound the error-body excerpt (4 KiB) and drain + close response bodies so the
  connection can be reused.
- Fix a shadowed `err` in `utils.SendPostJsonRequest`: a marshalling failure was
  swallowed and the function could return `(nil, nil)`, leaving the caller to
  dereference a nil response.
- Declare `Content-Type: application/json` when there is a payload.

### On the Content-Type change specifically

This is **not** fixing a live failure, and it would be wrong to describe it as
one. FastAPI falls back to parsing the body as JSON when no `content-type` header
is present at all — `fastapi/routing.py`, the `if not content_type_value` branch
(verified in the installed 0.128.8; the same branch exists in the 0.115.8 floor
mlnode pins). That is why the PoC v2 path has worked without it all along.

It is sent because the request should describe what it carries, it matches what
`SendDeleteJsonRequest` already does, and it is what keeps these calls working if
strict content-type checking is ever enabled upstream (a proxy, or a stricter
framework default).

`SendPostJsonRequest` is shared, so the header now goes out from every call site
with a non-nil payload: `CheckModelStatus`, `DownloadModel`, `DeleteModel`,
`InferenceUp`, and the PoC v2 path (`InitGenerateV2`, `GenerateV2`,
`StopPowV2`). `Client.Stop` passes nil, sends no body, and is unaffected —
covered by a test either way.

## Tests

New `mlnodeclient/client_test.go`:

- `/inference/up` 200 / 409 / 500 / 422, asserting both the error and its
  transient classification
- `/stop` 200 and 500, asserting the error is a `StatusError` carrying the code
- the Content-Type matrix: payload sends the header, nil payload sends none, and
  an unmarshalable payload returns an error rather than `(nil, nil)`
- the full transient/permanent status table

## Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean
- `go test` green for `mlnodeclient` and `utils`, plus `broker`,
  `internal/server/mlnode` and `poc` — the packages that actually drive these
  calls

### DIFF
--- a/common/utils/http.go
+++ b/common/utils/http.go
@@ -5,6 +5,7 @@ import (
 	"common/logging"
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"time"
 
@@ -25,8 +26,11 @@ func SendPostJsonRequest(ctx context.Context, client *http.Client, url string, p
 		// Create a POST request with no body if payload is nil.
 		req, err = http.NewRequestWithContext(ctx, http.MethodPost, url, nil)
 	} else {
-		// Marshal the payload to JSON.
-		jsonData, err := json.Marshal(payload)
+		// Marshal the payload to JSON. Note: assign with = (not :=) so a
+		// marshalling or request-construction failure is not swallowed by a
+		// shadowed err, which previously could return (nil, nil).
+		var jsonData []byte
+		jsonData, err = json.Marshal(payload)
 		if err != nil {
 			return nil, err
 		}
@@ -38,7 +42,17 @@ func SendPostJsonRequest(ctx context.Context, client *http.Client, url string, p
 	}
 	if req == nil {
 		logging.Error("SendPostJsonRequest. Failed to create HTTP request", types.Server, "url", url, "payload", payload)
-		return nil, err
+		return nil, fmt.Errorf("failed to create POST request for %s", url)
+	}
+	// Declare the body we are actually sending. This is not fixing a live
+	// failure: FastAPI falls back to JSON parsing when no content-type header is
+	// present at all (routing.py, the `if not content_type_value` branch), which
+	// is why these calls have worked without it. Send it because it is correct,
+	// it matches what SendDeleteJsonRequest already does, and it is what keeps
+	// these calls working if strict content-type checking is ever enabled
+	// upstream (a proxy, or a stricter framework default).
+	if payload != nil {
+		req.Header.Set("Content-Type", "application/json")
 	}
 
 	return client.Do(req)
--- a/decentralized-api/mlnodeclient/client.go
+++ b/decentralized-api/mlnodeclient/client.go
@@ -6,8 +6,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/productscience/inference/x/inference/types"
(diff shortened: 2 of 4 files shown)


Ryanchen911:
Retargeted from `upgrade-v0.2.14` to `upgrade-v0.2.16`, since mainnet has already upgraded past 0.2.14 (applied at height 5195700) and that branch has diverged from the shipping one.

**One real fix was needed on the way, worth flagging rather than leaving for a reviewer to trip over.** `decentralized-api/utils/http.go` moved to `common/utils/http.go` between the two bases. The rebase carried `client.go` to the new import path on its own, but `client_test.go` still imported `decentralized-api/utils`, which no longer holds `SendPostJsonRequest` — so the package did not compile. That's the third commit here.

Note this also means the PR could not simply have its base switched in the GitHub UI: the branch was cut from the old base, so a bare retarget would have pulled the whole 0.2.14 history into the diff (548 files) and still left the test broken. It was rebased with `--onto` instead, so the diff is unchanged at **4 files, +406/-7**.

Verified on `upgrade-v0.2.16`:

- `go build ./...` and `go vet ./mlnodeclient/...` clean in `decentralized-api`
- `go test ./mlnodeclient/...` passes
- `go test ./utils/...` passes in `common` (that's where the modified `http.go` now lives)

The three open Copilot threads are unaddressed but unchanged in substance; happy to walk through them if a reviewer wants.

bonujel:
Flagging a dependency that has appeared since this was opened, in case it helps
this one get looked at.

I have been slicing #1296 into reviewable pieces — #1644, #1645 and #1646 are
out so far. The next slice is the pre-PoC MLnode test, and it does not build
without this PR: `mlnode_tester.go` classifies MLnode failures through
`mlnodeclient.IsStatusError` and `IsTransientStatus`, which only exist here. A
409 from `/inference/up` has to read as retryable and a bad request as terminal,
and that distinction is exactly what this PR introduces.

So this is no longer only a standalone fix on the broker's production path — it
is also what the remaining onboarding work sits on.

One thing I noticed while trying to build that slice: this currently shows
`CONFLICTING` again. The base moved after your Aug 21 retarget, which is the
same treadmill the rest of us are on.

Ryanchen911:
Closing as superseded. The core intent here — reporting MLnode HTTP failures instead of silently discarding non-2xx responses — has since landed independently on `upgrade-v0.2.16` via `requireHTTPSuccess(resp, operation)` in `mlnodeclient/client.go`, which `Stop()` and the start-inference path already call. Rebuilding this PR conflicts head-on with that helper on the same lines (functionally equivalent, structurally different), so there's no clean rebase.

The only delta not covered by `requireHTTPSuccess` is the typed-error classification (`StatusError` + `IsTransientStatus`) for retryable-vs-terminal HTTP failures. If we later want retry classification on top of the existing helper, that can go up as a small focused PR rather than reviving this one. Closing to keep the 0.2.16 queue honest.

PR #1519, opened 2026-07-29
Reward power-cap computed on cPoC-confirmed weight penalizes honest participants

staaason:
Fixes a reward-misallocation bug in Bitcoin-style epoch settlement. The per-participant power cap (anti-concentration, 30%) was measured against the sum of **effective** (Confirmation-PoC-confirmed) weights, while the reward denominator is the sum of **full** PoC weights. The two bases differ, so the effective ceiling drifts with network confirmation health: when other participants fail their cPoC, the effective total shrinks, the 30% ceiling drops below its intended value, and an honest participant whose full-weight share is under 30% is still capped — the excess minted to the governance account instead of paid out.


staaason:

PR #1520, opened 2026-07-29
BridgeContract group-key rotation isn't domain-separated 

staaason:
The bridge group-key rotation was the only one of the three BLS-signed messages (mint, withdraw, rotation) without an operation tag. A signature over `epoch || GONKA_CHAIN_ID || <256 arbitrary bytes>` was byte-identical to a rotation authorizing those bytes as the next group key.

Add a `keccak256("TRANSITION_OPERATION")` tag at offset 40 of the rotation validation message, mirroring `MINT_OPERATION` / `WITHDRAW_OPERATION`. Applied in lockstep across all three producers/verifiers:
1.  `x/bls` keeper `computeValidationMessageHash` (aggregate check)
2.  decentralized-api `computeValidationMessageHash` (off-chain signer) via a shared `types.TransitionOperationTag()` source of truth.
Also add a G2 point-validity gate (`_isValidG2Point`) in `submitGroupKey` as a defense against non-curve forged keys.

### DIFF
--- a/decentralized-api/internal/bls/group_validation.go
+++ b/decentralized-api/internal/bls/group_validation.go
@@ -137,7 +137,7 @@ func (bm *BlsManager) ProcessGroupPublicKeyGeneratedToSign(event *chainevents.JS
 }
 
 // computeValidationMessageHash computes the validation message hash using the same format as the chain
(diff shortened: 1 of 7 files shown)


staaason:

PR #1523, opened 2026-07-30
Self-declared hardware model list can relabel PoC weight into an unvalidated model subgroup

staaason:
`setModelsForParticipants` rebuilt each participant's `Models`/`MlNodes` from the self-reported hardware list, discarding the per-model binding produced by PoC validation and preserved-node carryover.

Hardware may now only constrain that binding, never extend it.


staaason:

PR #1526, opened 2026-07-30
0.2.15 versions path fix

akup:
Fixing /v1/versions path at dapi, that should be enriched with mlnodes versions


DimaOrekhovPS:
Superseded by #1752: leftover JSON-shape restore only; /v1/versions already shipped in #1528.

PR #1530, opened 2026-07-31
fix(devshardd): evict in-memory hosts on epoch prune

akup:
## Summary
- On epoch change (dapi long-poll or chain-poll fallback), advance phase + managed-storage horizon, prune DB, **`EvictBefore` in-memory hosts**, and drop expired payload epochs — fixing the case where storage pruned escrows but hosts kept serving them.
- Seed the same path at startup (initial params snapshot does not fire `OnEpochChange`); shut down via `manager.Close()` so hosts and store tear down together.
- Simplify `chain.Phase` to epoch-only; **remove Comet `OnNewBlock` phase/prune/payload cleanup**.
- Mount `/debug/pprof/*` on the admin server (`DEVSHARD_ADMIN_ADDR`); remove the separate loopback `:port+10000` listener.

## Motivation
After the standalone `devshardd` move, epoch prune only cleaned storage. In-memory hosts (and validation workers) could outlive pruned escrow IDs, so the gateway still routed to stale sessions. This restores the prune ↔ evict coupling and uses one epoch clock instead of a second per-block path.

## Why drop `OnNewBlock` epoch work
Per-block handlers were doing the same job as runtime-config **`OnEpochChange`**, but worse for host fleets:

- **Duplicate logic.** New-block already pruned storage and dropped payload epochs; epoch change is the correct clock for that. Keeping both meant two paths that could drift (and still missed `EvictBefore`).
- **Extra node traffic.** Every host’s `devshardd` subscribed to Comet new blocks and called **`GetCurrentEpoch` on each block** to refresh phase. Hosts already get epoch/params via **dapi** (`GetRuntimeConfig` long-poll, with chain-poll fallback). That path is built to fan out epoch updates without N hosts hammering the chain node every block.
- **Unused height.** `phase.BlockHeight` had no consumers in standalone `devshardd`, so the per-block height updates were pure overhead.

Comet WS remains for escrow settlement (and readiness); it is no longer an epoch/prune driver.

## Test plan
- [ ] `go test ./cmd/devshardd/...` and `./storage/...` (from `devshard/`)
- [ ] Confirm `TestHostManager_EvictBeforeClosesOldSessions` and related resolution/recover tests pass
- [ ] With `DEVSHARD_ADMIN_ADDR` set: `curl -sS http://$ADMIN/debug/pprof/` and `go tool pprof http://$ADMIN/debug/pprof/heap`
- [ ] Confirm public listener has no `/debug/pprof/`
- [ ] On epoch advance in a long-lived host: old-epoch sessions leave memory; storage prune cutoff matches `EvictBefore`
- [ ] Confirm hosts no longer issue per-block `GetCurrentEpoch` after new blocks; epoch still advances via dapi runtime-config / fallback

### DIFF
--- a/devshard/cmd/devshardd/server.go
+++ b/devshard/cmd/devshardd/server.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"net/http"
+	"net/http/pprof"
 
 	"github.com/labstack/echo/v4"
 	"github.com/labstack/echo/v4/middleware"
@@ -46,6 +47,17 @@ func buildAdminServer(lifecycle *lifecycleState, storageReady func() bool) *echo
 	})
 	e.GET("/drain/status", func(c echo.Context) error {
 		return c.JSON(http.StatusOK, lifecycle.Status())
+	})
+	// Canonical /debug/pprof/ paths so Index sub-profile links resolve.
+	// Bound only when DEVSHARD_ADMIN_ADDR is set (same network trust as /drain).
+	e.Any("/debug/pprof/", echo.WrapHandler(http.HandlerFunc(pprof.Index)))
+	e.Any("/debug/pprof/cmdline", echo.WrapHandler(http.HandlerFunc(pprof.Cmdline)))
+	e.Any("/debug/pprof/profile", echo.WrapHandler(http.HandlerFunc(pprof.Profile)))
+	e.Any("/debug/pprof/symbol", echo.WrapHandler(http.HandlerFunc(pprof.Symbol)))
+	e.Any("/debug/pprof/trace", echo.WrapHandler(http.HandlerFunc(pprof.Trace)))
+	e.Any("/debug/pprof/:name", func(c echo.Context) error {
+		pprof.Handler(c.Param("name")).ServeHTTP(c.Response(), c.Request())
+		return nil
 	})
 
 	return e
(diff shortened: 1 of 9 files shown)


devshard/cmd/devshardd/server.go
```diff
@@ -47,6 +48,17 @@ func buildAdminServer(lifecycle *lifecycleState, storageReady func() bool) *echo
 	e.GET("/drain/status", func(c echo.Context) error {
 		return c.JSON(http.StatusOK, lifecycle.Status())
 	})
+	// Canonical /debug/pprof/ paths so Index sub-profile links resolve.
+	// Bound only when DEVSHARD_ADMIN_ADDR is set (same network trust as /drain).
+	e.Any("/debug/pprof/", echo.WrapHandler(http.HandlerFunc(pprof.Index)))
```
DimaOrekhovPS, line 53: Is this connected to the fix somehow?

DimaOrekhovPS (approved):

devshard/cmd/devshardd/server.go
akup: It was at same commit, that was dropped during remerging v4 from 0.2.13 base to 0.2.14

We can drop it, but it was at 0.2.14

The only thing I've changed - moved this path behind admin to have authentication, at 0.2.14 it was bind to local address (not exposed to external calls)

PR #1531, opened 2026-07-31
mlnode for vLLM 0.25.1: metrics exporter + heartbeat liveness + RC image pin

baychak:
As requested by @vbgd0 in the chat: the mlnode side of the vLLM 0.25.1 update, collected onto `main` so the release branch does not have to redo it.

## What this carries

Three things, in the order they stack:

1. **Metrics exporter** (was #1501, targeted `upgrade-v0.2.15`) — schema-v1 `GET /api/v1/metrics` for ML nodes. Cherry-picked clean onto `main`; the two commits include the Copilot-review fixes already applied there.
2. **Heartbeat liveness** (was #1421, targeted `main`) — replaces the `/health`-based liveness that lies in both directions: a hung TP worker keeps returning 200 (node silently dead for 5–22 min), a merely saturated engine misses the 2 s deadline and gets restarted while healthy (8+/day observed). Liveness now derives from `vllm:iteration_tokens_total_count`, which only freezes on a real stall. Fleet-validated: false restarts → 0, real stalls detected in ~2 min. Pairs with the shm-ring knob merged as gonka-ai/vllm#82 — that one lowers stall frequency, this one fixes stall detection.
3. **Compose image pin** — `deploy/join/docker-compose.mlnode.yml` points at the release-candidate image (overlay-k9, pinned by digest): vllm `release/v0.25.1` @04a165c0 + `gonka-vllm-plugins` v0.1.1 + this branch's mlnode. Placeholder until a gonka-ai-built release tag exists — swap the line then.

## Verification

Exporter + heartbeat + proxy unit suites run together on this exact branch: **63/63 pass**. The two source PRs carry the longer histories and review context; closing both in favour of this one.

Supersedes #1421 and #1501.

PR #1533, opened 2026-07-31
mlnode for vLLM 0.25.1: metrics exporter + heartbeat liveness + reviewed fixes

vbgd0:
## Replacement for #1531

This PR is the reviewed replacement for #1531. It preserves the complete commit history from that PR and adds commit 1b07e5c6c with the fixes found during review.

Why this is not parallel duplicate work: the original head branch belongs to kaitakuai and could not be updated with the available credentials. This branch exists to carry the same work plus the reviewed fixes in one mergeable head. Once maintainers select this PR, #1531 should be closed as superseded.

## Inherited from #1531

- #1501 schema-v1 MLNode metrics exporter
- #1421 vLLM scheduler-heartbeat liveness and damped proxy probing
- documentation corrections describing the implemented DAPI/exporter behavior
- pinned vLLM 0.25.1 DeepSeek-V4-Flash release-candidate MLNode image

The original commits are retained unchanged before the review-fix commit.

## Review fixes added here

- scrape all vLLM replicas concurrently instead of multiplying the timeout by replica count
- reject HTTP-error heartbeat responses and require a complete first heartbeat baseline
- serialize heartbeat state mutation and use monotonic time for grace tracking
- keep grace=0 as a real hang-detection disable switch
- move the newly blocking inference health checks off the asyncio event loop
- protect exporter cache mutation and prevent older concurrent scrapes from overwriting newer replica data
- discard results from an old model generation without clearing a newer generation cache
- prevent an older completed metrics-stage waiter from removing its replacement task
- honor image-provided MLNODE_VLLM_MODULE so the gonka_poc routes are actually loaded
- honor image-provided VLLM_RUNNER_TIMEOUT while keeping the old 1200-second default
- reduce test duplication and remove obsolete or speculative documentation claims

## Deliberately not changed

- no runner lifecycle, shutdown, restart, survivor, or FlashInfer-cache changes
- no heartbeat response cache and no background node-state poller
- no Prometheus family or schema-v1 contract changes
- no redesign of the existing proxy health damping
- no multimodal or image-input changes
- no card-profile, host-driver, or deployment logic beyond the image pin inherited from #1531

## Validation

- 106 affected unit tests passed in the pinned K9 image on a B300 Linux host
- targeted Ruff F/E9 checks passed
- git diff --check passed
- DeepSeek-V4-Flash startup completed after 1217.44 seconds, proving the image timeout override works beyond the former fixed 1200-second ceiling
- process command confirmed gonka_poc.entrypoint.api_router rather than the default vLLM entrypoint
- /api/v1/pow routes were present and healthy
- regular chat-completions smoke test returned exactly OK
- PoC with batch 16 and seq_len 1024 processed 1488 nonces in 60.253 seconds: 1481.755 nonce/min
- PoC stopped cleanly and the MLNode remained healthy with the backend IDLE

Batch 32 is not valid with this image profile: 32 x 1024 requires 32768 batched tokens while the pinned profile forces max-num-batched-tokens=16384. The measured compatible profile therefore uses batch 16.

## Before merge

- run repository CI on this head
- verify the schema-v1 endpoint through the DAPI collector, including disabled and stale-source behavior
- run a multi-replica test on a node with enough GPUs to confirm per-replica heartbeat timing and exporter labels
- decide whether the B300 TP=1 RC image is intentionally suitable for the shared compose path; the image profile is not a fleet-wide B200, H200, or H100 configuration
- retain only one of this PR and #1531 to avoid merging the same inherited commits twice

## AI assistance disclosure

OpenAI Codex was used for review, test refactoring, implementation of the review fixes, B300 deployment verification, and drafting this description. The human submitter is responsible for reviewing every changed line and validating the commands and results above.

### DIFF
--- a/deploy/join/docker-compose.mlnode.yml
+++ b/deploy/join/docker-compose.mlnode.yml
@@ -1,7 +1,13 @@
 services:
   mlnode-308:
     # ghcr.io/product-science/mlnode:3.0.12-post6-blackwell
-    image: ghcr.io/gonka-ai/mlnode:3.0.14-post2
+    # Release-candidate for the vLLM 0.25.1 stack (B300 + DeepSeek-V4-Flash):
+    # every layer from upstream sources — gonka-ai/vllm release/v0.25.1
+    # @04a165c0 (#78 port + #79/#80/#82) + gonka-ai/gonka-vllm-plugins v0.1.1
+    # + this repo's mlnode with the metrics exporter and heartbeat liveness
+    # (the two commits carried by this branch). Swap for the gonka-ai-built
+    # tag when the release image is cut.
+    image: ghcr.io/kaitakuai/mlnode-b300-deepseek-v4-flash:0.2.14-vllm0.25.1-overlay-k9@sha256:ef5260aefa5cc46fbbdee6783d3492d399ae0a8fb649c03a12f731707edcc11b
     hostname: mlnode-308
     volumes:
       - ${HF_HOME:-${HOME}/.cache}:/root/.cache
--- a/docs/adr/mlnode-monitoring-metrics-v1.md
+++ b/docs/adr/mlnode-monitoring-metrics-v1.md
@@ -1,0 +1,176 @@
+# ADR: ML-node monitoring — metrics and interface contract v1
+
+Status: **accepted** (team-approved 2026-07-10)
+Related: [operator guide](../observability.md), [decision log](../decision-log.md)
+
+## Context
+
+ML-node monitoring architecture (v0.5): an exporter inside mlnode → a
+collector + public endpoint in dapi → consumers store data on their side.
+This ADR fixes the Phase 0 contract: the final allowlist, formats,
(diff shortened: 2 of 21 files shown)


baychak:
One question on the `not st["seen"]` branch: a first scrape that comes back incomplete — including the ConnectTimeout path, which returns `(None, None, None)` for saturation — now yields `False` immediately, where the previous version granted the grace window before its first baseline. The watcher's consecutive-count softens it, but a node that is saturated from the very first scrape (warm-up under queued load after a restart) can now accumulate toward a restart it would previously have ridden out. Was that trade-off intended? If yes — fine by us, the tests clearly encode it; just want it to be a decision rather than a side effect.

Everything else reads as strict improvement: `monotonic()`, the `_hb_lock`, the parallel scrape window, `raise_for_status` with the broad except, and moving `is_healthy` off the event loop.

vbgd0:
Branch renamed to vllm-0.25.1-upgrade. GitHub closed this PR during the rename; the identical replacement is #1534.

Issue #1506, opened 2026-07-27
Gateway host-health deadlock: mass quarantine → all hosts stuck as no-winner/suspicious


maria-mitina:
## Summary

I managed to get all hosts into a kind of deadlock on testnet.

### How everyone entered quarantine

I created an escrow with the wrong owner (CLI key ≠ `DEVSHARD_PRIVATE_KEY` the gateway signs with), registered it on the gateway, and sent chat. The gateway broadcast / raced the request across participants. Every host correctly rejected with **HTTP 403** (`restricted to escrow owner`). The gateway treated each 403 as a **host health signal** and applied a **~30 min** inference quarantine to each host that answered.

I understand there are multiple ways to enter quarantine (404 session missing, transport failures, empty streams, etc.). On a small testnet this was easy to trip for the whole set at once; on mainnet that mass fan-out should be harder, but the same mechanism still exists.

### How they got stuck after unquarantine

After I cleared quarantine via `POST /v1/admin/participants/unquarantine`, every host moved into **suspicious / no-winner probation** (`failure_strikes` left after quarantine; attempts marked `primary_no_winner` / `suspicious_winner_deferred`).

They never got out, because there were **no non-suspicious winners** left who could serve the client. Probation only clears when finished good attempts call `ObserveSuccessfulInference` (strike decrement). With every usable host on probation:

1. Hosts still run inference and return real content
2. Content is deferred (`suspicious_winner_deferred`, `winner_nonce=0`) — not crowned to the client
3. Request fails / times out
4. Strike recovery does not progress in practice when no clean host can complete the request for the client

So: **wrong-owner 403 → mass quarantine → mass unquarantine into probation → no clean crown path → permanent suspicious deadlock**.

### Observed on

- Host: testnet gateway `.79`
- Image: `ghcr.io/gonka-ai/devshard-gateway:testnet-v4-local` (built from `devshard-0.2.14-v4`)
- Log signals: `participant_limit_activated status=403`, `decision=primary_suspicious`, `primary_no_winner_reason=probation`, `suspicious_winner_deferred`, `picker_exhausted`

### Suggested product fixes

1. Do **not** quarantine participants on 403 “restricted to escrow owner” (misconfigured escrow / signer mismatch, not a bad host).
2. When **no clean host remains**, crown a suspicious winner that already returned good content (or accept deferred completions toward strike recovery).
[...]

PR #1541, opened 2026-08-04
fix(gateway): stop 403 quarantine and break no-winner probation deadlock (#1506)

staaason:
## Changes
- **No quarantine on inference 403** (`participant_limiter.go`): a 403 is a verdict on the request's signer, identical from every host — never a host-health signal. 404, drift-401, 429/503 and transport triggers are unchanged. 403s stay visible in transport-error metrics and `send_failed` logs.
- **Immediate suspicious-fallback crowning for non-streaming** (`redundancy.go`): the existing `fallbackSuspiciousWinner` promotion now fires as soon as all attempts are done and nothing is left to try, instead of being starved by the 30-minute non-streaming no-content window. Escalation to untried hosts still takes priority (`escalationC == nil` gate), and the empty-stream reduced-max-tokens fallback is untouched (`fallback == nil` keeps the old path). Settled requests then decrement probation strikes via the normal `ObserveSuccessfulInference` path, so the pool heals itself.
- **Admin full clear** (`gateway.go`): `POST /v1/admin/participants/unquarantine` with `{"full": true}` forgets the participant entirely (quarantine + probation strikes + persisted row). Soft unquarantine behavior is unchanged.

### DIFF
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -3729,6 +3729,7 @@ func (g *Gateway) handleAdminUnquarantine(w http.ResponseWriter, r *http.Request
 	}
 	var req struct {
 		ParticipantKey string `json:"participant_key"`
+		Full           bool   `json:"full"`
 	}
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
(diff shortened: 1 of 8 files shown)


staaason:

PR #1543, opened 2026-08-04
fix(devshard): bound executor-supplied confirmed_at

staaason:
## Problem

`applyConfirmStart` verified the executor receipt signature but stored `confirmed_at` with no semantic bounds. A signature proves authorship, not sanity - no forgery is needed, the executor signs with the key it is authorised to use.

## Fix

`types.ValidateConfirmedAt` bounds the value against the in-state `started_at` in both directions (±900s), rejecting non-positive values. Relative to `started_at` because `applyConfirmStart` runs on user, host and replay - there is no agreed "now" available, and any wall clock would split consensus.

Bounding alone would have opened a griefing path: a malicious executor parking one unapplicable `MsgConfirmStart` in its mempool blocks the refusal timeout forever, stranding the reserved cost in `Pending`. `VerifyRefusedTimeout` now treats such a receipt as absent and accepts once past the bound.

### DIFF
--- a/devshard/cmd/devshardctl/proxy_test.go
+++ b/devshard/cmd/devshardctl/proxy_test.go
@@ -1713,13 +1713,13 @@ func TestHandleDebugInferences_IncludesSealedInferences(t *testing.T) {
 	require.NoError(t, err)
 
 	start := &types.MsgStartInference{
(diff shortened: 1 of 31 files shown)


akup:
I think this will be replaced with a height-sync protocol, and all timings will be using block time.

No reason to accept the PR, as this is scheduled to be changed in nearest future

Issue #1545, opened 2026-08-04
[BUG] POST /v1/participants returns unclear error for malformed JSON


Parikalp-Bhardwaj:
## Description

`POST /v1/participants` returns a nested parser error when the request body contains malformed JSON.

## Current behavior

```json
{"error":{"message":"unexpected EOF"}}
```

## Expected behavior
```
{"error":"Invalid request body"}
```

PR #1546, opened 2026-08-04
fix: normalize malformed participant request errors

Parikalp-Bhardwaj:
## Summary

Fix the malformed JSON response for `POST /v1/participants`.

Previously, the handler returned a nested parser error:

```json
{"error":{"message":"unexpected EOF"}}
```

Related #1545

### DIFF
--- a/decentralized-api/internal/server/public/post_participant_handler.go
+++ b/decentralized-api/internal/server/public/post_participant_handler.go
@@ -15,7 +15,7 @@ func (s *Server) submitNewParticipantHandler(ctx echo.Context) error {
 
 	if err := ctx.Bind(&body); err != nil {
 		logging.Error("Failed to decode request body", types.Participants, "error", err)
-		return echo.NewHTTPError(http.StatusBadRequest, err)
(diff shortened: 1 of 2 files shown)


redstartechno:
I ran this locally (WSL, head `52fe1b294` merged onto `upgrade-v0.2.16` @ `c33c9eaa5`, which is where fixes are landing right now; it merges cleanly there).

The handler change does what it says. From the test's own `t.Logf`:

- before: `{"error":{"message":"unexpected EOF"}}`
- after: `{"error":"Invalid request body"}`

The new string form also matches the other 400s in `internal/server/public` (45 of them pass a string message; this handler was the only one passing `err`), so the shape change is a consistency fix. Still, anyone who parses `error.message` from this one endpoint will see it change.

**One gap:** `TestSubmitNewParticipantHandler_MalformedJSON` also passes on the unpatched code, because it only asserts `http.StatusBadRequest`, which the old code already returned. Asserting the body would make it guard the change, e.g.:

```go
require.JSONEq(t, `{"error":"Invalid request body"}`, rec.Body.String())
```

With that line, the test fails on the current code and passes with this PR.

PR #1544, opened 2026-08-04
fix(devshard): verify timeout votes before counting slot weight

staaason:
A malicious verifier could return an accept vote signed with its own key while claiming another validator's slot. `CollectTimeoutVotes` credited the claimed slot's weight without verifying the signature or binding the slot to the responder, so unowned weight could cross the threshold and end collection early. `applyTimeout` rejects the whole `MsgTimeoutInference` on the first bad vote, so one spoofed vote discarded the honest quorum alongside it, and the best-effort diff dropped the tx silently while still consuming the nonce and fee.

Votes are now verified and bound to the responding verifier before contributing weight, the tally dedupes by validator address to match `applyTimeout`, and `HandleTimeout` reports `timeout_not_applied` when the state machine drops the tx.

PR #1549, opened 2026-08-05
Fix/devshard timeout vote signed slot

akup:
## Summary

- Hardening on top of the gateway-only fix (https://github.com/gonka-ai/gonka/pull/1544): include `voter_slot` in signed `TimeoutVoteContent`, and verify the signature against the known responder during collection so a vote cannot be reused under a different claimed slot.
- The **minimal fix** (responder bind + count weight from the contacted host) lives in a separate PR on `ak/devshard-timeout-vote-responder-bind` and touches **only the gateway** collection path — that alone closes the reported issue.
- This PR is **defense-in-depth / protocol hardening**, not required to stop the attack if the gateway-only PR is merged. It changes the signed preimage and should ship under a new approved protocol name.

## Problem

`CollectTimeoutVotes` fans out `VerifyTimeout` to known verifiers. A byzantine verifier could return `accept=true` with its own valid signature while claiming another validator’s higher-weight `voter_slot`. Collection counted that weight, could stop early and cancel outstanding honest RPCs, then the state machine rejected the invalid vote — so an otherwise available honest timeout quorum was lost.

## Fix (this PR)

1. Add `voter_slot` to `TimeoutVoteContent`; host signs it; state machine verifies over that content.
2. Gateway verifies the signature against the contacted verifier after the responder-bind check (assumes / pairs with the gateway-only bind from `ak/devshard-timeout-vote-responder-bind`).

See also: [gateway-only PR](https://github.com/gonka-ai/gonka/pull/1544)

## Severity (issue): Medium

Authenticated liveness/griefing, not fund theft. Requires a separate byzantine group verifier (executor alone cannot block its own timeout). Early-stop is a race: the malicious reply must be counted before enough honest votes arrive. SM still rejects the spoofed vote, so no false timeout under another validator’s identity.

## Test plan

- [x] `go test ./user/ -run TestCollectTimeoutVotes_`
- [x] `go test ./user/ ./state/ ./protocol/ ./transport/ ./host/`
- [ ] Confirm release uses a new `approved_versions.name` for this signed-slot change
- [ ] Land / link the gateway-only PR on `ak/devshard-timeout-vote-responder-bind`

### DIFF
--- a/devshard/cmd/devshardctl/proxy_test.go
+++ b/devshard/cmd/devshardctl/proxy_test.go
@@ -592,6 +592,7 @@ func (c *verifierClient) VerifyTimeout(_ context.Context, inferenceID uint64, re
 		InferenceId: inferenceID,
 		Reason:      reason,
 		Accept:      true,
+		VoterSlot:   voterSlot,
 	}
 	data, err := proto.Marshal(content)
 	if err != nil {
--- a/devshard/internal/testutil/testutil.go
+++ b/devshard/internal/testutil/testutil.go
@@ -118,19 +118,21 @@ func SignExecutorReceipt(t *testing.T, signer signing.Signer, escrowID string, i
 	return sig
 }
 
-func SignTimeoutVote(t *testing.T, signer signing.Signer, escrowID string, inferenceID uint64, reason types.TimeoutReason, accept bool) *types.TimeoutVote {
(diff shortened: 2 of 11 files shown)


a-kuprin:
Should go only after https://github.com/gonka-ai/gonka/pull/1569 goes to main with gateway-v4. Also needs rebasing and remerging

PR #1552, opened 2026-08-06
fix(poc): gate PoC v2 messages by participant permission

staaason:
## Problem
The three PoC v2 transaction types are declared `NoPermission`, so any chain account — with no PoC history and no participant record — can write entries into the PoC stage prefixes:

- `MsgSubmitPocValidationsV2`
- `MsgPoCV2StoreCommit`
- `MsgMLNodeWeightDistribution`

The V1 equivalent, `MsgSubmitPocBatch`, has required `ParticipantPermission` since it was introduced (`permissions.go:113`). The V2 messages dropped that check during the V1→V2 migration.

## Fix

1. `permissions.go` and the three `CheckPermission` call sites move from `NoPermission` to `ParticipantPermission`. The check is already the first statement in each handler, before any storage read or gas charge.
2. A node joining the network is registered but not yet active in the current epoch, and must be able to run its first PoC. `ActiveParticipantPermission` would break onboarding.
3. `PocPeriodValidationDecorator` gains `checkPocMessageSender`, rejecting PoC duty messages from non-participants. The handler check runs only in `DeliverTx`, so on its own it stops the storage write but still lets a non-participant fill blocks for free. This decorator runs only under `ctx.IsCheckTx()`, so it is a mempool admission filter with no consensus impact.
4. It follows messages into authz `MsgExec`, which matters in practice: when a node runs a separate signer key from its main account, `tx_manager.go:828-836` wraps *every* PoC message in `MsgExec` with `grantee = warm key`, `Creator = participant`. Both gates read the granter.

### DIFF
--- a/inference-chain/app/ante_poc_period.go
+++ b/inference-chain/app/ante_poc_period.go
@@ -128,11 +128,37 @@ func (ppd PocPeriodValidationDecorator) checkPocMessageTooLate(ctx sdk.Context, 
 	return nil
 }
 
+func (ppd PocPeriodValidationDecorator) checkPocMessageSender(ctx sdk.Context, msg sdk.Msg) error {
+	if ppd.inferenceKeeper == nil {
+		return nil
+	}
+
+	signers, ok := msg.(inferencemodulekeeper.HasSigners)
+	if !ok {
+		return nil
+	}
+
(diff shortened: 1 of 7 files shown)


staaason:

PR #1556, opened 2026-08-06
fix(queryapi): skip validator-set lookup for zero creation height in epoch participants

redstartechno:

`GET /v1/epochs/{N}/participants` fails outright when the stored `ActiveParticipants` blob has `CreatedAtBlockHeight == 0`: `getEpochParticipants` passes that height straight to `GetValidatorSetByHeight`, CometBFT rejects any height ≤ 0 (`height must be greater than 0, but got 0`), and the error propagates to the HTTP layer — even though the participants data itself was already fetched and unmarshaled successfully.

This is the code-level fatal path identified while investigating #983. **It deliberately does not claim to fix that issue**: as established in the issue thread, the live symptom observed on public nodes today is a 404 for past epochs (reachability of the zero-height path is unresolved and looks operational rather than code-level). This PR closes the latent bug so the endpoint degrades gracefully if/when a zero-height blob is served, matching how the same function already handles other partial-data conditions.

- Reproducible in a regression test against unmodified code: a stubbed Comet service returning `ActiveParticipants{CreatedAtBlockHeight: 0}` plus a CometBFT-faithful `GetValidatorSetByHeight` error makes the handler return the error instead of the data (output below).
- Zero is a legitimately decodable value: blobs written before the `created_at_block_height` field was added to `activeparticipants.proto` decode with the zero value (the only production writer, `x/inference` EndBlock, always writes a positive height — so zero specifically indicates a pre-field blob).
- The handling is asymmetric within one function: `GetBlockByHeight` failure a few lines earlier is non-fatal (logged, degraded response), while the validator-set call at the same zero height kills the whole response.

In `getEpochParticipants`, when `CreatedAtBlockHeight <= 0`, skip the `GetValidatorSetByHeight` call, log a warning, and return the response with an empty `validators` array. Real call failures at valid heights are still fatal, unchanged. One nil-guard added where the validator response is consumed (`validatorsToRawJSON(nil)` already yields `[]`, never `null`). The `validators` schema comment in `openapi.yaml` now documents the degraded case, mirroring the existing `block` field annotation.

- **Contract:** consumers see `"validators": []` instead of an error. An empty array is already a reachable value today (a valid-height valset response with zero validators), the field is a required non-nullable array in the schema, and the `block` field already established the degraded-response precedent in this endpoint — so no consumer-visible type change.
- **Proof path untouched:** the diff starts after the ABCI proof query; proof semantics at height 0 (= "latest") are pre-existing and out of scope here.
- Height > 0 behavior is byte-identical to before (same call, same error handling).

Fail-first regression test, real outputs below: the new test fails on unmodified `epoch.go` with the exact CometBFT error surfacing, and passes with the fix (endpoint returns 200, participants present, `validators` empty, and the doomed RPC is never issued).

- `common/queryapi/epoch.go` — zero-height guard + nil-safe validator consumption (+18/−7)
- `common/queryapi/openapi.yaml` — one-line schema comment for the degraded `validators` case
- `common/queryapi/tests/epoch_participants_zero_height_test.go` — new regression test

## Testing & evidence

Environment: Windows native, Go 1.26.4 (`common/go.mod` pins go 1.25.9), `CGO_ENABLED=1 go test -tags nolink_libwasmvm ./queryapi/...`.

- New test `TestEpochParticipantsZeroHeightSkipsValidatorLookup`: proven to fail on unmodified code (below), passes with the fix.
- Full `./queryapi/...` suite: all packages pass (incl. golden and compatibility tests). `go vet` clean.
- `gofmt`: new test file clean; `epoch.go` was already flagged on `main` before this change (CRLF checkout artifact), not introduced here.

### Before

```
ERROR Failed to get validators subsystem=Participants error="rpc error: code = InvalidArgument desc = height must be greater than 0, but got 0"
--- FAIL: TestEpochParticipantsZeroHeightSkipsValidatorLookup
    code=400, message=height must be greater than 0, but got 0
```

### After

```
WARN Skipping validator set lookup for zero/negative height subsystem=Participants height=0
--- PASS: TestEpochParticipantsZeroHeightSkipsValidatorLookup
ok      common/queryapi/tests
```

### DIFF
--- a/common/queryapi/epoch.go
+++ b/common/queryapi/epoch.go
@@ -189,12 +189,19 @@ func (h *Handlers) getEpochParticipants(ctx context.Context, epoch uint64) (*gen
 	}
 
 	// Validators at the creation height.
-	valsResp, err := h.chain.CometServiceClient().GetValidatorSetByHeight(ctx, &cmtservice.GetValidatorSetByHeightRequest{
-		Height: activeParticipants.CreatedAtBlockHeight,
-	})
-	if err != nil {
-		logging.Error("Failed to get validators", inferencetypes.Participants, "error", err)
-		return nil, err
+	var valsResp *cmtservice.GetValidatorSetByHeightResponse
+	if activeParticipants.CreatedAtBlockHeight <= 0 {
+		// Non-fatal: CometBFT rejects height <= 0, so skip the call rather than fail the request.
+		logging.Warn("Skipping validator set lookup for zero/negative height", inferencetypes.Participants,
+			"height", activeParticipants.CreatedAtBlockHeight)
+	} else {
+		valsResp, err = h.chain.CometServiceClient().GetValidatorSetByHeight(ctx, &cmtservice.GetValidatorSetByHeightRequest{
+			Height: activeParticipants.CreatedAtBlockHeight,
+		})
+		if err != nil {
+			logging.Error("Failed to get validators", inferencetypes.Participants, "error", err)
+			return nil, err
+		}
 	}
 
 	// Derive participant addresses from validator keys.
@@ -214,7 +221,11 @@ func (h *Handlers) getEpochParticipants(ctx context.Context, epoch uint64) (*gen
 		return nil, err
 	}
 
-	validators, err := validatorsToRawJSON(valsResp.Validators)
+	var vals []*cmtservice.Validator
+	if valsResp != nil {
+		vals = valsResp.Validators
+	}
+	validators, err := validatorsToRawJSON(vals)
 	if err != nil {
 		logging.Error("Failed to encode validators", inferencetypes.Participants, "error", err)
 		return nil, err
--- a/common/queryapi/openapi.yaml
+++ b/common/queryapi/openapi.yaml
@@ -557,7 +557,7 @@ components:
         # cmtservice.ProofOps — absent when not requested
         proof_ops:
           $ref: '#/components/schemas/RawProtoJson'
-        # cmtservice.Validator
+        # cmtservice.Validator — empty when the creation height is unavailable
         validators:
           type: array
           items:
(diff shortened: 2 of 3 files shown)


x0152 (approved):

redstartechno:
Re-verified this against the current `upgrade-v0.2.16` head (`bc4e4912f`), since the file moved under it:

**Still needed.** The unguarded call is live at `common/queryapi/epoch.go:208` — `GetValidatorSetByHeight(Height: activeParticipants.CreatedAtBlockHeight)` with a fatal `return nil, err`, exactly as when this PR was opened. The approval and the 8/8 green run still describe the right change.

**But the handler grew two more uses of the same height** in #1752 (`5e729d61b`, merged 09-15), and this PR predates them:

- `:150-160` — the proof re-query uses `proofHeight := activeParticipants.CreatedAtBlockHeight` and passes it to `ABCIQuery(..., Prove: true)`. At height 0 CometBFT serves the *latest* state, so the returned proof is anchored to the latest root rather than the creation height.
- `:171` — `heightP1 := activeParticipants.CreatedAtBlockHeight + 1` becomes `1`, so `GetBlockByHeight(1)` succeeds and block 1's app hash is what the proof gets verified against at `:194`.

The verification is non-fatal, so this doesn't 500 — it does something quieter: the response carries proof ops anchored at the latest height next to block 1, and `VerifyUsingProofRt` / `VerifyUsingMerkleProof` log a failure every time. A client that checks the bundle it was handed gets a mismatch, and the `block` field is simply the wrong block.

I'd rather not widen an already-approved PR, so unless you'd prefer it all in one, I'll send that as a separate PR on the same base: skip the proof re-query and the block+1 fetch when the creation height is unavailable, and return the response without a proof bundle instead of with a misleading one. Same "skip, don't fail" shape as this PR. Happy to fold it in here instead if that's easier to review — just say which.

PR #1559, opened 2026-08-07
fix(devshard): reject bind against settled escrow

staaason:
## Problem

`EscrowInfo` dropped the on-chain `settled` flag at the bridge boundary, so devshardd could lazily bind a new session against an escrow that was already settled - work the host can never be paid for, since the chain rejects a second settle (`VerifyDevshardSettlement`, `devshard_settlement.go:86`).

## Changes

**Plumb the flag** — `EscrowInfo.Settled`, mapped in both bridge implementations (`bridge/grpc.go`, `cmd/devshardd/bridge/chain.go`).

**Gate the bind path** (`session/manager.go`)
- `create()` rejects settled escrows — the choke point for all bind entry points
- `BindOwnerChat` rejects early, after the owner check so escrow state doesn't leak to non-owners
- `ErrEscrowSettled` added to `isPermanentResolutionFailure` (10-min cache)
- `HandleSettlementFinalized` negative-caches before `MarkSettled`, covering hosts with no local row
- `storeSessionIfAbsent` re-checks the settled tombstone **under `sessionsMutex`**, closing a TOCTOU where a settlement landing mid-`create()` was erased by the unconditional `resolutionFailures` delete. On rejection it closes the server and marks the row settled so `RecoverSessions` can't resurrect it. Matches only `ErrEscrowSettled`, so a concurrent transient failure can't block a legitimate create.

**Close the stale-cache path** — `EscrowCacheInfo.Settled` (rides the existing `escrow_json` blob, no migration); the warm sink drops rather than caches a settled escrow; a settled cache row is refused instead of serving stale metadata during a chain outage.

**Gateway self-healing** — a bare 409 matched none of the existing terminal-escrow detection, so the gateway would have kept a dead devshard active, retrying every host in the group. Added `X-Devshard-Error: escrow_settled`, captured it in
`UpstreamStatusError` (code first, body match retained for old hosts), wired it into `checkEscrowMissing`, and taught `EscrowChecker` to deactivate when the chain itself reports `Settled` — previously it logged "false escrow-not-found" and kept the runtime alive, since a settled escrow *is* still found on chain.

### DIFF
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -280,6 +280,9 @@ func buildRuntime(cfg RuntimeConfig, deps runtimeBuildDeps) (*devshardRuntime, e
 	if err != nil {
 		return nil, fmt.Errorf("runtime %s: get escrow: %w", cfg.ID, err)
 	}
+	if escrow.Settled {
+		return nil, fmt.Errorf("runtime %s: %w", cfg.ID, bridge.ErrEscrowSettled)
+	}
 	model := resolveRuntimeModel(cfg.Model, escrow.ModelID, deps.defaultModel, cfg.ID)
 	routePrefix := resolveRuntimeRoutePrefix(cfg.RoutePrefix)
 	session, sm, err := user.NewHTTPSession(user.HTTPSessionConfig{
@@ -3941,7 +3944,7 @@ func (g *Gateway) attachEscrowChecker(rt *devshardRuntime) {
 	if g.escrowChecker != nil {
 		rt.proxy.redundancy.onEscrowMissing = func() {
 			go g.escrowChecker.TriggerCheck(escrowID, func(reason string) {
-				g.deactivateDevshardByID(escrowID)
+				g.deactivateDevshardByIDWithReason(escrowID, reason)
 				// Escrow is terminal on chain (missing or settled) -- nothing to settle.
 				g.retireRuntime(escrowID, reason)
 			})
@@ -4104,6 +4107,15 @@ func (g *Gateway) retireRotatedDevshard(ctx context.Context, id, reason string, 
 	}
 	log.Printf("escrow_rotation_settling escrow=%s reason=%q", id, reason)
 	if _, err := gatewaySettleDevshardOnChain(g, ctx, id, adminSettleEscrowRequest{}); err != nil {
+		// Already settled on chain: there is nothing left to broadcast, so
+		// retire instead of failing the rotation forever.
+		if errors.Is(err, bridge.ErrEscrowSettled) {
+			log.Printf("escrow_rotation_already_settled escrow=%s reason=%q", id, reason)
+			g.clearSettlementPending(id)
+			g.deactivateDevshardByIDWithReason(id, "escrow settled on chain")
+			g.retireRuntime(id, reason)
+			return true, nil
+		}
 		return false, err
 	}
 	g.retireRuntime(id, reason)
@@ -4222,6 +4234,15 @@ func (g *Gateway) scheduleAutoSettlement(id, reason string) {
 				g.clearSettlementPending(id)
 				log.Printf("auto_settle_confirmed escrow=%s reason=%s tx_hash=%s settler=%s",
 					id, reason, result.TxHash, result.Settler)
+				g.retireRuntime(id, reason)
+				return
+			}
+			if errors.Is(err, bridge.ErrEscrowSettled) {
+				// Terminal: the chain already recorded settlement, so retrying
+				// can only fail. Clear the pending flag so restarts stop
+				// reconciling an escrow that is already done.
+				g.clearSettlementPending(id)
+				log.Printf("auto_settle_already_settled escrow=%s reason=%s", id, reason)
 				g.retireRuntime(id, reason)
 				return
 			}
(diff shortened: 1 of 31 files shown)


devshard/cmd/devshardd/session/manager.go
x0152, line 189: Hi @staaason 
Edge case: if local session is already settled, SessionServerExisting can return ErrSessionNotActive and this path returns 500

devshard/cmd/devshardctl/gateway.go
```diff
@@ -3940,10 +3940,10 @@ func (g *Gateway) attachEscrowChecker(rt *devshardRuntime) {
 	modelID := rt.model
 	if g.escrowChecker != nil {
 		rt.proxy.redundancy.onEscrowMissing = func() {
-			go g.escrowChecker.TriggerCheck(escrowID, func() {
+			go g.escrowChecker.TriggerCheck(escrowID, func(reason string) {
 				g.deactivateDevshardByID(escrowID)
```
a-kuprin: g.deactivateDevshardByIDWithReason(escrowID, reason)

devshard/cmd/devshardctl/gateway.go
a-kuprin, line 1: Let's add early interrupt at gateway session init and we will not waste resources before first inference like this:

```go
        ...
        escrow, err := br.GetEscrow(cfg.ID)
	if err != nil {
		return nil, fmt.Errorf("runtime %s: get escrow: %w", cfg.ID, err)
	}
---> here we add
	// A settled escrow is as terminal as a missing one: it can never fund
	// another inference, so refuse the runtime instead of waiting for a host
	// to 409 the first request.
	if escrow.Settled {
		return nil, fmt.Errorf("runtime %s: %w", cfg.ID, bridge.ErrEscrowSettled)
	}
<---
	model := resolveRuntimeModel(cfg.Model, escrow.ModelID, deps.defaultModel, cfg.ID)
	routePrefix := resolveRuntimeRoutePrefix(cfg.RoutePrefix)
	session, sm, err := user.NewHTTPSession(user.HTTPSessionConfig{
		PrivateKeyHex:    keyHex,
		EscrowID:         cfg.ID,
		Bridge:           br,
		StoragePath:      cfg.StoragePath,
		RoutePrefix:      routePrefix,
		RequestAdmission: sharedParticipantRequestLimiter,
		Escrow:           escrow,
	})
        ...
```

And handle the error at devshardctl/main.go

```go
...
			brokenLocalState := errors.Is(res.err, user.ErrLocalStateUnrecoverable)
			if brokenLocalState || errors.Is(res.err, bridge.ErrEscrowNotFound) ||
				errors.Is(res.err, bridge.ErrEscrowSettled) || errors.Is(res.err, errRuntimePrivateKeyMissing) {
				reason := "runtime could not be loaded"
				if brokenLocalState {
					reason = "local state unrecoverable"
				} else if errors.Is(res.err, bridge.ErrEscrowNotFound) {
					reason = "escrow missing on chain"
				} else if errors.Is(res.err, bridge.ErrEscrowSettled) {
					reason = "escrow settled on chain"
				} else if errors.Is(res.err, errRuntimePrivateKeyMissing) {
					reason = "private key missing"
				}
...
```

a-kuprin:
### Host-events settlement should finalize the session

`OnEscrowSettled` only drops the warm cache. If the host misses the chain WS event, the session stays live/`active`, and `BindOwnerChat` serves it via `SessionServerExisting` without ever hitting the new `Settled` check. `RecoverSessions` then reloads it on restart.

Wire the long-poll through the same `HandleSettlementFinalized` already used for chain events (including when `WarmEscrow` sees `Settled`). Keep both channels: host-events is optional; Comet is the always-on path.

**p.s.:**
This path duplication also should be fixed, I will look at, to reduplicate like it was done at https://github.com/gonka-ai/gonka/pull/1530 for OnEpoch events

devshard/cmd/devshardd/hostevents_sink.go
```diff
@@ -34,6 +34,10 @@ func (s *escrowWarmSink) WarmEscrow(escrowID string) error {
 	if err != nil {
 		return fmt.Errorf("warm escrow %s: %w", escrowID, err)
 	}
+	if info.Settled {
```
a-kuprin, line 38: `WarmEscrow` is the only writer of `escrow_cache`, and this PR already skips caching when settled (`hostevents_sink.go` → `OnEscrowSettled` deletes the row). So `Settled: true` never lands in the cache, and the refuse branch in `CachingEscrowBridge.GetEscrow` (`caching.go`) can never run.

devshard/storage/interface.go
```diff
@@ -122,6 +122,7 @@ type EscrowCacheInfo struct {
 	ValidationRate            uint32   `json:"validation_rate"`
 	VoteThresholdFactor       uint32   `json:"vote_threshold_factor"`
 	EpochID                   uint64   `json:"epoch_id"`
+	Settled                   bool     `json:"settled,omitempty"`
```
a-kuprin: We skip caching when settled (`hostevents_sink.go` → `OnEscrowSettled` deletes the row). So we don't need this at cache at all

a-kuprin:
@staaason please retarget it to [devshard-0.2.15-v5](https://github.com/gonka-ai/gonka/tree/devshard-0.2.15-v5)
It's good to go at v5 devshard release that is separately released from main-chain 0.2.16 upgrade

staaason:
@a-kuprin retargeted to [devshard-0.2.15-v5](https://github.com/gonka-ai/gonka/tree/devshard-0.2.15-v5)

PR #1563, opened 2026-08-07
Silence thinking in the chat template

qdanik:
## Problem

Kimi-K2.6 requests with small `max_tokens` return `finish_reason=length` with **empty `content`** — the whole budget goes to hidden reasoning. The validator already forces `thinking_token_budget = 0` below 256 to prevent exactly this, and in production it worked on some hosts and not others.

## Root cause

`thinking_token_budget` is a declared top-level vLLM field and `0` there means "close the reasoning block now" — so both our placement and our value were correct. But it is implemented as a **logits processor**, and vLLM discards every built-in processor when the node runs speculative decoding:

```python
if vllm_config.speculative_config:
    logger.warning("min_p and logit_bias parameters won't work with speculative decoding.")
    return LogitsProcessors([MinTokensLogitsProcessor(...)])
```

`ThinkingTokenBudgetLogitsProcessor` is dropped and **not named in the warning**. Upstream confirms it end to end: *"Without MTP → thinking budget enforced. With MTP → thinking budget NOT enforced."* That is the entire difference between the two hosts above.

## Fix

Below the 256 cutoff the validator now also sets `chat_template_kwargs: {"thinking": false}`, which goes through the chat template rather than the sampler:

```jinja
{%- if thinking is defined and thinking is false -%}
<think></think>
```

No logits processor can drop it, and vLLM builds its reasoning parser from the same kwargs — it falls back to identity, so the whole answer lands in `content` with no stray `</think>`. The budget is still sent: it stays the right governor wherever the processor is live.

A client that put its own `thinking` into `chat_template_kwargs` keeps it — the validator fills the field, never overwrites it.

## Verified

- Read the pinned engine's source (`v0.20.0`) and the model's own `chat_template.jinja` at the mainnet-pinned commit, not release notes.
- The template reads a boolean `thinking` and reads **neither** `thinking_token_budget` nor `enable_thinking` — greped across six Kimi repos. Moving the budget into kwargs would have been a pure no-op, so it stays top-level.
- Three new cases: template flag set below the cutoff, absent above it, client's own value survives.
- Three mutants killed: flag never set, flag set for every budget, flag set to `true`.
- `go test ./cmd/devshardctl/...` green.

## Scope

`Kimi-K2-Thinking` has no `thinking` variable in its template at all — nothing disables its reasoning, and this change is a no-op there. Only the K2.6 route is affected.

Docs updated in `docs/chat-api/`: the parameter row in `kimi-k2.6.md`, the think-burn entry in `troubleshooting.md`, and four new primary sources in `references.md`.


a-kuprin:
`mirrorThinkingToTemplateKwargs` only sets `thinking` if the key is absent:

```78:87:devshard/cmd/devshardctl/paramvalidators/thinking.go
func mirrorThinkingToTemplateKwargs(document map[string]any, enabled bool) error {
	chatTemplateKwargs, err := getOrCreateChatTemplateKwargs(document)
	// ...
	if _, exists := chatTemplateKwargs["thinking"]; exists {
		return nil
	}
	chatTemplateKwargs["thinking"] = enabled
	return nil
}
```

Order of stages:

1. **PreValidation** — `ThinkingValidator` may already set `chat_template_kwargs.thinking = true` from `thinking: {type: "enabled"|"adaptive"|"auto"}`.
2. **PostLimits** — budget validator force-zeros `thinking_token_budget`, then calls `mirrorThinkingToTemplateKwargs(..., false)` → **no-op** if the key exists.

So for:

```json
{"max_tokens": 100, "thinking": {"type": "enabled"}}
```

or

```json
{"max_tokens": 100, "chat_template_kwargs": {"thinking": true}}
```

you still get `thinking_token_budget: 0` **and** `thinking: true` in the template. Under MTP, the budget processor is dead → thinking stays on → same empty-content burn this PR claims to fix.

Budget force-zero **does** override a client `thinking_token_budget`. Template force-off does **not** override a client `thinking: true`. That is inconsistent with “bypass thinking entirely,” and the new test *“a client that asked to think keeps its own template answer”* freezes the weaker behavior.

This PR only helps bodies that omit `thinking` entirely.

**Recommendation:** on the force-zero path, **overwrite** `chat_template_kwargs.thinking = false` (same force semantics as budget). Keep fill-only for the normal non-force path. Drop or rewrite that “keeps its own answer” test for `max_tokens < 256`.

a-kuprin (approved):

PR #1558, opened 2026-08-07
Ak/devshard observability e2e

a-kuprin:
# Merge order

The PR was branched from https://github.com/gonka-ai/gonka/pull/1547, so it should be merged first

## Summary

This PR lands gateway nonce accounting plus the observability stack that makes those dispositions *actionable*: every request can be reconstructed end to end from Grafana / Tempo / Loki, and failures or classification outcomes are reachable from a single `trace_id` / `request_id` instead of being guessed from counters alone.

Previously, accounting e2e proved that every consumed nonce lands in exactly one disposition — but that only asserted *counts*. When something went wrong, an operator (or a developer) still could not answer “which hop failed?”, “was it the ML node or the host?”, or “why did this nonce become a ghost / unfinished / unused?” without digging through uncorrelated logs.

This change binds **logging, tracing, and metrics** onto the same identity for a client request:

- Gateway (`devshardctl`), host (`versiond` / `devshardd`), and node manager (`mock-dapi` / NodeManager gRPC) share one `trace_id`.
- The client-facing `X-Request-Id` rides the same path (HTTP + gRPC metadata), so Loki lines and spans join on `request_id` as well.
- Disposition outcomes show up both as Prometheus labels *and* as span attributes with identical string values, so Grafana data links from a counter to the failing request actually work.
- On ML-node failures, payload / quarantine lines attach to that same trace, so the bytes (or hashes) behind a classification are one query away.

Net effect: finding the *reason* for a failure or a classification is a Tempo / Loki / Prometheus lookup, not a multi-service log archaeology session.

### What landed

- **Accounting:** counters-only gateway epoch accounting that stays off the settlement/sequencer hot path, with honest disposition outcomes. Focused e2e covers happy path, ghosts (throttled / capability / host-unavailable), timeouts, overscheduling, unfinished paths, and HostStats cross-checks.
- **Shared OTel + collectors:** common observability bootstrap; `tempo-alloy` as the testenv default; Jaeger/Promtail kept green as a regression profile.
- **Request shape in Tempo:** gateway `attempt` spans (including multi-host fan-out under redundancy / shadow quarantine); late `nonce.disposition` spans from the sink; host `mlnode.acquire` / `release` (+ ML HTTP) spans.
- **Payload capture (T4a):** ML-failure and quarantine payload lines gated by `DEVSHARD_LOG_PAYLOADS*`, kept off limiter mutexes and mid-flight attempt state so capture cannot distort racing or timing.
[...]

PR #1574, opened 2026-08-10
`gateway`: Same-nonce stream reconnect and host ML drain for streaming HA

a-kuprin:
## Summary
Implements mid-stream same-nonce reconnect and host↔ML lifetime decoupling on top of `gateway-v4`, addressing [#1466](https://github.com/gonka-ai/gonka/issues/1466) §2–§3. Operator/design overview: [gateway-streaming-ha-overview.md](https://github.com/gonka-ai/gonka/blob/ak/stream-same-nonce-reconnect/devshard/docs/gateway-streaming-ha-overview.md).

Today a gateway↔host drop mid-SSE either opens a new nonce (regenerating work) or fails the request after content was already delivered. After accept, a dead client/gateway writer can also cancel host→ML, so the executor may never land `MsgFinishInference`. This PR keeps ML running under a host-owned context after accept, and lets `devshardctl` resend the **same** nonce with a resume cursor so the client continues from the break instead of regenerating—when reconnect is enabled and the session protocol is ≥ `v5` (dormant on ≤v4; default-off).

```mermaid
flowchart TD
  A[Winner streaming to client] --> B{Gateway↔host drop}
  B --> C{Reconnect enabled<br/>AND protocol ≥ v5?}
  C -->|no| D[Escalate new nonce<br/>or winner_failed_after_content]
  C -->|yes| E[Same-nonce reconnect ladder<br/>budget 1s / max 2 tries]
  E --> F{Live attach or<br/>storage replay?}
  F -->|resume OK| G[Skip delivered prefix<br/>continue to client]
  F -->|budget expired| H[Start secondary nonce<br/>keep ladder racing]
  H --> I{Prefix already<br/>delivered?}
  I -->|yes| J[Winner continuity:<br/>only A reaches client]
  I -->|no| K[First content crowns]
  E -->|ladder exhausted + prefix| L[winner_failed_after_content]
```

**Host (§2):** after receipt/accept, inference runs to completion (or controlled failure) even if the outbound stream dies; stream write failure is delivery loss, not work cancellation. Mid-flight generations are kept in a disk-backed live log with a fixed RAM ring so reconnect can attach live or fall back to the completed body in storage.

**Gateway (§3):** on disconnect/retry, resend the same nonce + payload/params with `(deliveredEvents, deliveredPartial)`. Host rejects mismatched prompt/params, skips re-execute if already running/done, and returns receipt plus live or cached stream. Once any prefix reached the client, a secondary cannot splice a different generation onto it. Reconnect eligibility requires bytes that actually hit the client wire—not merely classified upstream content after the client left. Optional stream-reset failover (default off) can clear the crown after the ladder is exhausted so a hedge may take over from scratch.

Also includes streamed-usage / well-formed replay prerequisites, cheaper streamed id-rewrite, a gateway SSE event size cap, unit coverage, and `make citest-attempt-reconnect`.

## Deferred (follow-up PRs)
- **Richer reconnect observability:** Prometheus/OTel for reconnect ladder runs, live-log window/spool/lag, hop timestamps along gateway↔host↔ML, and dashboards; plus periodic persistence of soft signals (reconnect blips, rolling stream bytes/sec) into the gateway store so they survive restarts. Today blips are recorded observe-only and do not affect routing. (after https://github.com/gonka-ai/gonka/pull/1568 and https://github.com/gonka-ai/gonka/pull/1558 lands)
- **Latency observability:** add to stream data when `devshardd` got chunk from `MLNode` and when it responded to `gateway`, so gateway can add latency metrics
- **Always-stream upstream completion:** force host→ML to always use a streamed completion regardless of the client `stream` flag, aggregate to a single JSON response when the client asked for non-streaming, and finish the remaining gateway normalization / intent work that makes streaming the universal upstream shape for TTFT and sustained bytes/sec on every chat request.
- **Cross-instance host HA ([#1466](https://github.com/gonka-ai/gonka/issues/1466) §4):** if `devshardd` reboots or fails over mid-generation, another instance cannot yet reattach to an in-flight ML job or resume from durable per-instance live state; that needs shared payload/body durability and ML keep-alive/reattach.

## Test plan
- [ ] `go test` for `devshard/host`, `devshard/transport`, `devshard/cmd/devshardctl` reconnect suites, `devshard/cmd/devshardd/inference`
- [ ] `make citest-attempt-reconnect` (v2 refuses reconnect; v5 resumes after mid-stream detach)
- [ ] Manual: enable attempt-reconnect on a v5 session, kill gateway↔host mid-SSE, confirm silent resume from break without a new nonce
- [ ] Manual: client disconnect after accept — host still finishes and publishes proof
- [ ] Confirm default-off / ≤v4 sessions keep today’s escalation path


redstartechno:
Thanks for the detailed write-up — the reconnect ladder and the host-owned execution lifetime cover §2/§3 far more thoroughly than the slice I had sketched, and `gateway-streaming-ha-overview.md` made the design easy to follow.

On the deferred list: **Step 5f (trim / lag / spool metrics)** looks self-contained and I don't see it claimed — #1575 covers Step 5g (hop timestamps), and Steps 6 and 8 read as gated on the gateway store work (#1568) and the dashboards. **I'd like to take 5f**, scoped to the host-side gauges in §6.2 of the overview: RAM window bytes, spool bytes/lag, `spool_unavailable`, `subscriber_lagged`, `reader_lag_bytes`, and reads served from spool vs RAM.

Since it lives in files this PR introduces (`devshard/host/livestream.go`, `devshard/host/spool.go`), the timing and base are yours to call:

1. Would you prefer I branch off `ak/stream-same-nonce-reconnect` now and open against `gateway-v4` once this merges, or wait until it has landed?
2. Any naming/label convention you want the exports to follow — I would otherwise mirror `devshard/observability/metrics_lifecycle.go` — and should they sit behind the same default-off gate as the reconnect path, or always be recorded?

If you would rather keep this one in-house, or it is already covered by the observability e2e work, just say so and I will drop it — no work started yet.

a-kuprin:
@redstartechno 
I think 5f is a tiny work, and it should be done only after whole bunch of gateway changes are reviewed and accepted.
(https://github.com/gonka-ai/gonka/pull/1558, https://github.com/gonka-ai/gonka/pull/1568, https://github.com/gonka-ai/gonka/pull/1575, https://github.com/gonka-ai/gonka/pull/1574, https://github.com/gonka-ai/gonka/pull/1579)

I don't see a lot of meaning to force 5f

a-kuprin:
Will not go to gateway-4 release, to make it more finite and testable. Should land after [1581](https://github.com/gonka-ai/gonka/pull/1581)

redstartechno:
Understood — I will not push on 5f. It makes sense to let the gateway stack (#1558, #1568, #1575, this one, #1579) be reviewed and land first, and I saw the note that this now goes after #1581 rather than into the gateway-4 release.

No work started on my side, and I will leave it alone. If the metrics are still open once that stack has settled and you would like them picked up, ping me here and I will take it then.

devshard/testenv/Makefile
```diff
@@ -77,7 +84,9 @@ list-citest-targets: ## Print runnable citest suite targets (one per line) for t
 		| sort -u
 .PHONY: obs-up obs-down
 # versiond-0 produces devshard-versiond:latest shared by all versiond-* hosts.
-CITEST_BUILD_SERVICES := mock-chain mock-dapi mock-openai devshardctl versiond-router versiond-0
+# Root docker-compose.yml names mock-openai as mock-openai-0 (per-host); citest
+# gencompose stacks reuse the same image tags (devshard-mock-openai:latest, …).
+CITEST_BUILD_SERVICES := mock-chain mock-dapi mock-openai-0 devshardctl versiond-router versiond-0
```
snevolin, line 89: There's no such service as mock-openai-0 here

PR #1576, opened 2026-08-11
fix(ante): require signer authorization for fee-exempt network duties

### LINKED ISSUE #1539: Hardening: network-duty fee-bypass admits zero-fee txs from non-participants (no signer authorization at the ante layer)
## Summary

`NetworkDutyFeeBypassDecorator` waives fees + clears min-gas-price for ~12 "network duty" message types, and the exemption is decided **purely by Go message type** (`isExemptMessageType` in `inference-chain/app/ante_fee.go`) — there is no check at the ante/`CheckTx` layer that the signer is an active participant / authorized host. So **any funded account can submit structurally-valid, zero-fee duty-typed transactions.** They pass `CheckTx`, enter the mempool, occupy block space, incur secp256k1 signature verification on every validator, and are rejected only later in `DeliverTx` (e.g. `participant is not active`).

The real authorization for these types (allowlist / participant / dedup / deadline) lives in the message handlers, which run in `DeliverTx` — i.e. **after** mempool admission and block inclusion. `CheckTx` runs the ante only, so unauthorized duty-typed txs are admitted regardless.

## Where

- `inference-chain/app/ante_fee.go` — `NetworkDutyFeeBypassDecorator` / `isExemptMessageType` / `GonkaFeeChecker`. On `main`, the exempt set is 12 types: `MsgSubmitPocBatch`, `MsgSubmitPocValidationsV2`, `MsgMLNodeWeightDistribution`, `MsgSubmitSeed`, `MsgSubmitHardwareDiff`, `MsgClaimRewards`, `MsgSettleDevshardEscrow`, and the 5 BLS DKG types.
- `inference-chain/app/ante_poc_period.go` — `PocPeriodValidationDecorator` gates only **timing** (`checkPocMessageTooLate`) and only for the 4 PoC types; it does not check the signer. The other 8 exempt types have no ante-layer gate at all.

## Impact (bounded — hardening, not a critical exploit)

An attacker imposes **free, unauthenticated load**: zero-fee inclusion consuming validator sig-verify + gossip + mempool + block bytes, with no economic cost (balance never drains; one-time dust to fund accounts).

Honest severity limits, verified against live mainnet consensus params (`max_gas = -1`, `max_bytes = 22 MB`):
- **Not** a practical throughput DoS: with unlimited block gas, saturating a block is bounded only by the ~22 MB byte limit, which (given `CheckTx` enforces the account sequence → one pending tx per account) would require on the order of tens of thousands of concurrently-funded accounts.
- The `Priority: 10_000_000` boost is **inert** for block ordering: the app uses the default `NoOpMempool`, so `PrepareProposal` builds blocks in CometBFT FIFO order and never consults the priority. No priority-based censorship of paid traffic.

So this is a resource-abuse / free-spam hardening gap, not a network-halting exploit.

## Reproduction

A zero-fee `MsgClaimRewards` signed by a non-participant account: `CheckTx` returns code 0 (admitted), the tx is included in a block, and only then fails in `DeliverTx` with `participant is not active` — having consumed block gas at zero fee.

## Suggested fix

Add an ante-layer signer-authorization check for the exempt types (reject duty-typed txs from non-participants / non-allowlisted signers at `CheckTx`), and/or withhold the fee-bypass + priority until the signer is authorized. This mirrors the participant check that already protects `MsgValidation`, and closes the gap where the only authorization runs in `DeliverTx` after admission.

Ryanchen911:
Fixes #1539

## The problem, confirmed

`NetworkDutyFeeBypassDecorator` sits at ante index 11. `SigVerificationDecorator` sits at index 19. The bypass decides the fee waiver from `isExemptMessageType`, which switches on the Go type and nothing else — so the waiver is granted eight decorators before anyone has proven the transaction is even signed by the account it names.

The real authorization for those ~12 types lives in the message handlers (`MessagePermissions`: `ParticipantPermission`, `ActiveParticipantPermission`, `EscrowAllowListPermission`, or an in-handler scan of the epoch's BLS participant list). SDK 0.53 runs handlers only in DeliverTx. So the sequence for a funded non-participant was: zero-fee duty tx admitted to the mempool → gossiped → signature-verified by every validator → included in a block → *then* rejected with "participant is not active". Free block space, paid for by the network.

I also confirmed the report's mempool-priority observation: there is no `SetMempool` call anywhere in `app/`, so BaseApp uses `NoOpMempool` and the decorator's `Priority: 10_000_000` has no effect on block ordering.

## Three constraints that shaped the fix

These are not in the issue, and each one rules out an obvious-looking fix:

**The actor is not the signer.** In warm-key mode the DAPI wraps duty messages in authz `MsgExec` signed by the grantee, while `Creator` names the cold account that is the actual protocol participant (`tx_manager.go` `broadcastMessagesAtAttempt`; `cosmosclient.go` sets `Creator = icc.Address`). A "check the tx signer" fix would reject the production path. The check therefore reads the message body.

**Active-set membership is too narrow.** `ActiveParticipantsSet` retains roughly two epochs — `module.go` clears `epoch-2`. Gating on it would strip the waiver from legitimate previous-epoch reward claims and from BLS traffic for the epoch being rotated. Registration is the right predicate: every `ActiveParticipant` is constructed from a registered one, so registration is a strict superset of what any handler requires, and the ante layer cannot reject traffic DeliverTx would accept.

**`MinGasPriceNgonka` is 0 on the live network.** Set by the v0_2_12 upgrade handler ("temporary due to issue in gas estimations"), and `FeeParams` is absent from `DefaultParams()`. `GonkaFeeChecker` returns early when the minimum is zero, accepting any fee. So merely *withholding* the waiver — the fix that first suggests itself — keeps nothing out of the mempool today. Something has to reject.

## What this PR does

Two layers.

`NetworkDutyFeeBypassDecorator` now grants the waiver only when the message's actor is authorized: a registered participant, or an allowlisted settler for `MsgSettleDevshardEscrow`. It withholds rather than rejects, so a false negative here cannot become a liveness failure for consensus-critical PoC / BLS traffic. This is the defense-in-depth half — it becomes load-bearing the moment governance raises the minimum gas price.

`NetworkDutySignerDecorator` (new, CheckTx and ReCheckTx only) rejects the transaction. This is the half that closes the gap today. It is placed after `SigVerificationDecorator`, alongside the existing bridge early-reject, for two reasons: the participant / allowlist / grant state reads must not run on unauthenticated transactions, and for the direct path the ordering is what makes the check meaningful at all — `creator`/`settler` is the declared `cosmos.msg.v1.signer` for every duty type, so by that point it is an authenticated address rather than an attacker-chosen string.

Inside `MsgExec` the signature only proves the grantee signed, so the decorator additionally requires the authz grant that DeliverTx would require. Without it an attacker could wrap a duty naming a registered participant as `Creator` and inherit its exemption. `HasGrantForMsg` mirrors the acceptance rule in authz `keeper.DispatchActions` (self-execution implicitly accepted, otherwise a live grant for that exact type) and deliberately does **not** call `Authorization.Accept` — that can mutate grant state, which must not happen during CheckTx. DeliverTx still runs the full authz dispatch.

Errors are the registered `types.Err*` values (`ErrParticipantNotFound`, `ErrNotAllowedEscrowCreator`, `authz.ErrNoAuthorizationFound`) so API-side classifiers can match them structurally instead of on raw-log substrings.

CheckTx-only, matching `PocPeriodValidationDecorator` and `BridgeExchangeEarlyRejectDecorator`: the state machine is unchanged, only mempool admission is tightened.

## Compatibility

- No consensus-breaking change. Nothing in DeliverTx behaves differently, so no upgrade handler is needed and nodes running this can validate alongside nodes that are not.
- The escrow branch uses `IsAllowedEscrowCreator`, which returns true on an empty allowlist — exactly what the handler does, so ante and DeliverTx cannot disagree.
- 11 of the 12 exempt types appear in `InferenceOperationKeyPerms`, so the warm-key grant exists for them. The exception, `MsgSettleDevshardEscrow`, is sent directly by devshardctl (`common/chain/tx/manager.go`, `Settler: m.keyring.Address`, no `MsgExec`), so requiring a grant breaks no production path.
- Nil keeper: the bypass fails closed (no waiver); the signer decorator passes through, so a misconfigured ante chain cannot stall duty traffic.

## Tests

`app/ante_network_duty_checktx_test.go` drives the fully wired ante chain through `BaseApp.CheckTx` with real signed transactions and a zero fee, rather than testing the decorator in isolation:

| test | asserts |
|---|---|
| `RejectsUnregisteredActor` | the #1539 regression — non-zero code, `ErrParticipantNotFound` in the log |
| `AdmitsRegisteredActor` | no over-rejection: registered participant still admitted at zero fee |
| `MsgExec_RejectsForgedCreator` | outsider naming a registered participant as `Creator` without a grant is rejected |
| `MsgExec_AdmitsGrantedWarmKey` | the production warm-key path still works |
| `NonDutyUnaffected` | ordinary traffic is not touched by the duty check |

`app/ante_fee_test.go` covers the authorization mapping for all 12 exempt types, asserts `dutyAuthorizationFor` and `isExemptMessageType` agree on every type (so the two sets cannot drift), and pins the fail-closed behaviour.

Verified: `go build ./...` clean, `go vet ./app/... ./x/inference/...` clean, `go test ./app/... ./x/inference/...` all passing.


Ryanchen911:
Two of the three are real; I've fixed both. The third does not apply, and I've added a comment so the next reader doesn't have the same doubt.

**1. Nested `MsgExec` could evade the signer check — valid, fixed.**

Skipping nested `MsgExec` left a real hole, and it was worse than "the duty isn't checked": an attacker could wrap a duty inside a `MsgExec` naming the *legitimate* warm key, so the inner grant check would pass against a grant the attacker doesn't hold.

`checkMessage` now descends the whole chain, mirroring authz `DispatchActions` (which routes a nested `MsgExec` back into itself with the inner grantee). Each level requires a grant from the next level's signer to the level above, so the full delegation chain has to be intact. Grants are only required when the subtree actually carries a duty, leaving ordinary nested traffic to authz and the handlers.

Unwrapping is bounded at 5 levels. Ante work during CheckTx is not gas-metered, so an unbounded walk would itself be a spam vector. Past the limit the tx is rejected rather than passed through: at that depth the tree can't be inspected, so it can't be shown not to carry a duty.

Three new CheckTx tests cover this: broken grant chain rejected, fully granted chain admitted (guards against turning the fix into a liveness bug), and rejection past the depth limit.

Note the fee-bypass side (`isNetworkDuty` in `ante_fee.go`) still stops at one level, deliberately. There the failure mode is *withholding* the waiver, never granting it, so nesting cannot smuggle a duty into a free tx. I've added a comment at both sites explaining why the two differ.

**2. `AuthzKeeper.Grants` needs a wrapped context — this one doesn't hold.**

`sdk.Context` implements `context.Context`, and `UnwrapSDKContext` type-asserts it back before falling back to the `SdkContextKey` value:

```go
func UnwrapSDKContext(ctx context.Context) Context {
	if sdkCtx, ok := ctx.(Context); ok {
		return sdkCtx
	}
	return ctx.Value(SdkContextKey).(Context)
}
```

`runtime.kvStoreService.OpenKVStore` — which is what `getGrant` reads through — goes through that same path. So the bare `sdk.Context` reaches the same store the ante chain is reading.

`TestNetworkDuty_CheckTx_MsgExec_AdmitsGrantedWarmKey` demonstrates this end to end: it saves a real grant and asserts CheckTx returns code 0 through the wired ante chain. If the lookup were failing, that test would fail. I've added a comment on `HasGrantForMsg` recording the reasoning.

**3. Fee-waiver comment overstated what withholding accomplishes — valid, reworded.**

 `MinGasPriceNgonka` is 0 on the live network (set by the v0_2_12 handler, "temporary due to issue in gas estimations") and `GonkaFeeChecker` returns early when the minimum is zero, so nothing fails on `ErrInsufficientFee` today. The comment now says so explicitly: while the minimum stays 0, `NetworkDutySignerDecorator` is what rejects unauthorized duty txs, and withholding the waiver only starts failing them on `ErrInsufficientFee` once governance raises the minimum above zero.

`go build ./...`, `go vet`, and the full `./app/...` and `./x/inference/...` suites are green. CheckTx coverage is now 8 end-to-end tests.

inference-chain/app/ante_network_duty_signer.go
```diff
@@ -0,0 +1,216 @@
...
+//
+// Why rejecting and not merely withholding the waiver: FeeParams
+// .MinGasPriceNgonka is 0 on the live network (set by the v0_2_12 upgrade
+// handler, "temporary due to issue in gas estimations"). With a zero minimum,
+// GonkaFeeChecker accepts any fee whether or not the bypass flag is set, so
+// withholding the waiver on its own does not currently keep anything out of the
+// mempool. Withholding is kept as defense in depth for when governance raises
+// the minimum; this decorator is what closes the gap today.
+//
+// CheckTx-only, matching PocPeriodValidationDecorator and
```
kAIPraxisBot, line 32: While you're bounding nested `MsgExec` unwrapping here, the sibling this comment references — `PocPeriodValidationDecorator` in `ante_poc_period.go` — recurses on the same nested `MsgExec` structure with **no** depth bound. Its `checkMessage` (`ante_poc_period.go:131`) calls itself at line 153 for every inner message, with no equivalent of the `maxMsgExecNestingDepth` guard you added here. It runs at the same pre-signature, CheckTx-only position, so it inherits the exact reasoning in your comment: ante work during CheckTx isn't gas-metered, so an unbounded unwrap is a spam surface rather than a correctness issue.

To be clear on severity: this is pre-existing, not introduced by this PR, and low-impact — the recursion is linear in tx bytes (each level is an encoded `Any`, so depth is bounded by the tx size limit), so it's a hardening gap, not an exploitable amplification. But since you're already in this area and explicitly modelled the new decorator on `PocPeriodValidationDecorator`, it seems worth mirroring the same bound there in the same change: thread a `depth int` through `checkMessage`, reject past `maxMsgExecNestingDepth`, and pass `depth+1` on the recursive call.

Ryanchen911: Good catch — mirrored it in 19bc4d3.

You're right that it inherits the exact reasoning: same CheckTx-only position, same ungated-ante argument, and I did model the new decorator on this one. `PocPeriodValidationDecorator.checkMessage` now takes `depth int`, rejects past `maxMsgExecNestingDepth`, and passes `depth+1` on recursion, with `AnteHandle` starting at 0.

Rather than defining a second constant I reused the one here, since both decorators walk the same structure at the same point in the chain — and generalized its comment accordingly, as it no longer describes only the duty check.

Agreed on your severity read: recursion is linear in tx bytes, so this is closing a hardening gap, not an amplification bug. I kept the reject-rather-than-pass-through behaviour for the same reason as the sibling: past the limit the tree can't be inspected, so it can't be shown not to carry a late PoC message.

Three tests in `ante_poc_period_test.go`:
- rejection past the limit, wrapping a `MsgSend` so the rejection is provably from depth alone rather than from the PoC check
- admission up to the limit, so the bound doesn't become a liveness bug for legitimate warm-key traffic (production uses one level)
- the walk starts at depth 0 from `AnteHandle`, pinning that the bound counts MsgExec levels and isn't offset by the caller

I sanity-checked the first and third against a build with the guard removed to confirm they actually fail without it.

PR #1578, opened 2026-08-11
fix(devshard): reject non-canonical escrow ids

staaason:
An escrow owner could bind several independent devshard sessions to one on-chain escrow by spelling the id with leading zeros. `ChainBridge.GetEscrow` parses the path segment with `strconv.ParseUint`, so `123`, `0123` and `00123` all resolve to the same chain escrow, but the raw string was kept as the devshard identity: it became the `HostManager` session key, the singleflight key, the Postgres row key, the state-machine `EscrowID`, and the payload directory. Every alias created a fresh session seeded with the full `escrow.Amount` and reached `engine.Execute`.

The work those sessions do cannot be paid for. Hosts sign `StateSignatureContent{EscrowId: "00123"}` while the chain rebuilds the preimage from `fmt.Sprint(escrow.Id)` in `VerifyDevshardSettlement`, so the quorum never verifies and the escrow cannot be settled for that session. An alias also re-rolls every host's validation-sampling seed, which is derived from `signer.Sign([]byte(escrowID))` — giving the owner a cheap resampling oracle over which inferences get validated.

A chain escrow now has exactly one string identity, `strconv.FormatUint(id, 10)`:
- `devshard/escrowid.go`: `ParseEscrowID` / `ValidateEscrowID` reject non-digits, `0`, overflow, and any spelling that is not the canonical form.
- `server/routes.go`: middleware returns 400 on `/sessions/:id/*` before auth and before session lookup.
- `bridge/grpc.go`, `cmd/devshardd/bridge/chain.go`: canonical parse, and `GetEscrow` returns the canonical id instead of echoing the caller's string. `caching.go` validates before the escrow_cache fallback.
- `session/manager.go`: validated in `BindOwnerChat` and `create`; `RecoverSessions` skips and logs non-canonical stored rows.
- Client side (`devshardctl` runtime config and settlement encode, `user` `NewHTTPSession`), so an operator or SDK user never signs a non-canonical id.

### DIFF
--- a/devshard/bridge/grpc.go
+++ b/devshard/bridge/grpc.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"common/chain"
+	devshardpkg "devshard"
 
 	inferencetypes "github.com/productscience/inference/x/inference/types"
 )
@@ -47,11 +48,7 @@ func NewGRPCBridgeFromURL(grpcURL string) (*GRPCBridge, error) {
 }
 
 func parseEscrowID(escrowID string) (uint64, error) {
-	id, err := strconv.ParseUint(escrowID, 10, 64)
-	if err != nil {
-		return 0, fmt.Errorf("invalid escrow id %q: %w", escrowID, err)
-	}
-	return id, nil
+	return devshardpkg.ParseEscrowID(escrowID)
 }
 
 func (b *GRPCBridge) GetEscrow(escrowID string) (*EscrowInfo, error) {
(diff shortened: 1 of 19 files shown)


a-kuprin (approved):

a-kuprin:
As suggested by report author:

Suggested PR comment:

---

One leftover gap: `/stats/shards/:escrow_id` sits outside `canonicalEscrowIDMiddleware`, and `SessionServerExisting` → `recoverAndStoreSession` does not validate the ID. A pre-existing active row like `"00123"` (skipped by `RecoverSessions`) can still be resurrected via `GET /stats/shards/00123`. That does not reopen unpaid inference budgets, but it can reload legacy alias state into memory.

Suggested hardening: validate in `SessionServerExisting` (covers stats + any other recover path), and ideally mark non-canonical active rows inactive in `RecoverSessions` / a one-shot migration so they cannot linger.

```go
func (m *HostManager) SessionServerExisting(escrowID string) (*transport.Server, error) {
	if err := devshardpkg.ValidateEscrowID(escrowID); err != nil {
		return nil, err
	}
	// ...
}
```

staaason:
@a-kuprin I have resolved the comment above

x0152 (approved):

PR #1579, opened 2026-08-11
`gateway`+`devshard` Always-stream upstream + shared scratch spool + citest

a-kuprin:
## Merge order

Merge **after**:

1. [#1574](https://github.com/gonka-ai/gonka/pull/1574) — same-nonce reconnect + host ML drain + LiveStream spool  
2. [#1575](https://github.com/gonka-ai/gonka/pull/1575) — streamed hop timestamps (`: devshard-ts`)

Those PRs land through `e708fb3`. This branch continues from that tip with always-stream forcing, shared `devshard/spool`, and force-upstream citest.

---

## Motivation

Non-streaming clients today wait for a single opaque JSON body. That hides the signals we need for smarter routing and host quality:

- **TTFT / first-token** — no progressive chunks → no first-token timer that matches reality  
- **Inter-chunk liveness** — a stuck host looks the same as a slow one until the whole response arrives  
- **Sustained stream speed (bytes/sec)** — soft signal for Decide / quarantine later; useless if the wire is one blocking read  

Always-stream flips the **upstream** shape to SSE for every chat request (when the admin flag is on), while the **client** still gets the shape it asked for: SSE for `stream:true`, one aggregated `chat.completion` for `stream:false`. Routing and soft signals can observe every request the same way; silent non-stream waits stop being a blind spot.

Same-nonce reconnect ([#1574](https://github.com/gonka-ai/gonka/pull/1574)) and hop timestamps ([#1575](https://github.com/gonka-ai/gonka/pull/1575)) already assume a streamed host path. This PR finishes the gateway half: force upstream stream, fold for JSON clients, keep client intent honest across double-normalization, and bound aggregate memory with a shared scratch spool.

**Default remains off** (`ForceUpstreamStreaming` false) until Step 14 soak; this PR lands the machinery and e2e that gate flipping it on.

---

## Always-stream flow

```mermaid
sequenceDiagram
  participant C as Client
  participant G as Gateway
  participant H as Host
  participant M as ML

  C->>G: chat.completions stream true or false
  Note over G: Snapshot client stream usage and logprobs intent
  Note over G: ForceUpstreamStreaming wires stream true include_usage and forced logprobs
  G->>H: HostRequest plus nonce
  H->>M: POST streamed completion
  M-->>H: SSE chunks
  H-->>G: SSE plus receipt plus meta
  alt Client wants SSE
    Note over G: Forward chunks and strip forced fields client did not ask for
    G-->>C: text event-stream until DONE
  else Client wants JSON
[...]

PR #1581, opened 2026-08-11
gateway accounting

qdanik:
## Summary

This branch closes a set of gaps where the gateway's own bookkeeping, its timeout votes and its escrow addressing could each disagree with what the chain and the hosts actually saw. Accounting gains a vocabulary for *why* a host failed rather than only *how often*; burned nonces become answerable to the group; and several paths that silently diverged from the protocol are brought back in line.

Two switches, both defaulting to the behaviour an unconfigured gateway should have: accountability on, accounting on. The join template opts a joining node out of accounting explicitly.

---

### A burned nonce cost the host nothing

**Was.** `HostStats.Missed` was incremented only in `applyTimeout`, which ran only for nonces the gateway had dispatched. A nonce the gateway declined to send — throttled host, or a host that could not serve the request at all — was spent silently. The nonce left the escrow either way, but nobody was answerable for it, so a participant refusing everything finished an epoch looking cleaner than one that accepted work and occasionally dropped it.

**Now.** `raiseGhostAccountability` raises a timeout for every accountable burn, carrying a payload built exactly as a real dispatch would, so it verifies against the nonce's committed `PromptHash`. The gateway does not decide the outcome — verifiers decide it through the normal `ChallengeReceipt` path: an unreachable executor is accepted and becomes a miss, a reachable one answers and is not charged. The accusable perimeter is deliberately narrow: `accountable()` is true only for throttled and capability-blocked burns. PoC windows and an empty queue are excluded, since the first is protocol-sanctioned and the second is our own scheduling.

### The refusal deadline could never elapse

**Was.** The accountability probe stamped `StartedAt` in milliseconds while every other `StartedAt` on the wire is seconds. A verifier computes the refusal deadline as now-in-seconds minus that stamp, so the difference was permanently negative, the deadline never elapsed, and every raised timeout was rejected. The mechanism was live and produced nothing.

**Now.** `ghostProbeParams` stamps `time.Now().Unix()`. The unit is asserted by tests that derive their clock independently of the field under test and check both sides of the boundary, so a test cannot pass by being unit-agnostic.

### The log reported the inverse of what happened

**Was.** `HandleTimeout` signals an applied timeout by returning an error, and the caller branched on `err != nil`. Applied rounds were therefore logged as failures.

**Now.** The caller branches on `result.Applied` and emits `ghost_timeout_applied`, keeping `ghost_timeout_failed` for rounds that genuinely reached no verdict.

### The ledger counted nonces but could not name a fault

**Was.** A participant record said how many nonces were assigned, finished or lost. It could not separate a slow host from a broken one, nor a host's own defect from a burn the gateway itself chose.

**Now.** Every counter carries a failure origin, a disposition and a reason drawn from one vocabulary, and findings are derived from the record rather than recomputed — so a finding and the numbers printed beside it cannot come from different reads. Added: `logprobs_not_token_ids` (a serving-stack defect that costs the host on every validated inference), the timing faults (`slow_receipts`, `slow_chunks`, `slow_decode`, `clock_drift`), and `blocked_by_capability`.

### Findings that fired but could not discriminate

**Was.** `challenges_unresolved` was a bare count flagged at any value above zero, reading identically whether a handful of disputes hung across a large epoch or a large share of a small one. Rounds that reached no verdict at all were surfaced by nothing.

**Now.** It carries a denominator and is flagged on the same scale as `chain_recorded_invalid`, since an open challenge is an invalid that has not landed. A new `timeouts_undecided` covers rounds that reached no verdict.

### A capability limit learned on one model blocked another

**Was.** Context limits and tool-support verdicts were keyed by participant, so a small context learned while serving one model suppressed routing to that host for every other model it served.

**Now.** Keyed by `(participant, model)`. The protocol-version verdict stays per-participant, since the build is one per host and covers every model it serves.

---

### A new escrow reached its hosts one nonce at a time

**Was.** A rotated escrow reached a host only when `nonce % groupSize` selected it. Until then that host held no session, so every catch-up, signature and vote it was asked for failed — for as long as it took the counter to come round.

**Now.** Creation sends one probe and then catches up every host in the group from it, spending one nonce per escrow rather than the sixteen a slot-by-slot sweep would cost. It runs only for escrows the gateway just created, never at startup, and is bound to the runtime's stop.

### The receipt and the vote it justified moved apart

**Was.** `MsgConfirmStart` was queued from the executor receipt while `confirmedAt` was set on a different path, so the two could disagree. The session then voted a refusal against a record the chain had already advanced, and every verifier rejected it. The receipt was also queued only after the stream closed, so a host holding an empty stream could outlast the last diff able to carry it.

**Now.** Both halves move together in `confirmStartOnReceipt`, at the moment the receipt event arrives.

### An escrow followed the gateway onto a version its hosts refused

**Was.** `route_prefix` was left empty on rotation-created escrows and re-resolved from process-global state on every runtime build, so a gateway redeploy moved every existing escrow onto a new protocol version at once. A host binds an escrow to the first version that reaches it and refuses every other one for the escrow's life, so the move was one-way and unrecoverable. Two separate resolvers derived the stamped protocol version and the wire prefix, and could disagree.

**Now.** The prefix is recorded at creation and the protocol version is derived from it, so the stamp and the wire cannot name different versions. Pending-commitment recovery keeps the version its escrow was minted under. The second resolver is removed. Escrows created before this change remain bound to whichever version reached their hosts first and need rotation; nothing on the gateway can rewrite a host's stored binding.

### A version conflict was not read as a refusal

**Was.** Only the "version not found" refusal was recognised as permanent. A host that had already bound the escrow under another version answers with a conflict instead, which was treated as an ordinary transient failure and retried.

**Now.** Both bodies are recognised. Neither is fixed by waiting, so neither is backed off from as a fault.

### One error event bought a host out of the vote it owed

**Was.** The long-response exemption gated on the chunk counter, which advances on error events as well as content. A host that emitted a single error and then held the stream open earned the same exemption as one still generating, and kept its timeout vote.

**Now.** It gates on the content source, which only a real content chunk sets. An empty stream is unaffected — it has neither — so a host is still charged for holding a stream it never filled.

### Catch-up failures were reported as success

**Was.** `sendCatchUp` logged a failed chunk and returned nil, so a caller was told the host held the escrow when it had just refused. `SyncHosts` abandoned the whole sweep on the first failure, leaving the remaining hosts unsynced.

**Now.** A failed chunk is returned as an error, and `SyncHosts` collects failures and continues, joining them at the end.

### A capability verdict held a host out of the rota

**Was.** A parsed capability refusal kept a host out of routing for the rest of the process, with nothing to re-test it.

**Now.** Capability verdicts are historical counters. Routing still screens each request against the refusal it matches, in the picker; what no longer happens is a host being withheld wholesale, so a host that replaced its build re-enters the rota on its own.

---

## Observability

- `GET /api/v1/epochs/{epoch}/participants` — per-participant record: dispositions, counters, findings, capability block. Per-participant rollup replaces the previous per-nonce tallies, which grew without bound.
- `GET /api/v1/epochs/{epoch}/participants/{participant}` — one host, same shape.
- `GET /api/v1/epochs/{epoch}/events` and `/events/{participant}` — per-nonce feed of the two verdicts that cost a host, each naming the escrow, nonce, slot and originating `request_id`. Bounded ring per escrow, so it cannot reintroduce the growth the rollup removed.
- Findings are exported through the Prometheus collector.
- Every finding code is explained once in `devshard/docs/accounting-findings.md` rather than crossing the network with each response.


a-kuprin:
Just a note.
Fix from https://github.com/gonka-ai/gonka/pull/1563 doesn't land here. Anyway it is preferable to merge #1563 merge first, when it is reviewed with someone else

a-kuprin:
According to commit [feat(accounting): record what a host actually delivered on a nonce, s…](https://github.com/gonka-ai/gonka/pull/1581/commits/8b58b5cfcc2f301d3b34c307bb804e8220daeab5)

### 1. Model-burn vs host empty are collapsed (please fix)

`isEmptyStreamAttempt` and `isModelBurnEmpty` are different blame:

- host empty → fault / quarantine path (`ObserveEmptyStreamForModel`)
- Kimi model burn → `usage.completion_tokens > 0`, telemetry only (`ObserveModelBurnEmpty`)

`gatewayAttemptFailureReason` maps both to `"empty_stream"`, so the ledger/Prometheus delivery series cannot tell them apart. After the thinking-template fix burn should be rarer, but residual burn will still pollute host delivery scores.

### 3. No finding for empty delivery — optional

Findings never read `delivery_reasons`. That only hurts if ops triage solely from `findings[]`. Finished+empty is the disposition hole (looks “used/healthy” otherwise); unfinished empties already show up via timeouts / failure origins. Fine to defer if dashboards already alert on the delivery metric or limiter quarantine. If findings *are* the ops surface, add a warning on finished + `empty_stream` (excluding burn once split).

### 3. Overlap with `DetailReason` — leave for later

It is mostly cleanup and could be deffered for next PRs.
Same allowlist / many of the same strings. Delivery’s unique value is tagging **finished usage** where timeout detail may be skipped. Duplication is messy but not wrong; tidy the taxonomy after the burn split if it still confuses.

a-kuprin:
According to [feat(devshardctl): measure the silence between chunks, so a slow host…](https://github.com/gonka-ai/gonka/pull/1581/commits/af98a8afb7959082867e9414541e5009cb53845b) commit

**Also measure decode throughput, not only inter-chunk gaps.**

Gaps answer “alive vs stalled.” They don’t answer “how much useful data moved.” Mean gap ignores chunk size, so tiny keepalives can look as healthy as real decode.

At attempt end, next to cadence, observe payload / time after first content — content bytes or completion tokens over `(end − first_content)`. Gaps stay the stall signal; throughput is the speed signal. End-of-attempt histogram is enough.

a-kuprin:
According to always-stream feature

Accounting, cadence, findings, `replaceNonFiniteNumbers`, and the two-pipeline force look good! And should be kept as is.

The always-stream **body path** should not ship as `bytes.Buffer` + `assembleSSEBody`. I suggest to take following design from `ak/only-streaming-to-gateway` (write-up: [devshard/docs/pr-1581-followups.md](https://github.com/gonka-ai/gonka/blob/ak/only-streaming-to-gateway/devshard/docs/pr-1581-followups.md)). Those are two layers; both are required.

I think I can cherry-pick them and add new PR targeted to your branch to keep changes clear, and close https://github.com/gonka-ai/gonka/pull/1579 PR 

## What to take

**Spool** replaces the unbounded in-memory SSE buffer. Design is here [spool-shared-library.md](https://github.com/gonka-ai/gonka/blob/ak/only-streaming-to-gateway/devshard/docs/spool-shared-library.md)
Winner bytes stay in RAM up to 2 MiB, then spill to an unlinked temp file, with a 16 MiB per-request ceiling and a process cap on concurrent files. Oversize becomes an error, not a 30-minute RAM soak. The fold then reads with `OpenReader()` (one SSE line at a time) instead of `ReadAll`.

**The local fold** replaces `assembleSSEBody`. Spool does not do this job — it only stores bytes. `assembleSSEBody` is a generic JSON merge with silent caps: at 65,536 events it `break`s and returns HTTP 200 with a prefix. With forced `logprobs` / `top_logprobs: 5`, chunks are roughly one per token, so a long generation can look complete and be truncated. An event with `"object": "chat.completion"` also short-circuits and discards deltas already merged.

The local fold is an OpenAI completion builder with a budget: oversize → error body, not a plausible 200; logprobs the client did not ask for are dropped at decode instead of built and stripped later; emit keeps `logprobs: null` and defaults `message.role` to `"assistant"`. Host errors are classified (passthrough / fail closed / drop trailing) instead of merged as another JSON object.


**p.s.:**
Also I would land tests including e2e tests from that branch

a-kuprin (approved):

PR #1583, opened 2026-08-12
weight override

gmorgachev:
Our security model requires inference quality to be identical to the governance model. Sometimes the same model can use weights in a different format for better performance while keeping the same quality.

This PR lets a host use another Hugging Face repository and commit for a governance model. The model is still served under the governance model ID, so inference routing and PoC continue to use the expected name.

Example:

```json
"MiniMaxAI/MiniMax-M2.7": {
  "args": [],
  "model_override": {
    "hf_repo": "host/custom-minimax-weights",
    "hf_commit": "0123456789abcdef0123456789abcdef01234567"
  }
}
```

DAPI adds `--revision` and `--served-model-name` automatically. It downloads the override before replacing a healthy deployment and keeps the current model available while the new weights are downloading.

The override is local to the host. It does not change the governance model, network routing, pricing, or PoC model ID. Existing PoC and inference validation still check that the alternative weights produce compatible results.

Tests:

- `go test ./...` in `decentralized-api`
- race tests for the changed DAPI packages

### DIFF
--- a/common/queryapi/tests/dapi_contract_test.go
+++ b/common/queryapi/tests/dapi_contract_test.go
@@ -15,20 +15,20 @@ import (
 // public handlers for read-only query routes. edge-api must expose the same keys
 // so existing clients keep working when the proxy steers traffic to edge-api.
 var dapiTopLevelKeys = map[string][]string{
-	"GET /v1/status":                          {"status"},
-	"GET /v1/models":                          {"object", "data"},
-	"GET /v1/governance/models":               {"models"},
-	"GET /v1/governance/models-legacy":        {"model"},
-	"GET /v1/pricing":                         {"unit_of_compute_price", "models"},
-	"GET /v1/versions":                          {"api_version", "node_version", "timestamp"},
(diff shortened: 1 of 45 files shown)


GLiberman:
4 fixes:

Current deployment per node — Store one applied {model_id, fingerprint} record per MLNode instead of per-model history. A later assignment to model B, then back to A, must not treat A’s old fingerprint as “already running.”

Reject non-2xx stop/start — Persist a fingerprint only after MLNode actually accepted stop/start. HTTP 4xx/5xx used to look like success, so DAPI could record B while A was still serving.

Pre-download before redeploy — If the node is healthy, download the target (override or governance rollback) before stopping inference. Avoid taking a good replica offline while weights are still missing.

Reconcile generation — Stamp each in-flight reconcile and its result with a generation. OriginalTarget == INFERENCE is not enough: a cancelled A result can land after B’s reconcile has started, persist A, clear pending, and drop B. Commit-only A→B changes are invisible to repo/served-name health checks.

PR #1562, opened 2026-08-07
Add gateway mockenv coverage tests

aikuznetsov:
## Summary

- Add a Gateway mockenv test layer that runs the real gateway HTTP stack with fake in-memory devshard runtimes.
- Cover the main Gateway flows: pooled chat, direct devshard routing, model access, cache behavior, limiter behavior, status, admin auth, private key redaction, and selected security-sensitive edge cases.
- Add Make targets for mockenv tests and wire them into CI with `gotestsum` and a separate JUnit report.
- Add a coverage document mapping Gateway logic to the new mockenv tests.

## Testing

- `make gateway-mockenv-test`
  - `DONE 43 tests`
- `make gateway-mockenv-test-junit`
- `make test DEVSHARD_TEST_PACKAGES=./cmd/devshardctl/messagevalidators`
- `git diff --check`

PR #1584, opened 2026-08-12
`Devshard v5`: Height-sync protocol + general devshard fixes

a-kuprin:
# Aggregated v5: Height-sync protocol + general devshard fixes

This is the updated [#1584](https://github.com/gonka-ai/gonka/pull/1584) body: original height-sync write-up, plus the feature-block index of everything else that has merged onto `devshard-0.2.15-v5`.

Inventory (per-block test plans): [`devshard/docs/v5-merged-features-test-plan.md`](devshard/docs/v5-merged-features-test-plan.md)  
Testing manuals: [`v5-manual-height-sync.md`](devshard/docs/v5-manual-height-sync.md) (block 1), [`v5-manual-residual.md`](devshard/docs/v5-manual-residual.md) (blocks 3, 5, 9)  
Release notes: [`release-0.2.15-v5.md`](devshard/docs/release-0.2.15-v5.md)  
On-chain submit JSON: [`gov-devshard-v5-and-bounty-payouts.json`](gov-devshard-v5-and-bounty-payouts.json)  
Bounties: [Proposed Bounties](#proposed-bounties) (91,300 USDT)

---

## Feature-block index

What landed on this branch after the v4 line, grouped by feature.

| # | Block | Headline | PRs |
| - | ----- | -------- | --- |
| 1 | [Height-sync protocol](devshard/docs/v5-merged-features-test-plan.md#1-height-sync-protocol) | Courier alignment, log-plane stamps, heartbeats, detection of lag / future / fabricated hash | [#1615](https://github.com/gonka-ai/gonka/pull/1615), [#1625](https://github.com/gonka-ai/gonka/pull/1625), [#1621](https://github.com/gonka-ai/gonka/pull/1621), [#1649](https://github.com/gonka-ai/gonka/pull/1649), [#1669](https://github.com/gonka-ai/gonka/pull/1669), [#1697](https://github.com/gonka-ai/gonka/pull/1697) |
| 2 | [HA overlay: router, catalog, Postgres](devshard/docs/v5-merged-features-test-plan.md#2-ha-overlay-router-catalog-postgres) | HAProxy L7, governance catalog, fail-closed HA storage, live PG readiness | [#1599](https://github.com/gonka-ai/gonka/pull/1599), [#1601](https://github.com/gonka-ai/gonka/pull/1601), [#1602](https://github.com/gonka-ai/gonka/pull/1602), [#1655](https://github.com/gonka-ai/gonka/pull/1655), [#1603](https://github.com/gonka-ai/gonka/pull/1603), [#1606](https://github.com/gonka-ai/gonka/pull/1606), [#1604](https://github.com/gonka-ai/gonka/pull/1604), [#1607](https://github.com/gonka-ai/gonka/pull/1607), [#1642](https://github.com/gonka-ai/gonka/pull/1642) |
| 3 | [Warm cutover and session recovery](devshard/docs/v5-merged-features-test-plan.md#3-warm-cutover-and-session-recovery) | `/ready` split, overlap waits `recovery_complete`, snapshot restore, epoch prune | [#1653](https://github.com/gonka-ai/gonka/pull/1653), [#1712](https://github.com/gonka-ai/gonka/pull/1712), in-tree warm-cutover, local `ak/devshardd-epoch-evict` |
| 4 | [Gateway v4 → v5](devshard/docs/v5-merged-features-test-plan.md#4-gateway-v4--v5) | Always-stream aggregation, height-seed loop, epoch stats | [#1681](https://github.com/gonka-ai/gonka/pull/1681), [#1684](https://github.com/gonka-ai/gonka/pull/1684) |
| 5 | [Host ping observability](devshard/docs/v5-merged-features-test-plan.md#5-host-ping-observability) | Gateway `/clock` pings and dapi mlnode pings; no routing effect | [#1580](https://github.com/gonka-ai/gonka/pull/1580) |
| 6 | [Validation, misses, settlement](devshard/docs/v5-merged-features-test-plan.md#6-validation-misses-settlement) | Payload withhold → vote false; error-stream miss; pending drain credit | [#1647](https://github.com/gonka-ai/gonka/pull/1647), [#1651](https://github.com/gonka-ai/gonka/pull/1651), [#1652](https://github.com/gonka-ai/gonka/pull/1652), [#1626](https://github.com/gonka-ai/gonka/pull/1626), [#1634](https://github.com/gonka-ai/gonka/pull/1634), [#1512](https://github.com/gonka-ai/gonka/pull/1512) |
| 7 | [Security and admission](devshard/docs/v5-merged-features-test-plan.md#7-security-and-admission) | Host cannot inject user txs; settled bind; SSRF; canonical escrow ids | [#1664](https://github.com/gonka-ai/gonka/pull/1664), [#1627](https://github.com/gonka-ai/gonka/pull/1627), [#1629](https://github.com/gonka-ai/gonka/pull/1629), [#1578](https://github.com/gonka-ai/gonka/pull/1578) |
| 8 | [Payload size and chat shape](devshard/docs/v5-merged-features-test-plan.md#8-payload-size-and-chat-shape) | Compress at rest; `n=1`; logprobs only if asked; stream parse hardening | [#1650](https://github.com/gonka-ai/gonka/pull/1650), [#1676](https://github.com/gonka-ai/gonka/pull/1676), [#1686](https://github.com/gonka-ai/gonka/pull/1686), [#1713](https://github.com/gonka-ai/gonka/pull/1713) |
| 9 | [Proxy and version stamp](devshard/docs/v5-merged-features-test-plan.md#9-proxy-and-version-stamp) | Atomic TLS install; strip `v` prefix on rotation protocol | [#1661](https://github.com/gonka-ai/gonka/pull/1661), [#1662](https://github.com/gonka-ai/gonka/pull/1662) |
| 10 | [Testenv E2E on v5](devshard/docs/v5-merged-features-test-plan.md#10-testenv-e2e-on-v5) | Re-apply Docker citest to this protocol line | [#1564](https://github.com/gonka-ai/gonka/pull/1564) |

Operator manuals in this tree: height-sync → [`v5-manual-height-sync.md`](devshard/docs/v5-manual-height-sync.md); residual 3 / 5 / 9 → [`v5-manual-residual.md`](devshard/docs/v5-manual-residual.md). HA overlay (2) and gateway (4, 8) land in files from Stas and Danya.

The height-sync section below is the original protocol landing. Operator metrics ([#1621](https://github.com/gonka-ai/gonka/pull/1621)) and the other blocks in the index have since merged onto this branch.

---

## Proposed Bounties

On-chain payouts from the community-sale contract as IBC USDT (`withdraw_ibc`, same pattern as [proposal #76](https://gonka.gg/network/proposals/76)). Total: **91,300 USDT**.

| GitHub | USDT | Explanation | Address |
| --- | ---: | --- | --- |
| [@akup](https://github.com/akup) | 36,000 | devshard v4.0.1, v4.0.2, v4.0.3, v4.1.0, v5 + height-sync | `gonka1ejkupq3cy6p8xd64ew2wlzveml86ckpzn9dl56` |
| [@aikuznetsov](https://github.com/aikuznetsov) | 6,500 | testing, PR [#1508](https://github.com/gonka-ai/gonka/pull/1508), [#1547](https://github.com/gonka-ai/gonka/pull/1547), [#1564](https://github.com/gonka-ai/gonka/pull/1564), [#1562](https://github.com/gonka-ai/gonka/pull/1562), [#1631](https://github.com/gonka-ai/gonka/pull/1631), [#1649](https://github.com/gonka-ai/gonka/pull/1649), [#1651](https://github.com/gonka-ai/gonka/pull/1651), [#1653](https://github.com/gonka-ai/gonka/pull/1653), [#1555](https://github.com/gonka-ai/gonka/pull/1555)/[#1571](https://github.com/gonka-ai/gonka/pull/1571) | `gonka1frlfyz2wtltdy47dq3w9pwc8ruvjvlthp2lh53` |
| [@snevolin](https://github.com/snevolin) | 12,000 | high-availability: [setup instruction](https://github.com/gonka-ai/gonka/blob/devshard-0.2.15-v5/docs/devshard-host-ha-setup.md), [test plan](https://github.com/gonka-ai/gonka/blob/devshard-0.2.15-v5/devshard/docs/devshard-host-ha-test-plan.md); HAProxy versiond router, catalog routes, independent fleet (#1599, #1606, #1610); fail-closed Postgres HA, live /readyz, persistent PGDATA, storage proofs (#1642, #1601, #1602, #1603, #1607, #1730); versiond/edge-api drain (#1604, #1605); public HAProxy + crash-safe TLS (#1609, #1661); update-devshard.sh safe migrate to HA layout (#1611) | `gonka1vnupswg7qz2w5k5ax6zrp02mxmln6arnvjc87h` |
| [@qdanik](https://github.com/qdanik) | 23,000 | Validation and H1 report fixes: token floor, logprob distance, n>1 ([#1391](https://github.com/gonka-ai/gonka/pull/1391), [#1436](https://github.com/gonka-ai/gonka/pull/1436), [#1676](https://github.com/gonka-ai/gonka/pull/1676)), Payload compression and logprobs gating ([#1650](https://github.com/gonka-ai/gonka/pull/1650), [#1686](https://github.com/gonka-ai/gonka/pull/1686)), Gateway accounting, delivered-answer serving, pre-floor diff replay ([#1581](https://github.com/gonka-ai/gonka/pull/1581), [#1682](https://github.com/gonka-ai/gonka/pull/1682), [#1595](https://github.com/gonka-ai/gonka/pull/1595)), Kimi, DeepSeek and stop_reason fixes ([#1563](https://github.com/gonka-ai/gonka/pull/1563), [#1613](https://github.com/gonka-ai/gonka/pull/1613), [#1713](https://github.com/gonka-ai/gonka/pull/1713)), Gateway v4 release, review of the previous release, H1 report reviews | `gonka1j3f2xkapx8cmczpjqcsrh7cc3peyj3ngkjv4p8` |
| [@redstartechno](https://github.com/redstartechno) | 300 | PR [#1336](https://github.com/gonka-ai/gonka/pull/1336), [#1512](https://github.com/gonka-ai/gonka/pull/1512), [#1460](https://github.com/gonka-ai/gonka/pull/1460) | `gonka105ce4495mj0mwkxqeasgdzqfq5jjrfq32eza5l` |
| [@shd](https://github.com/shd) | 11,500 | devshards protocol review and formalization, [Discussion #1340](https://github.com/gonka-ai/gonka/discussions/1340), includes devshard v4 | `gonka1p2hjjf63dqhpf5qmyaaq6u73zrzsrf3ra3qxk2` |
| [@Ryanchen911](https://github.com/Ryanchen911) | 2,000 | PR [#1491](https://github.com/gonka-ai/gonka/pull/1491), Issue [#1470](https://github.com/gonka-ai/gonka/issues/1470) | `gonka1zqss46r6jf6dhhyaa777kc2ppvjhn0ufkx4y57` |
| **Total** | **91,300** | | |

---

## Height-sync protocol (v5)

**Spec:** [`HEIGHT_SYNC_PROTOCOL_PROPOSAL.md`](devshard/docs/proposals/HEIGHT_SYNC_PROTOCOL_PROPOSAL.md)  
**Plan:** [`height-sync-implementation-plan.md`](devshard/docs/height-sync-implementation-plan.md)  
**Params:** [`height-sync-params.md`](devshard/docs/height-sync-params.md)  
**Tests:** [`height-sync-tests.md`](devshard/docs/height-sync-tests.md)

Sibling (not this PR): `ak/height-sync-protocol-dapi` mounts hash-only `/block/*` on production dapi.

---

## Summary

This PR lands height-sync on 0.2.15 / v5: every host and the sequencer keep a signed, replayable view of mainnet height inside the escrow log, even when there is no inference traffic.

It has two planes, plus a new kind of time in the log:

1. **Transport plane (Phases A–D).** Anchor envelopes on a `K`-nonce cadence, forced turns, courier peer tips, origin signatures, and `(C-quorum)`. Rewired onto `devshard/chainoracle/blocks` instead of cherry-picking `devshard-testenv`. Default compose stays off; opt in with `DEVSHARD_CHAINORACLE_URL` or `DEVSHARD_HEIGHTSYNC`.
2. **Log plane (Phase E, E0–E7 + E9).** Heartbeat turns, host acks, L0–L5a / L7 ingest checks, repair probes that never assign blame, close-ready arming, and `observed_height` stamps on inference txs so a busy escrow never heartbeats.
3. **Logical time.** Every Diff-resident height — heartbeat, ack, and inference `start` / `confirm` / `finish` — is a *reference height*: `max(own_tip, F(m))`. `F` is the escrow's logical clock, a function of the applied log. Later releases can fold timeouts, `USER_TIMEOUT`, and cPoC bands onto those heights instead of host/user wall-clock timestamps. This PR **carries** the stamps; it does **not** switch those decisions yet.

**Deferred to later releases (Phase F / dispute layer):**

* **Strong / `LightBlock` proving.** Envelope field 9, `D`-band escalation (`|Δ| > D` ⇒ Strong required), `(C-strong)` / `(C-hybrid)`, and `CATCHING_UP` forcing the next heartbeat to Strong. Until then `CATCHING_UP` is a label only.
* **On-chain disputes.** Marks (`DISPUTE_ORIGINATOR` / `DISPUTE_CARRIER`, vector contradiction, `DEFERRED_FAIL`) are recorded locally. Evidence packets, `MsgHeightSyncEvidence`, slashing, and cross-session equivocation wait on Strong, because the canonical half of a packet is a `LightBlock`.
* **L6 adjudication.** L6 reconciles a Diff-resident `(height, hash)` against the verifier's own oracle. That check runs *after* the pair is already in `Diff`, so it must not `INVALID` the diff (two verifiers whose followers have not reached `H` would split). The implementation records a `DEFERRED_FAIL` mark when the oracle already has block `H` and the hash mismatches. A durable recheck queue for heights that have not mined yet, and turning those marks into disputes, ship with the dispute layer. `F` is not unwound if a later L6 fails.

Also not in the original height-sync landing: cPoC skip carriers, flipping `timeout.go` / the seal clock onto heights, and default-on compose. Gateway operator metrics (plan §8.12, E8) later landed as [#1621](https://github.com/gonka-ai/gonka/pull/1621) on this same branch.

---

## Why

cPoC bands and `USER_TIMEOUT` finalization both need a height that every honest verifier recomputes from `Diff`. Envelope Anchors only exist on the request/response that carried them; a courier user has no follower of its own; a quiet escrow never syncs.

Height-sync puts a signed `(height, hash)` into the log on a wall-clock cadence. Hosts answer. The record of who answered is a pure function of `Diff`. Hosts that hear nothing from the user arm close-ready locally and still emit nothing — closing stays finalization's job.

The schedule is in **milliseconds**, not blocks. Mainnet height is the *result* of a height-sync turnover, so a block-denominated heartbeat is circular: a quiet courier cannot notice a block passed until it syncs, and a partitioned host sees a frozen tip and would never count enough blocks to arm.

---

## Height stamps and logical time

This is the load-bearing new feature. The rest of the protocol landing exists so these stamps keep moving when inference does not.

### Two planes, one meaning in the log

| Plane | Where | Height means | Who signs |
| ----- | ----- | ------------ | --------- |
| **Transport** | HeightSyncSection on the HTTP envelope | first-party follower tip of *this hop* | origin signature on the response; request leg unsigned |
| **Log** | `observed_height` / `observed_block_hash` on Diff-resident txs | **reference height** — logical time of the escrow | user (diff sig) on heartbeat / start; host (`host_sig` / `executor_sig` / `proposer_sig`) on ack / confirm / finish |

The envelope never enters `Diff`. Two verifiers replaying the journal would not see it, so it cannot be the clock for cPoC or `USER_TIMEOUT`. The stamp in the tx can.

There is **no second height in the log**. A lagging host does not write its raw tip into Diff: it writes `max(own_tip, F(m))`, or omits when the floor is more than `W_conf` above its tip. Its real reading stays first-party in the response-leg Anchor and in `sync_state`.

### What a stamp is

`observed_height` on a message means “the height its signer observed when producing this message” — exactly what `started_at_height` / `confirmed_at_height` would mean on the wire, so those names live only on the **record**, copied from the stamps of `MsgStartInference` and `MsgConfirmStart`.

A stamp is only as strong as the key on it:

| Stamp | Signer | Strength |
| ----- | ------ | -------- |
| MsgHeartbeat, MsgStartInference | user (diff signature) | a **claim** — attributable, checkable against the chain later via L6 |
| MsgHeightAck (`host_sig`), MsgFinishInference (`proposer_sig`), MsgConfirmStart (`executor_sig`, mirrored into ExecutorReceiptContent) | host | an **attestation** — rewriting it is forgery of that host |

The confirm trap: `executor_sig` is over `ExecutorReceiptContent`, a signing input that never enters Diff. A field on `MsgConfirmStart` that is missing from that content is outside the executor's signature, and the sequencer could rewrite the height. This PR mirrors `observed_height` / `observed_block_hash` into the receipt and copies them back before recovery.

Unstamped records omit proto3 zeros, so `post_state_root` bytes do not change (`v2` root).

### `F` is the escrow's logical clock

`F(m)` is the reference height the log had established at nonces `< m`. L0 holds every Diff-resident stamp to it. It never lowers.

Raise rules (spec §14), landed:

1. An envelope `(H, hash)` cannot be a future height (hash unknown in advance). Catch/punish is L6 / Phase F.
2. `F` never exceeds the max reported first-party **host** envelope `H`. Lifts (`stamp = F`) do not raise. Honest compose drops a host raise that does not match this hop's response-leg envelope.
3. Sequencer-composed stamps (`MsgHeartbeat`, `MsgStartInference`) never raise `F` and never count toward quorum `Q`. `Observe` is host-only.

A single party claiming `1<<40` therefore cannot become the escrow's time: the floor only follows host-signed first-party tips, unaided by at most `W_conf`, or a host-only `Q` for a larger jump.

### What this enables, and what it does not switch yet

| Decision today | Height form (later) |
| -------------- | ------------------- |
| Execution timeout against each host's clock | `h − confirmed_at_height` |
| Refusal timeout against user-controlled started_at | `h − started_at_height` (checkable) |
| Seal clock = max confirmed_at | max `confirmed_at_height` |

Migration is two-step on purpose: this PR **carries** the heights; folding them into a state root (timeouts, seals) needs a protocol-version gate, or two participants compute different roots. `timeout.go` / `stateClockLocked` stay on wall time.

---

## What landed

| Phase | What |
| ----- | ---- |
| **A** | `devshard/heightsync` transport: Omit/Anchor cadence, inbound classify, confirmation, origin signing, audit. Envelope + `MsgForceHeightSyncTurn` (oneof 9). Host/user/state seams on `chainoracle/blocks`. |
| **B** | In-process e2e in `testenv/scenarios/heightsync_anchor_e2e_*.go`. |
| **C** | Container `citest-height-sync` against mock-dapi `/block/*`. Default compose unchanged; the citest harness patches `DEVSHARD_CHAINORACLE_URL`. |
| **D** | Hash-only Tendermint observer, direct-chain adapter, host failover when dapi has no `/block/*` or is down. Production dapi mount is the sibling PR. |
| **E0–E3** | `MsgHeartbeat` / `MsgHeightAck` (oneof 10/11; 12/13 reserved for cPoC). User `MaybeHeartbeat` spans, concurrent and non-aborting. Host-signed acks into the mempool from the same oracle read as the response-leg Anchor. Gateway starts the heartbeat loop. |
| **E9** | Session-open seed: fan `POST /sessions/:id/height-sync` once so nonce 1 already has a tip. Consumes no nonce, does not advance `h_last`, never fails the session on a miss. |
| **E4** | `ValidateDiff` L0–L5a / L7. Envelope mismatches are marks, not INVALID. Historical replay without an envelope skips L4/L5a. L4 on both legs uses the producer rule (`max(anchor, F)`), not strict equality — a lagging honest sequencer lifting to the floor is not a dispute carrier. |
| **E5** | Signed host→host repair probe for missing acks. Outcomes `HEIGHT` / `UNREACHABLE` only — never `USER_CHEATING`, never a mark. |
| **E6** | Close-ready: host arms after `IdleTimeout` of wall-clock silence and emits nothing. |
| **E7** | `observed_height` / `observed_block_hash` on Start / Confirm / Finish. Confirm stamps are mirrored in `ExecutorReceiptContent`. Unstamped records keep the v2 root. |

Post-landing audit (steps 12–24 in `height-sync-audit-findings.md`): heartbeat loop on the gateway, fail-closed origin cache, bounded open-turn retain, no local-oracle fold into the SM tracker, compose through the log plane, capped marks, recovered `turn_seq`, host-only floor raise, oracle I/O off mutexes, repair budget/cancel, wire caps, saturating `HReq+D_ack`, `peer_seen` from host claims only.

### Cadence (the E6–E7 rewrite)

Scheduling is wall clock; evaluation stays in blocks so replay never consults a clock.

| Knob | Default | Job |
| ---- | ------- | --- |
| Interval | 6s (shipped; original landing used 3s) | Longest gap between full turnovers |
| TurnTimeout | 12s (2 × Interval) | Patience on one open turn; one dead slot cannot silence the cadence |
| IdleTimeout (`T_idle`) | 24s (4 × Interval) | Host silence budget before arming |
| AckDeadlineBlocks (`D_ack`) | **derived** | ⌈(Interval + TurnTimeout) / BlockTime⌉ + 1. Stamp slack; lateness is `observed_height > h_req + D_ack`. `D_ack = 1` made honest span acks late by construction. |
| DeltaBlocks (`D`) | 2 | SYNCED vs CATCHING_UP; Strong escalation is Phase F |
| WindowBlocks (`W_conf`) | 256 | Confirmation window, unaided floor raise, and how far a producer will carry F |

A **turnover** is `Q` distinct **host-signed** height claims: `MsgHeightAck`, or an executor stamp on ordinary inference. The user's own stamp is self-signed and does not count. `ORACLE_UNAVAILABLE` acks are **required**, echo `F(m)` from the log, and **count** toward `Q` (reachability, not a height witness). Busy inference therefore emits **zero** heartbeats (H2); a quiet session heartbeats every `Interval` (H1). Compose seed does not raise `F`; host-signed confirm/finish of the first inference land on the **next** nonce, so a quiet-heartbeat citest needs two chats before `Interval`.

`K_hb` and `MinRoundsPerBlock` are gone. Snapshot overlay: proto fields 13/15 kept their numbers and became `height_sync_interval_ms` / `height_sync_idle_timeout_ms`; field 18 is `height_sync_turn_timeout_ms`. Inference-chain still publishes zeros, so compiled defaults apply. Overlay scheduling knobs are validated and clamped; evaluation knobs (`D_ack`, `D`, `W_conf`, `BlockTime`) stay compiled.

---

## Wire

`DevshardTx` oneof:

| Field | Message |
| ----- | ------- |
| 9 | `MsgForceHeightSyncTurn` (transport; already allocated) |
| 10 | `MsgHeartbeat` |
| 11 | `MsgHeightAck` |
| 12/13 | reserved for cPoC `MsgSkipProbe` / `CarrySkip` |

Inference stamps are additive proto3 fields on existing txs (zero / empty = unstamped). `InferenceRecordProto` fields 18/19 (`started_at_height` / `confirmed_at_height`) are omitted when zero, so unstamped `post_state_root` bytes do not change.

Nothing on the scheduling side — Interval, TurnTimeout, IdleTimeout, last turnover time — is folded into `Diff` or a `SyncTurnRecord`. Turn state remains a pure function of the log.

---

## Log plane vs edge

| Plane | Height means | Rule | Replay |
| ----- | ------------ | ---- | ------ |
| Log (`Diff`) | reference height / logical time | L0 against `F(m)`; L0b same-executor confirm ≤ finish; L1–L3 framing/sig/causality; L7 vector | yes — same INVALID |
| Edge (envelope) | first-party follower tip | L4 binding (producer rule); L5a D band; (C-quorum) | **no** — skipped on catch-up / gossip (`sec = nil`) |
| Oracle (deferred) | pair vs chain | L6 hash-at-H | mark only; not a diff verdict |

L5b (in-log `D` band) is **withdrawn**. Divergence between followers is monitoring. L5a may refuse an exchange and never invalidates a Diff.

---

## Behaviour that must not regress

* **Opt-in (hosts).** Unset `DEVSHARD_CHAINORACLE_URL` / `DEVSHARD_HEIGHTSYNC` ⇒ no envelopes, no heartbeats, existing inference path unchanged. Gateway stamps stay courier-only; `DEVSHARD_GATEWAY_CHAIN_ORACLE` defaults **off** and must not skip host seed.
* **Busy escrow pays nothing.** Host-stamped Start/Confirm/Finish discharges the cadence (H2, H33). The user's own stamp does not (H2 companion).
* **Quiet escrow still syncs.** A heartbeat turn opens within `Interval` of the last turnover (H1), **after** host-signed `F` exists. The gateway loop calls `MaybeHeartbeat`; span dispatch is concurrent and continues on one slot's error.
* **Repair never blames.** Missing ack ⇒ `HEIGHT` or `UNREACHABLE`; the turn stays `degraded` (H17–H20).
* **Arming emits nothing.** Silence past `T_idle` sets a local flag only (H21–H23). A missing ack is never a reason to arm.
* **Replay is deterministic.** Two verifiers with the same `Diff` compute the same `SyncTurnRecord`, the same `F(m)`, and the same L0–L3 / L7 verdicts. L4 and L5a need the envelope and are skipped on catch-up / gossip ingest. Admission refusals do not feed the floor (attack 24e).
* **Floor is host-only.** Sequencer heartbeats never raise `F`. Honest compose omits a host raise that does not match this hop's envelope.
* **No Strong, no slash.** Marks are local. Nothing in this PR produces a `LightBlock`, an evidence packet, or a dispute tx.

---

## Out of scope / next releases

| Deferred | Why it is not this PR |
| -------- | --------------------- |
| Strong / `LightBlock` / `\|Δ\| > D` escalation | Envelope field 9, `(C-strong)` / `(C-hybrid)`, and `CATCHING_UP` forcing the next heartbeat to Strong. Until then `CATCHING_UP` is a label only. |
| Evidence packets, on-chain `MsgHeightSyncEvidence`, slashing, cross-session equivocation | Canonical half of a packet is a LightBlock. Marking lands here; adjudication does not. |
| L6 as a verdict or a dispute | The pair is already in Diff when L6 can see it. INVALID would split verifiers whose oracles lag. Marks exist; a persistent recheck queue and dispute promotion do not. `F` is not rolled back. |
| cPoC skip-probe carriers (oneof 12/13 reserved only) | Separate protocol. |
| Switching `timeout.go` / the seal clock onto stamped heights | State-root change; needs a version gate after the stamps have been carrying. |
| Enabling height-sync on default compose | Opt-in until the sibling dapi mount and ops surface exist. |
| Production dapi `/block/*` | Sibling [`ak/height-sync-protocol-dapi`](https://github.com/gonka-ai/gonka/tree/ak/height-sync-protocol-dapi). |

Gateway `/metrics` and `GET /v1/debug/heightsync` (plan §8.12, E8) **did** land later on this branch: [#1621](https://github.com/gonka-ai/gonka/pull/1621). H26/H27 container + H39–H49.

---

## Config

| Env | Default | Effect |
| --- | ------- | ------ |
| `DEVSHARD_CHAINORACLE_URL` | unset | HTTP oracle; empty ⇒ no height-sync on the host |
| `DEVSHARD_HEIGHTSYNC` | unset | 1/true enables chain-only height-sync without a dapi URL |
| `DEVSHARD_HEIGHTSYNC_K` | 10 | nonce cadence of Anchors (transport `K`, not the heartbeat interval) |
| `DEVSHARD_HEIGHTSYNC_SLOTS` | 1 | sync-turn width |
| `DEVSHARD_HEIGHTSYNC_PROBE_INTERVAL` | 30m | re-probe dapi after a transport miss |
| `DEVSHARD_GATEWAY_CHAIN_ORACLE` | unset / off | Gateway local follower for trust labels only; must not skip host seed |

Runtime-config overlay (`Snapshot.HeightSync`, proto 13–18): zero means keep the compiled default. `Validate` requires `D_ack · BlockTime ≥ Interval + TurnTimeout`, `T_idle > Interval + TurnTimeout`, and `2 · Interval ≤ F` (originator freshness). An overlay that would fail is clamped to compiled defaults.

---

## Test plan

```bash
cd devshard
GOMODCACHE="$HOME/go/pkg/mod" GOCACHE="$HOME/Library/Caches/go-build" \
  go test ./heightsync/... ./transport/... ./user/... ./host/... ./state/... ./chainoracle/...

# in-process e2e (held-response cases need -tags=dev)
GOMODCACHE="$HOME/go/pkg/mod" GOCACHE="$HOME/Library/Caches/go-build" \
  go test ./testenv/scenarios/ -run HeightSync -count=1
GOMODCACHE="$HOME/go/pkg/mod" GOCACHE="$HOME/Library/Caches/go-build" \
  go test -tags=dev ./testenv/scenarios/ -run HeightSync -count=1

# container (opt-in compose patch); two chats seed F before quiet Interval
make -C testenv citest-height-sync
```

Also `cd common && go test ./runtimeconfig/... ./nodemanager/...` for the ms overlay mapping.

Catalog: [`height-sync-tests.md`](devshard/docs/height-sync-tests.md). Phases A–D are §2–§6. Log plane is §7. Floor / stamp: H2, H28–H31, H33, H54–H58, H89–H91. L6 marks only: H13e, H14, H57. Host-claim overlays A/B/C: `citest-height-sync` + [`heightsync_host_claims.feature`](devshard/testenv/scenarios/heightsync_host_claims.feature).

Operator Gherkin: [`v5-manual-height-sync.md`](devshard/docs/v5-manual-height-sync.md). Residual (warm cutover, host ping, proxy stamp): [`v5-manual-residual.md`](devshard/docs/v5-manual-residual.md).


bonujel:
Focused review of the height-sync **log-plane security core** (height floor `F(m)`, signature coverage, replay determinism, and quorum). Overall, the core looks solid: user/sequencer-composed stamps do not raise the floor; executor, ack, and finish heightstamps are signature-covered; `Observe` folds only applied diffs; and L4/L5a are skipped without an envelope.

I noted one recovery-path concern that may affect deterministic L0 validation, plus one low-severity freshness-hardening opportunity. Details inline.

### F1: `devshard/state/heightsync.go:229`

**Could a `GetDiffs` failure during restore leave L0 operating without the previously established floor?**

`heightSyncFloor` is derived state and is rebuilt from diffs `1..LatestNonce`. On this error path, `h_last` and `turn_seq` are preserved, while the newly initialized floor remains empty and `RestoreState` returns normally.

This appears reachable when the snapshot is already at `meta.LatestNonce`: the host recovery path performs no subsequent replay, and the user recovery path also has a snapshot-only success path.

With an empty floor, `AsOf(m)` has no established value, so `checkL0` skips the regression check. This may allow a recovered verifier to accept a stamp below the floor reconstructed by a healthy verifier from the same pre-snapshot log, leading the two nodes to produce different L0 verdicts from the same history.

Would it make sense to treat floor reconstruction failure as fatal or retriable and propagate it through `RestoreState`, rather than proceeding with the snapshot scalars and an empty floor? A regression test could establish a nonzero floor, inject a restore-time `GetDiffs` failure, and verify that recovery does not proceed until the floor has been reconstructed.

### F2: `devshard/heightsync/inbound.go:109`

**Non-blocking: consider bounding future-dated originator timestamps in the freshness check.**

`now.Sub(time.UnixMilli(ts)) <= freshness` also evaluates to true for timestamps in the future because the resulting duration is negative. A sufficiently future-dated `OriginatorTimestampMs` may therefore remain fresh until that future time plus the configured freshness window. The same predicate also appears in `HeightSyncPeerTips.isFreshLocked`.

Request-leg carry-forward does not verify the origin signature—the existing test explicitly permits an unsigned carry—so the carrier can choose this timestamp. Signature coverage still protects verified response-origin evidence, but does not constrain this request-path freshness check.

This appears low severity: the `(height, hash)` must still pass the local `D`-band, the timestamp does not directly raise the durable floor, and it cannot create valid host-signed dispute evidence. The likely effect is limited to keeping a locally plausible carried tip eligible longer than intended.

Would it be worth allowing a small clock-skew tolerance while bounding both sides, for example `age >= -maxClockSkew && age <= freshness`, consistently in inbound validation and peer-tip selection? Tests for timestamps beyond the permitted future skew could help keep the two checks aligned.

a-kuprin:
@bonujel 


> ### F1: `devshard/state/heightsync.go:229`
> **Could a `GetDiffs` failure during restore leave L0 operating without the previously established floor?**

This is not divergent L0 on the same history.

The snapshot already contains diffs `1..N` and their state root. Restore does not re-run L0 on that prefix, and an empty floor cannot invent a higher `F` than that log already committed.

The replica is also not stuck with an empty floor. `F` in the verifier is a fold of diffs, but what each host *writes* is `max(oracle tip, F)` (or omit). After a failed `GetDiffs`, `F` is unknown, so that degenerates to the host’s live oracle height. That height is the current chain tip — at least the historical floor (reorg is omission / `FloorOutOfReach`, not a lower honest stamp). The next host-signed ack/confirm/finish puts that tip in `Diff`, and `Observe` re-seeds `F` from it.

Example: before crash, stamps have established `F = 1000`. Snapshot is at nonce `N`. `GetDiffs` fails; this replica comes up with an empty fold. Hosts still see oracle tip `1005`. They stamp `1005`, not a height from the missing diffs. Recovered replica skips L0 and then raises `F` toward `1005`; a never-restarted replica checks `1005 ≥ 1000` and accepts. Same honest verdict.

The only differing case is a regressing stamp `h = 500`: recovered replica would apply it; a healthy replica returns `INVALID(height_regression)`. That needs a malicious or buggy producer, and it cannot settle — close still needs `>2/3` sampled weight from hosts whose oracles and floors are at `~1000` and will not sign that fork.

So the hole is “this replica is more permissive after a store that loaded a snapshot but cannot read diffs.” That was the H102 liveness choice. Fail-closed would be tidier ops; it is not a cheaper attack on the escrow and not a log-plane security break.




> ### F2: `devshard/heightsync/inbound.go:109`
> **Non-blocking: consider bounding future-dated originator timestamps in the freshness check.**

We are not targeting at current release reworking originator timestamps. We have currently added height-stamps, that should replace timestamps in further releases.

Height-stamps are protected with height-sync protocol, and are consensus and could be trusted. Also deffered work on height-sync is disputes mechanisms. And any party who fake a (height, hash) pair could be caught and slashed

bonujel:
@a-kuprin Thanks for the detailed reply — the `max(oracle tip, F)` framing is the piece I was missing, and it holds up.

I went back through it on the branch. `referenceStamp` is `max(own_tip, F(m))` or omit (`devshard/host/heightsync.go:190`), `referenceStampLocked` has the same shape on the sequencer side (`devshard/user/heartbeat.go:331`), and your re-seeding argument follows from that: with `F` unknown the expression degenerates to the host's live oracle height, the next host-signed leg carries it into `Diff`, and `Observe` folds it back. Your worked example is right, and I'd argued F1 from the verifier side without checking the producer path. F2 as deferred work makes sense to me too — if height-stamps replace originator timestamps, hardening the freshness predicate now is wasted effort.

One thing I'd still like your read on, though: the "malicious or buggy producer" step.

A stale-floor replica seems to be a third case, and `checkL0`'s own comment lists it as one (`devshard/heightsync/logplane.go:320`):

```go
// Honest producers lift to F(m) or omit, so this firing is a
// shipping bug, a stale-floor replica, or authored misbehaviour.
```

It's the same `!known` branch you described, just read from the producer's side. A host whose own fold is missing takes `if !known || floor <= height { return height, hash }` and stamps its raw tip with no lift. A lagging oracle doesn't stop it — `evaluateSyncState` returns `CATCHING_UP` rather than `ORACLE_UNAVAILABLE` (`devshard/heightsync/syncstate.go:56`), and `buildHeightAckLocked` only gates on `ORACLE_UNAVAILABLE`. So a host that restarted, is behind, and has no fold emits a height below the floor the roster holds, without being malicious or buggy.

I ran it on `devshard-0.2.15-v5` at `03078d6e`, reusing the `failGetDiffsStore` already in `heightsync_snapshot_test.go`. Two replicas restored from the same snapshot over the same durable log, one with `GetDiffs` succeeding and one failing:

```
replica A (journal read ok): ApplyDiff(nonce 4) -> INVALID(height_regression):
                             heartbeat height 10 < floor 50 as of nonce 4;  LatestNonce=3
replica B (GetDiffs failed): ApplyDiff(nonce 4) -> nil;                     LatestNonce=4
next diff at nonce 4:        A -> nil,  B -> invalid nonce: expected 5, got 4
```

Since `applyCore` rejects the whole diff on an L0 failure (`devshard/state/machine.go:653`), this isn't one tx dropped from a set — it's one replica advancing the log while the other refuses it, after which neither can take the other's next diff.

Two caveats I'd rather state than gloss over:

- The state root doesn't cover `LatestNonce`, and heartbeat/ack apply is admission-only, so the split above isn't automatically a root mismatch. It becomes one as soon as an accepted diff touches root-covered state — with `FeePerNonce = 1` the same fixture gives the two replicas different roots straight away. So the verdict split and the root split are one failure with a delay, not two claims.
- I haven't shown a real deployment reaching "producer's own fold missing *and* its oracle behind" at the same time. The mechanism reproduces once that state exists; reachability is the part I can't demonstrate, and it may be rare enough that the current behaviour is the right trade.

What kept me on it rather than dropping it is that `observeHeightSyncLocked` justifies folding only on apply in these terms: "two honest verifiers different floors and therefore different L0 verdicts for every later diff — an escrow split" (`devshard/state/heightsync.go:199`). The restore branch appears to reach that state by a different route. And #1649 and #1653 have since restored `pendingTxKeys`, `committedEntries` and `sealedNonces` for snapshot-only recovery, which leaves `heightSyncFloor` as the one derived index still left empty by design — so I wanted to check that's deliberate rather than an oversight.

If it is deliberate and fail-closed costs more liveness than this is worth, that's a call I'm happy to defer to you on. In that case a line in the restore comment saying so might save the next reader from filing it again. I have the two tests locally — happy to paste them here if they'd be useful.

No rush on any of this — thanks for taking the time on the first round.

a-kuprin:
Added enhancement note:
https://github.com/gonka-ai/gonka/issues/1665

a-kuprin:
> @a-kuprin Thanks for the detailed reply — the `max(oracle tip, F)` framing is the piece I was missing, and it holds up.

@bonujel thanks for your review. Yes it is really better to fail when restoring with no diffs, then omit the height floor, derived from these diffs.

Please take a look on a fix:
https://github.com/gonka-ai/gonka/pull/1669

devshard/state/heightsync.go
```diff
@@ -0,0 +1,356 @@
...
+	defer sm.mu.RUnlock()
+	h, hash, known := sm.heightSyncFloor.AsOf(nonce)
+	return h, append([]byte(nil), hash...), known
+}
+
+func (sm *StateMachine) observeHeightSyncLocked(nonce uint64, txs []*types.DevshardTx) {
+	if sm.turnTracker == nil {
+		return
+	}
+	sm.turnTracker.Observe(nonce, txs, heightsync.LogResidentHeight(txs, sm.turnTracker.LastCompletedHeight()))
```
Ryanchen911, line 213: **P1 (high): a single user-signed stamp can degrade every open heartbeat turn.**

`observeHeightSyncLocked` folds **every** Diff-resident stamp into the height passed to `TurnTracker.Observe`:

```go
sm.turnTracker.Observe(nonce, txs, heightsync.LogResidentHeight(txs, sm.turnTracker.LastCompletedHeight()))
```

and `LogResidentHeight` includes user-signed `MsgHeartbeat` / `MsgStartInference` stamps (`stamp.go` iterates `TxStamp` over all txs). Meanwhile `TurnTracker` admits any stamp into `stampHeight` via `inferenceTxStamp`, and `Observe` ratchets `hNow` up to `max(hNow, stampHeight)` before `AdvanceHeight`. `windowClosed` then tests `hNow > h_req + D_ack`.

**Attack:** a malicious user puts a `MsgHeartbeat` with `observed_height = 1<<40` and an arbitrary 32-byte hash into any diff. L0 is satisfied (floor is 0 or lower, and the check is `h < floor`), L1–L3 pass (only framing / ack signature / ack causality — a lone heartbeat has no ack to constrain it). On apply, `LogResidentHeight` returns `1<<40`, `Observe` passes it to `AdvanceHeight`, and **every open turn degrades immediately** (`hNow > h_req + D_ack`).

I verified this with a focused PoC on this head: one `MsgHeartbeat{ObservedHeight: 1<<40}` turns turn 1 from `open` to `degraded`, and `ArmingContext()` reports `[1]` degraded.

**Consequences:**
- All heartbeat turns degrade in lockstep → close-ready arming is blocked (`ArmingContext` only arms on no-degraded-turns) and repair probes fire against honest hosts that did nothing wrong.
- It bypasses the PR's own headline property — "the floor is host-only" is true for the floor, but the logical clock that actually drives turn completion (`h_last` / turn state) is user-controlled.
- No state split (all verifiers derive the same degraded turns from the same log), so no direct fund loss today, but the cadence/reachability machinery this PR is built on is silently broken, and with L6 not wired (P3) there is no mark-based disincentive either.

**Suggested fix:** bound how far a user-signed (non-executor, non-ack) stamp may raise `hNow`/`stampHeight` — e.g. only host-signed stamps (`MsgHeightAck`, `MsgConfirmStart`, `MsgFinishInference`) may independently raise the turn clock, and user heartbeats may only carry up to `F(m) + W_conf`. This aligns the turn clock with the host-only floor rule.

devshard/heightsync/logplane.go
```diff
@@ -0,0 +1,715 @@
...
+	prevTurn := uint64(0)
+	if st.Tracker != nil {
+		prevTurn = st.Tracker.MaxTurnSeq()
+	}
+	for _, ref := range hbs {
+		hb := ref.hb
+		if hb.TurnSeq == 0 {
+			return fmt.Errorf("%w: turn_seq 0", ErrBadFraming)
+		}
+		if prevTurn > 0 && hb.TurnSeq < prevTurn {
```
Ryanchen911: **P2 (medium): `turn_seq` has no upper bound, so a malicious user can prune every honest turn record and capture the turn-sequence space.**

`checkL1` enforces only `TurnSeq == 0` and `TurnSeq < prevTurn`. A user can emit a heartbeat with `turn_seq = 1<<60`; the checks pass. `TurnTracker.prune` then computes `cutoff = maxTurnSeq - retain = 1<<60 - 64`, which deletes **all** honest turn records (seq 1..N), and `maxTurnSeq` is permanently pinned at `1<<60`.

**Consequences:**
- Honest hosts are thereafter forced to compose `turn_seq >= 1<<60` to pass L1 — the sequence space is captured by the attacker.
- `checkL7`'s `tracker.Record(turnSeq - 1)` now resolves to an empty/malicious record, so sync-vector reconciliation degrades.
- `HeartbeatNonceForSlot` / arming context all key off a poisoned `maxTurnSeq`.

**Suggested fix:** in `checkL1`, require `turn_seq` to be within a bounded window of `prevTurn` (e.g. `prevTurn < turn_seq <= prevTurn + slots_num` or a small cap), rather than merely `>= prevTurn`. Note the same lack of an upper bound exists on `MsgHeightAck.TurnSeq`.

devshard/state/heightsync.go
```diff
@@ -0,0 +1,356 @@
...
+		return types.ErrEmptyTx
+	}
+	if sm.state.Phase != types.PhaseActive {
+		return types.ErrSessionFinalizing
+	}
+	return nil
+}
+
+func (sm *StateMachine) checkLogPlaneLocked(nonce uint64, txs []*types.DevshardTx) error {
+	res := heightsync.CheckDiffLogPlane(logPlaneCtx(), heightsync.LogPlaneInput{
```
Ryanchen911, line 124: **P3 (medium): L6 (`DEFERRED_FAIL` attribution) is dead code in every production call path.**

`CheckDiffLogPlane` runs `checkL6` only when `in.Oracle != nil`, but both call sites here pass no `Oracle`, and `CheckEnvelopeBinding` (used at the transport edge) never calls `checkL6` at all. So the `DEFERRED_FAIL` marks and `carriedFrom` origin-attribution this PR describes as the accounting mechanism for bad actors can never fire in production.

This matters most in combination with P1: the only disincentive a malicious user faces for the `1<<40` heartbeat claim is a mark that is never recorded.

**Suggested fix:** either wire a real oracle into these call sites, or state plainly in the code (and PR description) that L6 is intentionally not wired until the dispute layer, so operators do not assume `DEFERRED_FAIL` is active.

devshard/chainoracle/blocks/client/lookup.go
```diff
@@ -0,0 +1,190 @@
...
+func NewLookup(cfg HTTPConfig) (*Lookup, error) {
+	if cfg.BaseURL == "" {
+		return nil, fmt.Errorf("blockoracle/client: empty base url")
+	}
+	if _, err := url.Parse(cfg.BaseURL); err != nil {
+		return nil, fmt.Errorf("blockoracle/client: invalid base url: %w", err)
+	}
+	hc := cfg.HTTPClient
+	if hc == nil {
+		hc = &http.Client{Timeout: 10 * time.Second}
```
Ryanchen911: **P4 (low): `Lookup` dials with a bare `http.Client` instead of the dial-time SSRF guard.**

`NewLookup` falls back to `&http.Client{Timeout: 10 * time.Second}`, i.e. the default transport, unlike `transport/client.go`, which builds its dialer through `httpguard.NewDialer()`. The BaseURL today comes from operator env (`DEVSHARD_CHAINORACLE_URL`), so this is lower risk than #1470, but the `httpguard` dial control is process-wide and cheap to apply — wiring it here keeps the new chain-oracle path consistent with the existing participant-controlled-URL handling and future-proofs it if the URL source ever moves on-chain.

Ryanchen911 (changes requested):
Thanks for your work! I reviewed this PR against the current head (`e28b27af`) and verified each finding by building the branch and running a focused PoC against `TurnTracker`/`LogResidentHeight`. CI is green and the design is strong overall — host-only floor, deterministic L0–L3, and the state-root split between committed `HeightSyncEscrowCommit` and derived runtime turn state are all right, and the executor-receipt stamp mirroring correctly closes the confirm blind spot. Four issues below, ordered by severity.

**Summary**

1. **P1 (high)** — a single user-signed stamp can degrade every open heartbeat turn, consistently across all verifiers. The log-plane claim that "logical time is host-only" holds for the floor but not for the clock that drives turn completion.
2. **P2 (medium)** — `turn_seq` has a lower bound but no upper bound, so a malicious user can prune every honest turn record and capture the turn sequence space.
3. **P3 (medium)** — L6 (`DEFERRED_FAIL` attribution) is dead code in every production call path: neither `CheckDiffLogPlane` call site passes an oracle, and `CheckEnvelopeBinding` never runs L6. The marks this PR advertises as the accounting mechanism for bad actors are not wired in.
4. **P4 (low)** — `chainoracle/blocks/client.Lookup` uses a bare `http.Client` instead of the dial-time `httpguard` that `transport/client.go` uses for participant-controlled URLs.

Details are inline.

devshard/state/heightsync.go
a-kuprin: Thanks a lot for finding. It will be addressed in separate PR soon

devshard/state/heightsync.go
a-kuprin: This is already documented, not a missing apply hook. We are deferring L6 for next releases

- Spec `devshard/docs/proposals/HEIGHT_SYNC_PROTOCOL_PROPOSAL.md`, **Log-plane checks (heartbeat pair)** L6 row; **Evaluation tiers — what may affect diff validity** (“Local oracle, deferrable”); **Garbage cleanup after L6 (Phase F)**; milestones table “Floor garbage cleanup after L6” (⏸).
- Plan `devshard/docs/height-sync-implementation-plan.md`, intro (“marking lands in E, adjudication lands with F”); **§8.7 Verifier — L0–L7 and the marking/adjudication split**; **E4 — Verifier L0–L7 + marks** (Deferred = L6); **Stay out of E**; **§12. Out of scope**.
- Code: `devshard/heightsync/logplane.go` `LogPlaneInput.Oracle` (“nil leaves pairs pending”), `CheckDiffLogPlane` (Deferred: L6), `CheckEnvelopeBinding` (“L4 and L5a only”), `checkL6` (`if in.Oracle == nil { return }`); `devshard/state/heightsync.go` `logPlaneCtx` (“neither call site supplies one”).

a-kuprin:
> Thanks for your work! I reviewed this PR against the current head (`e28b27af`) and verified each finding by building the branch and running a focused PoC against `TurnTracker`/`LogResidentHeight`. CI is green and the design is strong overall — host-only floor, deterministic L0–L3, and the state-root split between committed `HeightSyncEscrowCommit` and derived runtime turn state are all right, and the executor-receipt stamp mirroring correctly closes the confirm blind spot. Four issues below, ordered by severity.
> 
> **Summary**
> 
> 1. **P1 (high)** — a single user-signed stamp can degrade every open heartbeat turn, consistently across all verifiers. The log-plane claim that "logical time is host-only" holds for the floor but not for the clock that drives turn completion.
> 2. **P2 (medium)** — `turn_seq` has a lower bound but no upper bound, so a malicious user can prune every honest turn record and capture the turn sequence space.
> 3. **P3 (medium)** — L6 (`DEFERRED_FAIL` attribution) is dead code in every production call path: neither `CheckDiffLogPlane` call site passes an oracle, and `CheckEnvelopeBinding` never runs L6. The marks this PR advertises as the accounting mechanism for bad actors are not wired in.
> 4. **P4 (low)** — `chainoracle/blocks/client.Lookup` uses a bare `http.Client` instead of the dial-time `httpguard` that `transport/client.go` uses for participant-controlled URLs.
> 
> Details are inline.

@Ryanchen911 Please check the fixes here:
https://github.com/gonka-ai/gonka/pull/1697

All addressed except P3, as it is already documented https://github.com/gonka-ai/gonka/pull/1584#discussion_r3905104657

devshard/heightsync/logplane.go
a-kuprin: I've just removed `turn_seq` field, so user can't forge it. As it is anyway can be derived from nonce

https://github.com/gonka-ai/gonka/pull/1697

devshard/state/heightsync.go
a-kuprin: Finally I've just pinned it exactly to F that is logical time floor. Logical time is reported only by hosts, that are chain hosts (user isn't). So user just can't set logical time on his own, as he is not the source.

User stamp could close ack windows. Not `F`, not funds, not close-ready. **Landed:** sequencer stamps pinned to exactly `F` (above it is `MARK(height_unbacked)` at error level; no `F` yet means no heartbeat); turn clock is host-signed only.

a-kuprin: Also it is important that `F` not limited by `W_conf` now. `W_conf` parameter was a mistake, as height already limited by envelope height-sync protocol.
The idea is very simple, any height reported by host adjusts the floor, also if any host gives height larger then some host (oracle_height + D) this leads to strong prooving that is deffered to next release. So height is always bounded by `F` and `oracle_height + 2`, or source of height should proove it with inference chain artifacts.

So using here `F` just gives no space for malware user behavior

a-kuprin:
@Ryanchen911 fixes for your findings was merged from https://github.com/gonka-ai/gonka/pull/1697
Could you please approve, as requested changes landed to the branch

Ryanchen911 (approved):
Issues addressed, LGTM

gmorgachev (approved):
Approving as these version is on mainnnet already

PR #1566, opened 2026-08-09
Simple Dynamic Coefficients for Multi-Model PoC

gmorgachev:
## Motivation

To ensure reliable model availability and predictable throughput, Gonka needs a way to maintain target hardware allocations. Models like DeepSeek V4 Flash have highly hardware-specific performance profiles. For example, B300 outperforms H100 by over 7x due to native FP4 support. Manual tuning can temporarily balance incentives. Proposing a 0.21 coefficient attracts B300s to DeepSeek without disrupting other hardware ([calculations](https://docs.google.com/spreadsheets/d/1Tw4V7xEXR2p5MbCHqzqjS9vHXQ0eI1IHVXC6guEHnio/edit?gid=0#gid=0)).

However, static flat coefficients cannot maintain target allocations as network size evolves. Because one model always gives the highest weight, there is no clear incentive to have different models deployed on similar hardware. Similar hardware classes fully migrate to a single model simultaneously at the slightest threshold change. 

This proposal is an intermediate step toward demand-driven allocation, but in minimalistic form. Governance defines target shares and coefficient bounds, and the protocol dynamically adjusts base coefficients to align actual hardware distribution with those targets.

While this phase relies on direct governance, replacing active voting with aggregated allocation preferences submitted directly by hosts is low-hanging fruit.

## Summary

Flat coefficients fail to properly incentivize a multi-model network. Because relative nonce throughput on different hardware varies by model, flat coefficients cause entire hardware classes to migrate to a single model. Hardware distribution becomes a direct function of static coefficients rather than demand.

Fully automated demand-driven coefficients are complex, vulnerable to manipulation, and cause high redeployment churn. This PR proposes a simplified transitional approach:

- Governance sets target compute allocation shares per model (share of total compute capacity, which directly defines model throughput when network size is stable) based on demand signals (revenue, data from brokers, expectations from open source data, etc).
- Governance defines a bounded range [coeff_i_min, coeff_i_max] per model.
- The protocol dynamically adjusts base coefficients coeff_i within these bounds to align actual hardware distribution with target shares.
- To prevent sudden hardware jumps, any compute share exceeding the target is diluted at coeff_i_min within the epoch.

To avoid active governance voting for exact target shares, target determination can be delegated:

- Short term: direct target preferences submitted by hosts via transactions, aggregated and weighted on-chain.
- Long term: automated statistics derived from real inference usage and protocol revenue.

[...]

PR #1588, opened 2026-08-13
weightcap v0.2.16

DimaOrekhovPS:
## Summary

Adds `CapWeight` for governance, BLS, and PoC/cPoC validation voting. `Weight` remains the real weight used for rewards, confirmation, pricing, and selection.

For epoch N+1:

```
CapWeight_(N+1) = min(Weight_(N+1), P_N)
```

`P_N` is the participant's confirmed effective weight from epoch N. A new or returning participant has no `P_N`, so its first `CapWeight` is 0. It can earn rewards on `Weight`, but its governance, BLS, and validation voting power remain 0 until its compute is confirmed for one epoch.

A failed settlement `MissedStatTest` sets `P_N` to 0 for the PoC-derived participant set. A skipped test or test error leaves `P_N` unchanged. `INVALID` and `INACTIVE` participants are already excluded through live membership.

## Weight usage

- Governance and CometBFT power use `CapWeight`.
- BLS slot assignment uses `CapWeight`.
- PoC/cPoC validation voting powers and snapshots use `CapWeight`.
- Group eligibility and non-initial group caps use `P_N`.
- Rewards use `Weight` after the same participant-wide cPoC confirmation fraction used by `P_N`.

## Fallbacks

- If there is no previous epoch, `CapWeight` defaults to `Weight`.
- If no fresh PoC node survives seating, formation reuses eligible participants from epoch N.
- If every final `Weight` is 0, formation retries once with eligible epoch-N participants and halts if the retry is also 0.
- Epoch-N fallback ignores failed `MissedStatTest` results.
- If every `CapWeight` is 0 while some `Weight` is positive, formation sets `CapWeight = Weight` before persistence.
- Concentration capping runs on `Weight` before the previous-epoch cap and on `CapWeight` afterward.
- Missing previous-epoch group or live-member state halts formation.
- Guardians receive normal `CapWeight`. Temporary governance and BLS protection is applied afterward without changing stored weights or cPoC validation voting power.

## Testing

Tests cover cap calculation, participant-wide confirmation, weight routing, model eligibility, concentration, guardian ordering, BLS assignment, fallback and halt paths, error propagation, and pre-upgrade epochs.

See [`proposals/previous-epoch-weight-cap/README.md`](https://github.com/gonka-ai/gonka/blob/dl/upgrade-v0.2.16-weightcap/proposals/previous-epoch-weight-cap/README.md) for the full design.


gmorgachev (approved):

vitaly-andr:
Following the weight-cap change in `CalculateParticipantBitcoinRewardsWithTransfers`: a few lines below what you touched, at `bitcoin_rewards.go:918-923`, `workCoins` is taken from `CoinBalance` only when the status is exactly ACTIVE, while `accountsettle.go:272-273` clears `CoinBalance` regardless. A participant who flips to INVALID or INACTIVE mid-epoch gets no `SettleAmount` record at all, so the earned amount cannot be claimed afterwards, and unlike the RewardCoins remainder in the same function it is not sent to governance either.

Details and a failing test are in #1746. Raising it here because you were last in this function and the 0.2.16 milestone is still open.

vitaly-andr:
## chain-halt: zero-weight guard aborts finalization when the carry-over weight is zero

@DimaOrekhovPS, the guard keeps a zero-weight epoch from being sealed, which is what kept inference alive. One path out of it halts the chain instead. When the carry-over is seated but the retried pipeline still yields no positive weight, the error returns at `module.go:869` and `module.go:877` propagate out of `EndBlock`: no block is finalized and the height stops advancing.

bonujel:
On `upgrade-v0.2.16` (`e86e489`) the `*Halts` tests pass, so `onEndOfPoCValidationStage` does return these errors, and `EndBlock` (`:490`) passes them straight up. CometBFT treats a `FinalizeBlock` error as fatal, so every validator would stop at the same height. `:864` is the same path when the first pass already used the carry-over. The PR description calls this out ("halts if the retry is also 0"), so the open question is reachability.

I ran the formation fixture with PoC loss and collateral separated, and with mainnet's collateral params (`base_weight_ratio` 0.2, `grace_period_end_epoch` 180, already past), assuming the worst case of nobody holding collateral:

- no fresh PoC, inside the grace period → forms; the carry-over keeps its weight
- mainnet params, with or without fresh PoC → forms with 20% of the weight
- `FailedRetryHalts` setup with base 0.2 → recovers
- original test (base 0) → halts

So in the fixture a missed PoC round doesn't halt, and the trigger the tests use doesn't hold under current params. What I'd still look at:

1. Safety rests on `base_weight_ratio > 0`. At 0, any epoch where no seated participant holds collateral halts — that's the last case above.
2. #1566 added two more halt paths when it made `runWeightPipeline` return an error (`:855` and `:874`). They aren't in this PR's design, and I haven't traced what triggers them.
3. `onSetNewValidatorsStage` (`:597`) keeps the current validator set on zero power. @DimaOrekhovPS, any reason formation shouldn't do the same rather than halt?

@vitaly-andr is there a trigger you have in mind that works with current params?

PR #1590, opened 2026-08-13
fix(mlnode): accept DAPI's JSON POSTs sent without Content-Type

baychak:
DAPI posts JSON to `/api/v1/inference/up` and `/api/v1/inference/pow/init/generate` without a Content-Type header. Whether FastAPI tolerates that depends on its version: since `strict_content_type` (default **True** in current releases) a header-less body reaches `model_validate` unparsed and the request 422s before any handler runs — this is what production hit on the July images. The api lock currently pins a forgiving fastapi 0.115, but the constraint is `>=0.115.8`, so whether the node accepts the network's requests silently rides the pin.

A plain ASGI middleware injects `application/json` on `POST /api/*` when the header is missing; anything carrying a Content-Type passes through untouched. Tests assert at the ASGI level (what the inner app sees in scope) — a handler-level test would pass on the forgiving pin with no middleware at all.

We have shipped this as a build-time layer on top of the published mlnode image since July 24. The root cause lives in DAPI (Go client not setting the header); this keeps mlnode tolerant either way and outlives a DAPI-side fix harmlessly.

### DIFF
--- a/mlnode/packages/api/src/api/app.py
+++ b/mlnode/packages/api/src/api/app.py
@@ -31,7 +31,7 @@ from api.metrics.routes import router as metrics_router, metrics_enabled
 from api.metrics.allowlist import SCHEMA_VERSION, VLLM_ALLOWLIST
 from api.metrics import xid_source
 from api.watcher import watch_managers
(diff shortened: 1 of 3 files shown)


baychak:
Closing in favour of fixing the sender: `SendPostJsonRequest` in `common/utils/http.go` builds the JSON body and never sets `Content-Type`. A two-line change there covers every dapi call to mlnode; the middleware here only papered over it on the receiving side. PR follows.

PR #1595, opened 2026-08-14
fix(devshard): let a node replay its own pre-floor diffs on startup

qdanik:
## Problem

`enforce inference tokens floor (#1391)` added a `MinTokensFloor` check to `applyStartInference`. That check also runs on **replay**, so a node that served inferences before the floor landed cannot restart — it re-runs a rule written after its own diffs were made:

```
replay nonce 12: max_tokens below min_tokens floor: max_tokens 1, floor 64
```

Refusing the diff undoes nothing. It is already part of a recorded state root; the refusal only keeps the node down.

## Fix

Separate *authoring new work* from *replaying work already accepted*.

- `state.ApplyPersisted` sets `replayingPersisted` under the same mutex as the apply and clears it via `defer`
- the floor check is skipped only on that path
- both replay call sites switch to it: `user/recover.go` (gateway) and `cmd/devshardd/session/manager.go` (host daemon)

Exactly one check is relaxed, and only the one guarding creation of new work. Live traffic still has to meet the floor.

## Tests

| Test | Package | Covers |
|---|---|---|
| `TestApplyPersisted_ReplaysAStartWrittenBeforeTheFloor` | `state` | the policy split, and that the relaxation does not outlive the call |
| `TestRecoverSession_ReplaysADiffWrittenBeforeTheMaxTokensFloor` | `user` | the gateway's recovery path, from a seeded sub-floor diff |
| `TestRecoverSessions_ReplaysADiffWrittenBeforeTheMaxTokensFloor` | `cmd/devshardd/session` | the host daemon's recovery path |

Each fixture asserts `ApplyLocal` still rejects the diff before seeding it, so the test cannot silently stop reproducing the failure.

**Mutation-tested — 4/4 killed**, each verified to compile first: floor ignores the replay flag; flag never cleared; each replay site reverted to `ApplyLocal`.

`state`, `user` and `cmd/devshardd/session` pass in full.

### DIFF
--- a/devshard/state/machine.go
+++ b/devshard/state/machine.go
@@ -91,6 +91,8 @@ type StateMachine struct {
 	// validation lives in committedEntries (and on disk in the snapshot).
 	sealedNonces   map[uint64]uint64
 	inferenceStore storage.Storage
+	// replayingPersisted is written only under mu, by ApplyPersisted.
+	replayingPersisted bool
 
 	// Lookup maps derived from group at construction time.
 	slotToAddress      map[uint32]string
@@ -311,6 +313,16 @@ func (sm *StateMachine) verifyDiffUserSig(diff types.Diff) error {
 func (sm *StateMachine) ApplyLocal(nonce uint64, txs []*types.DevshardTx) ([]byte, error) {
 	sm.mu.Lock()
 	defer sm.mu.Unlock()
+	return sm.applyCore(nonce, txs, nil, "user")
+}
+
+// ApplyPersisted replays a diff this node already accepted. A diff is part of a recorded state root,
+// so a rule that tightened since it was written cannot undo it by refusing it, only fail to start.
+func (sm *StateMachine) ApplyPersisted(nonce uint64, txs []*types.DevshardTx) ([]byte, error) {
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+	sm.replayingPersisted = true
+	defer func() { sm.replayingPersisted = false }()
 	return sm.applyCore(nonce, txs, nil, "user")
 }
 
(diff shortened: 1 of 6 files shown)


devshard/state/machine.go
```diff
@@ -314,6 +316,16 @@ func (sm *StateMachine) ApplyLocal(nonce uint64, txs []*types.DevshardTx) ([]byt
 	return sm.applyCore(nonce, txs, nil, "user")
 }
 
+// ApplyPersisted replays a diff this node already accepted. A diff is part of a recorded state root,
+// so a rule that tightened since it was written cannot undo it by refusing it, only fail to start.
+func (sm *StateMachine) ApplyPersisted(nonce uint64, txs []*types.DevshardTx) ([]byte, error) {
+	sm.mu.Lock()
+	defer sm.mu.Unlock()
+	sm.replayingPersisted = true
+	defer func() { sm.replayingPersisted = false }()
```
a-kuprin, line 326: Seams it should be before `defer sm.mu.Unlock()`, as it can be set to false after unlock and race is possible

devshard/state/machine.go
```diff
@@ -314,6 +316,16 @@ func (sm *StateMachine) ApplyLocal(nonce uint64, txs []*types.DevshardTx) ([]byt
 	return sm.applyCore(nonce, txs, nil, "user")
 }
 
+// ApplyPersisted replays a diff this node already accepted. A diff is part of a recorded state root,
+// so a rule that tightened since it was written cannot undo it by refusing it, only fail to start.
+func (sm *StateMachine) ApplyPersisted(nonce uint64, txs []*types.DevshardTx) ([]byte, error) {
```
a-kuprin: I would better name it `ApplyLocalPersisted` to keep some mental link with `ApplyLocal`

and keep the comment from `ApplyLocal`:

```go
// ApplyLocal applies txs without signature verification. Used by the user
// to compute the post_state_root before signing the diff.
```

devshard/state/machine.go
qdanik: Multiple defer calls in a single function operate on a LIFO (Last In, First Out) basis.

a-kuprin: right, my miss

a-kuprin (approved):

PR #1596, opened 2026-08-14
Stabilize ML node mock assertions in event listener tests

aikuznetsov:
## Summary
- fix a flaky API wrapper test failure where `MockClient.WithTryLock` failed immediately on transient lock contention
- serialize test assertions with async mock client calls instead of treating an in-flight background worker as a test failure

Tes failure link: https://github.com/gonka-ai/gonka/actions/runs/31797682156/job/94758260624?pr=1594

## Root cause
`TestRegularPocScenario` can race with `nodeStatusQueryWorker`: the worker may hold the mock client mutex while the test reads call counters. The previous `TryLock` helper failed immediately with `TryLock called more than once`, even though this is expected transient contention in async tests.

## Verification
- attempted `go test ./internal/event_listener -run TestRegularPocScenario -count=5 -v` locally; local Docker runner failed before test execution on `github.com/supranational/blst` (`undefined: Message`), which appears to be local toolchain/arch-specific and not this change.

### DIFF
--- a/decentralized-api/mlnodeclient/mock.go
+++ b/decentralized-api/mlnodeclient/mock.go
@@ -84,13 +84,11 @@ func NewMockClient() *MockClient {
 	}
 }
 
+// WithTryLock serializes test assertions with async mock calls. The name is
+// historical; blocking here avoids flaky failures on transient lock contention.
 func (m *MockClient) WithTryLock(t *testing.T, f func()) {
-	lock := m.Mu.TryLock()
-	if !lock {
-		t.Fatal("TryLock called more than once")
-	} else {
-		defer m.Mu.Unlock()
-	}
+	m.Mu.Lock()
+	defer m.Mu.Unlock()
 
 	f()
 }


a-kuprin (approved):

redstartechno:
Same root cause and effectively the same fix as #1495, which has been open against `main` since 22 July — flagging it so the same flake doesn't get reviewed twice.

Your PR is on the better base (`upgrade-v0.2.16`), so I'm happy for this to be the one that lands. I'll close #1495 as absorbed once it merges.

Two small things #1495 also does, in case you want to fold them in here:

- With the `TryLock` branch gone, the `t *testing.T` parameter is unused, so the helper can become `WithLock(f func())` and `mock.go` no longer needs to import `testing` — it's a regular source file, not a `_test.go`.
- That's an 11-call-site rename, all in `decentralized-api/internal/event_listener/integration_test.go`.

For prior evidence that this is a recurring CI flake rather than a one-off: two independent occurrences on diffs that can't touch this package — #1460's CI on 2026-07-16 (diff is `devshard/host` only) and https://github.com/gonka-ai/gonka/actions/runs/29468383638 — plus a junit artifact from a red run on `main` itself. Details in #1495.

On your verification note: the `blst` `undefined: Message` failure is a local toolchain/arch issue and not caused by your change — I hit the same wall building natively on Windows and run this package under WSL instead. If it's useful I can run `TestRegularPocScenario` against this branch there and post the output.

aikuznetsov:
Closing this PR because #1495 already covers the same flaky test fix and has been approved. 
Thanks @redstartechno

PR #1587, opened 2026-08-13
Add rolling-safe HA for versiond routing

snevolin:
> [!IMPORTANT]
> GitHub requires the base branch to belong to the target repository, so this
> PR is opened against `gonka-ai/gonka:upgrade-v0.2.15`. Functionally it is
> stacked on `snevolin/gonka:sn/versiond-host-evacuation-0.2.15` and should be
> reviewed relative to that branch.
>
> - Conceptual base: `b1f28c12c5157bc49775b5bfa689d0d67a27b28d`
> - Net diff: `b1f28c12c5157bc49775b5bfa689d0d67a27b28d..c0342b4837f5e79af1e08908fa11f308b1796bde`
> - Feature commits: `740ad0902a9ef7b76557bc1179a2a22e590129b1^..c0342b4837f5e79af1e08908fa11f308b1796bde` (41 commits)
> - [Fork comparison](https://github.com/snevolin/gonka/compare/sn/versiond-host-evacuation-0.2.15...sn/versiond-router-ha)
>
> GitHub's Files changed view also includes the prerequisite host-evacuation
> series between `upgrade-v0.2.15` and `b1f28c12c`; that series is outside this
> PR's conceptual scope.

## Summary

This change turns `versiond-router` from one process into a replicated,
stateless routing tier. Three independent router slots derive the same
version-aware pools and escrow-based consistent-hash placement from Docker DNS,
active checks, and DAPI's existing `/versions` catalog. They require no shared
routing database, leader, or replica-to-replica coordination.

The PR also supplies the operational machinery needed to use that redundancy
safely: rolling fleet reconciliation, end-to-end admission checks, automatic
route discovery, crash-recoverable host upgrade, exact rollback, release
preflight, and production acceptance coverage.

## Architecture

```text
clients
   |
   v
proxy-router (public HAProxy, :80/:443)
   |
   +--> proxy-policy A/B (private nginx: TLS, HTTP/2, CORS, limits, rewrites)
           |
           +--> ordinary API and node routes
           +--> /devshard/* --> proxy-router :18081
           |                       |
           |                       +--> versiond-router slot 0
           |                       +--> versiond-router slot 1
           |                       +--> versiond-router slot 2
           |                                  |
           |                                  v
           |                         versiond --> devshardd
           |                                  |
           |                           shared PostgreSQL
           |
           +--> Tier A /v1/* --> proxy-router :18082 --> edge-api pool
```

HAProxy owns membership, health, and replica selection. nginx keeps the
[...]

PR #1601, opened 2026-08-18
Fail closed on unsafe HA storage

snevolin:
> [!NOTE]
> This focused PR extracts the HA storage guard that was previously implemented
> inside the broader [#1517](https://github.com/gonka-ai/gonka/pull/1517) and
> later carried through [#1587](https://github.com/gonka-ai/gonka/pull/1587).
> Those umbrella PRs were closed so their independent safety, routing,
> lifecycle, and deployment changes could be reviewed and merged separately.

## Merge Order

Merge this PR after #1642, **Unify devshard boolean environment parsing**.

This PR uses the shared parser from #1642 for `GONKA_HA`. That prerequisite is already merged into the base branch, so this PR contains only the HA storage guard changes.

## Summary

Fail closed when a multi-instance devshard deployment is configured with storage that cannot safely be shared between replicas.

- mark the versiond HA Compose topology with `GONKA_HA=true`;
- refuse to start `devshardd` in that topology unless `DEVSHARD_STORAGE_MODE=postgres` and `PGHOST` is set;
- apply the same guard during `--print-storage-mode` so versiond rejects an unsafe child before starting it;
- parse `GONKA_HA` through the shared devshard boolean grammar and reject unknown values;
- parse `Devshard-Ha` request headers strictly and reject malformed or repeated values with `503`;
- retain the request-time storage guard as defence for partial rollouts and accidental pool expansion.

Single-instance deployments are unchanged: with `GONKA_HA` disabled, SQLite remains supported.

## Why

Several `devshardd` replicas may receive requests for the same escrow in an HA pool. SQLite and hybrid storage retain local fallback state, so accepting that traffic can split one logical session across independent databases.

The existing request middleware intended to prevent this, but an unknown `Devshard-Ha` value was interpreted as non-HA and a missing router header left no deployment-level protection. A misconfigured HA process could therefore boot successfully and bypass the storage guard.

This change makes the deployment declaration authoritative at startup and keeps the request marker as an independent second line of defence.

## Scope

This PR validates storage **mode**, not PostgreSQL operation. It does not probe database readiness, compare database identities, change the PGDATA layout, or migrate data; those concerns remain independently reviewable changes.

## Validation

- `go test -race ./storage/mode` from `common/`
- `go test -race ./cmd/devshardd` from `devshard/`
- `go test -race ./internal/process` from `versioned/`


bonujel:
Went looking for a way past the guard and didn't find one. Listing what I checked, in case it saves the next reviewer the trip:

- Compose setting `GONKA_HA=true` on the shared anchor can't force a single-instance stack onto Postgres — the overlay brings up versiond2 and the router itself, and the anchor already pins `DEVSHARD_STORAGE_MODE=postgres` and `PGHOST`.
- `devshardstorage.NewStorage` has one other caller, `cmd/devshard-host`, which is the e2e image rather than a pool member. `buildHostManager` is the production path and the guard sits in it.
- Payload storage resolves through the same `common/storage/mode`, so it can't end up in a different mode than the sessions the guard checks.
- `proxy_set_header Devshard-Ha $devshard_ha_header` overrides whatever a client sends, and single-host renders the map to `""`, which nginx drops rather than forwards — so the header stays a router assertion.
- The `""` asymmetry between the two parsers (present-empty header is HA, empty env is not) looks deliberate: presence is the router's signal, absence is the default deployment shape. Both fail in the safe direction.

Two small things, neither blocking.

`RequireHADeploymentStorage()` runs inside `buildHostManager` (app.go:247), after `payloads.Open` has already created the payload dir at app.go:98. Only an `MkdirAll` gets there first, so nothing is at risk — but a boot guard reads more naturally before anything touches storage.

The comment on `HADeployment()` says the grammar is shared with the router's entrypoint. That holds once #1599 lands, where `bool_env` mirrors it down to the error string. Since the note says this one is meant to stand alone, it's worth a "(from #1599)": today's versiond-router derives HA from `HOST_COUNT > 1` and never reads `GONKA_HA`, so following the comment leads somewhere confusing.

The assertion I wouldn't have thought to write is `TestHADeploymentGrammar` requiring the error to contain `not a boolean`. The original bug was a value that parsed to the wrong answer without saying anything, so pinning that the refusal names its reason is the part that keeps it from coming back.

common/storage/mode/ha.go
```diff
@@ -58,3 +74,49 @@ func RequireConfiguredForHA() error {
...
+	switch strings.ToLower(strings.TrimSpace(raw)) {
+	case "1", "true", "yes":
+		return true, nil
+	case "", "0", "false", "no":
+		return false, nil
+	default:
+		return false, fmt.Errorf("%s=%q is not a boolean; use 1/true/yes or 0/false/no",
+			EnvHADeployment, raw)
+	}
+}
```
a-kuprin: Common devshard-wide behaviour for bool parsing is:
empty/0/false/no/off → false;
1/true/yes/on/t → true;

It's better keep it. You can reuse `strconv.ParseBool` from `common/storage/mode`

a-kuprin:
Versiond treats `--print-storage-mode` as "is this binary + env safe to start"
But RequireHADeploymentStorage isn't checked here, so wrong GONKA_HA will not prevent devshardd from start

At `devshard/cmd/devshardd/version.go` extend `maybePrintVersion` like:

```go
case printStorageModeFlag:
    storageMode, err := mode.Resolve()
    if err != nil {
        fmt.Fprintln(stderr, err)
        return 1, true
    }
    // Here we add the check
    if err := mode.RequireHADeploymentStorage(); err != nil {
        fmt.Fprintln(stderr, err)
        return 1, true
    }
    fmt.Fprintln(stdout, storageMode)
    return 0, true
```

a-kuprin:
## Advised fixes in tests

1. `devshard/testenv/cmd/gencompose/compose.go` (or `local-test-net/docker-compose.versiond.yml`):

`GONKA_HA` is only set on the join HA overlay. Testenv gencompose and local-test-net stay unmarked, which looks like the sqlite→HA citest still needs a running sqlite child in a multi-host pool (`sqlite_ha_migration_test.go` phase 2). That’s fine, but it means this PR has no e2e of the boot guard — and the next person who copies `GONKA_HA=true` into gencompose will break that scenario. Worth a one-line comment here that the omission is deliberate.

2. `devshard/cmd/devshardd/app.go` around the `RequireHADeploymentStorage()` call:

The boot-guard tests all live in `common/storage/mode` and call `RequireHADeploymentStorage()` directly. Deleting these two lines leaves `go test -race ./storage/mode` and `./cmd/devshardd` green. A `devshardd` test that `buildHostManager` / boot fails when `GONKA_HA=true` and sqlite would pin the wiring.

devshard/cmd/devshardd/app.go
```diff
@@ -247,6 +248,10 @@ func buildHostManager(
 		thresholds,
 	)
 
+	if err := mode.RequireHADeploymentStorage(); err != nil {
+		return nil, err
+	}
```
a-kuprin: Call order before `RequireHADeploymentStorage()` at `app.go:251`:

1. `MkdirAll` data dir
2. keyring from disk
3. chain gRPC + chain-ID query
4. `payloads.Open` (mkdir `payloads/`; PG connect if postgres)
5. `seedAvailabilityFromChain` (up to 3s)
6. params provider sync refresh
7. two background pollers `Start()`

The check is two `os.Getenv` calls. It belongs in `run()` before `buildApp`,
and in `--print-storage-mode` (finding 2). Nothing is corrupted on the
misconfigured path (`payloads.Open` in sqlite is file-only;
`closers.Close()` unwinds). Combined with finding 2 it is a slow noisy fail
inside a restart loop.

@Bonujel noted the `payloads.Open` half of this; the chain dial and params
query are also in front.

common/storage/mode/ha.go
snevolin: But `strconv.ParseBool` from `common/storage/mode ` doesn't recognise yes/no/on/off. Ok, I'll add it there.

snevolin: I implemented it in a separate PR: https://github.com/gonka-ai/gonka/pull/1642

a-kuprin: Yes, I have merged it to the target "[devshard-0.2.15-v5](https://github.com/gonka-ai/gonka/tree/devshard-0.2.15-v5)"

Here we still need to use it

snevolin: Added a usage of configenv.ParseBool here.

devshard/cmd/devshardd/app.go
bonujel: Thanks for laying the sequence out — I walked through it and it checks out. Everything ahead of `RequireHADeploymentStorage()` at app.go:251:

- `os.MkdirAll(cfg.DataDir)` — app.go:78
- keyring read from disk — app.go:142, inside `buildChainRuntime`
- chain dial and `resolveChainID` — app.go:154
- `payloads.Open` — app.go:101
- `seedAvailabilityFromChain` — app.go:223, bounded by `availabilitySeedTimeout = 3 * time.Second`
- `newParamsProvider` — app.go:226
- `mgr.Start(ctx)` and `cache.Start(ctx)` — app.go:236-237

All seven, so the guard is indeed two `os.Getenv` calls sitting behind a keyring read, a chain dial and two pollers.

Good catch on finding 2 — I'd missed that one. I think I was looking in the wrong shape: I went hunting for paths that open storage without passing the guard, and `--print-storage-mode` never touches storage. It only answers "is this safe to start", which is of course the question the guard is there for.

One small thing that might help the case for your fix: the path you'd be extending is already wired for this. versiond's `preflightChildWithAdminProbe` treats a nonzero exit from `--print-storage-mode` as a preflight failure — `versioned/internal/process/manager_test.go:265` pins it with a stub that exits 1 and asserts preflight fails. So the check you sketched wouldn't just surface the misconfiguration; it would keep the child from starting at all, ahead of everything in the list above.

snevolin: 1. The orderding issue is already fixed in 2a3ec249. The HA storage guard is now the first operation.
2. We intentionally keep --print-storage-mode read-only. versiond now checks the effective HA mode separately and rejects an unsupported or non-PostgreSQL HA child during replacement preflight, before stopping the serving child.

deploy/join/docker-compose.versiond.yml
```diff
@@ -15,6 +15,9 @@ x-versiond: &versiond
   environment:
     - VERSIOND_ORACLE_URL=http://api:9100/versions
     - VERSIOND_BINARY_NAME=devshardd
+    # This overlay can run multiple devshardd instances for one escrow. Make
+    # every child fail closed unless it uses shared PostgreSQL storage.
+    - GONKA_HA=true
```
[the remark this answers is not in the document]
snevolin: The missing GONKA_HA values are intentional in these two test-only fixtures. devshard/testenv must boot SQLite first and later enable HA at the router to exercise the request-time Devshard-Ha guard. local-test-net also supports generic testapp and SQLite scenarios.

a-kuprin (approved):

PR #1602, opened 2026-08-18
Track live PostgreSQL readiness

snevolin:
## Summary

> This focused PR is based on the PostgreSQL live-readiness work originally implemented in the closed umbrella PRs #1517 and #1587.

Track PostgreSQL availability after devshard storage has started:

- keep one monitor-owned PostgreSQL connection outside the application pool;
- reuse that connection for bounded health checks and reconnect it when needed;
- give connection setup the same 5-second budget as normal PostgreSQL connections;
- schedule each probe 5 seconds after the previous probe finishes;
- mark storage unready after two consecutive database failures;
- restore readiness after two consecutive successful probes;
- observe application-pool saturation independently from database health;
- stop and join the monitor before closing the application pool;
- keep SQLite-only storage behavior unchanged.

## Why

Previously, PostgreSQL readiness represented only successful startup and completion of the initial session-index rebuild. If the database became unavailable later, devshardd stayed alive and could continue reporting ready from stale startup state.

The process should remain alive so a transient database outage does not cause a restart loop, while readiness-aware routing can withdraw it until PostgreSQL recovers.

The health check deliberately does not borrow an application-pool slot. A full pool means that application work is using its configured capacity; it does not prove that PostgreSQL is unavailable. Conversely, a real database failure must still be detected while that pool is full.

## Behavior

This applies whenever a live PostgreSQL backend is attached. In fail-closed `DEVSHARD_STORAGE_MODE=postgres`, two confirmed database failures make `/readyz` return `503` without terminating devshardd. Two successful probes restore readiness. Hybrid storage uses the same PostgreSQL signal while attached. SQLite-only mode is unaffected.

The monitor uses one PostgreSQL connection in addition to the application pool. Application `MaxConns` remains entirely available to application work and is not changed by this PR.

Pool saturation does not by itself change readiness. It is exposed independently through:

- `devshard_postgres_health_probe_total{result="success|database_error"}`;
- `devshard_postgres_pool_saturated`.

This PR does not change storage-mode selection, compare database identities, migrate SQL data, set application-pool capacity, or change the PostgreSQL data-directory layout.

## Validation

- `go test -race ./storage ./observability`
- `go test ./...`
- `go vet ./storage ./observability`
- state-machine coverage for database-health hysteresis and independent saturation;
- Docker coverage proving that a dedicated health check succeeds with `MaxConns=1` fully occupied;
- Docker coverage proving that database loss is still detected while that pool remains occupied;
- reconnect coverage for a closed dedicated connection;
- failure-safe connection cleanup in the saturated-pool test.

## History

This focused PR extracts the PostgreSQL live-readiness work previously developed in the closed umbrella PRs #1517 and #1587 so it can be reviewed and merged independently.

### DIFF
--- a/devshard/docs/high-availability-architecture.md
+++ b/devshard/docs/high-availability-architecture.md
@@ -197,6 +197,10 @@ Operator signals (versiond / host logs and Prometheus):
 - `postgres readiness lost` / `postgres readiness recovered` — the live database
   failed or recovered across the two-probe readiness hysteresis. The devshardd
   process stays alive; `/readyz` returns `503` while database readiness is lost.
+  With the default 5-second interval and 5-second probe deadline, a fully
+  blackholed database can take about 20 seconds after the last successful probe
+  to make `/readyz` return `503`. An upstream health checker adds its own polling
+  delay before it removes the replica from traffic.
 - `devshard_postgres_health_probe_total{result="success|database_error"}` —
   outcomes from the dedicated PostgreSQL health connection. A single
   `database_error` can be a transient connection reset and does not make the
--- a/devshard/docs/high-availability-architecture.md
+++ b/devshard/docs/high-availability-architecture.md
@@ -194,6 +194,14 @@ Operator signals (versiond / host logs and Prometheus):
 - `reconcile_fast_forward` — expected on failover onto a lagging replica.
 - `diff_persist_retry` / `devshard_diff_persist_retry_total` — transient Postgres blips.
 - `diff_fork_detected` / `devshard_diff_fork_detected_total` — **must stay 0** in healthy HA; non-zero means real divergence and needs alert investigation.
+- `postgres readiness lost` / `postgres readiness recovered` — the live database
+  failed or recovered across the two-probe readiness hysteresis. The devshardd
+  process stays alive; `/readyz` returns `503` while database readiness is lost.
+- `devshard_postgres_health_probe_total{result="success|database_error"}` —
+  outcomes from the dedicated PostgreSQL health connection.
+- `devshard_postgres_pool_saturated` — whether all application-pool connections
+  were in use at the latest probe. Saturation is reported independently and does
+  not by itself make `/readyz` fail.
 
 Gateway catch-up and sticky failover remain independent: router
 `proxy_next_upstream` moves the HTTP request; host reconcile heals RAM from
--- a/devshard/docs/high-availability-architecture.md
+++ b/devshard/docs/high-availability-architecture.md
@@ -198,7 +198,10 @@ Operator signals (versiond / host logs and Prometheus):
   failed or recovered across the two-probe readiness hysteresis. The devshardd
   process stays alive; `/readyz` returns `503` while database readiness is lost.
 - `devshard_postgres_health_probe_total{result="success|database_error"}` —
-  outcomes from the dedicated PostgreSQL health connection.
+  outcomes from the dedicated PostgreSQL health connection. A single
+  `database_error` can be a transient connection reset and does not make the
+  service unready. Alert on `postgres readiness lost` or a sustained error
+  rate, not on one counter increment.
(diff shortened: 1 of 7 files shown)


bonujel:
`Pool.Ping` acquires a pooled connection before it pings (`pgxpool/pool.go:823`), so the probe competes with real traffic for the same `MaxConns`. Nothing sets `MaxConns` here — `ParseConfig("")` reads libpq env vars and `pool_max_conns` isn't one — so it's pgx's default of `max(4, NumCPU)`.

That makes a saturated pool hard to tell apart from an unreachable database. `Acquire` blocks until `probeCtx` expires rather than failing fast, which arrives as a probe failure, and two in a row take `Ready()` to false.

I checked the blocking half against puddle directly, since that part needs no database:

```go
p, _ := puddle.NewPool(&puddle.Config[int]{Constructor: ..., MaxSize: 2})
// hold both resources, the way real traffic would during a burst
ctx, cancel := context.WithTimeout(context.Background(), time.Second)
_, err = p.Acquire(ctx)
```

```
pool saturated: 2/2 acquired
probe Acquire returned after 1s: err=context deadline exceeded
probe succeeds again as soon as a connection is released
```

Running two of those through the hysteresis in this PR flips `healthReady` on the second.

What I can't tell from here is whether a real node's pool ever stays saturated across two consecutive probes — that's load-dependent and I have no traffic numbers for it. So this is a question about intent more than a defect: the PR describes readiness in terms of the database becoming unavailable, but the signal also withdraws a node whose Postgres is healthy and whose pool is merely busy. If shedding traffic off a saturated node is wanted, this gets it. If not, pinging on a connection outside the pool, or telling an `Acquire` deadline apart from a `Ping` error, would keep the two separate.

Two smaller things.

`Close()` waits up to `postgresHealthTimeout` for `healthDone` and then calls `pool.Close()` either way. Since `probeCtx` derives from the context `healthCancel()` cancels, the monitor should always return promptly — which seems like an argument for waiting on `healthDone` without a deadline instead of racing an in-flight probe.

`postgresHealthInterval` and `postgresHealthTimeout` are both a second, so a probe slow enough to matter consumes its whole interval and the next tick is already queued. Against a degraded database the probes end up back to back, holding a pooled connection continuously at the point connections are scarcest. A timeout below the interval would leave a gap.

snevolin:
@bonujel thanks for the detailed review.

The dedicated pgx.Conn is intentionally outside pgxpool, so the health probe neither consumes MaxConns capacity nor derives database health from a racy pool snapshot. Pool saturation is now reported independently and never changes readiness by itself.

The timer is reset only after each completed probe, and shutdown first cancels and joins the context-aware monitor before closing the application pool. The test now covers a full MaxConns=1 pool, monitor reconnection, and a real database outage while the application pool remains saturated.

I also addressed the two documentation notes:

- 5dc2ffd29 explains that one database_error may be transient and should not trigger an alert by itself.
- efbb32a55 documents the roughly 20-second worst-case readiness delay for a blackholed database, plus upstream health-check latency.

I kept postgresHealthProbeResult because explicit result names make the readiness state transitions easier to read. No additional runtime behavior was added for that point.

bonujel:
Thanks — I read through the new version and all four points check out.

The probe now holds its own `pgx.Conn` outside the pool and reconnects on
failure, so it no longer competes for `MaxConns`. Saturation is sampled through
`pool.Stat()` and stored separately: the switch that drives `healthReady` only
looks at Success and DatabaseError, so a busy pool can't withdraw a healthy node
on its own. The timer resetting after each completed probe closes the
back-to-back gap, and `Close()` now joins the monitor unbounded before touching
the pool, which is cleaner than the deadline I was picking at.

The test I wouldn't have thought to ask for is the one holding the application
pool saturated *while* the database is genuinely down — that's the case where
the two signals could still have been conflated, and it's good to see it pinned.

Keeping `postgresHealthProbeResult` makes sense to me. A bool would have had to
absorb saturation eventually, and the named results are what let it stay a
separate dimension.

PR #1603, opened 2026-08-18
Persist and migrate devshard PostgreSQL data

snevolin:
## Summary

> This focused PR is based on the PostgreSQL persistence and migration work originally implemented in the closed umbrella PRs #1517 and #1587.

Move the devshard PostgreSQL cluster from an image-owned anonymous Docker volume to an explicit persistent host directory, while preserving existing installations:

- store `PGDATA` under `DEVSHARD_POSTGRES_DATA_DIR` (default `./devshards/postgres`);
- detect and copy an existing PostgreSQL 16 cluster from the legacy anonymous volume on first replacement;
- publish the copied cluster atomically through a staging directory;
- preserve the legacy volume as an unchanged rollback source;
- resume safely after an interrupted copy;
- provide a recovery overlay for an already detached legacy volume;
- require explicit opt-in before initializing an empty database on an existing installation.

## Motivation

The PostgreSQL image declares `/var/lib/postgresql/data` as a Docker volume. Because the existing Compose deployment does not assign that volume a stable name or host path, Docker stores the cluster in an anonymous volume.

That layout works for the current container, but the relationship between the deployment and its data is implicit. After the service is removed, the volume can remain detached, and a later `docker compose up` can create a different volume. Using an explicit bind directory makes the data location predictable across container replacement, `docker compose down`, backup, and recovery operations.

PostgreSQL contains the shared session and escrow state used by versiond replicas, so an existing installation also needs a migration path to the new directory. Changing `PGDATA` and the Compose mount alone would leave the existing cluster in the legacy anonymous volume and present an empty target to the replacement container.

## Upgrade behavior

The wrapper entrypoint handles the storage-layout transition before PostgreSQL starts:

1. If the persistent target already contains a valid cluster, it is used directly.
2. Otherwise, an attached legacy PostgreSQL 16 cluster is selected as the migration source.
3. The source and target are validated and available disk space is checked.
4. Data is copied into a staging directory, validated, synchronized, and then moved into place atomically.
5. The legacy volume is left unchanged and remains available as a rollback copy.
6. Subsequent starts use the persistent target without repeating the migration.

If a previous copy was interrupted after completion of the staging phase, the entrypoint finishes publishing it on restart. If the legacy volume is already detached, the recovery overlay allows that volume to be attached explicitly as the source. When existing versiond artifacts are present but neither a valid target nor a migration source can be found, initialization stops with a diagnostic instead of creating a new cluster; a deliberate empty initialization remains available through `DEVSHARD_POSTGRES_ALLOW_EMPTY_INIT=true`.

## Validation and recovery

- only a valid legacy cluster with `PG_VERSION` is accepted as a migration source;
- the target must be empty or already contain a valid cluster;
- free space must cover the source plus a 10% reserve;
- an incomplete copy is never exposed as `PGDATA`;
- stale `postmaster.pid` is removed from the copy so PostgreSQL can recover from WAL;
- existing versiond artifacts make empty initialization fail closed unless `DEVSHARD_POSTGRES_ALLOW_EMPTY_INIT=true` is explicitly set;
- PostgreSQL receives configurable startup and shutdown budgets.

This is a storage-layout migration only. It does not change the PostgreSQL major version, run a SQL/schema migration, select the devshard storage mode, or alter runtime database readiness.

## Validation

- `shellcheck deploy/join/devshard-postgres-*.sh`
- `deploy/join/devshard-postgres-entrypoint_test.sh`
- `deploy/join/devshard-postgres-migration-preflight_test.sh`
- `deploy/join/devshard-postgres-upgrade_test.sh`

The final test creates a real PostgreSQL 16 cluster in the legacy anonymous volume, writes data, recreates the service with the persistent layout, and verifies that the data survives. The same checks run in the dedicated `Test devshard PostgreSQL persistence migration` CI job.

## History

This focused PR extracts the persistent-PGDATA migration work from #1517 and #1587 so it can be reviewed, tested, and merged independently from routing, graceful lifecycle, database identity, and host-updater changes.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -401,3 +401,17 @@ jobs:
           name: Devshard E2E Tests
           path: devshard/build/test-results/e2e/junit.xml
           reporter: java-junit
+
+  test-devshard-postgres-migration:
+    runs-on: ubuntu-latest
+    name: Test devshard PostgreSQL persistence migration
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Check migration scripts
+        run: |
+          shellcheck deploy/join/devshard-postgres-*.sh
+          deploy/join/devshard-postgres-entrypoint_test.sh
+          deploy/join/devshard-postgres-migration-preflight_test.sh
+      - name: Preserve a real PostgreSQL cluster across replacement
+        run: deploy/join/devshard-postgres-upgrade_test.sh
--- a/deploy/join/config.env.template
+++ b/deploy/join/config.env.template
@@ -54,6 +54,13 @@ export QUERY_GAS_LIMIT=10000000
 # export DEVSHARD_POSTGRES_DB=devshardd
 # export DEVSHARD_POSTGRES_USER=devshardd
 # export DEVSHARD_POSTGRES_PASSWORD=<FILLIN>
+# Persistent bind target for PGDATA. On the first replacement, the entrypoint
+# copies an existing v4 anonymous volume here without modifying its rollback copy.
+# export DEVSHARD_POSTGRES_DATA_DIR=./devshards/postgres
+# Set true only for a genuinely new HA installation with no legacy database.
+# export DEVSHARD_POSTGRES_ALLOW_EMPTY_INIT=false
+# export DEVSHARD_POSTGRES_STOP_GRACE_PERIOD=5m
+# export DEVSHARD_POSTGRES_START_PERIOD=30m
 
 # PostgreSQL tuning (optional)
 # export POSTGRES_MIN_WAL_SIZE=4GB
--- a/deploy/join/config.env.template
+++ b/deploy/join/config.env.template
@@ -57,10 +57,10 @@ export QUERY_GAS_LIMIT=10000000
 # Persistent bind target for PGDATA. On the first replacement, the entrypoint
 # copies an existing v4 anonymous volume here without modifying its rollback copy.
 # export DEVSHARD_POSTGRES_DATA_DIR=./devshards/postgres
-# Set true for the first start only when enabling HA on an installation that
-# has never used devshard PostgreSQL. Remove the setting immediately after the
-# cluster is created. Never use it to recover a missing anonymous volume.
-# export DEVSHARD_POSTGRES_ALLOW_EMPTY_INIT=false
+# For confirmed first-time HA enablement only, pass this override inline to the
+# first `docker compose up` command; do not persist it in config.env:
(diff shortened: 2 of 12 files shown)


a-kuprin:
## Findings to fix before merge

**1. Medium — logic.** Empty-init guard keys off `./devshards/bin`, not `.pg-bound` in `./devshards/data` and `./devshards2/data`. Any host that already ran versiond (including SQLite-only) refuses empty init. First-time HA enablement and “lost the anonymous volume” print the **same** error. The dangerous reply is `DEVSHARD_POSTGRES_ALLOW_EMPTY_INIT=true`, which creates a new empty cluster.

devshardd already writes `.pg-bound` when Postgres actually holds sessions. Use that, or document this decision tree in the 22-line doc and `config.env.template`:

| Situation | Do this | Do not |
|---|---|---|
| Already running HA; upgrading compose | `up -d --force-recreate` (keep the container so the anonymous volume is carried) | `compose down` first |
| Never ran HA postgres; binaries already on disk | Set `ALLOW_EMPTY_INIT=true` **once**, then unset | Treat the error as lost data |
| Had HA postgres, already ran `down` | Recovery overlay + exact dangling volume name | `ALLOW_EMPTY_INIT=true` |

**2. Medium — security / ops.** `:Z` on `./devshard-postgres-entrypoint.sh` privately relabels a git-tracked host file on SELinux. Use `:ro,z` (or no flag) on the script; keep `:Z` only on the private PGDATA bind. `./devshards/bin` is already `:z` because versiond shares it.

**3. Medium — ops docs.** Supported upgrade is **recreate in place**. `compose down` before the first new-layout start requires the recovery overlay and the exact volume name. That is easy to get wrong in a real upgrade.

## Lower-severity notes

- After migration the image `VOLUME` still creates a **new empty anonymous volume** on every `down`/`up` (the upgrade test even asserts this). Disk leak, and `docker volume prune` can delete the rollback copy.
- Interrupted `cp -a` starts over. HA is down for the full copy; `start_period: 30m` may be tight for a large DB and also applies to later starts.
- Preflight interpolates volume name and target path into a Docker mount CSV; a comma can inject extra mount options (operator-local).
- `validate_cluster` accepts any numeric `PG_VERSION` (pin 16 while the image is 16).
- `.gonka-copy-complete` is renamed into live `PGDATA`; move it next to `.migrated-from-v4`.

snevolin:
@a-kuprin thanks for review. Fixed in 474011f19. The guard now derives the PostgreSQL major from the image, fails closed when evidence mounts are missing, and has coverage for both versiond roots. The docs also describe explicit recovery when the old volume is permanently lost and keep the one-time override out of config.env.

a-kuprin (approved):

PR #1209, opened 2026-05-20
`devshard` height sync protocol, testenv + e2e for height sync

a-kuprin:
Detailed protocol description is at `devshard/docs/proposals/HEIGHT_SYNC_PROTOCOL_PROPOSAL.md`

## Height-sync protocol + oracle

- `devshard/blockoracle`: HTTP client, canonical headers, validator-signed commits
- `devshard/transport`: height-sync envelopes, peer tips, server ingest paths
- `devshard/user`: session/oracle integration for anchor/omit/strong modes

Height-sync end-to-end coverage so the devshard stack (hosts, devshardctl, block oracle) can be exercised against a mocked mainnet without a live chain.

It is still isolated from dapi (tested with mock), and final version should bake blockoracle into dapi.

### E2E tests

- **In-process** suite (`scenarios/heightsync_anchor_e2e_*`) — courier, confirm, origin-sig, cadence, etc.
- **Container** suite (`scenarios/container/`, tag `testenvci`) — real compose stack: smoke, cadence, lost-first-response, force anchor, cheating trail, feed-stopped recovery (Phases A–C)
- Plan/docs: [`CONTAINER_E2E_PLAN.md`](scenarios/CONTAINER_E2E_PLAN.md), [`SCENARIOS.md`](scenarios/SCENARIOS.md)

## `devshard/testenv` — integration harness

Also introduces a Docker Compose–based devshard integration test environment.

- Compose stack: **`mock-chain`** (gRPC chain mock), **`heightsyncd`** (signed block oracle over HTTP/SSE), **`devshardd-testenv` × N** (real host wiring with mocked inference/validation/bridge), and **`devshardctl`**
- Config generation (`gencompose`, `config.yaml`), Makefile targets (`make e2e`, `e2e-phase-b/c`, `e2e-keep`), CI-friendly scripts
- Optional observability overlay (Alloy → Loki / VictoriaMetrics → Grafana)
- Dev overlay with live-reload / debug (`DEVELOPMENT-MODE.md`)

### Protocol proposals (draft specs)

Normative / design docs added or expanded on this branch. The testenv harness in this PR is the foundation for exercising them end-to-end as they are implemented.

| Proposal | Role |
| -------- | ---- |
[...]

PR #1615, opened 2026-08-20
`devshard v5`: Height-sync protocol

a-kuprin:
# Height-sync protocol (v5)

**Spec:** [`HEIGHT_SYNC_PROTOCOL_PROPOSAL.md`](https://github.com/gonka-ai/gonka/blob/ak/height-sync-protocol/devshard/docs/proposals/HEIGHT_SYNC_PROTOCOL_PROPOSAL.md)
**Plan:** [`height-sync-implementation-plan.md`](https://github.com/gonka-ai/gonka/blob/ak/height-sync-protocol/devshard/docs/height-sync-implementation-plan.md)
**Params:** [`height-sync-params.md`](https://github.com/gonka-ai/gonka/blob/ak/height-sync-protocol/devshard/docs/height-sync-params.md)
**Tests:** [`height-sync-tests.md`](https://github.com/gonka-ai/gonka/blob/ak/height-sync-protocol/devshard/docs/height-sync-tests.md)

Sibling (not this PR): [`ak/height-sync-protocol-dapi`](https://github.com/gonka-ai/gonka/tree/ak/height-sync-protocol-dapi) mounts hash-only `/block/*` on production dapi.

---

## Summary

This PR lands height-sync on 0.2.15 / v5: every host and the sequencer keep a signed, replayable view of mainnet height inside the escrow log, even when there is no inference traffic.

It has two planes, plus a new kind of time in the log:

1. **Transport plane (Phases A–D).** Anchor envelopes on a `K`-nonce cadence, forced turns, courier peer tips, origin signatures, and `(C-quorum)`. Rewired onto `devshard/chainoracle/blocks` instead of cherry-picking `devshard-testenv`. Default compose stays off; opt in with `DEVSHARD_CHAINORACLE_URL` or `DEVSHARD_HEIGHTSYNC`.
2. **Log plane (Phase E, E0–E7 + E9).** Heartbeat turns, host acks, L0–L5a / L7 ingest checks, repair probes that never assign blame, close-ready arming, and `observed_height` stamps on inference txs so a busy escrow never heartbeats.
3. **Logical time.** Every Diff-resident height — heartbeat, ack, and inference `start` / `confirm` / `finish` — is a *reference height*: `max(own_tip, F(m))`. `F` is the escrow's logical clock, a function of the applied log. Later releases can fold timeouts, `USER_TIMEOUT`, and cPoC bands onto those heights instead of host/user wall-clock timestamps. This PR **carries** the stamps; it does **not** switch those decisions yet.

**Deferred to later releases (Phase F / dispute layer):**

- **Strong / `LightBlock` proving.** Envelope field 9, `D`-band escalation (`|Δ| > D` ⇒ Strong required), `(C-strong)` / `(C-hybrid)`, and `CATCHING_UP` forcing the next heartbeat to Strong. Until then `CATCHING_UP` is a label only.
- **On-chain disputes.** Marks (`DISPUTE_ORIGINATOR` / `DISPUTE_CARRIER`, vector contradiction, `DEFERRED_FAIL`) are recorded locally. Evidence packets, `MsgHeightSyncEvidence`, slashing, and cross-session equivocation wait on Strong, because the canonical half of a packet is a `LightBlock`.
- **L6 adjudication.** L6 reconciles a Diff-resident `(height, hash)` against the verifier's own oracle. That check runs *after* the pair is already in `Diff`, so it must not `INVALID` the diff (two verifiers whose followers have not reached `H` would split). The implementation records a `DEFERRED_FAIL` mark when the oracle already has block `H` and the hash mismatches. A durable recheck queue for heights that have not mined yet, and turning those marks into disputes, ship with the dispute layer. `F` is not unwound if a later L6 fails.

Also not in this PR: cPoC skip carriers, flipping `timeout.go` / the seal clock onto heights, default-on compose, and the gateway operator metrics in plan §8.12 (E8).

---

## Why

cPoC bands and `USER_TIMEOUT` finalization both need a height that every honest verifier recomputes from `Diff`. Envelope Anchors only exist on the request/response that carried them; a courier user has no follower of its own; a quiet escrow never syncs.

Height-sync puts a signed `(height, hash)` into the log on a wall-clock cadence. Hosts answer. The record of who answered is a pure function of `Diff`. Hosts that hear nothing from the user arm close-ready locally and still emit nothing — closing stays finalization's job.

The schedule is in **milliseconds**, not blocks. Mainnet height is the *result* of a height-sync turnover, so a block-denominated heartbeat is circular: a quiet courier cannot notice a block passed until it syncs, and a partitioned host sees a frozen tip and would never count enough blocks to arm.

---

## Height stamps and logical time

This is the load-bearing new feature. The rest of the PR exists so these stamps keep moving when inference does not.

### Two planes, one meaning in the log

| Plane | Where | Height means | Who signs |
| ----- | ----- | ------------ | --------- |
| **Transport** | `HeightSyncSection` on the HTTP envelope | first-party follower tip of *this hop* | origin signature on the response; request leg unsigned |
| **Log** | `observed_height` / `observed_block_hash` on Diff-resident txs | **reference height** — logical time of the escrow | user (diff sig) on heartbeat / start; host (`host_sig` / `executor_sig` / `proposer_sig`) on ack / confirm / finish |

The envelope never enters `Diff`. Two verifiers replaying the journal would not see it, so it cannot be the clock for cPoC or `USER_TIMEOUT`. The stamp in the tx can.

There is **no second height in the log**. A lagging host does not write its raw tip into Diff: it writes `max(own_tip, F(m))`, or omits when the floor is more than `W_conf` above its tip. Its real reading stays first-party in the response-leg Anchor and in `sync_state`.

### What a stamp is

`observed_height` on a message means “the height its signer observed when producing this message” — exactly what `started_at_height` / `confirmed_at_height` would mean on the wire, so those names live only on the **record**, copied from the stamps of `MsgStartInference` and `MsgConfirmStart`.

A stamp is only as strong as the key on it:

| Stamp | Signer | Strength |
| ----- | ------ | -------- |
| `MsgHeartbeat`, `MsgStartInference` | user (diff signature) | a **claim** — attributable, checkable against the chain later via L6 |
| `MsgHeightAck` (`host_sig`), `MsgFinishInference` (`proposer_sig`), `MsgConfirmStart` (`executor_sig`, mirrored into `ExecutorReceiptContent`) | host | an **attestation** — rewriting it is forgery of that host |

The confirm trap: `executor_sig` is over `ExecutorReceiptContent`, a signing input that never enters Diff. A field on `MsgConfirmStart` that is missing from that content is outside the executor's signature, and the sequencer could rewrite the height. This PR mirrors `observed_height` / `observed_block_hash` into the receipt and copies them back before recovery.

Unstamped records omit proto3 zeros, so `post_state_root` bytes do not change (`v2` root).

### `F` is the escrow's logical clock

`F(m)` is the reference height the log had established at nonces `< m`. L0 holds every Diff-resident stamp to it. It never lowers.

Raise rules (spec §14), landed:

1. An envelope `(H, hash)` cannot be a future height (hash unknown in advance). Catch/punish is L6 / Phase F.
2. `F` never exceeds the max reported first-party **host** envelope `H`. Lifts (`stamp = F`) do not raise. Honest compose drops a host raise that does not match this hop's response-leg envelope.
3. Sequencer-composed stamps (`MsgHeartbeat`, `MsgStartInference`) never raise `F` and never count toward quorum `Q`. `Observe` is host-only.

A single party claiming `1<<40` therefore cannot become the escrow's time: the floor only follows host-signed first-party tips, unaided by at most `W_conf`, or a host-only `Q` for a larger jump.

### What this enables, and what it does not switch yet

| Decision today | Height form (later) |
| -------------- | ------------------- |
| Execution timeout against each host's clock | `h − confirmed_at_height` |
| Refusal timeout against user-controlled `started_at` | `h − started_at_height` (checkable) |
| Seal clock = max `confirmed_at` | max `confirmed_at_height` |

Migration is two-step on purpose: this PR **carries** the heights; folding them into a state root (timeouts, seals) needs a protocol-version gate, or two participants compute different roots. `timeout.go` / `stateClockLocked` stay on wall time.

---

## What landed

| Phase | What |
| ----- | ---- |
| **A** | `devshard/heightsync` transport: Omit/Anchor cadence, inbound classify, confirmation, origin signing, audit. Envelope + `MsgForceHeightSyncTurn` (oneof 9). Host/user/state seams on `chainoracle/blocks`. |
| **B** | In-process e2e in `testenv/scenarios/heightsync_anchor_e2e_*.go`. |
| **C** | Container `citest-height-sync` against mock-dapi `/block/*`. Default compose unchanged; the citest harness patches `DEVSHARD_CHAINORACLE_URL`. |
| **D** | Hash-only Tendermint observer, direct-chain adapter, host failover when dapi has no `/block/*` or is down. Production dapi mount is the sibling PR. |
| **E0–E3** | `MsgHeartbeat` / `MsgHeightAck` (oneof 10/11; 12/13 reserved for cPoC). User `MaybeHeartbeat` spans, concurrent and non-aborting. Host-signed acks into the mempool from the same oracle read as the response-leg Anchor. Gateway starts the heartbeat loop. |
| **E9** | Session-open seed: fan `POST /sessions/:id/height-sync` once so nonce 1 already has a tip. Consumes no nonce, does not advance `h_last`, never fails the session on a miss. |
| **E4** | `ValidateDiff` L0–L5a / L7. Envelope mismatches are marks, not `INVALID`. Historical replay without an envelope skips L4/L5a. L4 on both legs uses the producer rule (`max(anchor, F)`), not strict equality — a lagging honest sequencer lifting to the floor is not a dispute carrier. |
| **E5** | Signed host→host repair probe for missing acks. Outcomes `HEIGHT` / `UNREACHABLE` only — never `USER_CHEATING`, never a mark. |
| **E6** | Close-ready: host arms after `IdleTimeout` of wall-clock silence and emits nothing. |
| **E7** | `observed_height` / `observed_block_hash` on Start / Confirm / Finish. Confirm stamps are mirrored in `ExecutorReceiptContent`. Unstamped records keep the `v2` root. |

Post-landing audit (steps 12–24 in [`height-sync-audit-findings.md`](https://github.com/gonka-ai/gonka/blob/ak/height-sync-protocol/devshard/docs/height-sync-audit-findings.md)): heartbeat loop on the gateway, fail-closed origin cache, bounded open-turn retain, no local-oracle fold into the SM tracker, compose through the log plane, capped marks, recovered `turn_seq`, host-only floor raise, oracle I/O off mutexes, repair budget/cancel, wire caps, saturating `HReq+D_ack`, `peer_seen` from host claims only.

### Cadence (the E6–E7 rewrite)

Scheduling is wall clock; evaluation stays in blocks so replay never consults a clock.

| Knob | Default | Job |
| ---- | ------- | --- |
| `Interval` | 3s | Longest gap between full turnovers |
| `TurnTimeout` | 6s (`2 × Interval`) | Patience on one open turn; one dead slot cannot silence the cadence |
| `IdleTimeout` (`T_idle`) | 12s (`4 × Interval`) | Host silence budget before arming |
| `AckDeadlineBlocks` (`D_ack`) | **10** (derived) | `⌈(Interval + TurnTimeout) / BlockTime⌉ + 1`. Stamp slack; lateness is `observed_height > h_req + D_ack`. `D_ack = 1` made honest span acks late by construction. |
| `DeltaBlocks` (`D`) | 2 | `SYNCED` vs `CATCHING_UP`; Strong escalation is Phase F |
| `WindowBlocks` (`W_conf`) | 256 | Confirmation window, unaided floor raise, and how far a producer will carry `F` |

A **turnover** is `Q` distinct **host-signed** height claims: `MsgHeightAck`, or an executor stamp on ordinary inference. The user's own stamp is self-signed and does not count. `ORACLE_UNAVAILABLE` acks are **required**, echo `F(m)` from the log, and **count** toward `Q` (reachability, not a height witness). Busy inference therefore emits **zero** heartbeats (H2); a quiet session heartbeats every `Interval` (H1).

`K_hb` and `MinRoundsPerBlock` are gone. Snapshot overlay: proto fields 13/15 kept their numbers and became `height_sync_interval_ms` / `height_sync_idle_timeout_ms`; field 18 is `height_sync_turn_timeout_ms`. Inference-chain still publishes zeros, so compiled defaults apply. Overlay scheduling knobs are validated and clamped; evaluation knobs (`D_ack`, `D`, `W_conf`, `BlockTime`) stay compiled.

---

## Wire

`DevshardTx` oneof:

| Field | Message |
| ----- | ------- |
| 9 | `MsgForceHeightSyncTurn` (transport; already allocated) |
| 10 | `MsgHeartbeat` |
| 11 | `MsgHeightAck` |
| 12/13 | reserved for cPoC `MsgSkipProbe` / `CarrySkip` |

Inference stamps are additive proto3 fields on existing txs (zero / empty = unstamped). `InferenceRecordProto` fields 18/19 (`started_at_height` / `confirmed_at_height`) are omitted when zero, so unstamped `post_state_root` bytes do not change.

Nothing on the scheduling side — Interval, TurnTimeout, IdleTimeout, last turnover time — is folded into `Diff` or a `SyncTurnRecord`. Turn state remains a pure function of the log.

---

## Log plane vs edge

| Plane | Height means | Rule | Replay |
| ----- | ------------ | ---- | ------ |
| Log (`Diff`) | reference height / logical time | L0 against `F(m)`; L0b same-executor `confirm ≤ finish`; L1–L3 framing/sig/causality; L7 vector | yes — same `INVALID` |
| Edge (envelope) | first-party follower tip | L4 binding (producer rule); L5a `D` band; `(C-quorum)` | **no** — skipped on catch-up / gossip (`sec = nil`) |
| Oracle (deferred) | pair vs chain | L6 hash-at-`H` | mark only; not a diff verdict |

L5b (in-log `D` band) is **withdrawn**. Divergence between followers is monitoring. L5a may refuse an exchange and never invalidates a Diff.

---

## Behaviour that must not regress

- **Opt-in.** Unset `DEVSHARD_CHAINORACLE_URL` / `DEVSHARD_HEIGHTSYNC` ⇒ no envelopes, no heartbeats, existing inference path unchanged.
- **Busy escrow pays nothing.** Host-stamped Start/Confirm/Finish discharges the cadence (H2, H33). The user's own stamp does not (H2 companion).
- **Quiet escrow still syncs.** A heartbeat turn opens within `Interval` of the last turnover (H1). The gateway loop calls `MaybeHeartbeat`; span dispatch is concurrent and continues on one slot's error.
- **Repair never blames.** Missing ack ⇒ `HEIGHT` or `UNREACHABLE`; the turn stays `degraded` (H17–H20).
- **Arming emits nothing.** Silence past `T_idle` sets a local flag only (H21–H23). A missing ack is never a reason to arm.
- **Replay is deterministic.** Two verifiers with the same `Diff` compute the same `SyncTurnRecord`, the same `F(m)`, and the same L0–L3 / L7 verdicts. L4 and L5a need the envelope and are skipped on catch-up / gossip ingest. Admission refusals do not feed the floor (attack 24e).
- **Floor is host-only.** Sequencer heartbeats never raise `F`. Honest compose omits a host raise that does not match this hop's envelope.
- **No Strong, no slash.** Marks are local. Nothing in this PR produces a `LightBlock`, an evidence packet, or a dispute tx.

---

## Out of scope / next releases

| Deferred | Why it is not this PR |
| -------- | --------------------- |
| **Strong / `LightBlock` / `|Δ| > D` escalation / `(C-strong)` (Phase F)** | Needs commit signatures, IAVL `Prove()`, and envelope field 9. Hash-only oracles must degrade, never fake Strong. |
| **Evidence packets, on-chain `MsgHeightSyncEvidence`, slashing, cross-session equivocation** | Canonical half of a packet is a `LightBlock`. Marking lands here; adjudication does not. |
| **L6 as a verdict or a dispute** | The pair is already in `Diff` when L6 can see it. `INVALID` would split verifiers whose oracles lag. Marks exist; a persistent recheck queue and dispute promotion do not. `F` is not rolled back. |
| cPoC skip-probe carriers (oneof 12/13 reserved only) | Separate protocol. |
| Switching `timeout.go` / the seal clock onto stamped heights | State-root change; needs a version gate after the stamps have been carrying. |
| Enabling height-sync on default compose | Opt-in until the sibling dapi mount and ops surface exist. |
| Gateway `/metrics` and `GET /v1/debug/heightsync` (plan §8.12, E8) | Operator views; H26/H27 container + H39–H49. |
| Production dapi `/block/*` | Sibling [`ak/height-sync-protocol-dapi`](https://github.com/gonka-ai/gonka/tree/ak/height-sync-protocol-dapi). |

---

## Config

| Env | Default | Effect |
| --- | ------- | ------ |
| `DEVSHARD_CHAINORACLE_URL` | unset | HTTP oracle; empty ⇒ no height-sync |
| `DEVSHARD_HEIGHTSYNC` | unset | `1`/`true` enables chain-only height-sync without a dapi URL |
| `DEVSHARD_HEIGHTSYNC_K` | 10 | nonce cadence of Anchors (transport `K`, not the heartbeat interval) |
| `DEVSHARD_HEIGHTSYNC_SLOTS` | 1 | sync-turn width |
| `DEVSHARD_HEIGHTSYNC_PROBE_INTERVAL` | 30m | re-probe dapi after a transport miss |

Runtime-config overlay (`Snapshot.HeightSync`, proto 13–18): zero means keep the compiled default. `Validate` requires `D_ack · BlockTime ≥ Interval + TurnTimeout`, `T_idle > Interval + TurnTimeout`, and `2 · Interval ≤ F` (originator freshness). An overlay that would fail is clamped to compiled defaults.

---

## Test plan

```bash
cd devshard
GOMODCACHE="$HOME/go/pkg/mod" GOCACHE="$HOME/Library/Caches/go-build" \
  go test ./heightsync/... ./transport/... ./user/... ./host/... ./state/... ./chainoracle/...

# in-process e2e (held-response cases need -tags=dev)
GOMODCACHE="$HOME/go/pkg/mod" GOCACHE="$HOME/Library/Caches/go-build" \
  go test ./testenv/scenarios/ -run HeightSync -count=1
GOMODCACHE="$HOME/go/pkg/mod" GOCACHE="$HOME/Library/Caches/go-build" \
  go test -tags=dev ./testenv/scenarios/ -run HeightSync -count=1

# container (opt-in compose patch)
make -C testenv citest-height-sync
```

Also `cd common && go test ./runtimeconfig/... ./nodemanager/...` for the ms overlay mapping.

Catalog: [`height-sync-tests.md`](https://github.com/gonka-ai/gonka/blob/ak/height-sync-protocol/devshard/docs/height-sync-tests.md). Phases A–D are §2–§6. Log plane is §7 (H1–H104 on this branch; H26/H27 container and H39–H49 gateway observability are E8). Floor / stamp: H2, H28–H31, H33, H54–H58, H89–H91. L6 marks only: H13e, H14, H57.

---

## Commits (this branch)

```
9cfcb8666 Prepared docs and implementation plan for height-sync
4e58fa3bd Add height-sync transport plane on chainoracle/blocks, env-gated on host and gateway.
2799c0938 Keep height-sync anchoring when dapi has no /block/* or is down, via a hash-only direct-chain oracle and host failover.
d017f91b4 Prove height-sync still works when dapi has no /block/*, via a mock-dapi 0.2.15 stand-in and citest against the 0.2.15-v5 /block/* path.
27fd7d3d3 Land height-sync log-plane heartbeat and host acks (E0–E3).
955f81f78 Seed a height from the roster before the first outbound round (E9).
72e875008 Reject invalid height-sync diffs on ingest and keep L4 mismatches as marks (E4).
2b647c1cb Probe missing height-sync acks over a signed host-to-host path without assigning blame (E5).
ea2013624 Schedule height-sync heartbeats and close-ready arming on wall clock (E6–E7).
43029b5ab Give Diff one height semantics: every stamp is max(own_tip, F(m)).
12dbaf076 Harden height-sync after landing: restore heartbeat turn_seq on recover, bind L4 to the producer rule, and keep oracle I/O off host/confirm/scheduler mutexes.
```

Plus uncommitted on this branch: host-only floor raise (spec §14 rules 1–3) and remaining hardening (saturating ack deadline, `peer_seen` from host claims, L3 heartbeat nonce, restore keeps snapshot `h_last`).


a-kuprin:
At earlier height-sync protocol version there was the comment by @vitaly-andr:

https://github.com/gonka-ai/gonka/pull/1209#issuecomment-4642723921

> **Review note — height-sync confirmation quorum: request-leg originator authentication**
> 
> Went through the height-sync trust path (`blockoracle/verifier`, `heightsync/*`, `transport/*`). The verifier + canonical encoding look solid (length-prefixed canonical bytes, recovered-address gate, strict >2/3, dup-signature guard). One thing on the confirmation side I wanted to raise before it grows a consumer.
> 
> The response leg signs and verifies the origin (`attachResponseOriginSignature` → `SignOrigin`, verified on receipt in `transport/client.go` via `heightsync.VerifyOrigin` — the only call site). The request leg is different: carry-forward sections are unsigned by design (`SenderSignature` doc: *"Omitted on courier request-leg carry-forward"*), and `HandleInference` never calls `VerifyOrigin`. Inbound auth (`transport/auth.go` `VerifyRequest`) signs `sha256(escrow_id || body || timestamp)` and recovers the **carrier's** address; the in-body `OriginatorSenderID` isn't bound by it. The endpoint is escrow-owner-restricted (`isOwner`), so the actor here is a host's own escrow owner.
> 
> `recordInboundAnchorIfAnchor` stores the attestation with `OriginatorSenderID` taken verbatim from `hs.OriginatorSenderID`, and `ConfirmationIndex.RecordAttestation` keys the quorum by that claimed id (`originatorKey`, falling back to `PeerID` — so no per-carrier dedup), gated only by `inRoster` + `Trust != TrustDisputeCarrier`. A fresh carry-forward anchor isn't forced to `TrustDisputeCarrier` — that's assigned only on freshness failure, and freshness is evaluated against the attacker-supplied `OriginatorTimestampMs`/`TimestampUnixMs` — so it lands as `TrustUntrustedPeer`/`TrustPeerAligned`, both of which count. `IsStrictlyConfirmed` then tallies one vote per distinct roster originator.
> 
> Net: `OriginatorSenderID` is documented *"must not be overwritten by the carrier"*, but that invariant isn't enforced on the request leg — so one escrow-owner carrier could submit anchors attributed to several distinct roster originators (fresh timestamps, height ≤ tip) and reach `ConfirmConfirmed` without those originators actually attesting.
> 
> Impact today looks effectively nil — I couldn't find any non-test consumer of `IsStrictlyConfirmed`/`ConfirmationView` (every `IsStrictlyConfirmed` caller is in `*_test.go`; the two `ConfirmationView()` methods just re-expose the view). So this is a "harden before wiring" flag, not a live issue. Two directions, whichever matches the intent:
>  - require an origin signature on request-leg carry-forward before an attestation is quorum-eligible, or
>  - restrict quorum counting to oracle-backed/self-attested trust and treat unverified carry-forward as evidence-only (like `TrustDisputeCarrier`).
> 
> Is request-leg origin verification already planned before `ConfirmationView` gets a consumer? Happy to send a small test that exercises the forged-originator path if useful.



**Request-leg carry stays unsigned by design** (§15). `Carry()` rewrites hop-local fields, so the originator’s response signature cannot be forwarded as-is; provenance for disputes lives in the courier’s cached `(blob, sig)` and is produced on demand (`HeightSyncEvidenceFor`), not on every hop.

**Admission is optimistic / weak-path:** a carry-forward Anchor is only accepted when it sits near the receiver’s known tip (`|Δ| ≤ D`; beyond that Strong / `INVALID(strong_required)`). Freshness `F` still applies. So a courier cannot freely invent arbitrary heights under fake originator labels — only claims close to what the host already believes.

**Forgery is attributable after the fact:** if `(H, hash)` later disagrees with the oracle, the host opens a dispute; the courier must produce the origin blob. Verify OK → `DISPUTE_ORIGINATOR`; missing/invalid → `DISPUTE_CARRIER`. That is the intended control, not request-leg `VerifyOrigin`.

**`(C-quorum)` is withdrawn** for this reason and others: counting unsigned carry-forward originator ids as votes would be the laundering you described. Downstream (cPoC) gates adversarial votes on each verifier’s **local oracle** (plus verifier vote quorum), not on an envelope-originator tally. `Q` remains only for turn reachability and floor corroboration.

So: no, we are not planning request-leg origin verification before wiring consumers; the weak path + dispute is enough for carry-forward, and confirmation no longer relies on those labels.

PR #1618, opened 2026-08-20
feat: Trainshards v0 (stage 2)

x0152:
### Stage 2: off-chain runtime (trainshard)

This stage adds the off-chain part that runs the work on the borrowed nodes. The flow: the host frees a reserved node and prepares it, the coordinator wires the prepared nodes into one mesh and deploys the container, the researcher starts the run and drives it through the coordinator, and at close the run is wiped and the node goes back to normal work inside the return buffer the chain gives it. This stage runs in two loops: a readiness loop (while trainshardd is up) and a run loop (after the node is reserved).
1. The host runs a training runtime on every GPU machine and opts a node in when it is ready
2. Once a node is reserved, the host drains it, pulls the base image the proposal names, and prepares it on its own
3. The coordinator sends out the mesh config and the container, and the run goes on with the nodes that are ready. A failed node is autokicked, cleaned up, and returns after the return buffer
4. The work cycle goes through the coordinator: deploy, start, stop, status, logs and shell
5. The container can be replaced during the run, with no vote and no full re-download
6. The container sees only its GPUs, the mesh and its own disk
7. At the end, the coordinator collects a result to settle with
8. On settle or timeout the shard closes and nodes return after the buffer; a failed node is autokicked, cleaned up, and returns after the same buffer while the run continues

**Result**: the researcher can run a job on the reserved nodes, interact with them, and collect a result to settle

Details: https://docs.google.com/document/d/1LLZngQ7VoIL3DVT8St40XLE8HcRcxyNXZueyzoQfWuE/edit?tab=t.no7vrbsyhyhk


Ryanchen911:
Reviewed at `8a2a2f8f7`.

### 1. The shared secret, and merge ordering

I know this is the placeholder the real dAPI adapter replaces, and `hmac.go` is upfront about it — "whoever holds it can sign as any address". Laying out the operator-side consequence anyway, because I think it argues for merge ordering rather than a code change.

The secret is symmetric and every participating host plus the coordinator must share the same value, so any one host in a shard can sign as the shard creator and issue commands to the other hosts in that shard.

Worth being precise about how far that goes. The base image is digest-pinned by governance (`ValidatePinnedImage`) and the run image must derive from it (`VerifyImage` / `DerivesFrom`), so an unrelated image cannot be run. But `DerivesFrom` only checks layer ancestry, so a `FROM <approved-base>` image with arbitrary layers on top satisfies it. The practical capability is: run arbitrary code in a container on someone else's leased machine, plus shell into it. That container is hardened (`CapDrop: ALL`, non-root, nothing privileged), so reaching the machine itself would additionally require a container escape — not direct control of the host.

Even so, given that `trainshardd` legitimately holds `docker.sock`, `pid: host`, `NET_ADMIN`, and `SYS_ADMIN` (not questioning those — it needs them), and that `/trainshard/` is publicly routed through the proxy, I'd suggest Stage 2 land after the real signing rather than before it. Nothing here is an implementation defect; a shared symmetric secret is just not something operators should be asked to deploy to mainnet, even briefly. Since the real adapter is next, this may already be the plan — flagging it so the merge order is deliberate.

### 2. `/trainshard/` is on the loosest rate-limit tier

`entrypoint.sh` puts it in `exempt_zone`: 500 rps, burst 2000, 300 concurrent connections — the same tier as chat and inference. Authentication happens inside `trainshardd` rather than at the proxy, so unauthenticated traffic reaches signature verification at that rate, each request costing up to a 1 MB body read plus an HMAC. For a machine whose job at that moment is running a training run, its own `limit_req` zone would fit better than the chat tier.

### 3. `SigningPayload` does not bind the target host

It covers version, method, path, timestamp, request ID, and body digest, but not which host the request is for. Combined with a shared secret, a signed request captured against host A replays against host B, since `Once` keeps its served-ID map per process. The path does carry shard and node IDs, which limits it in practice, but adding the host's own participant address to the payload closes it cheaply — and stays useful after the real signing lands.

### 4. The nftables ordering has no test

I checked whether `deploy --source` could be pointed at an internal address to reach the host's private network. It cannot: in `fence()` the denied CIDRs are added to the output chain before the allowances, and nftables takes the first matching verdict, so a denied range always wins.

That ordering is load-bearing for the SSRF boundary but nothing covers it. A future refactor that reorders those two loops would reopen the path to the host's internal network with every test still green. A small test asserting denied-before-allowed would lock it in.

### 5. Notes for when the real adapters go in

Both concern the trust boundary the fakes currently stand in for:

- Everything authorization-related reads shard records through `shard.ChainReader`. The real implementation will want to apply the same `plainName` validation to chain-supplied node IDs that `ParseNodeRef` applies to API-supplied ones, and to fail closed on chain read errors.
- `Drain` returning `true` is what lets `Plan` proceed to mesh and container creation. The real `nodemanager` needs to guarantee the dAPI has actually stopped routing inference to that node before returning `true` — otherwise requests keep landing on a drained node for the remainder of the epoch, which the reservation shield does not cover on the requester side.

vbgd0:
Reviewed the trainshard runtime end to end (daemon config/main, signedhttp guard + HMAC signer, hostd run/session/node endpoints and use cases, shard policy, docker adapter, netns/wireguard/nftables, xfsquota, fake chain, coordinator prepare_mesh/ops, proxy + compose wiring). The engineering quality here is high: the container hardening (CapDrop ALL, no-new-privileges, read-only rootfs, private ipc/cgroupns, pids/mem/cpu limits, project quotas), the atomic nftables ruleset swap with default-drop, and the wireguard link being created host-side and then moved into the sandbox netns (so the sandbox can never own the mesh port alone) are all exactly right. Findings, by severity.

### 1. The signed payload does not bind the target host — fix the contract before real signing lands

`contract.SigningPayload` covers `method, path, timestamp, requestID, body` — not the participant the request is aimed at. Paths like `/trainshard/v0/shards/{shard_id}/deploy` are identical across hosts, and node ids like `node1` will collide across participants in practice. So a signed deploy/stop/shell request for host A verifies unchanged on host B, and host A (who necessarily sees the signed request) can replay it to host B within the signature window.

Today this is moot because of the shared secret (see #2), but this is the *contract*, and it's the thing that must already be right when real signatures replace HMAC — otherwise the replay hole survives the signer swap. Adding the target participant address (and the query string, while you're at it, even though nothing uses one yet) to the signing payload is cheap now and painful later.

### 2. Shared-secret trust model: every host in a shard can drive every other host

`hmac.Shared.Recover` trusts the address embedded in the signature; the only check is the shared secret. Since `config.env.template` hands the same `TRAINSHARD_SHARED_SECRET` to all hosts, any participating host can sign as the shard creator or run key and pass `Actor.AuthorizedFor` on any *other* host: deploy, stop, stream logs, and — the expensive one — open a shell inside another host's run container and walk the researcher's code, data, and checkpoints.

The daemon logs a loud stand-in warning and the code comment ("worth no more than the hosts sharing it") shows this is understood. Two asks:

- Put the same warning in `config.env.template` next to `TRAINSHARD_SHARED_SECRET` — that's the only text an operator is guaranteed to read, and right now it reads like an ordinary API credential rather than "everyone holding this is everyone".
- Consider requiring an explicit `TRAINSHARD_INSECURE_SIGNING=yes`-style ack for `machine=docker`, so the stand-in cannot end up on production GPUs by inertia once the compose file circulates.

Is the plan to swap in dAPI account signatures before governance enables training on mainnet? Worth stating in the PR description, since the whole `CanAsk`/`RunKey` authorization model is decorative until then.

### 3. Deployed daemon runs against an empty fake chain — intended?

Shard records come from `chainfake` seeded by `TRAINSHARD_CHAIN_SEED`, but `docker-compose.trainshard.yml` never sets it, so the shipped daemon starts with an empty chain: readiness answers, everything else refuses (no shard, no reservation). If that's the intended scope of this stage — daemon deployable, runs driven only in seeded test setups — a one-line comment in the compose file / env template would stop operators from debugging a "broken" deploy.

### 4. Session replay dedup does not survive a daemon restart (signedhttp/once.go)

`Once.served` is in-memory, so a captured shell/logs request can be replayed after a daemon restart within the signature window (default 1m). Narrow, and under the current secret model irrelevant, but since sessions already have a store (`state.Sessions()`), persisting served ids there would close it for the real-signing era too.

### 5. Smaller points

- nginx routes `/trainshard/` through the **exempt** rate-limit zone. Signature verification is cheap but unauthenticated internet traffic still reaches the daemon unthrottled; the other authenticated services get a real zone. Was exempt chosen for the long-lived shell/log streams? If so, a comment; if not, a moderate zone plus the streaming exemption would be safer.
- `nftables` output chain: the blanket accept on the mesh interface precedes the `TRAINSHARD_DENIED_CIDRS` drops, so denied CIDRs are unenforceable over the mesh device (only via other interfaces). If denied CIDRs are meant to protect operator LANs from mesh-forwarded traffic too, the order matters; if they're only about the sandbox's own uplink, this is fine and worth a comment in the template where `TRAINSHARD_DENIED_CIDRS` is documented.
- `trainshardctl` `catalog()` builds a full assembly/ops module with zero-value deps just to enumerate command names — works, but a stray init in a future constructor turns `--help` into a crash. A static name list or a `Names()` method would be sturdier.
- Compose grants `pid: host`, docker.sock, `NET_ADMIN`+`SYS_ADMIN` — all genuinely needed (netns-by-pid, container lifecycle, wg/nft), just flagging that trainshardd is effectively root on the machine, which raises the stakes of #1/#2 accordingly.

Static review; Go isn't installed on my machine so I didn't run `go test ./...` locally — CI is green on both jobs.

x0152:
@vbgd0 @Ryanchen911 Thanks! Fixed

vbgd0:
Second pass at `fd06ef4e0`. The rework since round one is substantial and lands well: HMAC is gone in favour of secp256k1 recovery, `SigningPayload` now binds audience + query + a versioned schema + body digest, spent request ids are persisted (`localstate/served.go`) and held exactly until the signature stops verifying, `/trainshard/` has its own rate and connection zone, `Autokick` patience is counted from `UnpreparedAt` rather than the reservation, and `bindable` no longer trips over a leased node's own port. All good. Two things below block real deployments; the rest are small.

### 1. On an authz-mode host the daemon cannot start, and the only way to make it start is to put the cold key on the GPU machine

`docker-compose.trainshard.yml` hands the daemon `TRAINSHARD_KEY_NAME=${KEY_NAME}` — on mainnet that is the **warm** ML-ops key, with `ACCOUNT_PUBKEY` naming the cold participant account (`config_manager.go:646-653`, `tx_manager.go:993`). But `trainshardd/main.go` insists `signer.Address() == TRAINSHARD_PARTICIPANT`, and `mesh.VerifyIdentity` requires the identity signer to equal `Node.Participant`. The warm key's address is not the participant's, so every authz-mode host exits at startup with "the key signs as X and this daemon speaks for Y". A host can satisfy the check only by placing the cold account key into the trainshardd container — which also has `pid: host`, `docker.sock`, `SYS_ADMIN` — and that is exactly what the warm/cold split exists to prevent.

Suggested shape: let the daemon sign with the warm key (the compose file is already right about *which* key), and have verifiers accept a signer that is either the participant or holds an authz grant from the participant — the chain answers that directly (`authz` `Grants(granter=participant, grantee=signer)`), and both the coordinator (`mesh.Collect`) and the peer hosts (`signedhttp.Guard` → `Actor.AuthorizedFor`) already have a chain client in hand. The startup check then becomes "signer is the participant or its grantee". The `IdentityPayload`/`SigningPayload` contracts don't need to change.

### 2. `trainshardctl assemble` / `settle` can hang forever

`Signer.landed` polls `GetTx` until the context ends, and `trainshardctl` calls `Assemble`/`Settle` with `context.Background()` (cli/root.go:43,60). A transaction accepted by CheckTx but later dropped from the mempool is never found, so the CLI spins with no error. `cfg.timeout` already exists for the host calls; bound the chain calls with it too, or set a timeout height on the tx so the chain itself settles the question.

### 3. Grants for the new messages (cross-ref #1350)

`dapi.OptIn` / `dapi.Release` go through `/admin/v1/tx/send`, i.e. the warm key under `MsgExec`. Existing hosts have no authz grant for `MsgRefreshTrainingNodeOptIn` / `MsgAutokickTrainshardNode` (they were added to `InferenceOperationKeyPerms` in `b108da23d`, and the v0.2.16 upgrade does not backfill). I've asked for the backfill on #1350; until it lands, the run guide's "check the opt-in landed on chain" step should say to re-run `grant-ml-ops-permissions` first, or operators will read the silent failure as a daemon bug.

### 4. Smaller points

- `chain.inventory()` sums `Count` over every `Hardware` entry and names the model after the first one. The chain's own `CanonicalGpuProfileId` strips CPU/RAM/DISK/NIC/... types, which means those entries are expected in `HardwareNode.Hardware`; if an ML node reports any of them, `SameGPUs` compares 8 against 8+N and the node never becomes ready. Reuse the same non-GPU filter here.
- `TRAINSHARD_GPUS` / `TRAINSHARD_GPU_MODEL` appear in the template and in the run guide's per-machine block, but on a docker host they only feed the memory stand-in — `nvidia-smi` is the source of truth. Either drop them from the docker instructions or say they're memory-mode only, otherwise operators will "fix" a readiness mismatch by editing them.
- `.inference:/root/.inference` in the compose file should be `:ro`; the daemon only reads the keyring.
- `TRAINSHARD_CHAIN_ID` defaults to `prod-sim`. A wrong chain id fails as "signature verification failed" at broadcast, which points nowhere near the cause — read it from the node (`GetNodeInfo`) and drop the default.
- Comment drift: `netns/linux.go` says the mesh hands out `10.42` addresses; `mesh.Address` gives `10.<shard%256>.0.<rank+1>`.

Static review; CI green. Nice work on the round-one items — especially moving the audience into the signing contract before real signatures landed.

PR #1620, opened 2026-08-21
fix(versiond): grow readiness window instead of using a fixed one

maksimenkoff:
## Problem

`versiond` gives each child a fixed 60s to become ready. A child that needs longer is force-stopped and restarted, which discards the startup work it had already done. The next attempt gets the same 60s, so the window never grows and startup can never complete.

This hit two production nodes: `devshardd` v3 restoring a large session store restarted every ~2 minutes indefinitely and never served its port. The child reported no error — it was alive and working, just not finished within the window. Recovery required deleting the data directory, which loses unsealed sessions.

## Fix

Grow the readiness window on each failed attempt: **60s → 2m → 4m → 8m → 16m → 32m**, then keep retrying at that ceiling.

A slow-but-healthy start now gets a window large enough to finish, instead of being cut off at a constant unrelated to how much data the child has to load. Restarts never stop, so a child stuck for a reason that later clears still recovers without operator action.

## Changes

- `internal/config/config.go` — add `ReadyMaxWait` (`VERSIOND_READY_MAX_WAIT`, default 32m)
- `internal/process/manager.go` — grow the readiness window per attempt, capped at `ReadyMaxWait`; add `nextReadyWindow`
- `internal/process/ready_window_test.go` — new: tests for the backoff sequence, the cap, overflow, and config defaults

## Scope

`versioned/` only. Startup work is still discarded between attempts — persisting partial progress would require changes in `devshard/`.

### DIFF
--- a/versioned/internal/process/manager.go
+++ b/versioned/internal/process/manager.go
@@ -37,6 +37,7 @@ const (
 	storageModePostgres          = "postgres"
 	defaultDevshardShutdownGrace = 10 * time.Minute
 	installedVersionRetain       = 3
+
 )
 
 var (
@@ -114,6 +115,12 @@ func normalizeConfig(cfg config.Config) config.Config {
 	}
 	if cfg.ReadyTimeout <= 0 {
 		cfg.ReadyTimeout = 60 * time.Second
+	}
+	if cfg.ReadyMaxWait <= 0 {
+		cfg.ReadyMaxWait = 32 * time.Minute
+	}
+	if cfg.ReadyMaxWait < cfg.ReadyTimeout {
+		cfg.ReadyMaxWait = cfg.ReadyTimeout
 	}
 	if cfg.DrainPath == "" {
 		cfg.DrainPath = "/drain"
@@ -1112,6 +1119,11 @@ func (m *Manager) runChild(ctx context.Context, c *child) {
 	backoff := time.Second
 	lastStart := time.Now()
 
+	// Grows with each failed attempt, capped at ReadyMaxWait: a restart discards the
+	// child's startup work, so a fixed window can loop forever without ever letting
+	// startup finish.
+	readyWindow := m.cfg.ReadyTimeout
+
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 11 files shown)


a-kuprin:
@maksimenkoff please rebase it over "devshard-0.2.15-v5"
After I've changed target unneeded commits are at the PR (only [a467dab](https://github.com/gonka-ai/gonka/pull/1620/commits/a467dabbfb2afbacd7d019bd3cc8c0ecfd592b7b) relates to the PR)

Also please gofmt it

versioned/internal/process/manager.go
```diff
@@ -1112,6 +1119,11 @@ func (m *Manager) runChild(ctx context.Context, c *child) {
 	backoff := time.Second
 	lastStart := time.Now()
 
+	// Grows with each failed attempt, capped at ReadyMaxWait: a restart discards the
+	// child's startup work, so a fixed window can loop forever without ever letting
+	// startup finish.
+	readyWindow := m.cfg.ReadyTimeout
```
a-kuprin, line 1125: The window should be reset to `m.cfg.ReadyTimeout` at the point where the child transitions to `statusRunning`, so a post-crash restart gets fast failure detection again.

versioned/internal/process/manager.go
```diff
@@ -1358,6 +1374,15 @@ func waitForReady(ctx context.Context, port int, path string, timeout time.Durat
 // waitForChildServingReady gates the Starting -> Running transition. Modern
 // devshardd children must be logically ready on their admin listener and also
 // serve health checks on the public listener that receives proxied traffic.
+// nextReadyWindow doubles the readiness window, capped at max.
+func nextReadyWindow(current, max time.Duration) time.Duration {
```
a-kuprin, line 1378: Comment above should relate to function `waitForChildServingReady`

// waitForChildServingReady gates the Starting -> Running transition. Modern
// devshardd children must be logically ready on their admin listener and also
// serve health checks on the public listener that receives proxied traffic.

a-kuprin:
## Availability tradeoff is real and under-discussed

A child that is stuck previously got restarted every ~2 minutes; it now stays un-restarted for up to 32 minutes.

The PR treats "slow" and "hung" as indistinguishable, which is the underlying weakness. devshardd's `/ready` handler already returns a structured body distinguishing the reasons it isn't ready, but the probe in `readyEndpointReady` only looks at the status code.

It would be great to use `/ready` response body to gate the extension on evidence of progress — keep waiting while the process is alive and /ready reports "initializing", and restart promptly when the endpoint is unreachable — rather than blindly doubling.

## What should be fixed for consistency

1. **The rolling-upgrade path still has the original bug.** Because the swap uses restart=false, a devshardd that needs more than 60s to restore its session store will still fail the swap and be rolled back.

2. **Documentation not updated.** `devshard/docs/rolling-update.md` documents `VERSIOND_READY_TIMEOUT` in its settings table (line 276) but the new `VERSIOND_READY_MAX_WAIT` isn't added, and the flow diagram's readiness note is now inaccurate for the restart path.

a-kuprin:
All comments are addressed at https://github.com/gonka-ai/gonka/pull/1663 that is based on this PR

PR #1622, opened 2026-08-21
Mount hash-only /block/* on dapi so v5 hosts can consume chainoracle …

a-kuprin:
# Height-sync: dapi `/block/*` mount (hash-only)

**Companion to:** [#1615 — Height-sync protocol (v5)](https://github.com/gonka-ai/gonka/pull/1615) (`ak/height-sync-protocol` → `devshard-0.2.15-v5`)

Protocol, log plane, and failover live in #1615. This PR is the **production dapi extension**: the same `chainoracle/blocks` HTTP surface hosts already speak, mounted on decentralized-api.

---

## Summary

- Register `GET /block/latest`, `GET /block/:height`, and the rest of `blockserver.Mount` on dapi’s public Echo (root paths, not under `/v1/`).
- Wire a **hash-only** Tendermint / CometBFT RPC observer (`observer.NewTendermint`) from `chain_node.url`.
- Opt out with `DAPI_CHAINORACLE_DISABLED=1` (or `true` / `yes`).
- **Strong / `LightBlock` proving is not in this PR.** `Prove` stays unimplemented (stable 501 / `ErrProveNotImplemented`). Full prove support for Strong-mode height-sync is scheduled for later releases (Phase F in the height-sync plan).

---

## Why

#1615 points `DEVSHARD_CHAINORACLE_URL` at a chainoracle HTTP API. In testenv that is mock-dapi; in production it should be the node’s own dapi, not a separate heightsyncd process.

Without this mount, cadence still runs (Comet tip; direct-chain `GetLatestBlock` if the NewBlock cache is empty). `At(H)` against old dapi is a 404 → dummy header, and L6 does not mark. A **down** dapi (transport / 5xx) falls back to chain `At`. Landing the hash-only routes now is what makes L6 (and later Strong) able to reconcile a stamp against a real header; `/block/:height/prove` already exists on `Mount`, this observer simply does not implement it yet.

---

## What this PR does

| Change | Detail |
| ------ | ------ |
| `decentralized-api/.../public/chainoracle.go` | `mountChainOracle()` — skip if disabled / empty RPC URL; otherwise `observer.NewTendermint` + `blockserver.Mount` |
| `server.go` | Call `mountChainOracle()` at public-server startup |
| `go.mod` / `go.sum` | Depend on local `devshard` (and Comet client bits the observer needs) |

Hash-only means headers carry height + hash (and related metadata) with an **empty Commit** — enough for today’s height-sync Anchors and tip motion, not enough for Strong `(C-strong)` / dispute packets that need a `LightBlock`.

---

## What this PR does not do

- Strong-mode prove / `LightBlock` serving (next releases / Phase F)
- Height-sync protocol, heartbeats, acks, or gateway metrics (those are #1615 / obs follow-ups)
- Changing default host compose to require dapi `/block/*` (hosts still failover when routes are missing)
- Replacing mock-dapi citest coverage in #1615

---

## Relationship to #1615

| PR | Owns |
| -- | ---- |
| [#1615](https://github.com/gonka-ai/gonka/pull/1615) | `devshard` height-sync protocol, `chainoracle/blocks` library, hash-only Tendermint observer implementation, host failover, mock-dapi / citest |
| **This PR** | Production dapi **mount** of that library’s HTTP API |

Merge order: #1615 (or the `devshard` bits it lands) should be available to dapi’s `replace => ../devshard` before this mount is useful in a full tree build. The dapi change itself is intentionally small and isolated.

---

## Test plan

- [ ] Build decentralized-api against a tree that includes the hash-only `observer.NewTendermint` from #1615
- [ ] With `chain_node.url` set and `DAPI_CHAINORACLE_DISABLED` unset: `GET /block/latest` and `GET /block/<h>` return headers; tip advances
- [ ] `GET /block/<h>/prove?...` remains not implemented (501 / capability miss) — no Strong claim
- [ ] `DAPI_CHAINORACLE_DISABLED=1`: mount skipped; existing dapi routes unchanged
- [ ] Smoke: point a v5 host `DEVSHARD_CHAINORACLE_URL` at this dapi and confirm height-sync Anchors without a separate heightsyncd


vitaly-andr:
`handleAt` turns every `oracle.At` error into a 404 — pruned height, RPC timeout, node restart
mid-poll, all the same code. `client.Lookup.At` reads 404 as a capability miss and returns a
dummy with a nil error (`lookup.go:94-95`), and `failover.At` takes that as success (77-79),
so `o.chain.At` on 92 is never reached. `checkL6` then returns on the dummy, which is right
for the case its comment names — `// old dapi / no /block/:height; do not mark` — but a
timeout on the operator's own node arrives by the same route and is skipped the same way.
5xx for transport and internal errors would keep §7.3's two cases apart.

`store()` returns early on `hdr.Height <= o.latest.Height`, which is every historical lookup
once a tip exists, so `At()` results never land in `history` and the same height re-hits RPC.
Nothing is evicted either — one entry per tip for the life of the process. `tipcache.Cache`
already does both.

`pollOnce` and `At` pull whole blocks for four scalars. `rpc.Header` serves the same header
from `LoadBlockMeta` instead of `LoadBlock` (`rpc/core/blocks.go:98` vs `:130`), same hash.
Worth doing before the route is public: the proxy's generic `location /api/` strips the
prefix, so `/api/block/<h>` reaches `handleAt` from outside.

One question on the target. `upgrade-v0.2.16` still carries `devshard/chainoracle/**` with
`/block/latest`, `/block/stream` and a panicking `NewTendermint`, and mock-dapi mounts that
one, so this lands as a second copy. Is #1615's move coming to 0.2.16, or should this wait
for it?

a-kuprin:
@vitaly-andr 
Thanks for your review and findings, that is good cleanup

Transport/internal failures are now **502**; **404** is only `ErrHeaderNotFound` (pruned / unknown / Comet height-out-of-range). 5xx keeps the two §7.3 cases apart.

`store` was tip-advance-only, so historical `At()` never landed in `history`, and tips were never evicted. That path is gone. Unary `At()` is `tipcache` (window 100): NewBlock `Observe`s the tip; Header-RPC hits `Remember` and do not move `Latest()`.

`pollOnce`/`At` loaded full blocks for four scalars. Cache misses now use Comet `/header` (`LoadBlockMeta`). Same hash as `/block`.

Leaving `/block/:height` on Echo root is intentional, it can be made public and it is gated by rate limiter `api_zone`; `/api/block/<h>` is on GNKAPI. Hash-only headers are the same class of data as `/chain-rpc`.

**`/block/latest` / `/block/stream`** — agreed, those are not the event path. They were cleared


**Target `upgrade-v0.2.16`** — staying there. 0.2.16 is the chain+dapi upgrade; v5 is the devshard-only release. Production dapi mounts `common/chainoracle` (this PR).

The panicking `devshard/.../NewTendermint` and mock-dapi `/block/latest|stream` stay on this tree until v5 is
merged; they are not on the production dapi path.

gmorgachev (approved):

PR #1633, opened 2026-08-24
feat: lossless prompt-cache metadata passthrough + explicit vLLM prefix caching

### LINKED ISSUE #1632: OpenBroker strips provider prompt-cache metadata (prompt_tokens_details is always null)
# OpenBroker strips provider prompt-cache metadata (`prompt_tokens_details` is always `null`)

## Summary

OpenBroker performs prompt/prefix caching — latency collapses from 12–31 s on a cold 52k-token prompt to ~5 s on a changed-suffix reuse and 23–41 ms on a byte-identical replay — but every `/v1/chat/completions` response returns `"prompt_tokens_details": null`. The gateway normalizes away the underlying provider's cache token counts, making prefix caching invisible and unbillable to downstream consumers.

## Motivation

DeepSeek's API reports cache-served tokens (`prompt_cache_hit_tokens`) and prices cached input ~90% cheaper than fresh input — that discount is the core economics of reselling `deepseek-ai/DeepSeek-V4-Flash-0731`. Because OpenBroker strips the metadata, downstream OpenAI-compatible gateways cannot apply cached-input billing and cannot audit cache reuse. This blocks a real, measurable cost optimization for every API consumer and forces either overcharging (bill all as fresh) or undercharging (bill all as cached).

## Impact

- Who is affected (hosts, developers, validators): all developers and gateway operators using the OpenBroker API; downstream billing/analytics systems; hosts paying for cold inference when prefix reuse is not credited.
- Is effect network-wide or limited: **network-wide** — every response from the gateway carries no cache metadata, regardless of model or node.
- Likelihood: **common** — observed on 100% of requests across multiple runs and models (cold, changed-suffix, and exact-replay responses all return `null`).
- Severity [Impact x Likelihood]: **Medium** — high impact (billing correctness + pricing feasibility) x common likelihood, but no data loss or service failure.
- Affected components: OpenBroker gateway response normalization (`usage` object on `POST /v1/chat/completions`); upstream provider telemetry passthrough.

## Detailed description

Reproduce with a 3-request sequence against `https://api.openbroker.gonka.gg/v1/chat/completions` (model `deepseek-ai/DeepSeek-V4-Flash-0731`, a fixed 2,000-line system prefix ~52k prompt tokens, `max_tokens: 16`, `temperature: 0`):

1. **REQ 1 — cold prefix** (suffix A)
2. **REQ 2 — same prefix, changed suffix** (suffix B)
3. **REQ 3 — byte-identical replay of REQ 2** (suffix B again)

### Evidence — latency and cache metadata per run (direct to api.openbroker.gonka.gg, 2026-08-24)

| Run | REQ 1 — cold prefix | REQ 2 — changed suffix | REQ 3 — exact replay | `prompt_tokens_details` |
|---|---|---|---|---|
| Run 1 | 200 · 13,028 ms | 200 · 5,231 ms | 200 · 41 ms | `null` |
| Run 2 | 200 · 11,996 ms | 429 (rate limited) | 429 (rate limited) | `null` |
| Run 3 | 200 · 31,196 ms | 200 · 5,054 ms | 200 · 23 ms | `null` |

Request IDs (sample): `req-1787552573073481834-7795508`, `req-1787552586020653004-2215078`, `req-1787552591250553152-2215178`.

### Evidence — response usage object (identical shape on all requests)

| Field | Value |
|---|---|
| `prompt_tokens` | 52,027 |
| `completion_tokens` | 16 |
| `total_tokens` | 52,043 |
| `prompt_tokens_details` | `null` |

The latency signature proves caching is active (cold 12–31 s → changed-suffix ~5 s → replay tens of ms), yet no response carries the cached-token count.

Requested behavior — pass through the provider's cache metadata, e.g.:

```json
"usage": {
  "prompt_tokens": 52027,
  "prompt_tokens_details": { "cached_tokens": 48003, "uncached_tokens": 4024 }
}
```

and/or `usage.prompt_cache_hit_tokens`, and optionally `x-openbroker-prefix-cache: hit|miss` + `x-openbroker-cached-tokens: N` response headers.

Links to evidence: full dual-side test results are available on request; measurements taken on 2026-08-24 from an external host.

theaungmyatmoe:

Closes #1632. OpenBroker/network responses always return `usage.prompt_tokens_details: null`. Root cause (confirmed with the gateway team): vLLM writes `null` through as-is, and any typed round-trip through `completionapi.Usage` drops cache metadata if a host ever emits it — the struct only defines `prompt_tokens` and `completion_tokens`.

Live reproduction (2026-08-24, 3-request sequence, 52k-token prefix, `deepseek-ai/DeepSeek-V4-Flash-0731`):

| Request | Status | Latency | `prompt_tokens_details` |
|---|---|---|---|
| 1. cold prefix | 200 | 12,028–31,196 ms | `null` |
| 2. changed suffix | 200 | 5,054–5,486 ms | `null` |
| 3. exact replay | 200 | 23–41 ms | `null` |

Latency proves KV-cache reuse is happening per-host (same shard 60824 served #2 and #3), yet the metadata never reaches the client. Downstream gateways cannot apply cached-input billing or cache telemetry.

1. **`common/completionapi/openai.go`**: extend `Usage` with `total_tokens` and `prompt_tokens_details` (`cached_tokens`/`uncached_tokens`), preserved losslessly through every typed `Response` round-trip. `null` still passes through as `null` (no synthetic values).
2. **`mlnode` vLLM runner**: explicitly add `--enable-prefix-caching` to the launch args (was implicit vLLM default) and surface `prefix_caching` in the config summary, so hosts/operators can see and deliberately control the setting that makes KV reuse possible.

## What risks does this introduce? How can we mitigate?

- Field additions to a shared struct: `omitempty` keeps old serialization byte-identical when the provider sends nothing. `Usage.IsEmpty()` semantics unchanged.
- `--enable-prefix-caching`: on by default in vLLM; making it explicit changes nothing at runtime for hosts that kept the default, but guarantees consistent cache behavior across versions. Operationally reversible per-host.
- No billing logic changes: this is metadata passthrough only. Cached-input discounting remains a separate (commercial/routing) decision.

- `common/completionapi/usage_test.go`: round-trip test proves a provider payload with `prompt_tokens_details` survives unmarshal → re-marshal intact; streamed usage chunk carries `cached_tokens`; `null` details parse without error.
- Existing `responseprocessor_test.go` expectations updated for the new `total_tokens` field (test data contains `total_tokens: 41`).
- Verified: `go test ./completionapi/...`, `go build ./...` + full `go test ./...` in `common`, `go build ./...` + `go test ./cmd/devshardd/inference/...` in `devshard`.
- mlnode: new `tests/unit/test_runner_prefix_caching.py` (4 tests) — runs in CI where torch is installed; syntax/imports verified locally.

- `common/completionapi` (usage contract — shared by decentralized-api, devshard, validation)
- `mlnode/packages/api` (vLLM runner + tests)
- No chain, no protobuf, no consensus changes.

## Optional: ai-reviewer

Not run (no binary available locally); happy to run before merge if maintainers want it.

### DIFF
--- a/common/completionapi/openai.go
+++ b/common/completionapi/openai.go
@@ -133,8 +133,20 @@ type Logprob struct {
 }
 
 type Usage struct {
-	PromptTokens     uint64 `json:"prompt_tokens"`
-	CompletionTokens uint64 `json:"completion_tokens"`
+	PromptTokens        uint64               `json:"prompt_tokens"`
+	CompletionTokens    uint64               `json:"completion_tokens"`
+	TotalTokens         uint64               `json:"total_tokens,omitempty"`
+	PromptTokensDetails *PromptTokensDetails `json:"prompt_tokens_details,omitempty"`
+}
+
+// PromptTokensDetails carries provider-reported prompt-cache metadata. When a
+// provider serves a prompt prefix from cache, the cached token count is the
+// authoritative number for downstream cached-input billing and caching
+// telemetry. The field is preserved losslessly so that gateway round-trips
+// through the typed Response struct never drop it.
+type PromptTokensDetails struct {
+	CachedTokens   uint64 `json:"cached_tokens,omitempty"`
+	UncachedTokens uint64 `json:"uncached_tokens,omitempty"`
 }
 
 func (u *Usage) IsEmpty() bool {
--- a/mlnode/packages/api/src/api/inference/vllm/runner.py
+++ b/mlnode/packages/api/src/api/inference/vllm/runner.py
@@ -170,6 +170,13 @@ class VLLMRunner(IVLLMRunner):
         # summary and disable it deliberately.
         if "--enable-prefix-caching" not in self.additional_args:
             self.additional_args.append("--enable-prefix-caching")
+        # --enable-prompt-tokens-details makes vLLM emit
+        # usage.prompt_tokens_details.cached_tokens on cache hits — the
+        # telemetry this PR's completionapi contract preserves end-to-end.
+        # Required alongside prefix caching, otherwise cache reuse stays
+        # invisible to gateways and clients.
+        if "--enable-prompt-tokens-details" not in self.additional_args:
+            self.additional_args.append("--enable-prompt-tokens-details")
         self.processes: List[subprocess.Popen] = []
         self._hb = {}
         self._hb_lock = threading.Lock()
@@ -184,6 +191,8 @@ class VLLMRunner(IVLLMRunner):
             "model": self.model,
             "dtype": self.dtype,
             "prefix_caching": "--enable-prefix-caching" in self.additional_args,
+            "prompt_tokens_details": "--enable-prompt-tokens-details"
+            in self.additional_args,
             "max_num_seqs": self._get_arg_value("--max-num-seqs", default=0),
             "max_model_len": self._get_arg_value("--max-model-len", default=0),
             "tensor_parallel_size": self._get_arg_value("--tensor-parallel-size", default=0),
--- a/mlnode/packages/api/src/api/inference/vllm/runner.py
+++ b/mlnode/packages/api/src/api/inference/vllm/runner.py
@@ -164,6 +164,12 @@ class VLLMRunner(IVLLMRunner):
             self.additional_args.extend(
                 ["--worker-extension-cls", self.WORKER_EXTENSION_CLASS]
             )
+        # Prompt/prefix caching is the basis for KV-cache reuse across
+        # requests. Make it explicit instead of relying on the vLLM default
+        # so hosts and operators can see the effective setting in the config
+        # summary and disable it deliberately.
+        if "--enable-prefix-caching" not in self.additional_args:
+            self.additional_args.append("--enable-prefix-caching")
         self.processes: List[subprocess.Popen] = []
         self._hb = {}
         self._hb_lock = threading.Lock()
@@ -177,6 +183,7 @@ class VLLMRunner(IVLLMRunner):
         return {
             "model": self.model,
             "dtype": self.dtype,
+            "prefix_caching": "--enable-prefix-caching" in self.additional_args,
             "max_num_seqs": self._get_arg_value("--max-num-seqs", default=0),
             "max_model_len": self._get_arg_value("--max-model-len", default=0),
             "tensor_parallel_size": self._get_arg_value("--tensor-parallel-size", default=0),
--- a/mlnode/packages/api/src/api/inference/vllm/runner.py
+++ b/mlnode/packages/api/src/api/inference/vllm/runner.py
@@ -164,17 +164,11 @@ class VLLMRunner(IVLLMRunner):
             self.additional_args.extend(
                 ["--worker-extension-cls", self.WORKER_EXTENSION_CLASS]
             )
-        # Prompt/prefix caching is the basis for KV-cache reuse across
-        # requests. Make it explicit instead of relying on the vLLM default
-        # so hosts and operators can see the effective setting in the config
-        # summary and disable it deliberately.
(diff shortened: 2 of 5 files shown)


theaungmyatmoe:
@GLiberman @gmorgachev pls review

common/completionapi/openai.go
```diff
@@ -133,8 +133,20 @@ type Logprob struct {
...
+	TotalTokens         uint64               `json:"total_tokens,omitempty"`
+	PromptTokensDetails *PromptTokensDetails `json:"prompt_tokens_details,omitempty"`
+}
+
+// PromptTokensDetails carries provider-reported prompt-cache metadata. When a
+// provider serves a prompt prefix from cache, the cached token count is the
+// authoritative number for downstream cached-input billing and caching
+// telemetry. The field is preserved losslessly so that gateway round-trips
+// through the typed Response struct never drop it.
+type PromptTokensDetails struct {
```
qdanik, line 147: should we add `--enable-prompt-tokens-details` as a required vllm param?

theaungmyatmoe: Agreed — and I've added it. `--enable-prefix-caching` creates the KV reuse; `--enable-prompt-tokens-details` is what makes vLLM actually emit `usage.prompt_tokens_details.cached_tokens` on cache hits. Without it, the `completionapi` contract change stays dormant and responses keep returning `null`.

Pushed in `0067f10`, same pattern as the prefix-caching flag:

- `VLLMRunner.__init__` appends `--enable-prompt-tokens-details` when absent (explicit default, not duplicated if a host already passes it)
- `get_config_summary()` now exposes `prompt_tokens_details: true|false` alongside `prefix_caching` so operators see the effective setting
- Unit tests added: default-on, no-duplicate, config-summary reporting

One note: the flag only reports real cache hits, so on hosts that deliberately disable prefix caching it returns `cached_tokens: 0` rather than fabricating anything — the downstream contract treats 0 as a confirmed miss. Hosts can still opt out deliberately; it's just no longer the implicit default.

qdanik:
if the host will override the `cached_tokens` to 0 and `uncached_tokens` to `completion_tokens` just to increase his income? There is no mechanism to validate the real cached_tokens on the host side

mlnode/packages/api/src/api/inference/vllm/runner.py
```diff
@@ -164,6 +164,19 @@ def __init__(
             self.additional_args.extend(
                 ["--worker-extension-cls", self.WORKER_EXTENSION_CLASS]
             )
+        # Prompt/prefix caching is the basis for KV-cache reuse across
+        # requests. Make it explicit instead of relying on the vLLM default
+        # so hosts and operators can see the effective setting in the config
+        # summary and disable it deliberately.
+        if "--enable-prefix-caching" not in self.additional_args:
+            self.additional_args.append("--enable-prefix-caching")
```
qdanik: btw `--enable-prefix-caching` already enabled by default in vllm since V0 was replaced by V1

theaungmyatmoe: Good catch — prefix caching has been the default in vLLM since V0 was replaced by V1, so injecting the flag was redundant. Fixed in `66cb8cc`:

- Removed the `--enable-prefix-caching` injection from `VLLMRunner.__init__`
- Kept only `--enable-prompt-tokens-details` (the actual telemetry fix this PR needs)
- `get_config_summary()` now reports `prefix_caching: true` as the V1 default with a note instead of reading a flag we no longer inject
- Tests updated to assert the flag is NOT injected and that prefix caching is reported as the V1 default

The meaningful change — making vLLM emit `prompt_tokens_details.cached_tokens` on cache hits — is unchanged.

PR #1634, opened 2026-08-24
`devshard` Host challenge receipt async

a-kuprin:
Originally by @redstartechno
https://github.com/gonka-ai/gonka/pull/1460

Extended by

1. Add an execution deadline. `context.WithoutCancel` leaves the detached execution with no bound, so a wedged engine pins `h.executing[id]` forever (verified) and the DAPI-fallback loop in `inference/engine.go:242` spins infinitely. Wrap in `context.WithTimeout(..., ExecutionTimeout)`.
2. Emit `MsgConfirmStart` from `challengeReceiptLocked`, as `signReceipt` does. Without it the inference stays `StatusPending`, so the `MsgFinishInference` the goroutine produces can never apply — the work is unbillable.
3. Test asserts `execCtx.Done() == nil`, which `context.Background()` also satisfies — assert `request_id` survives to actually pin `WithoutCancel`.

### DIFF
--- a/devshard/cmd/devshardctl/proxy_test.go
+++ b/devshard/cmd/devshardctl/proxy_test.go
@@ -1908,6 +1908,13 @@ func TestRunInference_SpeculativeFallsThroughMultipleDeadHosts(t *testing.T) {
 	err := env.proxy.redundancy.RunInference(context.Background(), defaultParams(), &buf, nil)
 	require.NoError(t, err)
 
+	// RunInference returns once the live host's stream settles. Dead hosts
+	// may still be finishing on the background finalizer, which is where
+	// RecordRequest runs. Wait for that record before asserting on it.
+	require.Eventually(t, func() bool {
+		return len(env.proxy.perf.RecentRequests()) >= 1
+	}, time.Second, 10*time.Millisecond, "background finalizer should have recorded the request")
+
 	requests := env.proxy.perf.RecentRequests()
 	require.NotEmpty(t, requests)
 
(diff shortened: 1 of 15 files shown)


redstartechno:
Original author of #1460 here — thanks for picking this up and for the credit. Both gaps you closed are real.

The `MsgConfirmStart` one is the serious miss on my side: I detached the execution but never queued the confirmation, so the `MsgFinishInference` the goroutine produces had nothing valid to apply against. `TestHost_ChallengeReceipt_PublishesConfirmStartSoFinishCanApply` replaying the ConfirmStart/Finish pair onto a peer state machine and asserting `StatusFinished` is the right shape of proof — that is the check my original tests were missing, not just the assertion strength.

Two things I verified rather than take on faith, both of which hold:

- The "only reachable while the record is still pending" claim on the ConfirmStart block: `challengeReceiptLocked` returns early unless `rec.Status == types.StatusPending` (`host.go:1541-1543`), so the block is unreachable once a ConfirmStart has applied, and the `hasConfirmStart` mempool scan covers the still-queued case. No duplicate path that I can find.
- The budget anchor: `VerifyExecutionTimeout` measures from `rec.ConfirmedAt` (`timeout.go:120-121`), which is the same `confirmedAt` this path stamps at `host.go:1554`. So starting the budget when the challenge arrives lines up with the window verifiers actually enforce, rather than drifting from it.

One small note on `defaultExecutionBudget`. The comment derives it from settleability — work past `ExecutionTimeout` can no longer be settled — but that reasoning doesn't select a value in the branch where it applies. When `ExecutionTimeout` is 0, `nowUnix-rec.ConfirmedAt < config.ExecutionTimeout` is false immediately, so a timeout vote is acceptable from the start and there is no settleable window to match. The 32m constant is really bounding the wedge (keeping `h.executing[id]` from being pinned for the process lifetime), which is a different and perfectly good justification — it just isn't the one the comment gives. Non-blocking; the behavior looks right either way.

Also, your read on my write-up was correct: it overstated the caller surface. Corrected on my side.

a-kuprin:
Tested with e2e + unit from https://github.com/gonka-ai/gonka/pull/1634

PR #1643, opened 2026-08-25
`broker`: don't stall the broker command loop on node removal

a-kuprin:
## Summary

- Deleting a node while its worker is in an ML call parked the broker’s **only** command goroutine for up to 15 minutes, and could deadlock it forever if `highPriorityCommands` filled.
- `RemoveNode` now cancels the in-flight reconcile context, unregisters the worker immediately, and drains shutdown off that loop (5s cap).
- `Submit` after close returns `false` instead of panicking.
- Reconcile stamps a **broker-wide** `DeploymentGeneration` (not per-node) so a leftover result after same-id re-register cannot finalize the new node.

---

## Problem

The broker processes every command on a single goroutine (`processCommands`): `RemoveNode`, `LockAvailableNode` / `ReleaseNode` (inference assignment), result updates, sync, register.

### Actors

| Actor | Role |
|---|---|
| Broker loop | Serial `Execute` of every command |
| Node worker | `Execute` (HTTP to ML node, client timeout **15 min**) then **blocking** `QueueMessage(UpdateNodeResultCommand)` on `highPriorityCommands` (cap 100) |
| Reconcile loop | Submits work; stores `cancelInFlightTask` (no deadline) |
| Admin `DELETE /admin/v1/nodes/:id` | Queues `RemoveNode`, then waits on `<-response` |

### Data flow before (hang / deadlock)

```
Reconcile: intended ≠ current
  → ctx, cancel := WithCancel (no timeout)
  → cancelInFlightTask = cancel
  → worker.Submit(InferenceUp / Stop / …)

Worker: Execute(ctx)  →  HTTP to ML node (up to 15 min)
        then QueueMessage(result)  →  highPriorityCommands
        then wg.Done()

DELETE node-7:
  HTTP handler → QueueMessage(RemoveNode) → blocks on response
  Broker loop: RemoveNode.Execute
    → RemoveWorker → Shutdown → wg.Wait()   ← loop is stuck here
    → never calls cancelInFlightTask
```

Two outcomes:

**F1 — head-of-line stall (usual).** Worker is in `Execute`. Nobody cancels `ctx`. `Wait` lasts until the 15-minute HTTP timeout. For that entire time **no** broker command runs: `LockAvailableNode` stops, so the participant stops assigning inference even on healthy nodes. The DELETE request hangs too.

**F2 — deadlock (escalation).** `wg.Done()` runs only *after* a blocking send on `highPriorityCommands`. If that buffer is full (other workers’ results, sync, register batches during F1), `run()` never `Done`s, `Wait` never returns, and the only goroutine that would drain the buffer is the one in `Wait`. Restart is the only exit.

Retrying DELETE does not help: the second `RemoveNode` sits behind the first on the same loop.

`select { case commands <- cmd: default: }` only guards a **full** channel, not a **closed** one. After `Shutdown` closes `commands`, a racing `Submit` (reconcile still holding the worker pointer) panics `send on closed channel` and kills the process ([#1494](https://github.com/gonka-ai/gonka/pull/1494)).

---

## Fix (new data flow)

```
DELETE node-7
  Broker loop: RemoveNode.Execute
    1. under b.mu: copy cancelInFlightTask; unlock
    2. cancel()                         ← D1: HTTP should abort now
    3. RemoveWorkerAsync
         unregister from map immediately
         go ShutdownWithTimeout(5s)     ← D2/D3: wait is bounded and off-loop
    4. delete node from b.nodes
    5. response ← true                  ← loop is free

Worker: Execute sees ctx.Done()
        QueueMessage(result)            ← broker can drain the buffer
        wg.Done()

LockAvailableNode queued after RemoveNode
  → processed promptly (other nodes keep serving)
```

| Decision | What it does |
|---|---|
| **D1** Cancel before waiting | Reuses reconcile’s `cancelInFlightTask`; ML calls already thread `ctx`. Unlock before wait so `Execute` can take `b.mu`. |
| **D2** Bound the wait | Do not trust cancel (ignored `ctx`, or `QueueMessage` blocked). |
| **D3** 5s | Long enough for a cancelled HTTP round-trip; short enough not to look like a hang. |
| **D4 / step 2** Reaper goroutine | Command loop never waits, even for those 5s. |

`ShutdownWithTimeout`: set `closed`, close `shutdown` once, `Wait` with `NewTimer` (stopped on the success path). Timeout → log Error and return `false`; worker is already unregistered. Late `UpdateNodeResultCommand` is “unknown node” (`commands.go`).

`Submit` takes `closedMu` across the non-blocking send: after close it returns `false` (no panic, no `WaitGroup` Add/Wait race).

### Leftover result after same-id re-register (D5)

After a timed-out drain the old `run()` may still live until the 15-minute client timeout. If the operator re-adds the **same node id** before that, the leftover `UpdateNodeResultCommand` is no longer “unknown node”: a new `NodeState` is in `b.nodes`.

`upgrade-v0.2.16` already ignores a result whose `DeploymentGeneration` ≠ `ReconcileInfo.Generation`. That only helps when the numbers differ. Generation used to live on `NodeState` and increment on dispatch (`0 → 1` on the first reconcile). `RegisterNode` builds a **new** `NodeState`, so the counter reset to 0. The abandoned worker’s first attempt and the new node’s first attempt were both generation **1**, same intended status → the stale result applied.

Rejecting reconcile until the old worker dies would close that window, but it would park the re-added node for up to 15 minutes — the wait this PR moved off `DELETE`.

**D5 — broker-wide counter.** `Broker.deploymentGeneration` is an `atomic.Uint64` that only ever goes up for this process. `reconcile` calls `nextDeploymentGeneration()` (`Add(1)`) and copies that value onto `NodeState.DeploymentGeneration` and `ReconcileInfo.Generation`. Delete / re-register does not reset it.

```
DELETE node-7 (worker still in Execute, stamped gen 1)
POST same id
  new NodeState { DeploymentGeneration: 0 }
  first reconcile → nextDeploymentGeneration() → 2
  ReconcileInfo.Generation = 2

old Execute returns, result gen 1
  UpdateNodeResultCommand: 1 ≠ 2 → ignore
new Execute returns, result gen 2
  match → finalize
```

`NodeState.DeploymentGeneration` is only a snapshot of the last value assigned to that node object (JSON / debug). Uniqueness is the broker counter, not the per-node field.

Same-node cancel + redispatch (no delete) still works: 1 then 2 on the same object, as before.

---

## Tests

`decentralized-api/broker/node_worker_shutdown_test.go`  
CI: **Build and Test API Wrapper** (`go test ./...` in `decentralized-api`). Deterministic channel waits, no sleep-until-done.

| Test | What it pins |
|---|---|
| `TestRemoveNode_DoesNotBlockBrokerLoop` | D1: in-flight `Execute` sees cancel; `RemoveNode` returns quickly; `LockAvailableNode` after it is not blocked; worker is unregistered |
| `TestNodeWorker_ShutdownWithFullBrokerQueue` | D2: `ShutdownWithTimeout(100ms)` returns `false` while `QueueMessage` is blocked; draining the filler delivers the result; `Execute` treats it as unknown node |
| `TestNodeWorker_ShutdownWithTimeout_CleanDrain` | Normal path still runs all queued commands (deadline does not truncate) |
| `TestNodeWorker_ShutdownWithTimeout_ReturnsFalseOnHang` | Hung `Execute` → `false`; after release, second wait → `true` |
| `TestNodeWorker_ShutdownIdempotent` | Second `Shutdown` does not panic; `Submit` after close is `false` |
| `TestNodeWorkGroup_RemoveWorkerUnregistersOnTimeout` | Async remove drops the map entry before drain finishes |
| `TestRemoveNode_UnknownId` | Missing id → `false`, no panic |
| `TestRemoveNode_NilCancelInFlightTask` | No in-flight cancel → still unregisters and returns `true` |
| `TestReconcile_BrokerWideGenerationSurvivesSameIdReregister` | D5: delete + re-add same id gets gen 2, not 1; leftover gen-1 result is ignored; matching gen 2 finalizes |

Existing `TestNodeWorker_GracefulShutdown` still requires all queued commands to finish on `Shutdown()`.

Existing `TestNodeWorker_GracefulShutdown` still requires all queued commands to finish on `Shutdown()`.

---

## Test plan (reviewer)

- [ ] `go test ./broker/ -race -run 'TestRemoveNode_\|TestNodeWorker_Shutdown\|TestNodeWorkGroup_RemoveWorker\|TestReconcile_BrokerWideGeneration'`
- [ ] Staging: point a node at a host that accepts TCP but never answers, let reconcile dispatch, `DELETE /admin/v1/nodes/:id` — request returns quickly and inference assignment continues
- [ ] Before this PR the same DELETE hung and `LockAvailableNode` stopped

### DIFF
--- a/decentralized-api/broker/broker.go
+++ b/decentralized-api/broker/broker.go
@@ -140,6 +140,10 @@ type Broker struct {
 	configManager        *apiconfig.ConfigManager
 	lockMap              map[string]lockEntry
 	lockMapMu            sync.Mutex
+	// deploymentGeneration is broker-wide and never resets when a node is
+	// deleted or re-registered, so a leftover result from an abandoned worker
+	// cannot share a generation with a new node's first reconcile.
+	deploymentGeneration atomic.Uint64
 }
 
 type lockEntry struct {
@@ -231,8 +235,10 @@ type NodeState struct {
 	PoCValidationInference  bool      `json:"poc_validation_inference"`
 	DeploymentUpdatePending bool      `json:"deployment_update_pending"`
 	DeploymentRetryAfter    time.Time `json:"deployment_retry_after,omitempty"`
-	// DeploymentGeneration increments for each dispatched reconciliation so
-	// late results from a cancelled attempt cannot be applied to a newer one.
+	// DeploymentGeneration is the last broker-wide generation assigned to this
+	// node. Uniqueness comes from Broker.deploymentGeneration, not this field:
+	// a fresh NodeState after re-register starts at 0 and is assigned the next
+	// global value on dispatch.
 	DeploymentGeneration uint64 `json:"deployment_generation,omitempty"`
 
 	// Epoch data for this node, keyed by model_id.
@@ -969,6 +975,10 @@ func (b *Broker) checkAndRefreshClientsIfNeeded() {
 	}
 }
 
+func (b *Broker) nextDeploymentGeneration() uint64 {
+	return b.deploymentGeneration.Add(1)
+}
+
 func (b *Broker) reconcileIfSynced(triggerMsg string) {
 	epochPhaseInfo := b.phaseTracker.GetCurrentEpochState()
 	if epochPhaseInfo.IsNilOrNotSynced() {
@@ -1045,8 +1055,8 @@ func (b *Broker) reconcile(epochState chainphase.EpochState) {
 		ctx, cancel := context.WithCancel(context.Background())
 		intendedStatusCopy := currentNode.State.IntendedStatus
 		pocIntendedStatusCopy := currentNode.State.PocIntendedStatus
-		currentNode.State.DeploymentGeneration++
-		generation := currentNode.State.DeploymentGeneration
+		generation := b.nextDeploymentGeneration()
+		currentNode.State.DeploymentGeneration = generation
 		currentNode.State.ReconcileInfo = &ReconcileInfo{
 			Status:     intendedStatusCopy,
 			PocStatus:  pocIntendedStatusCopy,
--- a/decentralized-api/broker/deployment_state_test.go
+++ b/decentralized-api/broker/deployment_state_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"decentralized-api/apiconfig"
+	"decentralized-api/chainphase"
 	"decentralized-api/mlnodeclient"
 
 	"github.com/productscience/inference/x/inference/types"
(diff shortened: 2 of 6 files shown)


a-kuprin:
The better designed solving same problem:
https://github.com/gonka-ai/gonka/pull/1701

PR #1644, opened 2026-08-25
feat(selfcheck): offline broker PoC intended-state self-test

bonujel:
## What this is

A `decentralized-api selfcheck` subcommand: it drives the real broker against a
mocked chain (`MockChainBridge`) through synthetic PoC phases and asserts the
broker's intended-state transitions at each stage. Exit 0 on PASS, 1 on FAIL.

```
selfcheck: PASS
  [PASS] node-registered
  [PASS] epoch-models-populated
  [PASS] hardware-diff-submitted
  [PASS] poc-generate
  [PASS] poc-validate
  [PASS] inference-resumed
```

It is meant for operator confidence before joining the network — "does my
participant binary react to epoch events the way it should?" — without needing
a chain or an MLnode to answer it.

## Why it is a separate PR

Split out of the onboarding work in #1296, which is too large to review as one
change (51 files). #1296 stays open as the umbrella; further slices will follow
the same way #1518 did.

This slice is self-contained: a new `decentralized-api/selfcheck` package plus
26 lines in `main.go` for the subcommand. Nothing existing changes behaviour.

## Scope

It verifies broker PoC-phase wiring at the **intended-state** level. It
deliberately does not exercise a real MLnode worker, so it says nothing about
whether a node actually loads a model — that is what the MLnode test in the
next slice is for.

The approach follows what reviewers suggested on #866: leave the broker
untouched, drive it with synthetic events, and observe outcomes from a separate
evaluator.

## Verification

Both required suites, locally against this head:

- `Build and Test chain` — build clean, `go test ./...` green (38 packages)
- `Build and Test API Wrapper` — build clean, `go test ./...` green, container-backed tests included
- `decentralized-api selfcheck` run end to end against the mocked chain: all six assertions pass, exit 0

### DIFF
--- a/decentralized-api/main.go
+++ b/decentralized-api/main.go
@@ -18,6 +18,7 @@ import (
 	"decentralized-api/poc"
 	"decentralized-api/poc/artifacts"
 	"decentralized-api/poc/earlyshare"
+	"decentralized-api/selfcheck"
 	"decentralized-api/statsstorage"
 	"net"
 
@@ -86,6 +87,10 @@ func main() {
 	}
 	if len(os.Args) >= 2 && os.Args[1] == "pre-upgrade" {
 		os.Exit(1)
+	}
+	if len(os.Args) >= 2 && os.Args[1] == "selfcheck" {
+		runSelfcheck()
+		return
 	}
 
 	configManager, err := apiconfig.LoadDefaultConfigManager()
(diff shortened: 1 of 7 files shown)


bonujel:
One known limitation worth recording here: `selfcheck` builds a throwaway broker
and has no way to reclaim it. `NewBroker` starts four background goroutines plus
one per node, and there is no `Stop()` on `main` today, so those keep querying
MLnodes and the chain after `Run` returns.

It is harmless as shipped — `runSelfcheck` ends in `os.Exit`, so the process
takes the goroutines with it. It would matter to anyone calling `selfcheck.Run`
as a library, or constructing brokers repeatedly in tests.

#1646 adds the `Stop()` this needs. Once that lands I will add the call here, so
this PR does not have to grow a broker change to fix its own cleanup.

PR #1651, opened 2026-08-26
Validation retry tests + Fix validation retry recovery and lease release semantics

aikuznetsov:
## Changes

- Add validation retry coverage for stale lease reclaim and endpoint behavior.
- Add host cooldown and validation lease race coverage.
- Add e2e coverage for withholding executor scenarios.
- Replace the old session retry path with `ValidationRetryLoop`.
- Extend validation lease storage semantics for:
  - stale `pending` leases
  - stale `submitted` leases
  - epoch-scoped lease ownership
  - explicit lease release on non-submit paths
- Add bounded payload retrieval for validation fetches.
- Add validation fault/test hooks for deterministic retry and withholding tests.
- Add metrics/test coverage around validation lifecycle behavior.

## Fixes

- Fix validation retry so executor-attributable payload fetch failures produce `Valid:false` instead of leaving validation stuck.
- Fix validation leases being left `pending` after failed/non-submit retry paths.
- Fix stale `submitted` validation leases being ignored by retry recovery.
- Fix retry recovery to respect lease ownership before submitting validation results.
- Fix retry behavior after manager/session restart, including cases where validation tx existed only in volatile mempool.
- Fix withholding executor behavior so payload fetches are bounded and cannot hold a validation worker indefinitely.
- Fix retry release behavior for `Challenged` inferences so the hot path can publish the validation vote.

## Why stale `submitted` leases are retried

`submitted` means the retry loop already produced a validation result and pushed the validation tx into the host mempool. It does **not** mean the tx became durable.

If the process restarts after mempool submit but before the tx is applied into a diff, the Postgres lease can remain `submitted` while the actual validation tx is lost with the volatile mempool. Previously that state could become stuck forever: the lease was no longer `pending`, but there was also no durable validation result.

This PR treats stale `submitted` leases as retryable. On recovery, the retry loop checks the live host state first:

- if validation is already applied, it marks the lease skipped;
- if validation is still in mempool, it keeps/marks it submitted;
- if the mempool tx was lost, it revalidates and republishes safely under lease ownership.

That makes `submitted` a recoverable intermediate state instead of a terminal assumption.

### DIFF
--- a/devshard/cmd/devshardd/config_test.go
+++ b/devshard/cmd/devshardd/config_test.go
@@ -1,6 +1,10 @@
 package main
 
-import "testing"
+import (
+	"testing"
+
+	"devshard/cmd/devshardd/session"
+)
 
 func TestValidateBinaryLogVersion(t *testing.T) {
 	t.Parallel()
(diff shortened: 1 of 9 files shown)


a-kuprin:
**I propose add coverage for following scenario:**

Edge-case

It is possible that A didn't submit (gateway didn't took the validation but it is still in A's mempool) but ttl already passed and B can take the lease. So A has it in mempool, and B aquires the lease

## What the test should do

**1. A takes the stale lease and actually finishes the work.**

A’s retry loop steals the row, runs ML, gets `Valid: true`, puts a signed `MsgValidation` in **A’s mempool**, marks the lease `submitted`.

At this point the vote exists only in A’s RAM. Postgres only knows “submitted by replica-a.” B’s mempool is empty.

**2. Time passes, A is still running.**

The test ages `claimed_at` past TTL. It does **not** kill A. A still has the true vote in mempool, waiting for the gateway to pull it into a diff.

**3. B’s retry loop runs anyway.**

`AcquireOneStale` now matches `submitted` rows. B steals the lock (`status` back to `pending`, owner `replica-b`).

Before revalidating, B checks:

- live `ValidatedBy`? no — never applied
- **local** mempool? empty — B cannot see A’s mempool

So B thinks the work was lost. It fetches/ML again. In the test, B is wired to return `Valid: false` (same as “fetch failed, vote false”). B puts a **second** `MsgValidation` in **B’s** mempool.

Now two live processes hold two txs for the same slot, opposite bits. The lease belongs to B.

a-kuprin (approved):
I think it's good but there still is very rare edge-case described

PR #1652, opened 2026-08-26
`devshard` Account streamed host/ML errors as `TIMEOUT_REASON_ERROR` misses

a-kuprin:
Design: [proposals/error-finish-miss.md](https://github.com/gonka-ai/gonka/blob/ak/devshard-error-stream-finish/devshard/docs/proposals/error-finish-miss.md).

## Problem

A host can return an OpenAI-style error as the inference result and end up in a
state that is neither served, nor validated, nor charged as a miss.

This is an **HTTP 200 `text/event-stream`** whose body is an error envelope, not
an HTTP 5xx from the ML node. An HTTP 5xx never reaches the processor, so no
Finish would exist. The host treats the error event as a parseable completion,
stores the payload, and signs `MsgFinishInference` over `sha256(that payload)`.

Consequences today:

| Path | Behaviour |
| --- | --- |
| Client | `hostApplicationError`, HTTP 500. Not a success. |
| Miss vote | Skipped — the nonce is already finished |
| Chain | `StatusFinished`, host credited `ActualCost` |
| Validation | Eligible; if sampled, `Invalid` (no logprobs) |

The record is finished enough to block a miss while remaining unusable. The
signed `ResponseHash` is real evidence — the executor signed "this error is my
answer" — but nothing consumes it.

## Solution

Turn that artifact into an accountable miss, immediately.

- Reuse `MsgTimeoutInference`. Add `TIMEOUT_REASON_ERROR = 3`. No new messages,
  no new fields on any `Msg`.
- Apply **Finish then Timeout** in the same diff. Finish is the proof
  (`ProposerSig` over the error body). Timeout converts it into accounting:
  `StatusTimedOut`, `Missed++`, client refunded, host unpaid.
- Verifiers classify the same bytes the executor signed. No deadline, no RPC to
  the failing host.
- The client still receives today's `hostApplicationError`. Unchanged.

```text
error SSE (HTTP 200) → Finish queued
        │
        ▼
 gateway pins Finish, relays finish_tx + payload to verifiers
        │
        ▼
 VerifyErrorTimeout (local): sig, hash pin, IsTerminalErrorResponse
        │
        ▼
 same-diff compose: [Finish, Timeout{reason=ERROR, votes}]
        │
        ▼
 StatusTimedOut, Missed++, ActualCost unwound
```

### Why there is no wait

`REFUSED` and `EXECUTION` exist because the executor said nothing, so the only
evidence is that a deadline passed. `ERROR` is the opposite: the executor
answered and signed the answer. A 32-minute `ExecutionTimeout` would only push
the miss toward seal, where a finished record is evicted and the timeout is
rejected as `is sealed`.

### Classifier (fail closed)

`common/completionapi.IsTerminalErrorResponse` accepts only a streamed body
with a top-level `error` object, no content anywhere, and
`usage.completion_tokens == 0`. Unparseable extra `data:` JSON vetoes the miss.
Content-then-error stays on today's path. Ambiguous cases degrade to current
behaviour, not a false miss.

### Same-diff compose

A `StatusFinished` record is seal-eligible. If concurrent heartbeat /
`PrepareInference` / `SendPendingDiff` published Finish alone, the later Timeout
can miss the seal window and the miss is lost.

`handleErrorTimeout` pins that nonce's pending Finish (and any Timeout already
queued) for the vote round. Timeout is passed as `extraTxs` into the same locked
compose that includes the pinned Finish — it is not parked in the shared pending
queue first.

### Accounting

`ERROR` is the first timeout that applies **after** Finish. Surplus was already
returned, so unwind uses `ActualCost`, not `ReservedCost`:

- `Balance += ActualCost` (client gets the full reservation back)
- `HostStats.Cost -= ActualCost` (host is not paid)
- `Missed++`, not `Invalid`

Net: escrow whole, host Cost back to pre-Finish, miss recorded.

## Test plan

- [ ] Classifier fixtures, including EngineCore payload and unparseable-then-error
- [ ] State: Finish+ERROR same diff; Timeout-before-Finish rejected; sealed Finish
      rejects Timeout; `REFUSED`/`EXECUTION` vote bytes unchanged
- [ ] Verifier: no deadline, no executor contact; lying `OutputTokens` still
      accepts; content-then-error rejects
- [ ] Gateway: miss lands under a second with realistic `ExecutionTimeout`;
      concurrent compose cannot publish Finish without Timeout
- [ ] Citest A5 (`make citest-adversarial`) on a 3-host stack


anatoly-kuz-mntn (approved):
Looks solid to me. The original compile blocker is fixed, unit/integration coverage is green, and the A5 full-stack scenario is already added under `citest-adversarial`.

a-kuprin:
Also reviewed with @qdanik 
Concerns are fixed at 
https://github.com/gonka-ai/gonka/pull/1652/commits/51d88bb6bf95cdd3e04f0f11895ffb44c302425d
https://github.com/gonka-ai/gonka/pull/1652/commits/5892fc06c25b8629a181c2436a832f1b3e9ea9eb

PR #1666, opened 2026-08-28
feat(gateway): Bound always-stream aggregation (spool + fold)

a-kuprin:
**Supersedes [#1579](https://github.com/gonka-ai/gonka/pull/1579)** (`ak/only-streaming-to-gateway`). That PR mixed always-stream forcing with host LiveStream / same-nonce reconnect / hop timestamps. Always-stream itself landed in [#1581](https://github.com/gonka-ai/gonka/pull/1581) on an unbounded `bytes.Buffer` + `assembleSSEBody`. This PR is the deferred cherry-pick: shared scratch spool, local SSE fold, kill switch, and the two independent gates that should land before default-on soak. **Keep #1579 closed; do not reopen it.**

Noticed at comment https://github.com/gonka-ai/gonka/pull/1581#issuecomment-5273617406

Design notes: [pr-1581-followups.md](./pr-1581-followups.md), [spool-shared-library.md](./spool-shared-library.md), [gateway-always-stream-upstream-plan.md](./gateway-always-stream-upstream-plan.md) (Steps 12–13).

## Summary

- Fold winner SSE into a JSON `chat.completion` with `spool.Buffer` (RAM then spill) and a local `completionFolder` instead of `assembleSSEBody`, so a long always-stream body cannot OOM the gateway or return a silently truncated 200.
- Keep `ForceUpstreamStreaming` as a persisted kill switch (default **on**, matching #1581 production) so an admin can roll the force back without a deploy.
- Cap a single gateway↔host SSE line at 1 MiB (`ErrSSEEventTooLarge`) and the legacy JSON `ReadAll` at 16 MiB, so a host that never sends `\n` cannot grow `ReadBytes` for the inference deadline.
- Force explicit `n > 1` to `1` on the wire until reservation/settlement budget more than one `MaxTokens` output. Omitted `n` stays omitted.

Client-visible contract is unchanged from #1581: streaming clients still get SSE; non-streaming clients still get one JSON body. This PR changes how that JSON body is accumulated and how a hostile/oversize host is aborted.

### What this takes from #1579

| Taken | Not taken (stays out of this PR) |
|---|---|
| `devshard/spool` substrate (`Dir` / `File` / `Buffer` / `Index` / `Budget` / `Slots`) | Host LiveStream, disk resume, same-nonce reconnect |
| Gateway aggregate + logprobs spill onto `spool.Buffer` | Hop `:devshard-ts` comments |
| Local fold (`stream_aggregate.go`) replacing `assembleSSEBody` | Cross-instance ML reattach |
| Force-upstream + aggregate-spill citest | Step 14 default-on soak / dashboards |
| Step 12 SSE event cap, Step 13 `n=1` short-term gate | Proper multi-choice reservation (protocol/validation) |

#1579's reconnect/HA stack was meant to land through #1574 / #1575. This PR does not pull it.

### U1–U4 (body handling after #1581)

| ID | Change | Heals |
|---|---|---|
| U1 | `handleNonStreaming` writes into `aggregateResponseBuffer` → `spool.Buffer` (2 MiB RAM / 16 MiB ceiling / concurrent slots) | Unbounded RAM, no byte ceiling, `ReadAll` of a spilled body |
| U2 | `aggregateSSEStreamReader` + `completionFolder`; `assembleSSEBody` deleted | Silent 65k-event 200, `chat.completion` short-circuit, OpenAI `logprobs`/`role` drift, forced-logprobs trees |
| U3 | `replaceNonFiniteNumbers` on fold ingest | vLLM `-Infinity` / `NaN` dropping the content delta with the logprob |
| U4 | `force_upstream_streaming` admin setting, request snapshot, SQLite persist, default on | Unconditional force with no rollback |

Spool alone does not heal U2. The fold is a separate layer.

### Steps 12–13 (independent gates)

**Step 12.** `HTTPClient.parseSSEResponse` used unbounded `ReadBytes('\n')`. A selected executor can answer 200 + `text/event-stream`, write `data: `, and stream without a newline for the whole inference timeout. `readBoundedSSELine` aborts at `DefaultMaxSSEEventBytes` (1 MiB) with `ErrSSEEventTooLarge` → failure reason `sse_event_too_large` so redundancy escalates. The compat JSON `ReadAll` in `Send` is bounded at `MaxJSONResponseBytes` (16 MiB, same as the aggregate wire ceiling) so omitting `text/event-stream` cannot bypass the cap.

**Step 13 (short-term).** PostLimits `ForceLiteralParameter{Value: 1, OverwriteOnly: true}` on `n`. Explicit `n > 1` is rewritten to `1`; omitted `n` stays omitted. Proper fix (reserve `n × max_tokens`, validation multi-choice) remains open.

## Test plan

- [ ] `go test` under `devshard`: `./cmd/devshardctl/ ./spool/ ./transport/`
- [ ] Unit: `force_upstream_stream_test.go`, `stream_aggregate_test.go`, `aggregate_response_test.go`, `gateway_store_test.go` (persist kill switch), `transport/client_test.go` (`ErrSSEEventTooLarge` / near-limit event / JSON body cap)
- [ ] `make -C devshard/testenv citest-force-upstream-streaming` (shape, usage strip, logprobs, spill, oversize abort)
- [ ] Confirm default-on: without an admin flip, upstream is still forced stream (unchanged from #1581)
- [ ] Confirm kill switch: `POST /v1/admin/settings` `{redundancy:{force_upstream_streaming: false}}` restores client-shaped upstream; mid-flight flip does not change an in-flight snapshot
- [ ] Confirm `n > 1` is rewritten to `1` on the wire; omitted `n` is still omitted
- [ ] Do not reopen [#1579](https://github.com/gonka-ai/gonka/pull/1579)


qdanik (approved):

Ryanchen911:
Hi @a-kuprin — relaying a production observation from a 6Block devshard-gateway deployment that rolled v4 (this PR plus the always-stream / force-upstream-streaming work) up on 2026-08-31. Since the cutover, non-stream and stream failures both jumped sharply with very uniform timing, so I wanted to share the data and a concrete question before we decide how to proceed on our side.

## Symptom (post-cutover)
- **Non-stream** `/v1/chat/completions`: very frequent `client_aborted` at a flat **~125.01s**, `first token 0ms`, `tokens 0`, no `X-DEVSHARD-ID`.
- **Stream + non-stream**: frequent gateway 502 `inference: winner inference incomplete (nonce_finished=false)` and `inference: no non-probe attempt finished` at a flat **~300s**.
- Gateway **daily 502s** before cutover (v3): ~700–1,500/day → on 08-31: **43,261** → on 09-01: **120,446** (~60% of ~200k requests that day failed). `inference incomplete` went from ~300–650/day to ~**4,880/day**.

## What we think is happening
`force_upstream_streaming` (always-stream) forces even non-stream chat requests onto the gateway↔host SSE path, then folds the winner SSE into a JSON `chat.completion`. A non-stream request therefore only gets its **first byte after the entire inference is folded** — i.e. first-byte latency = full inference latency. Under slow hosts that is enormous: gateway logs show `since_first_token_ms` of **125–300s for 3416 attempts vs <30s for only 204** in a 2h window. So a non-stream client that gives up at 125s never sees a byte → `client_aborted`; the gateway abandons at 300s → `nonce_finished=false`.

Relatedly, `5a9554f8f` ("drop the non-stream ladder the forced stream made unreachable") removed `non_stream_response_floor`(20s) / `non_stream_no_content_timeout` / `non_stream_max_attempt_wait`. With those gone, a non-stream request has **no progress/no-content guard at all** on the folded path — it just blocks until the client (125s) or the gateway (300s) gives up.

## Requests / suggestions
1. Bound first-byte wait for non-stream requests (or restore a non-stream "ladder" on the folded path), so they don't sit for the full inference time with zero signal. Under slow hosts this is the difference between a recoverable slow request and a hard 125s/300s failure.
2. Consider an automatic fallback for `ForceUpstreamStreaming`: when observed first-token latency on the selected host exceeds some threshold (e.g. 30–60s), stop forcing for non-stream (or fall back to the non-stream path) instead of mandating always-stream cluster-wide. The override currently defaults **on** globally with no degradation path.
3. Publicly document the always-stream latency tradeoff for non-stream requests, and expose an operational switch so ops can disable per-model/per-scenario rather than being committed to it.

Happy to share raw logs / DB extracts if useful. This is meant as constructive feedback — happy to be corrected if I've misread the intended behavior.

a-kuprin:
> * Gateway **daily 502s** before cutover (v3): ~700–1,500/day → on 08-31: **43,261** → on 09-01: **120,446** (~60% of ~200k requests that day failed). `inference incomplete` went from ~300–650/day to ~**4,880/day**.

It is devshardd v4 problem, not the gateway switch to always-stream. It is investigated now. But one definite reason is that v4 stopped to get snapshots on catch-ups and can start for hours (catching diffs)
This catch-up is normal after long quarantine, and this leads that devshard not responding while catching-up and gets new quarantine. Something like this

It is definitely not the forcing non-stream problem

Ryanchen911:
Got it, thanks. If you need a hand with the logs or anything else, it’s my pleasure to help.

a-kuprin:
@Ryanchen911 
You can look at and review the fix:
https://github.com/gonka-ai/gonka/pull/1703

We gonna to update v4 first

PR #1669, opened 2026-08-28
`devshard` Persist the height-sync floor on snapshot restore

a-kuprin:
Following comment by [bonujel on #1584](https://github.com/gonka-ai/gonka/pull/1584#issuecomment-5448034487).

## Summary

- Persist `FloorIndex` on the snapshot envelope (host protobuf + gateway JSON), not in hashed `EscrowState`.
- Fail restore when `LatestNonce > 0` and neither the journal nor a valid blob can rebuild the fold (`ErrFloorNotRestored`).
- Refuse apply and omit producer stamps while `!floorReady`, so a replica cannot skip L0 with an empty index.
- Validate a present blob before installing it; omit `ExportHeightSyncFloor` unless the floor actually folded.

## Why

[The #1584 comment](https://github.com/gonka-ai/gonka/pull/1584#issuecomment-5448034487) agrees that `max(own_tip, F(m))` is the producer rule, then shows a restore hole that still splits the escrow.

`rebuildHeightSyncLocked` on a `GetDiffs` error used to swallow the error, keep snapshot `h_last`, and install an empty `FloorIndex`. L0 skips when `!known || floor == 0`. Two replicas from the same snapshot then diverged:

```
replica A (journal read ok): ApplyDiff(nonce 4) -> INVALID(height_regression):
                             heartbeat height 10 < floor 50 as of nonce 4
replica B (GetDiffs failed):  ApplyDiff(nonce 4) -> nil; LatestNonce=4
```

`applyCore` rejects the whole diff on L0, so this is not one tx dropped from a set. The state root does not cover `LatestNonce`, so the split is not automatically a root mismatch — it becomes one as soon as an accepted diff charges `FeePerNonce`. The comment also notes that #1649 / #1653 restored the other derived snapshot indexes, which left `heightSyncFloor` as the one still emptied by design.

That is the same split `observeHeightSyncLocked` already refuses to create by folding only on apply.

## Behavior

**Envelope, not root.** `StateSnapshotProto.height_sync_floor = 5` and JSON `height_sync_floor`. Fields 14–20 of `EscrowStateProto` are unchanged. Settlement still hashes only `HeightSyncEscrowCommit` (the seven forced-turn / cadence scalars). `F` is not in `restHash`.

**Journal first, blob as fallback.** Restore still replays diffs `1..LatestNonce` when `GetDiffs` succeeds, because that is what rebuilds the turn tracker. The blob is installed only when that replay cannot run. A missing blob (legacy snapshot) stays a journal rebuild.

**Fail closed.** `LatestNonce > 0` and no journal and no blob → `ErrFloorNotRestored`, `floorReady=false`. Apply and `localBestEffort` reject; host/user `referenceStamp` omit rather than write a raw oracle tip against a missing `F`.

**Blob is validated.** `FloorIndexFromProto` rejects anything no diff sequence can produce (nonces/heights not strictly increasing, missing stamps, nil entries, repeated signers). A rejected blob degrades to a journal replay; restore fails closed only if that cannot run either. `ExportHeightSyncFloor` returns nil unless `floorReady`, so a not-ready machine cannot persist an empty-but-authoritative fold.

**Not this change.** `GetDiffs` returning empty with a nil error is still a successful fold. On a self-consistent store that case does not arise in production (`PruneEpoch` drops sessions, diffs, and snapshots together). It remains the test seam for synthesized `RestoreState` with no journal. Seeding `F` from the oracle is not a substitute: `F` is a function of the applied log, and a local tip would split L0 the same way.

## Test plan

- [x] `go test ./heightsync/ ./types/ ./state/ ./host/ ./user/ ./cmd/devshard-host/ ./cmd/devshardctl/`
- [x] H64 — floor round-trips in the snapshot; root and `HeightSyncFloorAsOf` still match
- [x] H102 — no blob + `GetDiffs` error → restore fails; apply is rejected; export is nil
- [x] A/B split — blob present + `GetDiffs` error → both replicas L0-reject heartbeat height 10 at nonce 4
- [x] Unfoldable blobs (out-of-order nonce/height, nil entry, repeated signer) return `ErrFloorBlobInvalid`
- [ ] Confirm a host restart with a current protobuf snapshot still applies the next heartbeat after restore
- [ ] Confirm a gateway restart from a JSON snapshot with `height_sync_floor` omitted still rebuilds from the journal

### DIFF
--- a/devshard/cmd/devshard-host/main.go
+++ b/devshard/cmd/devshard-host/main.go
@@ -20,6 +20,7 @@ import (
 
 	devshardpkg "devshard"
 	"devshard/gossip"
+	"devshard/heightsync"
 	"devshard/host"
 	"devshard/internal/boolvalue"
 	"devshard/internal/e2econfig"
@@ -353,11 +354,14 @@ func recoverHostState(store storage.Storage, sm *state.StateMachine, escrowID st
 
 	replayFrom := uint64(1)
 	if snapNonce, snapData, snapErr := store.LoadSnapshot(escrowID); snapErr == nil && snapNonce > 0 && snapNonce <= meta.LatestNonce {
-		snapState, err := host.UnmarshalStateSnapshot(snapData)
+		snapState, _, _, floorProto, err := host.UnmarshalStateSnapshotWithCommitted(snapData)
 		if err != nil {
 			return fmt.Errorf("unmarshal snapshot nonce %d: %w", snapNonce, err)
 		}
-		sm.RestoreState(snapState)
+		floor := heightsync.FloorIndexFromProto(heightsync.FloorConfigFor(len(snapState.Group), sm.HeartbeatConfig()), floorProto)
+		if err := sm.RestoreStateWithFloor(snapState, floor); err != nil {
+			return fmt.Errorf("restore snapshot nonce %d: %w", snapNonce, err)
+		}
 		replayFrom = snapNonce + 1
 	} else if snapErr != nil && !errors.Is(snapErr, storage.ErrSnapshotNotFound) {
 		return fmt.Errorf("load snapshot: %w", snapErr)
--- a/devshard/cmd/devshard-host/main.go
+++ b/devshard/cmd/devshard-host/main.go
@@ -358,7 +358,16 @@ func recoverHostState(store storage.Storage, sm *state.StateMachine, escrowID st
 		if err != nil {
 			return fmt.Errorf("unmarshal snapshot nonce %d: %w", snapNonce, err)
 		}
-		floor := heightsync.FloorIndexFromProto(heightsync.FloorConfigFor(len(snapState.Group), sm.HeartbeatConfig()), floorProto)
+		// A rejected blob degrades to a journal replay; if that cannot run
+		// either, RestoreStateWithFloor fails closed rather than serving L0
+		// from a floor we could not verify.
+		floor, floorErr := heightsync.FloorIndexFromProto(
+			heightsync.FloorConfigFor(len(snapState.Group), sm.HeartbeatConfig()), floorProto)
+		if floorErr != nil {
+			log.Printf("recover_host escrow=%s snapshot_nonce=%d floor_blob_rejected=%v (rebuilding from diffs)",
+				escrowID, snapNonce, floorErr)
+			floor = nil
+		}
 		if err := sm.RestoreStateWithFloor(snapState, floor); err != nil {
 			return fmt.Errorf("restore snapshot nonce %d: %w", snapNonce, err)
 		}
--- a/devshard/cmd/devshardctl/gateway_test.go
+++ b/devshard/cmd/devshardctl/gateway_test.go
@@ -74,7 +74,7 @@ func gatewayTestRuntimeForLimits(t *testing.T, id string, balance, nonce uint64)
 	st := sm.ExportState()
 	st.Balance = balance
 	st.LatestNonce = nonce
-	sm.RestoreState(st)
+	require.NoError(t, sm.RestoreState(st))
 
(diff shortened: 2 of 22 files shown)


bonujel (commented):
This looks right to me. Persisting the fold, preferring the journal, failing closed when neither source works, and holding producers back while the index is unverified covers the case from #1584 properly.

I checked it out at `bb89f3e` and ran `heightsync`, `host`, `state`, `user` and `types` — all green. Since a couple of the changes relax or move a safety property, here is what I verified rather than took on trust, in case it saves someone a second pass:

**The `Clone` hash sharing is safe as described.** This is the one place the PR trades an isolation guarantee for speed, so I went looking for an aliasing write and did not find one: `appendEntry` copies the hash on the way in (`heightsync/floor.go:299`), the claim insert copies too, and the in-place raise assigns a whole new `floorEntry` rather than writing into a shared array. Nothing rewrites a stored hash, so a clone sharing those slices has nothing to alias into.

**`FloorIndexFromProto` is validating the right things in the right place.** Strict nonce and height monotonicity, non-zero heights, stamps present, no repeated signer. Since `AsOf` binary-searches and reads the last entry as a running maximum, an out-of-order blob would answer `F(m)` with a height the log never held and nothing downstream would notice — so restore really is the only place that check can live.

**Preferring the journal over the blob is the right precedence**, since it also rebuilds the tracker and leaves the blob for the case it was added for.

---

One non-blocking note, offered as defense in depth rather than as a bug:

`rebuildHeightSyncLocked` sets `floorReady` on `err == nil` from `GetDiffs` (`state/heightsync.go:262`), but `GetDiffs` has no completeness contract — a missing range is not an error, only a missing session is (`storage/memory.go:562`). So "no error" is slightly weaker than the invariant `HeightSyncFloorReady` documents ("a consistent fold of 1..LatestNonce"). If a store ever returned a short range, the fold would be marked authoritative anyway, and because `ExportHeightSyncFloor` also gates on `floorReady`, the under-folded index would be written into the next snapshot envelope as authoritative.

**I could not find a path that produces that state, though**, and I looked: appends are persist-first with retry (`host/host.go:673`), snapshots are only ever written and read locally, and there is no transport that ships a snapshot from a peer — so a node's snapshot always comes from history it applied and therefore persisted. That leaves storage-level loss underneath the persist-first guarantee, which would mostly surface as a `GetDiffs` error and hit the path you already handle. So this is comfortably a "if you want the belt as well as the braces" comment: checking that the folded records cover `1..LatestNonce` before setting `floorReady` would make the code enforce what the doc comment promises. Same check would also seed the turn tracker on an empty-records restore, where state and tracker currently disagree — though that predates this PR.

---

One question, on the operational side rather than the code: `ErrFloorNotRestored` means a host with an unreadable journal and no usable blob will refuse to start (`cmd/devshard-host/main.go:298`) and refuse every apply. I think that is the right trade over splitting the escrow. Is the way out restoring the journal from a peer, and would it be worth a line in the restore comment saying so? Mostly thinking of whoever meets this for the first time during an incident.

Nothing here blocks it from my side. Happy to share the tests I wrote while poking at the restore paths if they would be useful.

a-kuprin:
> Is the way out restoring the journal from a peer, and would it be worth a line in the restore comment saying so? Mostly thinking of whoever meets this for the first time during an incident.

I think we will revisit restoration paths in next releases. It's better to rethink it in complex

a-kuprin:
> but `GetDiffs` has no completeness contract — a missing range is not an error, only a missing session is (`storage/memory.go:562`). So "no error" is slightly weaker than the invariant `HeightSyncFloorReady` documents ("a consistent fold of 1..LatestNonce").

Same here. I would better take a time and make some complex work, understanding and reinventing of catch-ups

PR #1679, opened 2026-08-30
fix(queryapi): skip validator-set lookup for zero/negative CreatedAtBlockHeight

wrvnnull:
## Summary

Export `GRPCErrorToHTTP` in `common/queryapi/handlers.go` and remove the duplicated private copy from `decentralized-api/internal/server/public/get_participants_handler.go`, reusing the shared helper instead.

## Changes
- `common/queryapi/handlers.go`: export `grpcErrorToHTTP` → `GRPCErrorToHTTP`.
- `decentralized-api/internal/server/public/get_participants_handler.go`: drop local duplicate, call `queryapi.GRPCErrorToHTTP(err)`.
- Remove now-unused `google.golang.org/grpc/codes`/`status` imports in the dapi file.

## Test plan
- Existing `decentralized-api/internal/server/public/get_participants_handler_test.go` covers both handlers.
- Caller-visible response shape/code unchanged.

No issue linked.

### DIFF
--- a/common/queryapi/epoch.go
+++ b/common/queryapi/epoch.go
@@ -188,15 +188,6 @@ func (h *Handlers) getEpochParticipants(ctx context.Context, epoch uint64) (*gen
 		}
 	}
 
-	// Validators at the creation height.
-	valsResp, err := h.chain.CometServiceClient().GetValidatorSetByHeight(ctx, &cmtservice.GetValidatorSetByHeightRequest{
-		Height: activeParticipants.CreatedAtBlockHeight,
-	})
-	if err != nil {
-		logging.Error("Failed to get validators", inferencetypes.Participants, "error", err)
-		return nil, err
-	}
-
 	// Derive participant addresses from validator keys.
 	addresses := make([]string, len(activeParticipants.Participants))
 	for i, participant := range activeParticipants.Participants {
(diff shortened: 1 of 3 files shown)


redstartechno:
Cross-linking so this does not get merged twice: #1556 makes the same `epoch.go` change (skip `GetValidatorSetByHeight` when `CreatedAtBlockHeight <= 0`, degrade to an empty validators array), was opened on 2026-08-07 against `upgrade-v0.2.16`, and was approved by @x0152 the same day. Background is in #983.

Two things worth noting for whoever reviews this one:

- The PR title describes the zero-height fix, but the body describes an unrelated refactor (exporting `GRPCErrorToHTTP` and removing the duplicate in `get_participants_handler.go`). The diff contains both. Splitting the refactor out would make the review easier.
- #1556 also updates `openapi.yaml` and adds `epoch_participants_zero_height_test.go` covering the zero-height path; those are not in this diff.

If the maintainers would rather land the fix on `main` first, I am happy to retarget #1556 or let this one carry the `epoch.go` part — either way, only one of the two should merge.

redstartechno:
Heads-up on an overlap, so a maintainer doesn't end up reviewing the same guard twice.

The first hunk here is the same fix as #1556, down to the log message: skip `GetValidatorSetByHeight` when `CreatedAtBlockHeight <= 0` and degrade to an empty validators array. #1556 was opened 2026-08-06, targets `upgrade-v0.2.16` rather than `main`, carries a regression test (`common/queryapi/tests/epoch_participants_zero_height_test.go`), and has been approved (x0152, 2026-08-07) with all checks green. I'm not claiming priority — arriving at the same guard independently is expected when CometBFT's `height must be greater than 0` is right there in the logs — just flagging it so the duplicate is visible.

The part of this PR that #1556 does *not* cover is the second change: exporting `GRPCErrorToHTTP` and deleting the copy in `decentralized-api/internal/server/public/get_participants_handler.go`. That dedup is worth having on its own, and it's unrelated to the zero-height guard — splitting it out would let it land without waiting on the overlap to be resolved.

One thing that affects this PR either way: #1752 (`5e729d61b`, merged 09-15) added two more uses of the same height to that handler, both after the point this patch guards — `proofHeight := activeParticipants.CreatedAtBlockHeight` fed to `ABCIQuery(..., Prove: true)` (height 0 serves the latest state, so the proof is anchored to the wrong root), and `heightP1 := CreatedAtBlockHeight + 1`, which becomes block 1 and is what the proof then gets verified against. Neither is fatal, so the endpoint returns 200 with a proof bundle that doesn't verify. Worth guarding in the same pass, whichever PR ends up carrying this.

PR #1687, opened 2026-08-31
fix(bls): MsgSubmitDealerPart - unvalidated ECIES ciphertext length

GLiberman:
**Summary**

- Tighten MsgSubmitDealerPart encrypted-share size checks on chain (ValidateBasic and keeper shape): each ciphertext must be between 113 and 1024 bytes.
- On the API, treat a failed share decrypt as an invalid dealer and keep verification going with the remaining dealers.
- If dealer processing hits an unexpected error, abort that verification attempt instead of submitting a partial or incorrect vector.
- Fee-exempt MsgExec now runs the inner message ValidateBasic as well, so wrapped dealer parts use the same size checks.


a-kuprin:
The application-level gate here looks right, but the root cause is still live in `github.com/gonka-ai/cosmos-sdk`: `crypto/ecies` diverges from the go-ethereum code it was copied from in two places — `GenerateShared` dropped the `IsOnCurve` check, and `Decrypt`'s length gate is `rLen+hLen+1` (98) instead of `rLen+hLen+params.BlockSize` (113). Anything in the 98–112 band reaches `symDecrypt`, where `make([]byte, len(ct)-params.BlockSize)` is a negative length and panics.

That means this PR fixes one caller, not the primitive. `kr.Decrypt` / `ecies.Decrypt` stays panic-on-attacker-input for every other consumer, now and future, and the new `decryptKeyring` recover is what stands between us and a process kill. Better to fix it upstream in the fork — restore the `IsOnCurve` check in `GenerateShared` and raise the `Decrypt` floor to `rLen+hLen+params.BlockSize`, matching geth — and keep this PR as defense-in-depth. Worth adding the 98-byte regression test on the fork side too, since `ecies_short_ciphertext_test.go` currently reimplements the fork's KDF/MAC internals from outside the module just to reach the panic.

I think it is good to fix primitive, while PR sounds and closes what it should

PR #1701, opened 2026-09-01
fix(dapi): do not block the broker loop on RemoveNode

GLiberman:
RemoveNode ran on the shared processCommands loop and waited for the node worker to finish. Worker HTTP (inference/up, stop, PoC) can sit until the 15-minute ML-node client timeout. That blocked StartPocCommand for every node.

In the incident, node was deleted and re-registered while inference/up was hung. Generate ran with the loop still stuck on that wait, so PoC start was skipped.

RemoveNode now unregisters immediately, cancels the in-flight reconcile, shuts the worker down in the background, and drops queued worker commands instead of draining a 15-minute backlog.

Because shutdown no longer waits, the same node id can be registered again while the old worker is still in HTTP. DeploymentGeneration resets on register, so a late UpdateNodeResultCommand can match the new node. Each RegisterNode gets a process-local RegistrationSeq; results from a previous registration are ignored.

Because DELETE returns immediately, in-flight inference and a status scan can still complete against the old host. Lock leases and status updates also carry RegistrationSeq; a mismatch is ignored so the replacement’s concurrency accounting and health are not updated by the previous registration.

Applied-deployment SQLite delete stays on the command loop (an async delete can wipe a same-id re-register) and is capped at 1s.


decentralized-api/broker/node_worker_test.go
```diff
@@ -624,6 +682,54 @@ func TestNodeWorkGroup_AddRemoveWorkers(t *testing.T) {
 	assert.False(t, exists1, "Worker 1 should not exist after removal")
 }
 
+func TestNodeWorkGroup_RemoveWorkerDoesNotBlockOnInFlightHTTP(t *testing.T) {
```
a-kuprin, line 707: This covers hung HTTP. Please also pin the deadlock: `QueueMessage` is a blocking send on `highPriorityCommands`, and `wg.Done()` runs only after that send. Fill a cap-1 broker channel, park the worker in `QueueMessage`, then `RemoveWorker`. Assert it returns immediately, the worker is gone, queued commands did not run, and after you drain the filler the late result is delivered and ignored on a re-register (`RegistrationSeq` mismatch). On the old code this never returns.

decentralized-api/broker/node_admin_commands.go
```diff
@@ -299,24 +300,46 @@ type RemoveNode struct {
...
 		command.Response <- false
 		return
 	}
-	delete(b.nodes, command.NodeId)
 	if b.configManager != nil {
-		if err := b.configManager.DeleteAppliedDeploymentsForNode(context.Background(), command.NodeId); err != nil {
+		ctx, cancel := context.WithTimeout(context.Background(), appliedDeploymentDeleteTimeout)
+		err := b.configManager.DeleteAppliedDeploymentsForNode(ctx, command.NodeId)
+		cancel()
+		if err != nil {
```
a-kuprin, line 367: Keeping the delete on the loop is right, but on timeout you only Warn and continue. A same-id re-register then calls `refreshDeploymentUpdatePendingFromApplied` and can inherit the previous deployment and skip redeploy. Either retry, or refuse to trust a leftover row.

PR #1712, opened 2026-09-03
`devshard V5` Restore host snapshot load on v4 session recovery

a-kuprin:
It's just carrying same commits as https://github.com/gonka-ai/gonka/pull/1703 but targeted to be merged to v5


a-kuprin:
As this PR just merging already tested in production and reviewed PR https://github.com/gonka-ai/gonka/pull/1703 it will be merged

PR #1723, opened 2026-09-07
fix(devshard): reject incomplete stream cache entries

### LINKED ISSUE #1722: Incomplete HTTP 200 streams are cached and replayed as successful responses
## Problem

The devshard gateway can store an incomplete streamed chat response in the one-hour response cache. Identical requests then receive the same partial body immediately instead of starting a new inference.

## Observed behavior

A production DeepSeek streaming request returned partial reasoning for 1,205.663 seconds and then closed without `finish_reason` or usage. Eight identical retries returned the same 1,369,407-byte body in 0.871–2.758 seconds. The response fingerprint was identical through two broker endpoints.

A second response included `data: [DONE]` but had no `finish_reason`, usage, or final answer. The DONE marker made the response appear complete to clients even though generation did not finish.

No request prompts or credentials are included in this report.

## Cause

`gatewayChatCacheCapture.cacheEntry` calls `cacheableResponse`, which accepts any non-empty HTTP 2xx body unless it contains a recognized error. It does not validate semantic SSE completion.

This combines badly with streaming paths that can append `[DONE]` after `RunInference` returns without a terminal model chunk. A partial response can therefore become a deterministic cached result for one hour.

## Expected behavior

A successful streaming response must be cached only when:

- every observed choice has a non-empty `finish_reason`;
- `[DONE]` occurs after those terminal choice events;
- the body has no OpenAI-style error event.

Previously cached incomplete stream entries should also be rejected on lookup.

## Reproduction

Pass this HTTP 200 body to `gatewayChatCacheCapture.cacheEntry` with `stream=true`:

```text
data: {"choices":[{"index":0,"delta":{"reasoning":"still working"},"finish_reason":null}]}

data: [DONE]
```

Current result: the entry is cacheable.

Expected result: the entry is rejected because `[DONE]` is framing, not semantic completion.

## Proposed fix

Validate semantic stream completion in the shared cache eligibility path used by capture, Set, and Get. A focused PR with regression tests will follow.

As a separate operational escape hatch, the gateway could honor request `Cache-Control: no-cache, no-store`. This header must not replace completion validation.

aleksandr-cstl:
## Summary

- require semantic completion before caching a successful chat response, for both the streaming and the aggregated non-streaming shape;
- a choice is terminal when `finish_reason` or `stop_reason` carries a non-null value (matching `rewriteStreamingDataEvent`);
- streaming bodies additionally require `[DONE]` after every observed choice is terminal;
- deterministic error bodies keep their existing cacheability, including the streaming shape where a host rejection is emitted as a 200 SSE error event;
- validation runs once, in `cacheEntry`; `Set` and `Get` do not re-scan the body. An entry that does not fit the byte cap is reported as `skipped_too_large`.
- every skip is logged as `gateway_cache_skipped` with a reason and counted in `devshard_gateway_chat_cache_total{result="skipped_<reason>"}`.

## Problem

The current cache accepts any non-empty HTTP 2xx body. A hard timeout or other mid-stream termination can therefore store partial reasoning or content. Identical requests then receive the same truncated body for the one-hour cache TTL. A synthetic `[DONE]` marker does not prove that model generation finished, and the non-streaming aggregator folds the same truncation into a `chat.completion` whose choices have `finish_reason: null`.

Fixes #1722.

## Review focus

`trackChoiceCompletion` records every observed choice index (decoded as raw JSON, so unusually typed `index`/reason fields do not drop the event) and marks it finished on a non-null `finish_reason` or `stop_reason`. `completeStreamingChatResponse` requires `[DONE]` after all observed choices are finished; `completeChatChoices` applies the same scan to a single aggregated JSON payload. A body with no complete choices is still cacheable only if it is a deterministic error payload that `cacheableResponse` already vetted.

## Testing

- `go test ./cmd/devshardctl -count=1` on `gateway-v5`
- `TestGatewayPooledChatDoesNotCacheIncompleteResponse` reproduces the production shape end-to-end through `handlePooledChat` for both request modes: the second identical request must reach the runtime.
- Unit cases that fail on the previous validator: aggregated non-streaming truncation, `stop_reason`-only termination, string-typed `index`, multi-choice with one open choice, deterministic error stream.

### DIFF
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -1554,14 +1554,7 @@ func (g *Gateway) handlePooledChat(w http.ResponseWriter, r *http.Request) {
 	if capture := g.serveChatToRuntime(rt, "/v1/chat/completions", body, w, r); capture != nil {
 		sourceRequestID, _ := requestLogFromContext(ctx)
 		entry, reason := capture.cacheEntry(rt.id, stream, sourceRequestID, r.Context().Err())
-		if reason == "" {
-			g.chatCache.Set(cacheKey, entry, time.Now())
-			g.metrics.RecordChatCache(requestModel, "stored")
-			logRequestStage(ctx, "gateway_cache_stored", "escrow", rt.id, "model", requestModel, "stream", stream, "bytes", len(entry.Body))
-		} else {
-			g.metrics.RecordChatCache(requestModel, "skipped_"+reason)
-			logRequestStage(ctx, "gateway_cache_skipped", "escrow", rt.id, "model", requestModel, "stream", stream, "reason", reason)
-		}
+		g.storeChatCache(ctx, "gateway_cache", cacheKey, rt.id, requestModel, stream, entry, reason)
 	}
 }
 
@@ -1692,14 +1685,7 @@ func (g *Gateway) handleDevshard(w http.ResponseWriter, r *http.Request) {
 		if capture := g.serveChatToRuntime(rt, innerPath, body, w, r); capture != nil {
 			sourceRequestID, _ := requestLogFromContext(ctx)
 			entry, reason := capture.cacheEntry(rt.id, stream, sourceRequestID, r.Context().Err())
-			if reason == "" {
-				g.chatCache.Set(cacheKey, entry, time.Now())
-				g.metrics.RecordChatCache(limitModel, "stored")
-				logRequestStage(ctx, "gateway_devshard_cache_stored", "escrow", rt.id, "model", limitModel, "stream", stream, "bytes", len(entry.Body))
-			} else {
-				g.metrics.RecordChatCache(limitModel, "skipped_"+reason)
-				logRequestStage(ctx, "gateway_devshard_cache_skipped", "escrow", rt.id, "model", limitModel, "stream", stream, "reason", reason)
-			}
+			g.storeChatCache(ctx, "gateway_devshard_cache", cacheKey, rt.id, limitModel, stream, entry, reason)
 		}
 		return
 	}
@@ -1861,6 +1847,19 @@ func (g *Gateway) serveChatToRuntime(rt *devshardRuntime, path string, body []by
 	capture := &gatewayChatCacheCapture{ResponseWriter: w}
 	rt.handler.ServeHTTP(capture, req)
 	return capture
+}
+
+func (g *Gateway) storeChatCache(ctx context.Context, stage, cacheKey, escrow, model string, stream bool, entry cachedChatResponse, reason string) {
+	if reason == "" && !g.chatCache.Set(cacheKey, entry, time.Now()) {
+		reason = "too_large"
+	}
+	if reason != "" {
+		g.metrics.RecordChatCache(model, "skipped_"+reason)
+		logRequestStage(ctx, stage+"_skipped", "escrow", escrow, "model", model, "stream", stream, "reason", reason)
+		return
+	}
+	g.metrics.RecordChatCache(model, "stored")
+	logRequestStage(ctx, stage+"_stored", "escrow", escrow, "model", model, "stream", stream, "bytes", len(entry.Body))
 }
 
 func (g *Gateway) recordGatewayRequestOutcome(model, outcome, reason string) {
--- a/devshard/cmd/devshardctl/gateway.go
+++ b/devshard/cmd/devshardctl/gateway.go
@@ -1516,6 +1516,7 @@ func (g *Gateway) handlePooledChat(w http.ResponseWriter, r *http.Request) {
 	cacheKey := chatCacheKey(requestModel, body, clientIntent)
 	stream := chatRequestStream(body)
 	if entry, ok := g.chatCache.Get(cacheKey, time.Now()); ok {
+		g.metrics.RecordChatCache(requestModel, "hit")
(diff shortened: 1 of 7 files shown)


qdanik:
@aleksandr-cstl please, take a look into `gateway-v5` branch and prepare your fix there if an issue still exists

aleksandr-cstl:
Confirmed: the issue still exists on `gateway-v5` at `39240311`. Its cache eligibility still accepts any non-empty HTTP 2xx stream, and the streaming handler can still append `[DONE]` without a model `finish_reason`.

I rebased this PR onto `gateway-v5` and force-updated the branch. The only rebase conflict was the new `fmt` import used by the v5 cache key; the semantic completion validation applies cleanly to the v5 implementation.

Verification on the rebased branch:

```text
go test ./cmd/devshardctl -count=1
ok devshard/cmd/devshardctl 120.375s
```

qdanik:
Two gaps.

**Non-streaming replies skip the check.** `cacheableChatResponse` returns early on `!stream`, and `entry.Stream` is what the client asked for, not the forced upstream flag. So a `stream: false` client still gets a truncated answer replayed for an hour: the aggregator folds it into a normal-looking `chat.completion` with `finish_reason: null`, and the cache stores it. The same choice scan works on that body — it just needs to run for both shapes.

**`stop_reason` should count as terminal too.** Right now only `finish_reason` does, but `rewriteStreamingDataEvent` ends a choice on either one. A host that finishes with `stop_reason` alone will never be cached again.

Minor: `Index int` and `FinishReason string` fail the whole event decode when a host types either differently, and that choice is then never seen. `json.RawMessage` for both would be safer.

aleksandr-cstl:
Thanks, all three addressed in `42bf4702`.

- **Non-streaming:** `cacheableChatResponse` no longer returns early on `!stream`. The same choice scan (`completeChatChoices`) now runs on the aggregated `chat.completion` body, so a folded truncation with `finish_reason: null` is rejected. Test fixtures that returned bare `{"message": ...}` choices now carry `finish_reason: "stop"` to remain cacheable.
- **`stop_reason`:** a choice is terminal on a non-null `finish_reason` *or* `stop_reason`, matching `rewriteStreamingDataEvent`.
- **Typing:** `index`, `finish_reason`, and `stop_reason` are decoded as `json.RawMessage`; a missing `index` is treated as `0`.

New cases (aggregated truncation, `stop_reason`-only, string `index`, second choice open) fail against the previous validator and pass now. `go test ./cmd/devshardctl -count=1` is green on `gateway-v5`.

aleksandr-cstl:
Self-review pass, `f0cc7611`:

- **Regression I introduced:** `handleStreaming` emits a host rejection before the first chunk as a 200 SSE `data: {error}` + `[DONE]`. Deterministic ones were cacheable before and my completion scan rejected them (no choices). Now a body without complete choices is still cacheable when it is an error payload already vetted by `cacheableResponse`. Covered by the `deterministic error` case, which fails on `42bf4702`.
- **Hot path:** I had made `Get` re-run the completion scan on every hit. The cache is process memory and `Set` is the only producer, so that bought nothing and doubled the per-hit parse on multi-MB bodies. `Get` is back to the pre-existing error screen; `Set`/capture remain the gate (`TestChatResponseCacheSetRejectsIncompleteStream`).
- **End-to-end coverage:** `TestGatewayPooledChatDoesNotCacheIncompleteResponse` drives `handlePooledChat` twice with a truncated stream / a folded `finish_reason: null` completion and asserts the runtime is called twice.
- `docs/proxy-architecture.md` cache section updated to state the completion requirement.

qdanik:
### Gaps

**1. Truncated + an error event is still cached.** `cacheableChatResponse` ends with `return responseBodyHasErrorPayload(body)`, and the gateway writes such errors itself: `proxy.go:460` on an incomplete winner, `:431`/`:440` on a host error. Measured, body = one chunk with `finish_reason: null`:

| body | before | after |
| --- | --- | --- |
| truncated + `[DONE]` | cached | refused |
| truncated + `winner inference incomplete` | cached | **cached** |
| truncated + host error + `[DONE]` | cached | **cached** |
| truncated + `empty content stream` + `[DONE]` | cached | **cached** |
| truncated + `upstream timeout` | refused | refused |

Fall through to that path only when no choice was seen — `stream_aggregate.go:606` already does it that way.

**2. Complete answers refused when an event won't parse.** The check reads one `data:` line as one object, and an unparsed line means its choice is never counted. Two real shapes: an event split over two `data:` lines, which the package's reader joins (`stream_aggregate.go:169-176`), and a terminal chunk with `-Infinity`, which `replaceNonFiniteNumbers` handles only in the fold (`stream_aggregate.go:453`). Each refusal costs a full inference and a nonce.

**3. Tests don't exercise the fix.** Every "complete" fixture keeps content and `finish_reason` in one chunk; vLLM sends the terminal chunk separately, so the cross-event state the fix adds never runs. Two mutations survive the suite: a later `finish_reason: null` cancelling a finished choice, and dropping the empty-string check in `jsonValuePresent`.

### Also

- Refusal is silent: no log beside `gateway_cache_stored`, no cache series in `metrics.go`.
- The body is validated twice per store (`cacheEntry`, then `Set`): 4.3 ms → 9.7 ms per pass on a 620 KB stream.
- `jsonValuePresent` accepts `false` and `0` as a finish reason.
- The PR description says `Get` evicts incomplete entries; the diff doesn't touch `Get`.

qdanik:
@aleksandr-cstl could you please create a new PR into [devshard-0.2.x-v6](https://github.com/gonka-ai/gonka/tree/devshard-0.2.x-v6)?

PR #1729, opened 2026-09-07
fix(inference): key model capacities lookup by epoch index

### LINKED ISSUE #1725: x/inference: GetAllModelCapacities keys the epoch-group map with a block height, so the query always returns empty
`GetAllModelCapacities` reads `EpochGroupDataMap` with a block height where the map is keyed by epoch index, so the lookup never hits and the query returns an empty list on every call.

## The mismatch

The map is written keyed by `EpochIndex`:

```go
// inference-chain/x/inference/keeper/epoch_group_data.go:11
func (k Keeper) SetEpochGroupData(ctx context.Context, epochGroupData types.EpochGroupData) {
	k.EpochGroupDataMap.Set(ctx, collections.Join(epochGroupData.EpochIndex, epochGroupData.ModelId), epochGroupData)
}
```

and read with the parameter named `epochIndex`:

```go
// inference-chain/x/inference/keeper/epoch_group_data.go:16
func (k Keeper) GetEpochGroupData(ctx context.Context, epochIndex uint64, modelId string) (val types.EpochGroupData, found bool) {
	val, err := k.EpochGroupDataMap.Get(ctx, collections.Join(epochIndex, modelId))
```

but this call site passes a block height instead:

```go
// inference-chain/x/inference/keeper/query_dynamic_pricing.go:116
mainEpochData, found := k.GetEpochGroupData(goCtx, uint64(currentEpoch.PocStartBlockHeight), "")
```

`Epoch.Index` and `Epoch.PocStartBlockHeight` are different quantities on different scales — a sequential epoch counter versus a chain height — so they coincide only by accident. `found` is therefore false, and the handler takes its early return.

## Why it is quiet

The early return is not an error:

```go
// query_dynamic_pricing.go:117
if !found {
	k.LogError("Failed to get epoch group data for capacity query", types.Pricing)
	return &types.QueryGetAllModelCapacitiesResponse{
		ModelCapacities: modelCapacities,   // nil
	}, nil
}
```

so the RPC answers `200` with an empty list rather than surfacing a failure, and every consumer sees a well-formed "no capacities" response.

## Consumers that cannot tell the difference

`common/queryapi/models.go:227` collapses both cases into an empty map, by its own documented contract ("Returns an empty map if the query fails or returns no data"):

```go
resp, err := qc.GetAllModelCapacities(ctx, &inferencetypes.QueryGetAllModelCapacitiesRequest{})
if err != nil || len(resp.ModelCapacities) == 0 {
	return map[string]int{}
}
```

`decentralized-api/internal/server/public/get_pricing_handler.go:92` builds its per-model metrics map by ranging over the same empty slice, so `Capacity` and `Utilization` are absent for every model. That handler serves the live route registered at `decentralized-api/internal/server/public/server.go:112`:

```go
g.GET("governance/pricing", s.getGovernancePricing)
```

So the visible effect is that governance pricing reports no per-model capacity, and nothing distinguishes that from genuinely having none.

## Suggested fix

The immediate one is the key at the call site:

```go
mainEpochData, found := k.GetEpochGroupData(goCtx, currentEpoch.Index, "")
```

The durable one is a naming change nearby, because this call site is a reasonable misreading rather than a typo. `getEpochGroupWeightData(ctx, pocStartHeight uint64, ...)` in the reward path declares a parameter called `pocStartHeight` that its callers correctly fill with `msg.EpochIndex`. Anyone reading that signature would conclude the map is keyed by height. Renaming that parameter to `epochIndex` removes the trap; grepping for other callers passing a height is worth doing in the same pass.

Separately, the two consumers above would be more honest distinguishing a failed query from an empty result, so a future regression of this kind surfaces instead of rendering as zero.

## What I verified, and what I did not

Verified by reading `main` at `379bebced6`: the write key, the read key, the call site, the two consumers, and the route registration. I also searched open issues and pull requests for `GetAllModelCapacities` and `all-model-capacities` and found nothing covering this.

Not verified today: the live response. `node1.gonka.ai:8000` and `node2.gonka.ai:8000` both returned nginx `503` on `/chain-api/productscience/inference/inference/all_model_capacities` across repeated attempts while I was writing this, so I could not re-confirm the empty result against mainnet in this pass. An earlier observation of mine on 2026-08-28 recorded `{}` from that endpoint, but I am flagging it as dated rather than presenting it as a current measurement. The code path above does not depend on that confirmation.

redstartechno:

`GetAllModelCapacities` (`x/inference/keeper/query_dynamic_pricing.go`) looks up the root epoch group data with the effective epoch's `PocStartBlockHeight`, but `EpochGroupDataMap` is keyed by `EpochIndex`. The lookup never hits, so the query always returns `200` with an empty `model_capacities` list.

Fixes #1725

- `SetEpochGroupData` writes the map with `collections.Join(epochGroupData.EpochIndex, epochGroupData.ModelId)`; every other keeper caller of `GetEpochGroupData` passes an epoch index (`collateral.go`, `power.go`, `accountsettle.go`, `epoch_data_transient_cache.go`, `dynamic_pricing.go`). This query is the only one passing a block height.
- `Epoch.Index` and `Epoch.PocStartBlockHeight` differ by four orders of magnitude on mainnet (the issue reports epoch 386 vs height 5952281), so `found` is always false and the handler takes its non-error early return.
- The regression test below fails on the current code with `actual: []types.ModelCapacity(nil)`.

Pass `currentEpoch.Index` instead of `uint64(currentEpoch.PocStartBlockHeight)`, and include the index in the not-found log line so the next mismatch is diagnosable.

Query-only change; no state transitions or message handlers touched, so no consensus impact. The behavioral change is that the endpoint now returns the cached capacities it was always meant to return. Consumers already handle the populated case (`common/queryapi/models.go` builds a map from the response), and they already tolerated the empty case, so there is no new failure mode. Mitigation: two keeper tests pin the lookup key and the per-model skip behavior.

The new tests seed an effective epoch whose index and PoC start height differ, store the root epoch group under the index, cache capacities, and call the query. They fail on the old code and pass with the fix (output below). The full `x/inference/keeper` package passes and `go vet` is clean.

- `inference-chain/x/inference/keeper/query_dynamic_pricing.go`
- `inference-chain/x/inference/keeper/query_dynamic_pricing_test.go` (new)

Downstream readers of the query, unchanged but now receiving data: `common/queryapi/models.go` (`GetAllModelCapacities` wrapper) and its callers in `decentralized-api`.

## Testing & evidence

Run in WSL2 (Ubuntu 22.04, go1.24.2 per the module's `go` directive).

### Before

New tests against the unmodified `query_dynamic_pricing.go`:

```
=== RUN   TestGetAllModelCapacities_LooksUpEpochGroupByIndex
ERROR Failed to get epoch group data for capacity query subsystem=Pricing
    query_dynamic_pricing_test.go:37:
        	Error:      	Not equal:
        	            	expected: []types.ModelCapacity{...{ModelId:"model-a", Capacity:0x3e8}, {ModelId:"model-b", Capacity:0xfa}}
        	            	actual  : []types.ModelCapacity(nil)
--- FAIL: TestGetAllModelCapacities_LooksUpEpochGroupByIndex (0.00s)
=== RUN   TestGetAllModelCapacities_SkipsModelsWithoutCachedCapacity
ERROR Failed to get epoch group data for capacity query subsystem=Pricing
    query_dynamic_pricing_test.go:62:
        	Error:      	Not equal:
        	            	expected: []types.ModelCapacity{types.ModelCapacity{ModelId:"cached", Capacity:0x2a}}
        	            	actual  : []types.ModelCapacity(nil)
--- FAIL: TestGetAllModelCapacities_SkipsModelsWithoutCachedCapacity (0.00s)
FAIL
FAIL	github.com/productscience/inference/x/inference/keeper	0.095s
```

### After

```
=== RUN   TestGetAllModelCapacities_LooksUpEpochGroupByIndex
INFO Retrieved all model capacities totalModels=2 subsystem=Pricing
--- PASS: TestGetAllModelCapacities_LooksUpEpochGroupByIndex (0.00s)
=== RUN   TestGetAllModelCapacities_SkipsModelsWithoutCachedCapacity
WARN Failed to get cached capacity for model modelId=uncached error="capacity not found for model: uncached" subsystem=Pricing
INFO Retrieved all model capacities totalModels=1 subsystem=Pricing
--- PASS: TestGetAllModelCapacities_SkipsModelsWithoutCachedCapacity (0.00s)
PASS
ok  	github.com/productscience/inference/x/inference/keeper	0.154s
```

Full package:

```
$ go vet ./x/inference/keeper/ && go test ./x/inference/keeper/ -count=1
ok  	github.com/productscience/inference/x/inference/keeper	15.725s
```

### DIFF
--- a/inference-chain/x/inference/keeper/query_dynamic_pricing.go
+++ b/inference-chain/x/inference/keeper/query_dynamic_pricing.go
@@ -113,9 +113,10 @@ func (k Keeper) GetAllModelCapacities(goCtx context.Context, req *types.QueryGet
 		}, nil
 	}
 
-	mainEpochData, found := k.GetEpochGroupData(goCtx, uint64(currentEpoch.PocStartBlockHeight), "")
+	// EpochGroupDataMap is keyed by epoch index, not by PoC start height.
+	mainEpochData, found := k.GetEpochGroupData(goCtx, currentEpoch.Index, "")
 	if !found {
-		k.LogError("Failed to get epoch group data for capacity query", types.Pricing)
+		k.LogError("Failed to get epoch group data for capacity query", types.Pricing, "epochIndex", currentEpoch.Index)
 		return &types.QueryGetAllModelCapacitiesResponse{
 			ModelCapacities: modelCapacities,
 		}, nil
--- a/inference-chain/x/inference/keeper/query_dynamic_pricing_test.go
+++ b/inference-chain/x/inference/keeper/query_dynamic_pricing_test.go
@@ -1,0 +1,63 @@
+package keeper_test
+
+import (
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+
+	"github.com/productscience/inference/x/inference/types"
+)
+
+// The epoch group data map is keyed by epoch index. The effective epoch's
+// PoC start height is a block height on a different scale, so a lookup by
+// height only succeeds when the two happen to coincide.
+func TestGetAllModelCapacities_LooksUpEpochGroupByIndex(t *testing.T) {
+	k, ctx := setupTestKeeperWithDynamicPricing(t)
+	goCtx := sdk.WrapSDKContext(ctx)
+
+	effectiveEpoch := types.Epoch{
+		Index:               386,
+		PocStartBlockHeight: 5952281,
+	}
+	require.NoError(t, k.SetEpoch(ctx, &effectiveEpoch))
+	require.NoError(t, k.SetEffectiveEpochIndex(ctx, effectiveEpoch.Index))
+	k.SetEpochGroupData(ctx, types.EpochGroupData{
+		EpochIndex:          effectiveEpoch.Index,
+		ModelId:             "",
+		PocStartBlockHeight: uint64(effectiveEpoch.PocStartBlockHeight),
+		SubGroupModels:      []string{"model-a", "model-b"},
+	})
+	require.NoError(t, k.CacheModelCapacity(goCtx, "model-a", 1000))
+	require.NoError(t, k.CacheModelCapacity(goCtx, "model-b", 250))
+
+	resp, err := k.GetAllModelCapacities(goCtx, &types.QueryGetAllModelCapacitiesRequest{})
+	require.NoError(t, err)
+	require.Equal(t, []types.ModelCapacity{
+		{ModelId: "model-a", Capacity: 1000},
+		{ModelId: "model-b", Capacity: 250},
+	}, resp.ModelCapacities)
+}
+
+func TestGetAllModelCapacities_SkipsModelsWithoutCachedCapacity(t *testing.T) {
+	k, ctx := setupTestKeeperWithDynamicPricing(t)
+	goCtx := sdk.WrapSDKContext(ctx)
+
+	effectiveEpoch := types.Epoch{
+		Index:               7,
+		PocStartBlockHeight: 900,
+	}
+	require.NoError(t, k.SetEpoch(ctx, &effectiveEpoch))
+	require.NoError(t, k.SetEffectiveEpochIndex(ctx, effectiveEpoch.Index))
+	k.SetEpochGroupData(ctx, types.EpochGroupData{
+		EpochIndex:          effectiveEpoch.Index,
+		ModelId:             "",
+		PocStartBlockHeight: uint64(effectiveEpoch.PocStartBlockHeight),
+		SubGroupModels:      []string{"cached", "uncached"},
+	})
+	require.NoError(t, k.CacheModelCapacity(goCtx, "cached", 42))
+
+	resp, err := k.GetAllModelCapacities(goCtx, &types.QueryGetAllModelCapacitiesRequest{})
+	require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


redstartechno:
Adding the user-visible consequence, which the original description undersold — I traced the consumer chain today while reviewing a related PR:

`GetAllModelCapacities` is not only an internal query. `decentralized-api` calls it to build the public pricing response: `internal/server/public/get_pricing_handler.go:41-70` populates `Capacity` and `Utilization` per model from `getModelMetrics`, and serves them alongside the per-token price on the public pricing endpoint.

With the lookup keyed by `currentEpoch.PocStartBlockHeight` instead of `currentEpoch.Index`, `GetEpochGroupData` misses for every epoch where the two don't coincide — i.e. all of them in practice, since one is an epoch counter and the other a block height (the test uses index 386 / height 5952281 from a real epoch). So the handler logs `Failed to get epoch group data for capacity query` and the public endpoint reports no capacity and no utilization at all.

That also makes this a prerequisite for anyone trying to reason about dynamic pricing from outside the chain: utilization is the input to the whole stability-zone mechanism, and right now the published value is missing rather than wrong, which is easy to misread as "the feature is off".

No changes to the patch — still the one-line key fix plus the regression test, still 9/9 green. Rebased base is unchanged (`inference-chain/x/inference/keeper/query_dynamic_pricing.go` has had no upstream commits since this branched). Available to rebase or split if that helps it move.

Issue #1733, opened 2026-09-08
Extend e2e for HA devshard config with multiple routers and multiple versiond


a-kuprin:
# Proposal: Unjoined-network citest for multiple versiond-routers

**Status:** Draft  
**Related:** [PR #1610](https://github.com/gonka-ai/gonka/pull/1610) (`VERSIOND_POOL_ENDPOINTS_FILE`), existing `TestRouterStickiness` / `TestVersiondStickySessionFailover`  
**Scope:** One new testenv citest. Not a second `citest-stack`, not multi-host public ingress.

## Gap

Join HA after #1610 can run **several** `versiond-router` processes against versionds that **do not share a Docker network** with the routers (other machines, listed by address).

CI today covers the pieces separately:

| Coverage | Routers | Versionds | Network |
|---|---|---|---|
| testenv citest | 1 | 2 (real, Postgres HA) | one shared compose network, DNS `versiond-pool` |
| `test-fleet` / `test-routing` | 2–3 | fake / one stub | shared test networks |
| `endpoints-render_test.sh` | render only | n/a | n/a |

Nothing boots **two real routers + two real versionds with no shared Docker network** and checks that the same escrow still hashes to the same versiond.

That is the `#1610` other-host path: `VERSIOND_POOL_ENDPOINTS_FILE` replaces DNS.

## Topology

Two Compose projects on one CI Docker daemon. **No common network.**

```text
Compose A  (network: testenv-versiond)          Compose B  (network: testenv-routers)
  mock-chain, mock-dapi, mock-openai              versiond-router-0
  postgres, versiond-0, versiond-1               versiond-router-1
  publish each versiond :8080 → host            no versiond-pool DNS
                                                 VERSIOND_POOL_ENDPOINTS_FILE =
                                                   [{id, host, port}, …]
  test client (harness) ──HTTP──▶ both routers' published :8080
```

Reachability A→B is only **host-published ports**. Inside router containers, `localhost` is the router, so the endpoint list must use one canonical host both slots can dial, e.g. `host.docker.internal` (`extra_hosts: host-gateway`) plus the port from `docker compose port`. **The same `host:port` string must appear in every slot’s list** — the ring is `hash-key addr`. Different literals for the same versiond (`127.0.0.1` vs `172.17.0.1`) are a failed test, not a flake to paper over.

[...]

PR #1735, opened 2026-09-08
test(devshard): cover unjoined versiond router fleet

snevolin:
Existing real-stack tests exercise one router on a shared Docker network. They do not verify that independent routers agree on session routing when versiond membership comes from `VERSIOND_POOL_ENDPOINTS_FILE`.

Add `citest-unjoined-router-fleet` as a separate, automatically discovered CI suite. It starts two real HAProxy routers and two PostgreSQL-backed versionds in disjoint Compose networks, connected through host-published ports. The test checks network separation, failed `versiond-pool` DNS lookup, identical endpoint files, readiness, and exact `X-Upstream-Addr` agreement across repeated requests to both members. After stopping versiond-0, both routers must converge on versiond-1 within one shared 45-second window, while its existing session mapping stays stable.

The scenario uses synthetic session paths for hashing. Their upstream 404 is expected because devshardd registers `/healthz`, not `/sessions/:id/healthz`; child health and readiness must independently return HTTP 200. The scenario and run command are documented in `testenv/docs/scenarios.md`.

Validation:

- `TestUnjoinedRouterFleet` with freshly built versiond, devshardd, router, and mock images in an isolated Docker-in-Docker environment (Alpine/musl): **PASS, 27.09s**.
- `go test ./testenv/... -count=1`: **PASS**.
- Confirmed `list-citest-targets` discovers the new suite independently of `citest-stack`.

Refs #1733.

### DIFF
--- a/devshard/testenv/Makefile
+++ b/devshard/testenv/Makefile
@@ -71,6 +71,7 @@ dev-logs: ## Follow dev overlay logs.
 
 .PHONY: citest-images citest-stack citest-stack-build list-citest-targets
 .PHONY: citest-validation-lease-race citest-escrow-longpoll
+.PHONY: citest-unjoined-router-fleet
 .PHONY: citest-versiond-rolling-update citest-versiond-host-evacuation citest-adversarial citest-payload-withholding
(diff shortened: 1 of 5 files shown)


a-kuprin:
Is it still in draft?

PR #1737, opened 2026-09-08
Expose host ping probe race on escrow release

aikuznetsov:
## Summary

Adds a deterministic regression test exposing a remaining host-ping metrics lifecycle race after #1736.

No production code is changed.

## Scenario

1. The host-ping scheduler starts a real HTTP probe.
2. The test server blocks the response while the probe is in flight.
3. `ReleaseEscrow` removes the target and clears its Prometheus series.
4. The HTTP response is released.
5. The completed probe reaches `hostPingSink.Observe` and recreates metrics for the released host.

The blocking handler only makes the ordering deterministic; in production, the same sequence can occur when an escrow is released while a network request is still in flight.

## Expected behavior

A probe that completes after its escrow has been released must not recreate any host-ping series.

## Current behavior

The test fails because a deleted series is restored:

```text
devshard_gateway_host_clock_divergence_seconds still has
host="http://127.0.0.1:<port>" after cleanup
```

## Reproduction

```bash
go test ./cmd/devshardctl/ \
  -run '^TestHostPingInFlightProbeAfterReleaseDoesNotRestoreSeries$' \
  -count=1
```

### DIFF
--- a/devshard/cmd/devshardctl/hostping_test.go
+++ b/devshard/cmd/devshardctl/hostping_test.go
@@ -1,7 +1,11 @@
 package main
 
 import (
+	"context"
 	"errors"
+	"net/http"
+	"net/http/httptest"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


a-kuprin:
@aikuznetsov 1 test failed here
Could you fix or it will not go to v6 ?

PR #1738, opened 2026-09-09
feat(dapi): serve block headers and proofs over NodeManager gRPC

a-kuprin:
## Summary
- Add unary `GetBlockHeader` and `ProveBlockPath` on NodeManager gRPC as twins of HTTP `GET /block/:height` and `GET /block/:height/prove`.
- Both RPCs read the same `BlockOracle` as the HTTP mount (`WithBlockOracle`). No second producer, and no gRPC stream — live tip stays Comet NewBlock.
- Proto is additive only. Omit the oracle and both RPCs answer `FailedPrecondition`. `nmrpc` holds handlers + wire mapping; `nmclient` is a gRPC `BlockOracle` for hosts (no production caller yet). mock-dapi serves the same RPCs for testenv.

It's followup for https://github.com/gonka-ai/gonka/pull/1622 that adds grpc for same functionality endpoints as /block, at node manager server

### DIFF
--- a/common/chainoracle/blocks/nmclient/lookup.go
+++ b/common/chainoracle/blocks/nmclient/lookup.go
@@ -1,0 +1,111 @@
+// Package nmclient is the unary NodeManager consumer of GetBlockHeader
+// and ProveBlockPath. Live tip is Comet NewBlock, not a gRPC stream.
+package nmclient
+
+import (
+	"context"
+	"errors"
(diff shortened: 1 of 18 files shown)


gmorgachev:
@akup i think "common/chainoracle/blocks/nmrpc" is lost

gmorgachev (approved):

Issue #1691, opened 2026-08-31
GLM 5.3 Flash: vLLM 0.28 image + PoC fix


tcharchian:
Pitstop’s main track is MLNode images, experiments, and evaluations for bringing new models onto the network.

The current piece of work is GLM 5.3 Flash: a vLLM 0.28 image plus a PoC fix. PoC currently exceeds the allowed memory budget; that needs to be resolved before handover.

This issue also sets a public bar for how MLNode image work is specified and accepted, so anyone taking an image can see what was checked and what the limitations are.

Confirm the live on-chain set from poc_params.models if the list below has drifted.

---

### Process

- [ ] Experiments completed and written up (hardware, models, what passed, what did not)
- [ ] MLNode starts and reaches a working state (does not crash)
- [ ] Write-up reviewed (review does not have to block image built; once the two items above are done, the image can be built and backward compatibility can be checked in parallel)
- [ ] Image rechecked against the model list below before handover
- [ ] If the image is intentionally for a specific model: notes published before handover, listing which models are supported and which are not

---

### Per-model checks

For every model below: both PoC and inference must pass. Installing this image must not break models already on chain unless that is stated in the notes before handover.

`MiniMaxAI/MiniMax-M2.7` — base model, active

- [ ] PoC starts, finishes, and stays within the allowed memory budget
- [ ] Inference works
- [ ] Supported on this image, or unsupported and called out in notes before handover

`moonshotai/Kimi-K2.6` — active

- [ ] PoC starts, finishes, and stays within the allowed memory budget
- [ ] Inference works
- [ ] Supported on this image, or unsupported and called out in notes before handover

`deepseek-ai/DeepSeek-V4-Flash-0731` — active

- [ ] PoC starts, finishes, and stays within the allowed memory budget
- [ ] Inference works
- [ ] Supported on this image, or unsupported and called out in notes before handover

`zai-org/GLM-5.2-FP8` — registered on chain, not bootstrapped (penalty_start_epoch = 500)

- [ ] PoC starts, finishes, and stays within the allowed memory budget
- [ ] Inference works
- [ ] Supported on this image, or unsupported and called out in notes before handover

GLM 5.3 Flash — target of this image (vLLM 0.28)

- [ ] vLLM 0.28 image builds and runs
- [ ] PoC starts, finishes, and no longer exceeds the allowed memory budget
- [ ] Inference works
- [ ] Experiment notes attached (hardware, memory, PoC, inference)

---

### Handover

[...]

PR #1750, opened 2026-09-10
fix(mlnode): drop the stock ubuntu user on 24.04 bases so entrypoint.sh can create appuser

baychak:
`ghcr.io/gonka-ai/mlnode:3.0.17` does not start through its own `entrypoint.sh`: the vLLM 0.28 base is Ubuntu 24.04, which ships `ubuntu` at UID 1000, and the script creates `appuser` at `HOST_UID` (default 1000) under `set -e`. The container exits with `useradd: UID 1000 is not unique` before `exec "$@"`. 3.0.16 sat on 22.04.

Reproduced by @clanster on 4×B200 and 4×H200. One line after `FROM`; no-op on 22.04 bases. [#1691](https://github.com/gonka-ai/gonka/issues/1691).


baychak:
Superseded by [#1751](https://github.com/gonka-ai/gonka/pull/1751): the runtime user setup is dead code, so it goes rather than the base image's account.

PR #1753, opened 2026-09-10
fix(devshard): Fixes metrics growth from retired escrow IDs

GLiberman:
Fixes Prometheus cardinality growth from retired escrow IDs left in gateway HistogramVec/CounterVec children.

- Keep emitting `devshard_host_*` and `devshard_gateway_participant_*` (dashboard still uses only the participant families).
- Forget escrow-scoped host, slot-decision, and picker series when a runtime is deactivated, finalized, retired, cleaned, or replaced without settlement.
- Persist export switches `metrics.devshard_host` and `metrics.devshard_participant` (both default on; env on first init).
- Serialize disable/Reset against Observe so a disabled group cannot come back from an in-flight sample.
- Restore `ghost_no_send` on slot-decision metrics for dashboard skipped-slot panels.

Already-running gateways with active metrics still need a restart (or those ids to drop) to shed leftover series. After that, escrow rotation no longer piles them up.


qdanik:
Suggestion for a smaller fix that also stops the leak from coming back:

1. **Drop the `devshard_host_*` timing histograms.** They take the same four measurements from the same sample as `devshard_gateway_participant_*`. The only difference is the key: `devshard_id × host_idx` doesn't survive rotation, because `host_idx` is a slot index inside one escrow's group. Nothing in the repo reads them. Dropping them removes ~960 series per escrow (16 slots × 4 histograms × 15 series) and needs no lifecycle code. If something outside the repo depends on them, deprecate for a release; the participant group already has the same data under a stable key.
2. **Then drop the export switches** (store columns, admin fields, env vars) and `timingMu`. They have nothing left to protect, and the hot path stops paying for a global lock.
3. **Keep `ForgetEscrow` in `dropRegisteredRuntimeLocked`** for the escrow-labelled counters the dashboard does use (`slot_decisions_total`, `picker_choice_total`). Add `startup_skipped_escrow` to it, and close the paths that bring series back:
   - run `finishRaceOutcome` via `goTrackedRaceCleanup` instead of a bare `go func()`;
   - don't dispatch ghost probes from a retired runtime.

   Alternative: count these on the runtime and emit them from `gatewayMetricsCollector`, like `escrowWeightDesc`. Then there is no delete step to forget.
4. **Finalize: call `ForgetEscrow(id)` instead of `retireRuntime`.** That sheds the series without turning a repeated `POST /finalize` into a 404. Keep the retire on replacement without settlement: that retirement really was missing.
5. **Fix the race in `gateway_test.go:197`:** read `g.runtimes` under `g.mu`. `go test -race` fails on it today.

PR #1757, opened 2026-09-11
fix(devshard): count only sampled passes in invalidation SPRT

zpoken:
On-chain devshard settlement fed `assigned - missed - invalid` into the participant invalidation SPRT as *passes*, so every unsampled inference counted as evidence of honesty. An executor over-reporting `OutputTokens` on the ~90% of inferences that are never sampled could not reach INVALID, which made the slash/mining-forfeit deterrent inert.

This replaces the derived count with counters committed in the devshard state root and scored on chain.

## Devshard

- `HostStats.validated` (field 7): `+1` per passing sampled replay while `Finished` (one per validator address), `-n` on challenge (every other `ValidatedBy` bit on a `Finished` record is a credited pass), `+1` on `Challenged -> Validated`. Invalidated or unresolved challenges contribute no passes.
- `HostStats.finished` (field 8): `MsgFinishInference` applied per slot. 
- Self-validation guard is now per address, not per slot: a multi-slot executor could previously validate its own inference from another owned slot.
- Plumbed through hash, snapshot codec, both hand-rolled tx encoders (fields emitted only when non-zero), JSON views, accounting (`protocol_validated`, `protocol_finished`). Accounting merges host stats by nonce so a challenge-driven decrease is reflected in the preview.

## Chain

- `DevshardSettlementHostStats` / `DevshardHostStatsProto` fields 7–8, `DevshardHostEpochStats` 9–10.
- Settlement verifies `invalid <= finished <= assigned - missed` and `validated <= finished * slots`, then feeds the SPRT with `min(validated, 2 * finished * validation_rate + 2)`. The state machine cannot verify sampling eligibility (validator seeds are private until reveal), so this bounds what a colluding roster address can manufacture. The cap is on `finished`, not the nonce-derived bound, so control diffs cannot widen it.


qdanik:
@zpoken thanks, the direction is right, and the cap `2 * finished * rate / 10000 + 2` matches the sampling model. A few important things before merge:

**Blockers**

**Error-miss breaks settlement.** `applyErrorMiss` moves a `Finished` record to `TimedOut` and does `Missed++`, but doesn't undo `Finished++` or the credited passes. The chain checks `finished <= assigned - missed`, so on a busy escrow a single error-miss is enough to reject an honest settlement. If the escrow can't settle within its window, `distributeUnsettledEscrow` splits the amount across slots: the creator loses the refund and the invalidations never reach the SPRT. Suggested fix: `Finished--` and `Validated -= ValidatedBy.Count()` (clamped) in `applyErrorMiss`, plus a test.

**Major**

1. **The fix is off on mainnet by default.** It only applies when `reports_validated` is true. The flag defaults to false, mainnet already has approved versions, and nothing in the PR sets it. It needs an upgrade handler or at least a clear note in the description.
2. **An empty `approved_versions` list means opposite things.** The approval check treats it as "any version allowed", while scoring treats it as the strict policy. On dev/test chains an old binary's settlement with `Invalid > 0` gets rejected. Defaulting to legacy there would be safer.
3. **The scoring policy comes from governance params at settle time.** Everything else (`validation_rate`, timeouts) is frozen on the escrow, so flipping the flag re-scores escrows that are already live.
4. **`applyValidationVote` has no self-validation check.** The executor's address can still earn a pass through the vote path.
5. **`protocol_finished` in accounting is always 0.** `maxHostStats` doesn't merge `Finished`, and the recorder never refreshes it on finish.

**Tests**

These mutations still pass the suite:
- a challenge wiping all of `Validated` instead of subtracting;
- `Validated += weight` for multi-slot validators (every validator in the test group has weight 1);
- the snapshot codec dropping the new fields (the fixture doesn't set them).

Also, no settle test sets `ApprovedVersions`, so the legacy path and the cap are never exercised end to end.

zpoken:
@qdanik 

Things done after review:

- **Error-miss unwinds the finish.** `applyErrorMiss` now decrements `Finished` and withdraws the inference's credited passes, so `finished <= assigned - missed` holds and an error-miss no longer blocks an honest settlement.
- **Self-vote guard.** `applyValidationVote` rejects votes from any slot owned by the executor's address, same as phase-1 validations. Test fixtures that resolved challenges only through the executor's own vote were moved to 3-host groups.
- **Scoring policy is pinned per version.** `DevshardEscrowParams.version_policies` is an append-only record written the first time a version is approved (`MsgUpdateParams`, the v0.2.16 upgrade handler, `InitGenesis`). Settlement reads it instead of `approved_versions`, so removing and re-adding a version cannot change how live escrows are scored. Pre-existing production approvals are recorded as legacy at upgrade. Empty allowlist = legacy, matching the approval check. The upgrade handler keeps the recorded policy and logs instead of halting the chain if the proposal info re-declares a flag.
- **v0.2.16 upgrade handler** applies `approved_versions` from the proposal's `info` JSON; the new version's entry carries `"reports_validated": true`.
- **Accounting.** `Finished` merges from the newest nonce; the recorder refreshes the executor slot on finish/error-miss diffs without consuming the verdict dedup entry, so a challenge in the same diff is still recorded.
- **Tests.** Challenge subtracts only that inference's passes; multi-slot validator counts one pass, not weight; snapshot codec round-trips all `HostStats` fields; end-to-end `SettleDevshardEscrow` for the legacy path, the cap, and removal/re-approval; upgrade handler append, flip, and pre-existing approvals.

a-kuprin:
@zpoken 
As inference-chain part should be separated from devshard release and should land to inference-chain before new devshard version is released I have made a separate PR: https://github.com/gonka-ai/gonka/pull/1794

The important thing there that it sets default sprt policy to derived (current proposed policy) if version_policy isn't set in governance proposal.
Also it fills all existing versions with `DERIVED = 1; // legacy: assigned - missed - invalid`

PR #1768, opened 2026-09-14
fix(devshard): carry escrow timeouts through host bridge

### LINKED ISSUE #1762: devshard: host bridge drops escrow refusal_timeout/execution_timeout, so host and gateway can bind different SessionConfig
On `devshard-0.2.15-v5` (PR #1584), the gateway and the host build `SessionConfig` with the same function but fetch the escrow through different bridges. The gateway's bridge copies `refusal_timeout` / `execution_timeout` off the escrow row; the host's bridge does not, so the host silently falls back to the compiled defaults (60 / 1920). Both values feed consensus — auto-seal and timeout votes — so a governance value other than the default desynchronises the two sides.

All line references are at PR head `9d1f01472459f17b1516933951d2617027b0ad08`.

## Where the two paths diverge

The gateway uses `bridge.GRPCBridge` ([main.go:527](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/cmd/devshardctl/main.go#L527), [gateway.go:410](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/cmd/devshardctl/gateway.go#L410)), which carries both fields:

```go
// devshard/bridge/grpc.go:99-104
		ValidationRate:            e.ValidationRate,
		VoteThresholdFactor:       e.VoteThresholdFactor,
		RefusalTimeout:            e.RefusalTimeout,
		ExecutionTimeout:          e.ExecutionTimeout,
		EpochID:                   e.EpochIndex,
		Settled:                   e.Settled,
```

The host (`devshardd`) uses `CachingEscrowBridge` over `ChainBridge` ([app.go:276](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/cmd/devshardd/app.go#L276)), and that bridge's `EscrowInfo` literal ends without them:

```go
// devshard/cmd/devshardd/bridge/chain.go:130-137
		InferenceSealGraceNonces:  e.InferenceSealGraceNonces,
		InferenceSealGraceSeconds: e.InferenceSealGraceSeconds,
		AutoSealEveryNNonces:      e.AutoSealEveryNNonces,
		ValidationRate:            e.ValidationRate,
		VoteThresholdFactor:       e.VoteThresholdFactor,
		EpochID:                   e.EpochIndex,
		Settled:                   e.Settled,
	}, nil
```

The warm cache cannot restore them either — [`EscrowCacheInfo`](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/storage/interface.go#L130) has no such fields, so [`EscrowInfoFromCache`](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/cmd/devshardd/bridge/caching.go#L129) maps a cached row back without them as well.

Both sides then go through the same mapper, [`SessionConfigAtBind`](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/bridge/config.go#L19) → `SessionConfigFromEscrow`, where zero means "use the compiled default":

```go
// devshard/types/config.go:50-53
func DefaultSessionConfig(groupSize int) SessionConfig {
	return NormalizeSessionConfig(SessionConfig{
		RefusalTimeout:    60,
		ExecutionTimeout:  32 * 60,
```

## Why both values are consensus-relevant

Auto-seal reads `ExecutionTimeout` from the session config ([seal.go:452](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/state/seal.go#L452)) and folds it into the Finished clock gate:

```go
// devshard/state/seal.go:43-51
func FinishedClockRequiredSeconds(graceSeconds, executionTimeout int64) int64 {
	...
	return graceSeconds + executionTimeout
}
```

`autoSealLocked`'s own comment states the consequence: "Mixed binaries that disagree on this sum diverge on SealedAcc / post_state_root."

Host-side timeout votes use the same config — [timeout.go:109](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/host/timeout.go#L109) rejects a REFUSED vote while `now - StartedAt < RefusalTimeout`, and [timeout.go:183](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/host/timeout.go#L183) rejects an EXECUTION vote while `now - ConfirmedAt < ExecutionTimeout` — while the gateway computes its own deadline from its own config ([session.go:2824](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/user/session.go#L2824)).

## Reproduction

Same mapper, same group size, one field differing: the host's escrow (fields dropped by `ChainBridge`) against the gateway's (fields carried by `GRPCBridge`) for a governance `execution_timeout = 1200`, with `inference_seal_grace_seconds = 3600`:

```go
hostCfg := SessionConfigAtBind(16, &hostEscrow)        // RefusalTimeout/ExecutionTimeout zero
gatewayCfg := SessionConfigAtBind(16, &gatewayEscrow)  // RefusalTimeout 60, ExecutionTimeout 1200
state.FinishedClockRequiredSeconds(int64(cfg.InferenceSealGraceSeconds), cfg.ExecutionTimeout)
```

```
SUBJECT host    (ChainBridge drops the fields): refusal=60 execution=1920 sealThreshold=5520
CONTROL gateway (GRPCBridge carries them):      refusal=60 execution=1200 sealThreshold=4800
```

Two consequences follow from that gap:

- **Sealing.** For an inference whose `stateClock - ConfirmedAt` lands between 4800 and 5520, the gateway folds it into `SealedAcc` and the hosts do not, so `post_state_root` disagrees. Hosts answer `post_state_root does not match computed state root` ([types/errors.go:35](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/types/errors.go#L35)) and the gateway marks them `escrow_state_root_diverged` ([redundancy.go:3787](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/cmd/devshardctl/redundancy.go#L3787)). A raised timeout produces the mirror image.
- **Votes.** The gateway starts voting at its own (smaller) deadline while hosts still reject the vote as premature, so quorum is never reached and the record stays `Started`. At settlement a still-live record pays the executor its full reservation ([machine.go:1668](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/state/machine.go#L1668)):

```go
	case types.StatusStarted, types.StatusPending:
		rec.ActualCost = rec.ReservedCost
		rec.Status = types.StatusFinished
		sm.state.HostStats[rec.ExecutorSlot].Cost += rec.ReservedCost
```

## Why it is quiet today

Mainnet governance currently sets exactly the compiled defaults, so the two paths agree by coincidence. Read from the public REST endpoint (`/chain-api/productscience/inference/inference/params` on `node1.gonka.ai:8000`), 2026-09-13:

```
devshard_escrow_params.refusal_timeout = 60
devshard_escrow_params.execution_timeout = 1920
devshard_escrow_params.default_inference_seal_grace_seconds = 3600
```

Nothing detects the mismatch on its own: `SessionConfig` is not part of the state root ([hash.go:38](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/devshard/state/hash.go#L38) folds balance, host stats, inferences, phase, warm keys, fees and version), so a config divergence only ever surfaces later as a root mismatch or a stuck vote.

Stands do change these values through governance — `RuntimeConfigTests.kt` bumps [refusal_timeout](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/testermint/src/test/kotlin/RuntimeConfigTests.kt#L416) and [execution_timeout](https://github.com/gonka-ai/gonka/blob/9d1f01472459f17b1516933951d2617027b0ad08/testermint/src/test/kotlin/RuntimeConfigTests.kt#L435) — so a non-default value is reachable in test environments before it is ever reachable in production.

## Origin

Commit `88ebd4456` (#1564) added `refusal_timeout = 17` / `execution_timeout = 18` to `inference/inference/devshard_escrow.proto` together with the gRPC-bridge mapping and `SessionConfigAtBind`, and touched neither `devshard/cmd/devshardd/bridge/chain.go` nor `devshard/storage/interface.go`. On the branch's merge base `379bebced638aeb5e6077bfd51c986f898443832` none of the three files mentions either field.

## Suggested fix

1. Copy both fields in `ChainBridge.GetEscrow` (`devshard/cmd/devshardd/bridge/chain.go:120`), mirroring `GRPCBridge`.
2. Add both to `storage.EscrowCacheInfo` and to both cache mappers (`EscrowCacheFromInfo` / `EscrowInfoFromCache`) — without this half, a bind served from the warm cache still loses them.
3. Consider a test that feeds one bridge's real `EscrowInfo` into the other's bind path and asserts field-for-field equality, since each side's tests build their own fixtures today.
4. Two related points worth deciding separately:
   - `devshard/user/httpsession.go:156` lets gateway-side configuration override both timeouts after the bind mapping, so the two sides can still be desynchronised by a flag once the bridge is fixed.
   - `devshard/docs/params-dataflow.md` is now stale on this: the lane A table (line 27) does not list the two fields, line 45 still files them under lane C as devshardctl-only, and line 52 says `SessionConfig` carries the defaults.

## Not checked

- No end-to-end run on a stand: the sealing divergence above is code reading plus arithmetic, not an observed root mismatch.
- Whether any escrow row already on mainnet carries non-zero values in fields 17/18 — the chain release carrying this proto is not what mainnet runs today.

redstartechno:

The host binary (`devshardd`) and the gateway (`devshardctl`) build `SessionConfig` from the same on-chain escrow row through the same mapper, but fetch it through different bridges. The gateway's `GRPCBridge` copies `refusal_timeout` / `execution_timeout`; the host's `ChainBridge` did not, and the host's warm cache row (`storage.EscrowCacheInfo`) had no slots for them either. The host therefore always bound the compiled defaults (60 / 1920) while the gateway bound the governance values.

Fixes #1762

- `devshard/bridge/grpc.go:101-102` maps both fields; `devshard/cmd/devshardd/bridge/chain.go` (before this PR) ends its `EscrowInfo` literal at `VoteThresholdFactor` / `EpochID`, and `storage.EscrowCacheInfo` has no such fields, so `EscrowInfoFromCache` cannot restore them either.
- Both values are consensus-relevant: `state/seal.go` folds `ExecutionTimeout` into `FinishedClockRequiredSeconds` (auto-seal clock gate, so `SealedAcc` / `post_state_root` diverge), and `host/timeout.go:109,183` reject REFUSED / EXECUTION votes as premature against the host's own copy while the gateway starts voting at its own deadline (quorum never reached, record stays `Started` and pays the full reservation at settlement).
- The issue reporter and @aikuznetsov independently reproduced the divergence on the branch head (`execution_timeout=1200` gives seal thresholds 5520 on the host vs 4800 on the gateway). It is masked on mainnet only because governance currently sets exactly the compiled defaults.
- The two new tests below fail on the previous code with both fields read back as `0`.

Copy both fields in `ChainBridge.GetEscrow`, mirroring `GRPCBridge`; add them to `storage.EscrowCacheInfo` and to both cache mappers (`EscrowCacheFromInfo` / `EscrowInfoFromCache`) so a bind served from the warm cache keeps them too. Correct `devshard/docs/params-dataflow.md`, which filed these two fields under the live lane C only and said `SessionConfig` carries the defaults.

- **Behavior change on the host.** After this PR a host binds the escrow's frozen timeouts instead of the compiled defaults. That is the intended (gateway-matching) behavior; on mainnet today the values are identical, so there is no observable change until governance moves them. Hosts on mixed binaries would diverge only under a non-default governance value, which is exactly the situation the PR makes consistent.
- **Cache compatibility.** The cache row is stored as a JSON blob (`escrow_json` column via `storage/escrow_cache.go`), so no DDL or migration. Rows written by older binaries lack the keys and unmarshal as `0`, which `SessionConfigFromEscrow` already maps to the compiled default; newer rows read by older binaries carry two ignored keys.
- **Out of scope, left as is.** `devshard/user/httpsession.go:156` lets gateway-side configuration override both timeouts after the bind mapping, so a flag can still desynchronise the two sides after this fix; that is a separate decision and not touched here.

- `TestChainBridge_GetEscrow_MatchesGatewayBridge` serves one escrow over the in-process mockchain (same scaffold as `devshard/bridge/grpc_test.go`), fetches it through `ChainBridge` and `GRPCBridge` on the same connection, and requires identical `EscrowInfo` and identical `SessionConfigAtBind` output. It fails on the previous code (`expected: 90, actual: 0`) and passes with the fix.
- `TestEscrowCache_RoundTripKeepsSessionConfigFields` round-trips a fully populated `EscrowInfo` through both cache mappers and requires field-for-field equality plus identical `SessionConfigAtBind` output. It fails on the previous code (diff shows both timeouts as `0`) and passes with the fix.
- Full test run of `devshard/cmd/devshardd/...`, `devshard/storage` and `devshard/bridge` is green; `go vet` clean.

- `devshard/cmd/devshardd/bridge/chain.go` (host bridge mapping)
- `devshard/cmd/devshardd/bridge/caching.go` (both cache mappers)
- `devshard/storage/interface.go` (`EscrowCacheInfo` gains two JSON fields; no DDL)
- `devshard/cmd/devshardd/bridge/chain_test.go`, `caching_test.go` (new tests)
- `devshard/docs/params-dataflow.md` (lane A / lane C wording)

## Testing & evidence

Run in WSL2 (Ubuntu 22.04, go1.25.9 per the module's `go` directive) from the `devshard` module.

### Before

New tests against the unmodified `chain.go`, `caching.go`, `storage/interface.go`:

```
=== RUN   TestEscrowCache_RoundTripKeepsSessionConfigFields
    caching_test.go:152:
	Error:      	Not equal:
	            	Diff:
	            	--- Expected
	            	+++ Actual
	            	@@ -20,4 +20,4 @@
	            	  VoteThresholdFactor: (uint32) 67,
	            	- RefusalTimeout: (int64) 90,
	            	- ExecutionTimeout: (int64) 1200,
	            	+ RefusalTimeout: (int64) 0,
	            	+ ExecutionTimeout: (int64) 0,
	            	  EpochID: (uint64) 9,
--- FAIL: TestEscrowCache_RoundTripKeepsSessionConfigFields (0.00s)
=== RUN   TestChainBridge_GetEscrow_MatchesGatewayBridge
    chain_test.go:65:
	Error:      	Not equal:
	            	expected: 90
	            	actual  : 0
--- FAIL: TestChainBridge_GetEscrow_MatchesGatewayBridge (0.00s)
FAIL
FAIL	devshard/cmd/devshardd/bridge	0.090s
```

### After

```
=== RUN   TestEscrowCache_RoundTripKeepsSessionConfigFields
--- PASS: TestEscrowCache_RoundTripKeepsSessionConfigFields (0.00s)
=== RUN   TestChainBridge_GetEscrow_MatchesGatewayBridge
--- PASS: TestChainBridge_GetEscrow_MatchesGatewayBridge (0.00s)
PASS
ok  	devshard/cmd/devshardd/bridge	0.087s
```

Full affected packages (Docker-bound `TestPostgres_*` / `TestHybrid_*` suites skipped locally; `TestCheckStorageDDL_CleanTree` skipped because it shells out to `scripts/check-storage-ddl.sh`, which cannot run from a CRLF checkout, unrelated to this change):

```
ok  	devshard/cmd/devshardd	5.280s
ok  	devshard/cmd/devshardd/bridge	0.215s
ok  	devshard/cmd/devshardd/events	0.033s
ok  	devshard/cmd/devshardd/inference	2.360s
ok  	devshard/cmd/devshardd/session	15.681s
ok  	devshard/cmd/devshardd/tx	0.210s
ok  	devshard/storage	60.419s
ok  	devshard/bridge	0.245s
```


redstartechno:
Note on the red check: `Test race-releases host publish scripts` fails on the base branch itself, not on this diff (which touches only `devshard/`). shellcheck stops on SC2054 at `.github/scripts/devshard-release-skip-duplicate.sh:72` (`--json databaseId,headSha,headBranch` inside an array literal, introduced in 1878e2b48). The same job is red on #1744. Quoting the field list (`--json "databaseId,headSha,headBranch"`) clears it; happy to push that as a separate one-liner if useful.

redstartechno:
Superseded by 3d1b445 on `devshard-0.2.15-v5`, which fixes the same drop and goes further: extracting the shared `bridge.EscrowInfoFromQuery` removes the gateway/host divergence at its source, so the inlined mapping in this PR would now be a step backwards. Closing.

Two things 3d1b445 does not cover, which I'll send as a small follow-up against the same base:

- `devshard/docs/params-dataflow.md` is now out of date with the code it describes: the lane A table has no `refusal_timeout` / `execution_timeout` row, the lane C table still lists them as "Per inference attempt; not in state root", and the closing line still says `SessionConfig` carries *default* timeouts.
- The warm-cache mappers (`EscrowCacheFromInfo` / `EscrowInfoFromCache` in `cmd/devshardd/bridge/caching.go`) are still two hand-written field lists, which is the mechanism that dropped the fields in the first place. `bridge.EscrowInfo` has 20 fields and `storage.EscrowCacheInfo` carries 18; `ModelID` and `Settled` are dropped today (harmless, but nothing records that it is deliberate). A field-by-field round-trip test with an explicit exclusion list would have caught #1762 on the cache path and will catch the next field added.

PR #1776, opened 2026-09-15
Ci/contributor and triage workflows test mmd ps

maria-mitina-ps:
trying test from the unknown external contrib


maria-mitina:
test failed

PR #1777, opened 2026-09-15
Ci/contributor and triage workflows test mmd ps 1

maria-mitina-ps:
test 1


maria-mitina:
test failed - first time contributor send open PR but this never reaches the workflow because approval is needed. after approval is given, the open signal is nto resent to the workflow

maria-mitina:
reopening pr does not runt he acknowledge workflow

PR #1778, opened 2026-09-15
Ci/contributor and triage workflows test mmd ps 2

maria-mitina-ps:
test again trying to use approved path


maria-mitina:
second test failed because the approval is not enough - the pr need to be merged for a new contributor

PR #1535, opened 2026-07-31
Upgrade v0.2.16

DimaOrekhovPS:
The v0.2.16 upgrade includes protocol changes and bug fixes across the chain and API node.

## Trusted Weight

Before v0.2.16, a sudden increase in claimed compute immediately increased a participant's power in governance, BLS, and PoC validation. This gave unconfirmed capacity influence over consensus and over validation of other participants.

The upgrade limits that power to compute confirmed in the previous epoch. New or returning participants start at 0, and a failed confirmation sets the next baseline to 0. New compute still earns rewards immediately. Only trust-sensitive power waits one confirmed epoch.

## Dynamic Coefficients v1

The network needs predictable throughput for each model so users can rely on its availability. The target throughput for each model should follow expected demand.

With fixed coefficients, an entire hardware class tends to switch to the same model. This makes capacity per model difficult to balance and predict.

This upgrade lets governance set a target percentage of network compute and a coefficient range for each model. The protocol dynamically adjusts the coefficient inside that range to move compute toward the target. Compute above the target is scored at the minimum coefficient, which discourages oversupply.

Governance initially sets targets from demand estimates and data sources such as OpenRouter. Later versions can aggregate host estimates and eventually use on-chain model usage. This version changes incentives only. It does not automatically switch the models deployed by a host. At upgrade, the minimum and maximum both equal the current scale, so coefficients and rewards stay unchanged until governance opens the ranges.

## Fee

Historically, the protocol has charged no transaction fees. An attacker can therefore submit high-volume messages at little cost while every validator pays the processing and storage cost.

The upgrade groups transaction types and adds per-message gas rules so each group can be priced separately. All groups are disabled by default and charge nothing. Governance can enable and price them later. The upgrade proposal info can also enable groups at upgrade height.

## PoC Challenge

PoC and random Confirmation PoC prove a host's claimed capacity. During the rest of the epoch, inference statistics check that this hardware is used for work assigned by the protocol. A high rate of missed or invalid inferences can remove a host. This provides a strong ongoing check, but its sensitivity depends on inference volume and may not reveal every gap between claimed and available capacity. PoC Challenge adds an additional security layer for these cases.

[...]

PR #1786, opened 2026-09-16
chore(lint): exclude coefficients sim binary from forbidigo

redstartechno:

The `Check for panic or Must` lint job has been failing on `upgrade-v0.2.16` since 2026-09-08, which is one of the red checks on the train merge PR #1535.

The job fails with 5 forbidigo violations, all in the coefficients simulation tool added by #1566:

```
x/inference/coefficients/sim/main.go:116:3  use of `panic` forbidden
x/inference/coefficients/sim/main.go:162:3  use of `panic` forbidden
x/inference/coefficients/sim/main.go:168:9  use of `mathsdk.LegacyMustNewDecFromStr` forbidden
x/inference/coefficients/sim/main.go:182:3  use of `panic` forbidden
x/inference/coefficients/sim/svg.go:406:2   use of `panic` forbidden
```

`x/inference/coefficients/sim` is a standalone `package main` with its own `run.sh`, `config.json` and `README.md`. It is a simulation/plotting tool — it is never imported by, or linked into, the chain binary. The forbidigo rules exist to keep `panic` and `Must*` helpers out of consensus code ("panic forbidden in consensus code"), so they do not apply here.

Adds `^x/inference/coefficients/sim/` to the existing `linters.exclusions.paths` in `inference-chain/.golangci.yml`, next to the `^cmd/` entry that already exempts standalone binaries for the same reason.

I chose the exclusion over rewriting the tool's error handling deliberately: rewriting a simulation tool's `panic`-on-bad-config into error plumbing would be a larger change to someone else's new code, for no gain in consensus safety.

The exclusion is scoped to the `sim/` directory only. The real `x/inference/coefficients` package is still linted — if a `panic` or `Must*` is introduced there, the job still fails. No runtime behavior changes; this only affects lint scope. If you would rather have the tool rewritten to return errors, or moved under `cmd/`, say so and I will redo it that way.

Ran the same linter version the workflow uses (golangci-lint v2.6, `--timeout=15m`, working directory `inference-chain`) on `upgrade-v0.2.16` before and after the change — 5 issues before, 0 after. Also confirmed the package still builds and vets cleanly, since a lint exclusion must not hide a compile problem.

`inference-chain/.golangci.yml` (2 added lines). No Go source is modified.

## Testing & evidence

Local run against `upgrade-v0.2.16` at `bc4e4912f`.

### Before

```
$ golangci-lint run --timeout=15m
x/inference/coefficients/sim/main.go:116:3: use of `panic` forbidden because "panic forbidden in consensus code" (forbidigo)
		panic("hosts and max_passes must be positive; epochs must be non-negative")
		^
x/inference/coefficients/sim/main.go:162:3: use of `panic` forbidden because "panic forbidden in consensus code" (forbidigo)
		panic(value + " exceeds Decimal int64 storage")
		^
x/inference/coefficients/sim/main.go:168:9: use of `mathsdk.LegacyMustNewDecFromStr` forbidden because "Must* helpers forbidden in consensus code" (forbidigo)
	return mathsdk.LegacyMustNewDecFromStr(value)
	       ^
x/inference/coefficients/sim/main.go:182:3: use of `panic` forbidden because "panic forbidden in consensus code" (forbidigo)
		panic(err)
		^
x/inference/coefficients/sim/svg.go:406:2: use of `panic` forbidden because "panic forbidden in consensus code" (forbidigo)
	panic("model config not found: " + modelID)
	^
5 issues:
* forbidigo: 5
```

### After

```
$ golangci-lint run --timeout=15m
0 issues.
EXIT=0

$ go build ./x/inference/coefficients/... && go vet ./x/inference/coefficients/...
BUILD_VET_OK
```


redstartechno:
Closing as superseded by #1819, which fixed the `forbidigo` findings in `inference-chain/x/inference/coefficients/sim` at the source instead of excluding the directory. `lint` is no longer failing on the #1535 train, so this exclusion is no longer needed. Thanks!

PR #1787, opened 2026-09-16
fix(testermint): reference the real property in rehearsal prep patch

redstartechno:

The `Testermint Upgrade Rehearsal` job has been failing on `upgrade-v0.2.16` on a single Kotlin compile error, which is one of the red checks on the train merge PR #1535.

Every rehearsal run on the train fails with exactly one error:

```
e: file:///home/runner/work/gonka/gonka/old/testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt:67:51 Unresolved reference 'version'.
> Compilation error. See log for more details
```

`old/` is not a directory in this repo — `testermint-upgrade-rehearsal.yml` checks the candidate out into `path: new` and the previous release into `path: old` (currently `release/v0.2.15`). The file that fails to compile is created by `testermint/upgrade-rehearsal/previous-release-prep.patch`, which lives on this branch. So the fix belongs on the train, not in a released tag.

The reference is simply wrong. In `testermint/src/main/kotlin/data/devshard.kt`:

```kotlin
data class DevshardSettlementData(
    @SerializedName("escrow_id")
    val escrowId: String,
    // devshardctl finalize emits "version" (v1 wire compat); inferenced reads the same key.
    @SerializedName("version")
    val stateRootAndProtocolVersion: String,
    ...
```

`version` is the JSON wire key; the Kotlin property is `stateRootAndProtocolVersion`. There is no property named `version` on `DevshardSettlementData` — on `release/v0.2.15` or on this branch (the data class is byte-identical on both). The patch used the serialized name where it needed the property name.

One identifier in the prep patch:

```diff
-            settlementVersion = settlement.parsed.version
+            settlementVersion = settlement.parsed.stateRootAndProtocolVersion
```

Minimal. The change is confined to the rehearsal prep patch and affects only the rehearsal job; no production or chain code is touched. The value read is the same one either way — the property is the wire field's binding, so the JSON that `devshardctl finalize` emits is unchanged.

**Credit / overlap:** #1516 (@milkstoner, open against `main` since 2026-08-24) makes the identical change. This is not meant to duplicate it — `upgrade-v0.2.16` carries its own copy of this patch file, and the train branch is ahead of `main`, so merging #1516 into `main` does not turn this job green on the train. If you would rather land #1516 first and cherry-pick, close this and I will follow that route.

- The unmodified patch applies cleanly to `release/v0.2.15` (`git apply --check` passes), confirming the failure is at compile time, not apply time.
- After the change the patch still applies cleanly, and the resulting line 67 reads `settlementVersion = settlement.parsed.stateRootAndProtocolVersion`, which matches `val stateRootAndProtocolVersion: String` in that same checkout.
- The identifier being replaced does not exist on `release/v0.2.15`, which is exactly what the compiler reported.

**What I could not verify locally:** I have no JDK on this machine, so I did not run Gradle and cannot claim a green local `compileTestKotlin`. The rehearsal job in CI is the real verification here.

`testermint/upgrade-rehearsal/previous-release-prep.patch` (1 line). No Go code, no chain code, no production paths.

## Testing & evidence

Verified against a worktree of `release/v0.2.15` (`4d687ed67`), the release the rehearsal actually checks out.

### Before

```
$ git apply --check previous-release-prep.patch   # on release/v0.2.15
(applies cleanly — failure is at compile time)

CI:
e: .../old/testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt:67:51 Unresolved reference 'version'.

$ grep -nE "val stateRootAndProtocolVersion|val version" testermint/src/main/kotlin/data/devshard.kt
77:    val stateRootAndProtocolVersion: String,
(no `val version` — the old identifier never existed)
```

### After

```
$ git apply --check previous-release-prep.patch   # on release/v0.2.15
(applies cleanly)

$ git apply previous-release-prep.patch
$ sed -n '67p' testermint/src/test/kotlin/UpgradeRehearsalPrepTests.kt
            settlementVersion = settlement.parsed.stateRootAndProtocolVersion
```


redstartechno:
One caveat on verification, so nobody reads the green checks on this PR as proof of the fix: `testermint-upgrade-rehearsal.yml` triggers on `workflow_dispatch` and on `push` to `upgrade-v*`, not on `pull_request`. The rehearsal job therefore does not run against this PR — the checks here only cover the usual build/test workflows.

If you want it verified before merging, a `workflow_dispatch` of "Testermint Upgrade Rehearsal" with `candidate_ref` pointing at this branch (`redstartechno:fix/rehearsal-prep-settlement-version`) would exercise it. Otherwise the first real signal is the rehearsal run on the next push to `upgrade-v0.2.16`.

redstartechno:
Closing in favour of #1788, which makes the identical change to `previous-release-prep.patch` (`settlement.parsed.version` → `stateRootAndProtocolVersion`). One copy is enough for review, and either one clears `upgrade-rehearsal` on #1535. Thanks @DimaOrekhovPS.

PR #1828, opened 2026-09-24
x/inference: reject MsgCreatePoCChallenge when poc_v2_enabled=false

cyberdelamain:
`SubmitPoCChallengeValidations` refuses every vote when `poc_params.poc_v2_enabled` is false (`msg_server_poc_challenge.go:136`, `ErrNotSupported`), but `CreatePoCChallenge` never checks the flag. With V2 disabled a challenge can still be opened, but no vote can be stored while the flag stays false, so any punishable segment (>= `min_punishable_segment_blocks`) is evaluated with zero measured weight and fails (target removed from inference for the rest of the epoch, confirmation weight cut, challenger refunded).

This was noted by Copilot in the #1811 review (https://github.com/gonka-ai/gonka/pull/1811#discussion_r4075508966) and is still present on `upgrade-v0.2.16`. PoC Challenge ships with v0.2.16 (mainnet runs v0.2.15, where the feature is absent) and V2 is enabled on mainnet, so this only matters after that upgrade plus a governance change of the flag, but the create path should not accept a challenge that the vote path is guaranteed to reject.

Change: `CreatePoCChallenge` returns `ErrNotSupported` unless `PocParams.PocV2Enabled` is true, the same guard and error the vote handler uses. Test: `TestCreatePoCChallenge_RejectsWhenPocV2Disabled` in `keeper/poc_challenge_test.go` (fails on `upgrade-v0.2.16` without the change: the challenge is created; passes with it, together with the existing Challenge tests).

Not addressed here: challenges already OPEN when governance flips the flag true -> false (`SetParams` only reacts to false -> true).

Wallet for a possible bounty: gonka15u0r3mf6t7zsfuslusnyt7hsjrq357yumpe8st


gmorgachev (approved):
poc_v2_enabled is a dead code and can't be enabled on chain. But agree that it's cleaner this way until lefovers cleaned up

PR #1831, opened 2026-09-24
fee groups: charge ungrouped non-duty messages the highest enabled price

cyberdelamain:
### Problem

Since #1616 the ante fee check prices a tx as the max `min_gas_price` of the enabled groups its messages belong to (`FeeParams.EnabledPayingPrice`, `x/inference/types/fee_params.go`). A message type with no compiled group (`FeeGroupOf(msg) == ""`) is skipped, so a tx made only of such messages costs 0 whatever governance enables, and `Validate` rejects a `groups[].msgs` rule for an ungrouped type, so governance cannot close this without an upgrade.

Ungrouped today (inference module alone): `MsgUpdateParams`, `MsgSetClaimRecipients`, `MsgSubmitUnitOfComputePriceProposal`, `MsgRegisterModel`, `MsgDeleteGovernanceModel`, `MsgCreatePartialUpgrade`, `MsgRegisterBridgeAddresses`, `MsgRegisterLiquidityPool`, `MsgGovernanceCancelBridgeOperation`, `MsgMigrateAllWrappedTokens`, `MsgAddParticipantsToAllowList`, `MsgRemoveParticipantsFromAllowList`, `MsgPut/DeleteDevshardApprovedVersion`; plus `bls.MsgRequestThresholdSignature`, the other modules' `MsgUpdateParams`, and SDK types outside `cosmosTypeURLPrefixes` (`/cosmos.upgrade.`, `/cosmos.consensus.`, `/cosmos.auth.`, `/cosmos.mint.`). Most of these are gov-only and fail in the handler for any other signer, but signature verification and ante pass first, so they still reach a block for free.

The code already treats "ungrouped is free" as a hazard: `unwrapFeeMsgs` refuses to leave a `MsgExec` unclassified because "ungrouped would be free" (`app/ante_fee.go`), and `MsgRequestThresholdSignature` is deliberately left out of both the BLS group and `IsNetworkDuty` ("intentionally excluded") — yet the net effect is that it pays nothing. Its handler is deprecated and always fails (code 1103), so a funded non-participant can submit zero-fee failing txs (up to 32 KiB of data per message, many messages per tx) until blocks are full, even with every fee group enabled.

This is independent of #1539/#1576: that fix gates the fee-exempt duty types by signer; the path here needs no duty type and has no `GasCap`.

### Reproduction

App test through the full ante chain on `upgrade-v0.2.16` (not included here to keep the diff small; happy to add it to this PR): all 8 compiled groups enabled at 1000 ngonka, `Validate()` ok. A funded account that is not a participant signs `MsgRequestThresholdSignature` with fee = 0: `CheckTx` returns code 0 (admitted), `FinalizeBlock` returns code 1103 "message type is deprecated", sender balance unchanged. Control in the same setup: `bank.MsgSend` with fee = 0 gets code 13 (insufficient fee). With this change the same tx gets code 13 in `CheckTx`.

### Change

- `EnabledPayingPrice`: an ungrouped, non-exempt message raises the price to the new `MaxEnabledPrice()` (the highest enabled group price). With nothing enabled the result is 0 as before, so current behaviour is unchanged until governance enables a group. Network duties stay exempt, grouped types keep their own price.
- DAPI mirror `FeeTreeCache.PriceForMsgs` gets the same rule, so hosts sending `MsgSetClaimRecipients` / price proposals attach the fee the chain will ask for.
- Comment on `MessageFeeGroups` updated; test `TestEnabledPayingPrice_UngroupedPaysMaxEnabledPrice` in `x/inference/types/fee_params_ungrouped_test.go` (fails on `upgrade-v0.2.16`: price 0; passes with the change). `x/inference/types`, the `app` fee/ante tests and `decentralized-api/cosmosclient/tx_manager` pass.

### Behaviour change to agree on

Host-sent `MsgSetClaimRecipients` and `MsgSubmitUnitOfComputePriceProposal` will pay once any group is on. If they are meant to be free, the narrower alternative is to list them in the epoch group (or `IsNetworkDuty`) explicitly and keep fail-closed for everything else. Gov-executed messages are not affected: they run inside `x/gov`, not through ante.

Wallet for a possible bounty: gonka15u0r3mf6t7zsfuslusnyt7hsjrq357yumpe8st


GLiberman:
All mentioned msg types were mentioned in intentionallyUngrouped list.

cyberdelamain:
Thanks for looking. Agreed that the grouping decision is deliberate and recorded in `intentionallyUngrouped`; the PR was about a consequence of that decision rather than the list itself: since the ante fee check skips ungrouped messages, a tx made only of them is admitted to the mempool with fee = 0 even when every group is enabled, and the check that rejects it (authority, permission, or deprecation) runs later, in the handler. With all 8 groups enabled at 1000 ngonka, a non-participant's `bls.MsgRequestThresholdSignature` with fee = 0 gets `CheckTx` code 0 and is included, failing in `FinalizeBlock` with 1103, sender balance unchanged (app-level test through the full ante chain; happy to share it).

If the ledger should stay as it is, a narrower option that keeps these types ungrouped and only acts at mempool admission (`CheckTx`, no consensus change): after unwrapping `MsgExec` the same way `unwrapFeeMsgs` does (authz accepts an inner message signed by the executor itself without a grant), reject a gov-only ungrouped message (`MsgUpdateParams` of every module and the other governance-only registry/bridge/allowlist messages) whose signer is not the gov authority, and reject the deprecated `MsgRequestThresholdSignature`. Gov-executed messages run inside `x/gov` and never pass through ante, so no legitimate tx is affected. That narrows the gap rather than closing it: `MsgSetClaimRecipients`, `MsgSubmitUnitOfComputePriceProposal` and `genesistransfer.MsgTransferOwnership` would still be admitted free from any account unless the same check also requires the signer to be a participant (for `MsgTransferOwnership`: an eligible genesis account, i.e. listed in genesistransfer `allowed_accounts` while `restrict_to_list` is on, and not yet transferred). Happy to open that as a separate small PR if it is useful; otherwise no action needed.

